PR #19, opened 2014-05-08
Fixes for Clang/OS X build

mruegenberg:
I had a few problems building openvdb_maya with OS X 10.9 (which no longer ships with GCC) and Maya 2015 (which has a weird lib dir location on OS X).


0joshuaolson1:
@mruegenberg

PR #22, opened 2014-05-22
fixes for OpenVDBVisualize template

sdelhomme:
I had 2 issues with OpenVDBVisualize node:
1- grid menu value is always reset to *
2- changing grid value will always change first OpenVDBVisualize node created in Maya (because newOpenVDBVisualizeGridSelection is only called once).

Tested with Maya 2013 Linux x64


richhones:
This PR is now 10 years old and the openvdb_maya project officially unmaintained. I would recommend we close this. @kmuseth any objections?

kmuseth:
@richhones  - no objections

richhones:
@sdelhomme we don't have anyone looking after openvdb_maya anymore and it is only left in the repo as reference. As such, we are unable to verify these changes and so I am going to close this PR.  Apologies this never made it in.

PR #23, opened 2014-05-26
Check if vdb input is empty for OpenVDBReadNode

sdelhomme:
This is a fix that will avoid Maya to crash if a user try to connect an OpenVDBRead node to a OpenVDBVisualize node without first setting VdbFilePath.


OpenVDB-DevTeam:
This issue was revolved in a recent version.

PR #29, opened 2014-10-27
Cmake support

pigei:
Added CMAKE support. 
Added glew dependency to compile under Window


hevy-CC4477:
Do you intend to add the cmake support ?

phcerdan:
+1

PR #33, opened 2015-01-24
Makefile enhancements

davvid:
Make the build more configurable.
Follow the commonly-accepted meaning of DESTDIR and prefix variables in the Makefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+/openvdb/dependencies
+/openvdb/doc/html
+/openvdb/openvdb_viewer
+/openvdb/vdb_print
+/openvdb/vdb_render
+/openvdb/vdb_test
+/openvdb/vdb_view
+*.o
+*.so
+*.so.*
+*.swp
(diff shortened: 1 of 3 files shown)


MrMormon:
Uh, davvid?

davvid:
Hello.  It looks like this branch has gone stale.  It's been a while since I've taken a look at this PR.

Would you like me to take a stab at rebasing and updating this series to the latest "master" branch?  I guess a better question is whether these changes are desired.

I'd like to kick these changes upstream because so that we don't have to tweak openvdb at $dayjob for build-related stuff.  If this is desirable then I can take another stab.

The only caveat is that this *will* change the semantics of the existing build setup.  IMO that is okay because the existing semantics are ad-hoc and non-standard, so using the "standard" GNU makefile variables, etc. will only help packagers in the future.

Let me know if I should rebase this branch.  If so I'll take a stab at it again when I get a chance.

danrbailey:
I'm going to close this one out now this is pretty old and the focus is now on improving the CMake build. Thanks for the contribution nonetheless.

davvid:
Sweet, I'll have a few updates coming for the cmake build instead.  Namely, start using `GNUInstallDirs` so that we can do `install(.... DESTINATION ${CMAKE_INSTALL_LIBDIR})` etc. in the various `install()` places.  cheers

PR #35, opened 2015-05-18
LevelSetHDDA Improvements

danrbailey:


MrMormon:
Why was this closed?

caron:
`tools::LevelSetRayIntersector` uses `math::LevelSetHDDA` internally... the test was redundant?

MrMormon:
Sorry, I was just researching what parts of OpenVDB need to be merged before I fork it for my own documentation purposes.

danrbailey:
The confusion was that the ray-tracing in OpenVDB assumes vertex-centered rather than cell-centered transforms. This wasn't clear to me when I submitted this unit test.

I've suggested that ideally the ray-tracing should be able to handle both styles of transform, but as it stands this unit test isn't correct.

PR #36, opened 2015-06-07
glfw fix: request framebuffer size instead of window's

jehutymax:
In systems where the framebuffer and window sizes differ (such as on the OS X Mavericks and later), the framebuffer is the one we want. As it is, the viewport is about a quarter of the window size.


OpenVDB-DevTeam:
Great catch! /Mihai

PR #38, opened 2015-07-17
Viewer key fix

diekev:
Number keys (numpad or not) are not recognized here unless I use the keys definitions set in GLFW.

For the record here's my setup:
- Linux Mint 17 64-bit
- GLFW version 2
- Keyboard set to "French - Latin 9"


OpenVDB-DevTeam:
Thank you! We've integrated the changes.

PR #40, opened 2015-07-17
Compile fix: allow usage of float precision rays.

diekev:
It is currently impossible to compile client code making use of float precision rays, e.g. this code fails:

```
using namespace openvdb;
using namespace openvdb::math;
typedef Ray<float> RayType;
typedef tools::VolumeRayIntersector<FloatGrid,
                                                            FloatTree::RootNodeType::ChildNodeType::LEVEL,
                                                            RayType> IsectType;
```

This is mainly due to a misuse of the `RealT` or `RealType` template parameters and the globally defined `Real` typedef set to double, so some function called by the Ray or {Volume|LevelSet}RayIntersector classes would get different types whilst expecting a single type for all their parameters. For example `math::Min(T, T, T, T)` would get (float, double, double, double) when using float rays.

Also, one assert was failing in the Ray class when comparing floats, so `math::isRelOrApproxEqual` is used now.

### DIFF
--- a/openvdb/math/Ray.h
+++ b/openvdb/math/Ray.h
@@ -54,9 +54,9 @@ class Ray
 {
 public:
     BOOST_STATIC_ASSERT(boost::is_floating_point<RealT>::value);
-    typedef RealT      RealType;
-    typedef Vec3<Real> Vec3Type;
-    typedef Vec3Type   Vec3T;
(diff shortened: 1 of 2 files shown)


OpenVDB-DevTeam:
Thank you! We've applied these fixes

PR #41, opened 2015-07-17
Avoid using C++ built-in typeid keyword directly.

diekev:
This set of changes is required for the integration of OpenVDB in the Cycles render engine, from the Blender Foundation.

We can't use C++'s built-in RTTI system in Cycles, due to LLVM being used to JIT compile OSL source code (OSL is used as a shading backend in Cycles). So the 'typeid' keyword was replaced with `BOOST_SP_TYPEID`. By default, it's using the 'typeid' keyword, so nothing really changes, and if `BOOST_NO_TYPEID` is defined (like in Cycles' case), it'll fall back to a custom implementation.

There's also a `dynamic_cast` which gives issues here, so I changed it to a `reinterpret_cast`, but I'm not too sure about that (it could be a simple `static_cast` by the looks of it).

### DIFF
--- a/openvdb/Types.h
+++ b/openvdb/Types.h
@@ -264,7 +264,7 @@ enum MergePolicy {
 ////////////////////////////////////////
 
 
-template<typename T> const char* typeNameAsString()                 { return typeid(T).name(); }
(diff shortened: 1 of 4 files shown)


OpenVDB-DevTeam:
Thank you! We've integrated the changes.

PR #42, opened 2015-08-20
Duplicate source stealable updates

Idclip:
Updates to duplicateSourceStealable workflow in Filter, Filter Level Set and Convert SOPs to resolve issues.

 - all duplicateSourceStealable signatures now accessible in all compiled versions of houdini
 - Filter SOPs provide the current gdp to the duplicateSourceStealable call initiated on the "startNode"
 - Convert SOP unable to use duplicateSourceStealable, replaced with duplicateSource


OpenVDB-DevTeam:
Thank you for this contribution! We've integrated the changes.

PR #46, opened 2015-12-23
Added tbb::parallel_reduce functionality to NodeManager

bretttully-oxyntix:
We needed to do a map-reduce operation on the grid and found a huge performance hit using iterator ranges rather than the node manager.

NB: it there a way to ignore the CMake script? I couldn't work out how to exclude it from the pull request...


MrMormon:
This looks to be 16 commits behind.

OpenVDB-DevTeam:
I'm adding this to v3.2 (with some minor modifications) - thanks for your contribution

-Ken

PR #50, opened 2016-03-08
Multi-buffer I/O

danrbailey:


danrbailey:
Introduced a few extra small changes to this commit:

* #defines added with each feature - OPENVDB_HAS_SEEKABLE_IO and OPENVDB_HAS_MULTIPLE_LEAF_BUFFERS to help with enabling/disabling logic in dependent libraries (ie OpenVDB Points).
* 224 file format version number introduced to indicate in the header that the grid may use multiple leaf buffers.
* Support for a new environment variable to globally disable this feature on write - OPENVDB_DISABLE_DELAYED_LOAD. If this is set and the file format is 224, the version will be bumped back to 223 for all files when written to prevent any incompatibility warnings on reading these files in older versions of the library.

danrbailey:
Closing this Pull Request as this feature has now been accepted.

PR #51, opened 2016-03-09
Error in documentation (example code) at line 1029

MatteoRagni:
Error in one example of `Composite.h` operations (`cgsDifference(gridA,gridB)` instead of `cgsDifference(*gridA,*gridB)`)


OpenVDB-DevTeam:
Thanks - fixed is (and the csgIntersection) typos

-K

PR #60, opened 2016-04-08
Fix some compile errors due to mismatching parameter types

diekev:
This is an addendum to one of my previous PR (https://github.com/dreamworksanimation/openvdb/pull/40).

The previous patch would only address issues in the VolumeRayIntersector class, whilst this one takes care of the LevelSetRayIntersector class. The entire public API for the two classes were tested with both float and double type, so compilation should be fine now.


OpenVDB-DevTeam:
We've made these changes. Thank you!

Issue #59, opened 2016-04-05
TestTree.cc uses 'small' as a variable name, 'small' is already defined on Windows


caron:
I don't have a CLA but here is a tip for fixing the unittest for Windows builds...

https://github.com/caron/openvdb/blob/195735f531955119c9c4d57165dbdea3f34070d2/openvdb/unittest/TestTree.cc#L1662-L1680

small is defined in 'rpcndr.h' as...

`#define small char`

e4lam:
Please submit a pull request.

caron:
OK, but I need to learn how to submit a pull request for only a single changeset.

e4lam:
Thank you!

caron:
You're welcome! I personally have more fixes for Windows, some gathered from the community, that I think could also be considered. Right now the library itself builds nicely on Windows, but the viewer and the unittest don't. I am unable to build on Linux so I don't know when or if I break the Linux build. Some CI tool would be great here...ie Travis CI or AppVeyor

OpenVDB-DevTeam:
Thank you for reporting this! We've integrated your fix.

PR #61, opened 2016-04-08
Fix TestTree.cc for Windows build

caron:
#59 TestTree.cc uses 'small' as a variable name, 'small' is already defined on Windows


OpenVDB-DevTeam:
Thank you for the fix!

PR #66, opened 2016-06-05
openvdb/Makefile: Make python module install path a changable variable

dracwyrm:
The location to the installed python headers and modules are all variables
because they vary from system to system, so it makes sense to have the
install path a variable as well. It can't be the same as the system
location, because some distros need to compile and install to a temp
location before being moved -- Gentoo Linux, for example. Thus, the
temp location can be declared. Destdir cannot be used as it would cause
duplicated directory structure, for example:
DESTDIR=/tmp/usr
PYTHON_LIB_DIR=/usr/lib64/python2.7/site-packages
${DISTDIR}${PYTHON_LIB_DIR}=/tmp/usr/usr/lib64/python2.7/site-packages
DESTDIR needs to have usr in it otherwhise headers would be installed to
/include rather than /usr/include
Also, this makes it possible to compile 'make python' in a loop for producing
Python modules for different versions of Python installed on the system.

### DIFF
--- a/openvdb/Makefile
+++ b/openvdb/Makefile
@@ -158,6 +158,9 @@ PYCONFIG_INCL_DIR := $(PYTHON_INCL_DIR)
 # The directory containing libpython
 PYTHON_LIB_DIR := $(HFS)/python/lib
 PYTHON_LIB := -lpython$(PYTHON_VERSION)
+# The Directory to install the python module and includes to.
+PYTHON_INSTALL_INCL_DIR := $(DESTDIR)/python/include/python$(PYTHON_VERSION)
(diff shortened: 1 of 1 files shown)


arthurtiteica:
If I understand this right this will solve install paths like

    /usr/python/include
    /usr/python/lib

If so +1 from me.

dracwyrm:
@arthurtiteica Yeah. It's reusing their variables like DESTDIR, which is set to /usr (usually or prefix/usr), and most libs for python live in usr/some/path. The DESTDIR part can be dropped if people want to give the full path not relative to usr. Maybe they store the libs in /opt/ ? But on Gentoo Linux we have separate directories for all versions of python, so we can have multiple installed versions for maximum compatibility with all packages. Some may use python 3.5, but some may not be compatible with that, for example.

Issue #68, opened 2016-06-21
OpenVDB v3.2.0 Python Module Fails to Compile with Python v3.x 


dracwyrm:
Hi,

Using the source from master, at the date in the title, the python module fails with this message:

```
In file included from /usr/include/sched.h:28:0,
                 from /usr/include/pthread.h:23,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/x86_64-pc-linux-gnu/bits/gthr-default.h:35,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/x86_64-pc-linux-gnu/bits/gthr.h:148,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/ext/atomicity.h:35,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/bits/basic_string.h:39,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/string:52,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/5.3.0/include/g++-v5/stdexcept:39,
                 from /usr/include/boost/function/function_base.hpp:14,
                 from /usr/include/boost/function/detail/prologue.hpp:17,
                 from /usr/include/boost/function/function_template.hpp:13,
                 from /usr/include/boost/function/detail/maybe_include.hpp:13,
                 from /usr/include/boost/function/function0.hpp:11,
                 from /usr/include/boost/python/errors.hpp:13,
                 from /usr/include/boost/python/handle.hpp:11,
                 from /usr/include/boost/python/args_fwd.hpp:10,
                 from /usr/include/boost/python/args.hpp:10,
                 from /usr/include/boost/python.hpp:11,
                 from python/pyOpenVDBModule.cc:33:
python/pyOpenVDBModule.cc: In function ‘void init_module_pyopenvdb()’:
python/pyOpenVDBModule.cc:595:5: error: return-statement with a value, in function returning 'void' [-fpermissive]
     import_array();
```

However, Python v2.7.x will compile the module. If you need more details, please let me know.

Thanks,
Jon

Appendix: The compile settings:
[...]

PR #69, opened 2016-06-21
BUG: fix python3/2 numpy import_array macro build issue

dracwyrm:
This fixes the issue faced in issue #68.

There might be a cleaner fix, but this does work.

Tested with Python v2.7, 3.4, and 3.5.


dracwyrm:
I did some research and updated this PR with a much, much cleaner way of doing this. This is compile tested with Python v2.7, v3.4, and v3.5.

PR #71, opened 2016-08-24
Update recommended Doxygen version to online docs'

0joshuaolson1:
Although current 1.8.11 works for me too, so we could update the version project-wide instead.


0joshuaolson1:
Fixed by 4.0.0.

PR #75, opened 2016-09-21
Fix unit test warnings on Mac related to openvdb::Index64

danrbailey:
On Mac, openvdb::Index64 is uint64_t which is unsigned long long, not unsigned long. This patch fixes the warnings about the mismatching types in the unit tests by using Index64 instead of size_t or 0UL/1UL.


OpenVDB-DevTeam:
Funny - I made the exact same changes didn't push them yet :)

-K

PR #76, opened 2016-10-13
CMake configuration files and support macros

nyue:
Initial commit of CMake configuration files and support macros to build the OpenVDB core library and associated binaries.


meshula:
Nice!!! I can hardly wait to try this out.

caron:
Me too! Thanks Nicholas!

PR #85, opened 2016-11-04
Allow classes derived from LeafNode to optionally skip reading and writing the ValueMask

danrbailey:
No ABI or file format changes.


danrbailey:
Closing this pull request - it's more than two years old and I believe I've found a better way of getting around this issue when looking into delay-loading.

PR #90, opened 2016-11-10
Fix a crash in OpenVDB Combine SOP when using an unrecognised grid type

danrbailey:
This is only when using the OpenVDB Combine SOP for "Copy A" and "Invert A" modes with both inputs connected. This bug appears with any non-HDK supported grid types (such as mask grids, point index grids or point data grids).


OpenVDB-DevTeam:
This was resolved (in commit fcf1eb8) by removing the "unknown grid type" assertions from the various `UTvdbProcessTypedGrid*` functions in `UT_VDBUtils.h`.

PR #91, opened 2016-11-10
Fix a crash in OpenVDB To Spheres when passing in a VDB fog volume

danrbailey:


OpenVDB-DevTeam:
Thanks for reporting this issue! However the problem turns out to be unrelated to the fact your input is a FOG volume - I've identified the problem and will fix it in the next update.

-Ken

PR #93, opened 2016-11-11
Apply a std::max on grain size to ensure it is never zero.  

jmlait:
Since this is a simple fill of an array, place the minimum grain size as 1024.

This fixes an infinite loop if the number of things to fill falls below the number of threads.

### DIFF
--- a/openvdb_houdini/houdini/GEO_VDBTranslator.cc
+++ b/openvdb_houdini/houdini/GEO_VDBTranslator.cc
@@ -70,12 +70,16 @@
 #include <UT/UT_IStream.h>
 #include <UT/UT_Version.h>
 
+#include <FS/FS_IStreamDevice.h>
+
+#include <GA/GA_Stat.h>
 #include <GU/GU_Detail.h>
 #include <SOP/SOP_Node.h>
 #include <GEO/GEO_IOTranslator.h>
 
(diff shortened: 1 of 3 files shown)


openvdb_houdini/houdini/GEO_VDBTranslator.cc
```diff
@@ -70,12 +70,16 @@
 #include <UT/UT_IStream.h>
 #include <UT/UT_Version.h>
 
+#include <FS/FS_IStreamDevice.h>
```
danrbailey: Hi @jmlait - this header isn't available in Houdini 15.0 or earlier, would it be possible to add some #ifdefs around the new functionality please so we can still build VDB against earlier versions? (@OpenVDB-DevTeam)

jmlait: My apologies....  I should have broken this into two pull requests for exactly that reason.  I tried to avoid creating specific topic branches, but that then meant my second request got implicitly merged with the first :<  If you can get to the #ifdefs before me, I'll integrate your changes.  The fileStat() should work in 15.0, just the stream-aliasing is specific to 16.0  Otherwise I'll make a new update next week.  Sorry again for this situation.

PR #94, opened 2016-11-11
Enable Blosc in CMake C++ Unit Tests

danrbailey:
This fixes the issue where the library the unit tests are built against have OPENVDB_USE_BLOSC enabled, but when the unit tests themselves are built they include headers that don't.


danrbailey:
Tagging @nyue

PR #104, opened 2016-12-03
Introduce sequential pre-fetching from disk for faster performance

danrbailey:
When voxel and point attribute data is out-of-core and it is known that all the data will be read from disk, it is faster to sequentially pre-fetch the data using contiguous I/O rather than rely on on-demand loading in a multi-threaded context. Pre-fetching has minimal overhead if the data is already in-core.

Pre-fetching is exposed in the API and used in the Points Convert SOP.

### DIFF
--- a/openvdb/points/PointDataGrid.h
+++ b/openvdb/points/PointDataGrid.h
@@ -237,6 +237,15 @@ inline void
 setStreamingMode(PointDataTreeT& tree, bool on = true);
 
 
+/// @brief  Sequentially pre-fetch all delayed-load voxel and attribute data from disk in order
+///         to accelerate subsequent random access.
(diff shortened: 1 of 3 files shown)


OpenVDB-DevTeam:
Dan, can you please look into this [build failure](https://travis-ci.org/dreamworksanimation/openvdb/builds/201631735)?

danrbailey:
@OpenVDB-DevTeam I've just submitted a patch to fix the unit tests when building with Blosc disabled:

https://github.com/dreamworksanimation/openvdb/pull/133

Now that we have Travis CI, I'm going to re-submit all my outstanding pull requests to force a build so as to catch any further errors.

PR #105, opened 2016-12-03
Add Point Data Grid support to Clip SOP, Visualize SOP, Topology to Levelset SOP

danrbailey:
Adding trivial support for VDB Points to a few SOPs.
Clip SOP only supports inside bounding box clipping with VDB Points for now. Visualize and Topology to Levelset SOPs are fully supported with this change.


OpenVDB-DevTeam:
Merged (with minor changes)

PR #106, opened 2016-12-03
Extend GLSL shader patching to fix offset bugs in viewport decoration

danrbailey:
This fixes a long-standing bug in viewport decorations with VDB Points to apply GLSL shader patching for offsets for marker, normal and velocity decorations.

Disables viewport decorations for point number and point position, which are not yet supported.


danrbailey:
I've just added a new commit to this pull request that fixes viewport visualization for Houdini 16.

OpenVDB-DevTeam:
Merged (with minor changes)

PR #114, opened 2016-12-15
Workaround Visual Studio 2015 codegen bug that would crash ConnectNod…

e4lam:
…eMaskSegments


OpenVDB-DevTeam:
Merged (with minor changes)

PR #115, opened 2016-12-15
Workaround Visual Studio 2015 codegen bug that would crash ConnectNodeMaskSegments

e4lam:


OpenVDB-DevTeam:
Merged (with minor changes)

PR #117, opened 2016-12-21
Link against libHoudiniAPPS3 to fix issues loading GR primitive in Hython

danrbailey:


OpenVDB-DevTeam:
Fixed in commit 16419dc76d.

PR #122, opened 2017-01-10
Introduce install_lib rule to Makefiles for library and maya

danrbailey:
In the same style as the introduction of the recent install_lib rule for the Houdini makefile, this adds matching rules for the library and the Maya makefile to build and install the compiled libraries only. This is particularly useful for the library makefile as it avoids building the additional binaries (print, render, viewer, etc).


OpenVDB-DevTeam:
Merged, with some changes

PR #123, opened 2017-01-12
openvdb/CMakeLists.txt: Allow custom install directory for python mod

dracwyrm:
Some distros store python modules in a different directory than the
default one defined in CMakeFiles.txt file. This patch adds an
option to all this directory to be changed.

Compile and install tested on Gentoo Linux.

Signed-off by: Jonathan Scruggs (j.scruggs@gmail.com)

NOTE: This should also be compatible with the OpenVDB 3.3 branch.


nyue:
I would recommend changing the new CMake variable for controlling the python module installation path from
`PYOENVDB_INSTALL_DIRECTORY`
to
`OPENVDB_PYTHON_MODULE_INSTALL_DIR`
for naming consistency

dracwyrm:
This has been added so closing.

PR #124, opened 2017-01-13
Fix initialization bug with point data grid types

danrbailey:
Resolve initialization issues when calling openvdb::points::initialize()


OpenVDB-DevTeam:
Merged (commit 0010fab71e87) with minor changes

PR #127, opened 2017-01-18
Fix padding bug in stream compression

danrbailey:
Fix a critical bug in the compression of padded data arrays with a very specific byte size range.


OpenVDB-DevTeam:
Merged with minor changes

PR #128, opened 2017-01-31
Point Serialization Fixes

danrbailey:
Both of these changes do not change the file format or the ABI.

* Bug fix where writing out-of-core arrays could corrupt the metadata being written with an AttributeArray as the serialization flags are used to retain information about how the data is stored before being brought in-core. This issue can cause segmentation faults without this fix.
* Improvements to the multi-pass I/O in being able to correctly handle a variable number of passes per-leaf.

### DIFF
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1460,7 +1460,8 @@ Grid<TreeT>::readBuffers(std::istream& is)
         is.read(reinterpret_cast<char*>(&numPasses), sizeof(uint16_t));
         const io::StreamMetadata::Ptr meta = io::getStreamMetadataPtr(is);
         assert(bool(meta));
-        for (uint32_t pass = 0; pass < uint32_t(numPasses); ++pass) {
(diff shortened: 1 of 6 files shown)


OpenVDB-DevTeam:
Merged (commit 0010fab71e87) with minor changes

OpenVDB-DevTeam:
Dan, these changes seem to be incompatible with your earlier PointDataLeaf changes (though a bad merge is also a possibility).  Can you please take a look at this [build failure](https://travis-ci.org/dreamworksanimation/openvdb/builds/205086193)?

danrbailey:
@OpenVDB-DevTeam just pushed up a fix (https://github.com/dreamworksanimation/openvdb/pull/134). I built this through Travis when I saw this failing earlier today, so this fix should resolve things. If I'm not mistaken, I think this was just a merge issue.

OpenVDB-DevTeam:
Thanks for the quick fix!

PR #132, opened 2017-02-07
Introduce help files for OpenVDB SOPs

danrbailey:
These are based on a combination of SideFX help and DWA documentation.

Note that this pull request does not include CMake installation changes.


OpenVDB-DevTeam:
Resolved in commit a642f63

PR #145, opened 2017-04-13
Options to toggle implicit linking of Windows Boost

nyue:
On Windows, depending on the compiler version, implicit linking via header files can result in linkage failure of boost libraries.

To alleviate this problem, Boost implicit linking is disabled by default so that builds will have success on more variety of Visual Studio versions used to build OpenVDB


nyue:
Additional item added to pull request which fixes the linkage/missing-symbols error when building Houdini plugins on OS X.

PR #148, opened 2017-04-19
Changed the GUI_PrimVDBPointsHook priority to 1 to not clash with the houdini shipped one

danrbailey:


danrbailey:
This fix (contributed by @plattfot) gets rid of this warning on starting up Houdini when the open-source nodes are installed:

HDK Warning: "VDB Points" and "VDB Points" have the same priority (0)

PR #160, opened 2017-06-20
Some cmake tunes.

yangyanli:


Idclip:
@yangyanli thanks for these changes, they were used as a base to significantly improve the CMake python support in #382. Closing this as incorporated

PR #165, opened 2017-07-28
Adding support for PointDataGrid VDBs as masks in OpenVDB SOPs

danrbailey:
In Houdini 16, the UTvdbProcessTypedGridTopology() / GEOvdbProcessTypedGridTopology() methods do not include PointDataGrids. However many of the OpenVDB SOPs assume that all VDBs supported in Houdini are handled by this method, which in the worst cases (such as OpenVDB Create SOP) can cause a seg fault through dereferencing an invalid grid.

This patch adds PointDataGrid mask or topology support to the following SOPs  - Analysis, Clip, Create, Prune, Rasterize, Remove Divergence and Scatter. It also adds a warning to the Segment and Resample SOPs which are yet to be extended to support PointDataGrids.


danrbailey:
Closing this PR as @Idclip has re-submitted this here - https://github.com/dreamworksanimation/openvdb/pull/221.

PR #167, opened 2017-08-26
convert some boost functionality into c++11

dokipen3d:
Converted a lot of type_traits into c++11 (std::is_floating_point, std::is_same etc). Was a straight conversion and can then remove some boost headers.

Also changed a lot of boost::bind into std::bind. One issue there was that boost bind has the _1, _2 placeholders in a global namespace so there was ambiguous calling without the std::placeholders to qualify the names.

Another issue with boost bind was that I was getting errors with the bind not matching the overloaded function definition. Not sure if that was an issue with clang on OSX but the only way to get the right function to be bound was to change it's name and therefore making it explicit what function we want called. adding *Impl to the function name seemed an okay idea a the time.

All the tests in vdb_test pass and the SOP nodes seem to be working in houdini from my basic testing. 

I'm open to suggestion if this could be made any nicer (like removing the std::placeholders with namespace aliases?). 

Next step is to go through and change any boost::shared_ptr's to std::shared_ptrs (although there seems to be a comment about abi 3 compatibility which we want to keep for now I'm guessing?

### DIFF
--- a/openvdb/tools/ParticlesToLevelSet.h
+++ b/openvdb/tools/ParticlesToLevelSet.h
@@ -100,9 +100,6 @@
 
 #include <tbb/parallel_reduce.h>
 #include <tbb/blocked_range.h>
-#include <boost/bind.hpp>
-#include <boost/function.hpp>
-#include <boost/type_traits/is_floating_point.hpp>
 #include <boost/utility/enable_if.hpp>
 #include <boost/mpl/if.hpp>
 #include <openvdb/Types.h>
@@ -110,6 +107,8 @@
 #include <openvdb/math/Math.h>
 #include <openvdb/math/Transform.h>
 #include <openvdb/util/NullInterrupter.h>
+#include <type_traits>
+#include <functional>
 #include "Composite.h" // for csgUnion()
 #include "PointPartitioner.h"
 #include "Prune.h"
@@ -119,6 +118,7 @@ namespace openvdb {
 OPENVDB_USE_VERSION_NAMESPACE
 namespace OPENVDB_VERSION_NAME {
 namespace tools {
+
 
 namespace p2ls_internal {
 // This is a simple type that combines a distance value and a particle
@@ -134,7 +134,7 @@ template<typename SdfGridT,
 class ParticlesToLevelSet
 {
 public:
-    typedef typename boost::is_void<AttributeT>::type DisableT;
+    typedef typename std::is_void<AttributeT>::type DisableT;
     typedef InterrupterT                          InterrupterType;
 
     typedef SdfGridT                              SdfGridType;
@@ -143,7 +143,7 @@ public:
     typedef typename boost::mpl::if_<DisableT, size_t, AttributeT>::type  AttType;
     typedef typename SdfGridT::template ValueConverter<AttType>::Type AttGridType;
 
-    BOOST_STATIC_ASSERT(boost::is_floating_point<SdfType>::value);
+    BOOST_STATIC_ASSERT(std::is_floating_point<SdfType>::value);
 
     /// @brief Constructor using an exiting signed distance,
     /// i.e. narrow band level set, grid.
@@ -414,7 +414,7 @@ template<typename SdfGridT, typename AttributeT, typename InterrupterT>
 template<typename ParticleListT, typename GridT>
 struct ParticlesToLevelSet<SdfGridT, AttributeT, InterrupterT>::Raster
 {
-    typedef typename boost::is_void<AttributeT>::type DisableT;
+    typedef typename std::is_void<AttributeT>::type DisableT;
     typedef ParticlesToLevelSet<SdfGridT, AttributeT, InterrupterT> ParticlesToLevelSetT;
     typedef typename ParticlesToLevelSetT::SdfType   SdfT;//type of signed distance values
     typedef typename ParticlesToLevelSetT::AttType   AttT;//type of particle attribute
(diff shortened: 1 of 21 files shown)


dokipen3d:
Just making a note here for myself more than anything. 
Been looking into the bind usage in the level set filter. I was looking into a way to possibly use lambdas as a lot of the c++ community says that binds are a bad idea. Not sure why as they seem pretty cool to me. At first this usage was really confusing me due to the placeholders and the overloaded. But the placeholders are pure magic. The c++ standard says that the arguments are evaulated and if they dont fit (I assume it does all the usual lookup and deduction) then it is discarded. Then it goes to the next placeholder and any extra arguments and sees if that matches the overloaded function.

This seems like magic function glue! 

In the future, if we want to go full c++ geek, we might be able to look at variadic generic lambdas to do this. but that is a c++14 feature (VFX_PLATFORM 2018 + gcc 5.3.1 anyone?).

dokipen3d:
Another side note about bind. it seems with the c++bind, even though the arguments are supposed to be discarded, it can't find the right resolution. If you want to be able to do that you have to point it to the function and cast it to a function pointer with the right signature.

mTask = std::bind((int (Filter::*)(const LeafRange&, int))&Filter::median,  
                    std::placeholders::_1, 
                    std::placeholders::_2, 
                    width);

Not very readable. So if we do want to ditch boost, and switch over to c++11 bind, we either have to rename the function internally (like with the *Impl in the PR code) or add a cast like above (which IMHO is nasty)

dokipen3d:
I've converted all binds (std or boost) to c++11 lambdas. This should compile to better code and hopefully is more readable.

dokipen3d:
I see that some of the changes from boost to std have been done in [https://github.com/dreamworksanimation/openvdb/commit/ffb02d5527379a08c07449e39d3324ac2e5eff28](https://github.com/dreamworksanimation/openvdb/commit/ffb02d5527379a08c07449e39d3324ac2e5eff28)

....(even going so far as to add the *Impl to the function names, not sure if that was coincidence or not). 

Should I close this? Is the switch over to lambdas not a good idea? I was thinking that lambdas incur not runtime cost vs bind which I think has some (not that they are being used in performance sensitive places in the code). I might ask Ken when I get the chance.

OpenVDB-DevTeam:
Switching to lambdas is a good idea in general, but in these non-performance-critical cases they're perhaps less desirable given the added verbosity.

openvdb/tools/ParticlesToLevelSet.h
```diff
@@ -414,7 +414,7 @@ template<typename SdfGridT, typename AttributeT, typename InterrupterT>
 template<typename ParticleListT, typename GridT>
 struct ParticlesToLevelSet<SdfGridT, AttributeT, InterrupterT>::Raster
 {
-    typedef typename boost::is_void<AttributeT>::type DisableT;
+    typedef typename std::is_void<AttributeT>::type DisableT;
```
BenFrantzDale, line 417: Does this work? Looks like the boost version is either `true_type` or `false_type` whereas the std version is a value, `true` or `false`.
I think we aren't targeting C++17 yet, so sadly we can't use the `_v` and `_t` suffixes on these std metafunctions.

openvdb/tools/ParticlesToLevelSet.h
```diff
@@ -492,7 +494,9 @@ struct ParticlesToLevelSet<SdfGridT, AttributeT, InterrupterT>::Raster
                 mParent.mInterrupter->start(
                     "Rasterizing particles to level set using const spheres");
             }
-            mTask = boost::bind(&Raster::rasterFixedSpheres, _1, _2, SdfT(radius));
+            mTask = [radius](Raster* raster, const tbb::blocked_range<size_t>& range) {
+                return raster->rasterFixedSpheres(range, SdfT(radius));
```
BenFrantzDale, line 498: To be a pure refactor, shouldn't this capture the result of `SdfT(radius)` so `SdfT` isn't called in every call of `mTask`? That is, `mTask = [SdFTOfRadius = SdfT(radius)](Raster* ...`

openvdb/tools/ParticlesToLevelSet.h
```diff
@@ -758,11 +764,11 @@ struct ParticlesToLevelSet<SdfGridT, AttributeT, InterrupterT>::Raster
     getAtt(size_t n, AttT& a) const { mParticles.getAtt(n, a); }
 
     template <typename T>
-    typename boost::enable_if<boost::is_same<T,ValueT>, ValueT>::type
+    typename boost::enable_if<std::is_same<T,ValueT>, ValueT>::type
     Merge(T s, const AttT&) const { return s; }
 
     template <typename T>
-    typename boost::disable_if<boost::is_same<T,ValueT>, ValueT>::type
+    typename boost::disable_if<std::is_same<T,ValueT>, ValueT>::type
```
BenFrantzDale, line 771: Should this be `std::enable_if_v<!std::is_same<T, ValueT>::value, ValueT>`? (`std::enable_if_v` is C++14; `std::is_same_v` is C++17.)

In C++17, it looks like these two `Merge` functions could be combined and an `if constexpr` used to switch the behavior.

openvdb/tools/ParticlesToLevelSet.h
dokipen3d: I think the std::void () operator is a value, but ::type as far as I know is a true_type/false_type still (https://en.cppreference.com/w/cpp/types/is_void). If I'm wrong, then happy to be told so :)

openvdb/tools/ParticlesToLevelSet.h
dokipen3d: Yeah good point :)

openvdb/tools/ParticlesToLevelSet.h
dokipen3d: Looks like it would work. I share your wish of being able to use c++17 _v. It could probably be easily added to the std::namespace. 

Anyway, since this PR was made, I'm not exactly sure I'm allowed to contribute as I'd need to get it approved by the company I am contracting for, so feel free to make any changed yourself :)

Idclip:
I'm going to close this as a lot of these changes have since been incorporated (our new base standard is C++17!). The main thing that hasn't been changes is the std::bind -> lambda changes but it's probably not worth keeping this open just for that given the amount of conflicts. Thanks for highlighting all these!

PR #171, opened 2017-09-05
Change hardcoded paths to GNUInstallDirs variables.

dracwyrm:
According to standards, 64 bit files need to be installed to lib64
http://www.pathname.com/fhs/pub/fhs-2.3.html#LIB64

GNUInstallDirs enforces the standards on all systems, Linux,
Windows, MacOS, and BSD. These variables can be overridden.

Fix bug of not being able to manually set Python install dir.

Signed-off by: Jonathan Scruggs <j.scruggs@gmail.com>

NOTE: The only thing I'm unsure of is changing lib to CMAKE_INSTALL_LIBDIR
in openvdb_houdini/CMakeLists.txt. That needs to be double checked.

Thanks.


danrbailey:
Closing this PR as this suggestion has now been adopted as of this #498. Thanks!

PR #173, opened 2017-09-20
Fix Boost 1.65 compatibility

dracwyrm:
There may be a better way of doing this, but I'm not the best at using boost and python. At least, I hope this gives ideas on how to fix this properly.

### DIFF
--- a/openvdb/python/pyFloatGrid.cc
+++ b/openvdb/python/pyFloatGrid.cc
@@ -45,9 +45,11 @@ exportFloatGrid()
     // Add a module-level list that gives the types of all supported Grid classes.
     py::scope().attr("GridTypes") = py::list();
 
-    // Specify that py::numeric::array should refer to the Python type numpy.ndarray
-    // (rather than the older Numeric.array).
-    py::numeric::array::set_module_and_type("numpy", "ndarray");
+    #if BOOST_VERSION < 106500
+        // Specify that py::numeric::array should refer to the Python type numpy.ndarray
+        // (rather than the older Numeric.array).
+        py::numeric::array::set_module_and_type("numpy", "ndarray");
+    #endif
 
     pyGrid::exportGrid<FloatGrid>();
 #ifdef PY_OPENVDB_WRAP_ALL_GRID_TYPES
--- a/openvdb/python/pyGrid.h
+++ b/openvdb/python/pyGrid.h
@@ -81,6 +81,15 @@ using namespace openvdb::OPENVDB_VERSION_NAME;
 #pragma clang diagnostic pop
 #endif
 
+#if BOOST_VERSION < 106500
+    typedef typename py::numeric::array pyndarray;
+    namespace arrayFunc = boost::python::numeric;
+#else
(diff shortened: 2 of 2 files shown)


dracwyrm:
I'm just wondering if using the namespaces for the functions is the best course of action, or the best name for the namespace.
namespace arrayFunc = boost::python::numpy;
Because that doesn't actually point to the array function. I was wondering if there was a way to define arrayFunc so that it actually defines the function, then I would call arrayFunc(blah) instead of arrayFunc::array(blah).

Apart from that, I don't know if there is a cleaner way of doing this.

I appreciate any advice.

OpenVDB-DevTeam:
boost::python::numpy::array() takes a Python sequence object and, optionally, a numpy::dtype as arguments.  Code like
`py::object pointArrayObj = arrayFunc::array(py::list(), "float32");`
(pyGrid.h, line 1280) won't compile, because there is no implicit conversion from a string literal to a numpy::dtype.  In your testing, are you sure you're compiling the NumPy-related code?

dracwyrm:
@OpenVDB-DevTeam Yeah. It compiled the NumPy related code. Not sure how it actually worked with what you are saying. It even ran with the python fest script you have.

I think the function calls need to be wrapped in if statements where conversion code can be added, right? Is there a better way?

dracwyrm:
If I understand https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html correctly, would this work for numpy:
`py::object pointArrayObj = arrayFunc::array(py::list(), arrayFunc::dtype('f4'));`
f4 should be 32 bit float.

OpenVDB-DevTeam:
It's hard to say, but the preferred boost::python::numpy approach seems to be
`py::object pointArrayObj = arrayFunc::array(py::list(), arrayFunc::dtype::get_builtin<float>());`

The latest [commit](https://github.com/dreamworksanimation/openvdb/commit/ffb02d5527379a08c07449e39d3324ac2e5eff28) includes a (more elaborate) fix for Boost 1.65.  Can you give that a try?

dracwyrm:
@OpenVDB-DevTeam I just did. I only used the the parts that were for the python directory and cut out everything else.

I noticed a bug.
In openvdb/python/pyFloatGrid.cc, you have:
```
+#ifndef PY_OPENVDB_USE_BOOST_PYTHON_NUMPY
		      // Specify that py::numeric::array should refer to the Python type numpy.ndarray
		      // (rather than the older Numeric.array).
		      py::numeric::array::set_module_and_type("numpy", "ndarray");
 +#endif
```
However, PY_OPENVDB_USE_BOOST_PYTHON_NUMPY isn't defined, so it still tries to execute this line, thus failing.

I changed it to:
```
+#if BOOST_VERSION < 106500
     // Specify that py::numeric::array should refer to the Python type numpy.ndarray
     // (rather than the older Numeric.array).
     py::numeric::array::set_module_and_type("numpy", "ndarray");
+#endif
```
And it worked. I think pyFloatGrid.cc is called before pyGrid.h during the compile phase, thus that variable is not getting defined.

PR #186, opened 2017-11-03
Add zeroVal support for Mat3 and Mat3 typed attribute arrays

danrbailey:
* Fix a bug where zeroVal<Mat3R>() crashes.
* Add Mat3 as a TypedAttributeArray and support in the Houdini plugin.

### DIFF
--- a/openvdb/Types.h
+++ b/openvdb/Types.h
@@ -100,14 +100,18 @@ using math::Vec4d;
 
 // Three-dimensional matrix types
 using Mat3R = math::Mat3<Real>;
+using math::Mat3s;
+using math::Mat3d;
 
 // Four-dimensional matrix types
 using Mat4R = math::Mat4<Real>;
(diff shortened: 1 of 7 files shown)


dokipen3d:
Awesome!

OpenVDB-DevTeam:
Isn't it a little odd that `zeroVal<Mat>()` is `Mat::identity()` rather than `Mat::zero()`?  It's not clear why that was done for `Mat4` originally.

danrbailey:
@OpenVDB-DevTeam - yeah it is a bit odd. It's inconsistent as well in that the zeroVal specialisation for Quat is a zero quaternion, not the identity quaternion. 

Perhaps there should be a new identityVal<> method that resolves to zero for all PODs and vector types, but the identity for matrices and quaternions, then switch zeroVal<> for matrices and quaternions to be zero?

dokipen3d:
Would be great if we could merge this please!

PR #194, opened 2017-11-08
Add a missing endif that was lost.

jmlait:


OpenVDB-DevTeam:
Travis is reporting `Makefile:448: *** extraneous 'endif'` errors (see, e.g., job [#364.5](https://travis-ci.org/dreamworksanimation/openvdb/jobs/299258582#L854)).

jmlait:
That is bizarre, as it is intended to remove an insufficient number of endifs!  The change I made is very much missing an endif.

I wonder if the line 473,
ifeq (no,$(strip $(shared)))
    LIBHOUDINI := $(LIBHOUDINI_STATIC)
    LIBHOUDINI_IMPORT := $(LIBHOUDINI_STATIC)
else ifdef WINDOWS_NT
    LIBHOUDINI := $(LIBHOUDINI_SHARED)
else
    LIBHOUDINI := $(LIBHOUDINI_SHARED_VERSIONED)

    # Encode the path to libopenvdb.so into plugins, and because some
    # plugins might depend on Houdini libraries that are not linked into
    # hython (among other tools), encode the Houdini library path as well.
    HCUSTOM_EXTRA_LDFLAGS += -Wl,-rpath,$(OPENVDB_LIB_DIR) -Wl,-rpath,$(HDSO)
endif # shared

Might be confusing some Makefile analysis software?  That will make it look like insufficient endifs if they don't realize they are supposed to be shared?

OpenVDB-DevTeam:
I'm wondering if we're looking at the same file.  The [`openvdb_houdini/Makefile`](https://github.com/dreamworksanimation/openvdb/blob/master/openvdb_houdini/Makefile) on the master branch has no line that reads "`else ifdef WINDOWS_NT`".

jmlait:
Aha, I've located the discrepancies. 

When the line
houdini/GR_PrimVDBPoints.o: HCUSTOM_FLAGS += -L$(HDSO) -lHoudiniAPPS3
was added we had to guard that with ifdef LINUX on our version of the Makefile.  That change was never sent back upstream or not merged.  So when a new line was added after, the merge process wiped out the endif on my version of the Makefile.

The other else ifdef is again Windows specific code branches.

My apologies for submitting this red herring!  Thank you for reviewing it.

PR #198, opened 2017-11-09
Do not set app title as doing so is unexpectedly sticky.

jmlait:


OpenVDB-DevTeam:
Would calling `setAppTitle(nullptr)` or `setAppTitle("")` in `~Interrupter()` solve the problem?

jmlait:
The problem is the app title isn't behaving like a stack.   So if you clear it, you wipe out whatever legitimate title was set earlier.  To make matters worse, there is likely no prior legitimate title as the initial startup value isn't set properly.  So get/restore likely doesn't work.  And I'm not sure if you can restore it to the never-set default state.  So the best solution is to not use the setAppTitle at all.

The bug the user sees is the interrupt dialog gets named "Converting..." and seems to stick with that; since some implicit VDB operation pushes that title and it sticks.

OpenVDB-DevTeam:
In that case, perhaps it would be best to save the title and use it as the `opname` argument to `opStart`.  Like this, maybe:
```
class Interrupter
{
public:
    explicit Interrupter(const char* title = nullptr):
        mUTI{UTgetInterrupt()}, mRunning{false}, mTitle{title ? title : ""}
    {}

    ...

    void start(const char* name = nullptr) {
        if (!mRunning) { mRunning = true; mUTI->opStart(name ? name : mTitle.c_str()); }
    }

    ...

private:
    UT_Interrupt* mUTI;
    bool mRunning;
    std::string mTitle;
};
```

jmlait:
Yes, that would make more sense.  

However, looking at least at my version of SOP_OpenVDB_Fracture I'm not sure where start() is ever called.   (Note there is no need to call start; wasInterrupted() can pick up the parent's interrupt stack which implicitly wraps every node)   So, even with this change the title text will still never show up.

In any case, setting the title doesn't work the way one would expect, so should be avoided.

OpenVDB-DevTeam:
A number of tools make calls to `start()` (see `tools/Filter.h`, for example), but it looks like in all cases they pass an opname that would override the title.  Still, I'll make that change just to be safe.

PR #199, opened 2017-11-09
Add += support for bools in SumOp.

jmlait:
With thanks to Dan Bailey for the patch.


jmlait:
I can't quite figure out how to patch in the DCO so have resubmitted as PR #318.

PR #201, opened 2017-11-10
Fix Indirect Header Includes

danrbailey:
* Fix a bug where Travis CI was not failing on the first error
* Introduce a new rule to the Makefile and to Travis to catch any indirect header includes
* Resolve all existing indirect header includes


OpenVDB-DevTeam:
Merged with changes in commit 220851ae3527b804fb784869792fb20e8743a34d.

PR #202, opened 2017-11-10
Minor include and Point Delete fixes 

Idclip:


danrbailey:
@Idclip - these changes look good to me. Are the two new conditionals being evaluated by the unit tests, otherwise might be worth adding that?

PR #203, opened 2017-11-15
getUnload() does not return bool so cannot compare with true.

jmlait:
This generates an unsafe == warning on MSVC builds.


jmlait:
The real problem is clearly that Houdini's getUnload() still returns int when it should be returning bool.  But that isn't changing soon :>

MSVC will complain about a performance issue with some settings on that.  It would work for us as we have that warning disabled.

What we had done in our build prior to merging in your change was:

`if(node->getUnload() != 0) return true;`

I can't seem to find where we diverged here, but the 0 was done as `true` in both the official git and our public facing git repository.

OpenVDB-DevTeam:
Odd that MSVC would complain about the explicit cast to bool but not about the implicit cast in the boolean context of the "if" expression.  But testing against zero seems like a reasonable compromise.  Do you want to make that change?

jmlait:
Switched to != 0.

PR #204, opened 2017-11-15
OpFactory-based SOP verbification API

OpenVDB-DevTeam:
Jeff, would you mind to take a look at this alternative verbification API?  In particular, do `OpFactoryVerb` and `OpFactory::setVerb` in `ParmFactory.cc` seem sufficiently general to you, or are there use cases that they don't cover properly?

### DIFF
--- a/openvdb/CHANGES
+++ b/openvdb/CHANGES
@@ -5,8 +5,20 @@ Version 5.1.0 - In development
 
     Houdini:
     - Added output grid naming options to the Points Convert SOP.
+    - Added a SOP_NodeVDB::evalStdString() method that facilitates
+      string parameter evaluation in expressions, e.g.,
+      matchGroup(*gdp, evalStdString("group", time)).
     - Added a SOP_NodeVDB::matchGroup() overload that eliminates the need
       to const_cast the GU_Detail.
+      [Contributed by Jeff Lait]
+    - Grid transforms are now more aggressively simplified, making it
+      less likely to produce nonuniform voxels erroneously.
+      [Contributed by Jeff Lait]
+    - Added a houdini_utils::OpFactory::setVerb() method to register
+      compilable SOPs.
+    - Added SOP_VDBCacheOptions, a convenience base class for compilable SOPs.
+      [Contributed by Jeff Lait]
+    - The Advect SOP is now a compilable SOP.
       [Contributed by Jeff Lait]
 
 
--- a/openvdb/CHANGES
+++ b/openvdb/CHANGES
@@ -8,6 +8,8 @@ Version 5.1.0 - In development
     - Added a SOP_NodeVDB::evalStdString() method that facilitates
       string parameter evaluation in expressions, e.g.,
       matchGroup(*gdp, evalStdString("group", time)).
+    - Removed the deprecated openvdb_houdini::validateGeometry() function.
+      Use convertGeometry() instead.
     - Added a SOP_NodeVDB::matchGroup() overload that eliminates the need
       to const_cast the GU_Detail.
       [Contributed by Jeff Lait]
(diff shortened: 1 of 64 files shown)


jmlait:
Well, it is a bit of a quandary...   I made that #define SOP_CLASS for a very specific reason: to minimize the diff introduced by adding verb support to legacy nodes.   I have 27 nodes so converted in our version of the baseline (as VDB gained Verb support in 16.0)   But despite those changes, it is still possible to merge in your changes for 5.0 as the absolute change was minimal.  

The proposed change to SOP_OpenVDB_Advect is quite significant and would be very arduous to merge.  Is such a large change required by the new API?  Because if it is it will be a lot of work to port the rest of the nodes, as I can't use the existing code I've already done.  Ideally whatever API is used will have a way to move nodes over with as few lines as possible and have as little effect on merges.

If you do go through with this, I would have to maintain the #define DEFINE_VERB_IMPLEMENTATION on my version for other nodes.  And I'm not sure I'd have time to make that big of a merge any time soon.  (Keep in mind we have a lot of small but important differences; our version is SOP_OpenVDB_Advect_Level_Set rather than Advect; our version has parameters named advectspatial rather than advectSpatial, our version has "boost" swapped with "hboost" everywhere...)

As an API for defining new nodes it seems reasonable, and definitely nicer than the SOP_CLASS approach I made.  OPFactoryVerb isn't fully general, there are some other cases that one may want access to (such as nodes that only cook a subset of their inputs, or nodes that use local variables) but it looks like these exceptions could be done by building one's own local subclass with the extra methods set.

danrbailey:
I worry that the SOP_CLASS approach might add a little too much confusion to a new developer coming to the codebase and it generally concerns me that we might shy away from a more readable solution just for the sake of easing the one-time merge process. Nonetheless, I do appreciate there are a lot of nodes affected by this and pretty significant changes to each of them. 

Perhaps it would be better to try and do an initial pass to attempt to align the two codebases a little more before attempting a merge. For example, I think the DWA nodes should be converted to use the Houdini convention of all lowercase parameters (but with extra obsolete parameter hooks to maintain backwards compatibility). An hboost macro could perhaps remove some of the merge complexity on the Houdini end too. Finally a slightly more longer-term change, but I've thought for a while that there's too much logic in the SOP layer which should really be migrated up into the core library API. In this case, the AdvectionParms and AdvectOp class have very minimal ties to the HDK which could fairly easily be removed and benefit from having unit tested functionality and re-usability outside of Houdini.

jmlait:
The newer SOPs seem to work a lot better - they tend to avoid defining things inside their class; avoid member variables on the class, and strongly separate the parameter reading from the parameter use.  This makes having different names in those cases less frustrating.  (And they tend to be named nicely to start with!)

If I seem reluctant to accept a one-time merge cost; that would be because I just finished paying the one-time merge cost of 4.0 -> 4.2 :>  Trying to bring in all the tooltip changes without accidentally renaming a parm was... interesting.

Might be good to try verbifying the SOP OpenvDB Filter or Filter Level Set.   This is an interesting case if you want to retain the old behaviour as well, as in the non-verb case it relies on OP_NodeChain; and the verb-based approach relies on INPLACE being efficient.

My big fear is that to big a change will cause us to diverge from the published VDB version.  From Particles and From Polygons have both sadly diverged a bit; and I didn't get the time to bring the improvements to Clip forward.  If someone wants I can tar up our version of the nodes and email it to you so you can see what we are merging into.

OpenVDB-DevTeam:
Yes, it would help to see at least a few examples of the code differences between your versions of the nodes and ours, and I'm open to renaming our parameters if that would help to ease the burden of merging.  (And Dan is right that some code can probably be migrated into the core library.)

Idclip:
In regards to OP_NodeChain, an issue we've been having is the result from networks utilising both SESI and DW nodes. The parameter interface differences between the SESI/DW Filter SOPs breaks this! I'd recommend deprecating OP_NodeChain in favour of either unloading nodes (which then allows both filter SOPs to utilize "duplicateSourceStealable") or, as @jmlait says, using the verbified INPLACE option. Both results should be near identical to the node chain workflow and will help with any verb-based migration.

In any case, I have a temporary fix I've been using which just ensures the node chain breaks instead of collecting these mixed parameter interfaces:

https://github.com/Idclip/openvdb/commit/3ee4f04dd511fa39f6bf697173f071910ded81f8

PR #213, opened 2018-01-04
VDB Points Python Bindings

danrbailey:
A re-submission of PR163, but now with python builds and tests added to Travis CI.


danrbailey:
I've just rebuilt this PR to only include the Python binding changes, as Travis now builds the Python bindings so there are no longer any modifications required there.

danrbailey:
Rejecting this PR, going to re-submit as I'm not sure why Travis CI didn't get kicked off.

PR #214, opened 2018-01-16
Fix for building defaults because SizeTraits falsely returns 1 for Mat4.

jmlait:
We ran into an out of bounds write because SizeTraits in the Mat4 case was returning 1.  I couldn't deduce why the template magic was failing, but by rolling the GA_Defaults into the getValues you can avoid a lot of action-at-distance and just make the values array on the stack.  Thus it is more efficient and less error prone.

### DIFF
--- a/openvdb_houdini/houdini/PointUtils.cc
+++ b/openvdb_houdini/houdini/PointUtils.cc
@@ -727,98 +727,106 @@ createTypedMetadataFromAttribute(const GA_Attribute* const attribute, const uint
     return openvdb::TypedMetadata<ValueType>(value).copy();
 }
 
-
-template<typename T> struct SizeTraits                          {
-    static const int Size = openvdb::VecTraits<T>::Size;
-};
-template<typename T> struct SizeTraits<openvdb::math::Quat<T> > {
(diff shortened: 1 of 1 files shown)


danrbailey:
Thanks Jeff, I agree this is better. :)

jmlait:
That still leaves the action-at-distance of passing a raw T * into the getValues and hoping the caller has things sorted out.  So we are one mis-typed template away from re-introducing the problem.   This is a rather hard problem to detect & isolate as it is just random memory corruption; so I'd recommend going with the most robust solution rather than least code solution.  (The submitted bug is: "Bypass and unbypass this node 20-30 times and it crashes..."  I did a *lot* of clicking while working on this :>)  

The proposed patch isolates the result of SizeTraits to the getValue functions, which we have to specialize for each type anyways.  And also allows us to avoid memory allocation entirely, which is a good thing.

OpenVDB-DevTeam:
Those are certainly valid points.

Dan, OK to merge?

danrbailey:
Yes, I think this is a good change. We don’t have automated testing of this Houdini code, so less error-prone code is better.

PR #221, opened 2018-03-08
Improved VDB Points grid handling in some SOPs

Idclip:


openvdb_houdini/houdini/SOP_OpenVDB_Resample.cc
```diff
@@ -506,7 +507,7 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Resample)::cookVDBSop(OP_Conte
                 refGdp = gdp;
             }
             if (refGdp) {
-                refGroup = matchGroup(*gdp, evalStdString("reference", time));
+                refGroup = matchGroup(*refGdp, evalStdString("reference", time));
```
jmlait: Good catch!

pcucka:
>  I think it would be better if we had a UTvdbProcessFOO that handled Typed & Point grids for these super-generic.

Perhaps even better would be a more generic `UTvdbProcess<GridTypeList, OpType>` template.  Making the `GridTypeList` a `std::tuple` (or something similar) of allowed grid types would enable us to predefine type lists for the existing categories while giving us the flexibility to define custom lists as needed.

danrbailey:
Closing this one down, I've just submitted the same basic contents of this PR in #591 but using GEOvdbApply and GridBase::apply()

PR #227, opened 2018-03-21
Fix memory leaks in static variable pointers.

mortenbh:
Hi,

over at Autodesk we run OpenVDB under Valgrind, and we discovered that OpenVDB leaks a number of singletons in static variable pointers. This patch fixes these memory leaks.

Notice, we have used Meyers singleton pattern makes sure that the singletons are only allocated once the static function they reside in is called. This is also thread-safe in C++11 and above.

In AttributeArray,h it was not as straight-forward to use this pattern since the singleton's constructor depending on some code outside the class. In that case (and in Tree.h) we instead used std::call_once which is also thread-safe and serves the same purpose.

### DIFF
--- a/openvdb/Grid.cc
+++ b/openvdb/Grid.cc
@@ -74,35 +74,13 @@ struct LockedGridRegistry {
     GridFactoryMap mMap;
 };
 
-// Declare this at file scope to ensure thread-safe initialization.
-Mutex sInitGridRegistryMutex;
-
 
 // Global function for accessing the registry
 LockedGridRegistry*
 getGridRegistry()
 {
-    Lock lock(sInitGridRegistryMutex);
-
-    static LockedGridRegistry* registry = nullptr;
-
-    if (registry == nullptr) {
-
-#ifdef __ICC
-// Disable ICC "assignment to statically allocated variable" warning.
-// This assignment is mutex-protected and therefore thread-safe.
-__pragma(warning(disable:1711))
-#endif
-
-        registry = new LockedGridRegistry();
-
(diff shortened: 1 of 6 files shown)


mortenbh:
Rebased against latest master and signed off the commit.

jmlait:
Thank you for the PR!   This was not merged earlier because:
1) It isn't actually a memory leak as the resources are freed in either case on process termination.  Because static destruction order is a source of nightmares, there was a tendency to be conservative about triggering it unnecessarily.
2) While you are correct this is C++11 safe, sadly MSVC was rather late to the game in supporting static function variables.

We think we are safely beyond the non-conforming compilers, however.

We would like to use this PR as a test of using the CLA mechanism to ingest external code.  This will require us to wait still longer until the CLA system is enabled, so hopefully you will have continued patience with us!

mortenbh:
Sounds good. I'll be looking forward to it.

jmlait:
Thank you for your patience!   We think the EasyCLA system is in place and hopefully working.

Can you please do an empty commit to trigger the CI, DSO, and EasyCLA tests again?

mortenbh:
Hi Jeff, thaks for getting back to me. The PR has has a conflict now so it'll be a small amount of work to fix this. I'll have a look at this as soon as I have a moment. Stay tuned.

jmlait:
Excellent, thank you for following up!

mortenbh:
I've rebased against latest master and resolved the merge conflicts now.

jmlait:
Thank you for the contribution!  It is now in!

PR #230, opened 2018-04-19
points/PointSample.h for sampling numeric VDB volumes onto PointDataGrids

Idclip:
- Added point sample functionality for VDB Points to the Sample Points Houdini SOP
     - Moved filterPointCount functions on PointCount.h out of the internal namespace
     - Unit Tests

### DIFF
--- a/openvdb/points/PointCount.h
+++ b/openvdb/points/PointCount.h
@@ -117,6 +117,36 @@ Index64 inactiveGroupPointCount(const PointDataTreeT& tree, const Name& name,
     const bool inCoreOnly = false);
 
 
+/// @brief Total points in a tree which are included by a filter
+/// @param tree PointDataTree.
+/// @param filter Filter of the specified template type
+/// @param inCoreOnly if true, points in out-of-core leaf nodes are not counted
+template <typename PointDataTreeT, typename FilterT>
+Index64 filterPointCount(const PointDataTreeT& tree,
+    const FilterT& filter,
+    const bool inCoreOnly = false);
+
+
+/// @brief Total active points in a tree which are included by a filter
+/// @param tree PointDataTree.
+/// @param filter Filter of the specified template type
+/// @param inCoreOnly if true, points in out-of-core leaf nodes are not counted
+template <typename PointDataTreeT, typename FilterT>
+Index64 filterActivePointCount(const PointDataTreeT& tree,
+    const FilterT& filter,
+    const bool inCoreOnly = false);
+
+
+/// @brief Total inactive points in a tree which are included by a filter
+/// @param tree PointDataTree.
+/// @param filter Filter of the specified template type
+/// @param inCoreOnly if true, points in out-of-core leaf nodes are not counted
+template <typename PointDataTreeT, typename FilterT>
+Index64 filterInactivePointCount(const PointDataTreeT& tree,
+    const FilterT& filter,
+    const bool inCoreOnly = false);
+
+
 /// @brief Generate a new grid with voxel values to store the number of points per voxel
 /// @param grid             the PointDataGrid to use to compute the count grid
 /// @param includeGroups    a vector of VDB Points groups to be included (default is all).
@@ -200,34 +230,33 @@ Index64 threadedFilterPointCount(   const PointDataTreeT& tree,
     return tbb::parallel_reduce(leafManager.leafRange(), Index64(0), pointCountOp, PointCountOp::join);
 }
 
+/// @note  The following functions have been moved out of the point_count_internal
+///        namespace
 
 template <typename PointDataTreeT, typename FilterT>
-Index64 filterPointCount(const PointDataTreeT& tree,
+OPENVDB_DEPRECATED Index64 filterPointCount(const PointDataTreeT& tree,
                          const FilterT& filter,
                          const bool inCoreOnly = false)
 {
-    using ValueIterT = typename PointDataTreeT::LeafNodeType::ValueAllCIter;
-    return threadedFilterPointCount<  PointDataTreeT, FilterT, ValueIterT>(tree, filter, inCoreOnly);
+    return points::filterPointCount<PointDataTreeT, FilterT>(tree, filter, inCoreOnly);
 }
 
 
 template <typename PointDataTreeT, typename FilterT>
-Index64 filterActivePointCount( const PointDataTreeT& tree,
+OPENVDB_DEPRECATED Index64 filterActivePointCount(const PointDataTreeT& tree,
                                 const FilterT& filter,
                                 const bool inCoreOnly = false)
 {
-    using ValueIterT = typename PointDataTreeT::LeafNodeType::ValueOnCIter;
-    return threadedFilterPointCount<  PointDataTreeT, FilterT, ValueIterT>(tree, filter, inCoreOnly);
+    return points::filterActivePointCount<PointDataTreeT, FilterT>(tree, filter, inCoreOnly);
 }
 
 
 template <typename PointDataTreeT, typename FilterT>
-Index64 filterInactivePointCount(   const PointDataTreeT& tree,
+OPENVDB_DEPRECATED Index64 filterInactivePointCount(const PointDataTreeT& tree,
                                     const FilterT& filter,
                                     const bool inCoreOnly = false)
 {
-    using ValueIterT = typename PointDataTreeT::LeafNodeType::ValueOffCIter;
-    return threadedFilterPointCount<  PointDataTreeT, FilterT, ValueIterT>(tree, filter, inCoreOnly);
+    return points::filterInactivePointCount<PointDataTreeT, FilterT>(tree, filter, inCoreOnly);
 }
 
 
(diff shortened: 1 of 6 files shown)


openvdb/points/PointSample.h
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    void operator()(const typename LeafManagerT::LeafRange& range) const
+    {
+        PointFilterT filterCopy(mFilter);
+
+        for (auto leaf = range.begin(); leaf; ++leaf) {
+            // get the relevant point data attribute arrays for the voxel in question
+            typename PositionAttributeHandleT::Ptr positionHandle =
```
danrbailey: Some autos would clean this up a bit.

openvdb/points/PointSample.h
```diff
@@ -0,0 +1,379 @@
...
+
+///////////////////////////////////////////////////
+
+
+namespace point_sample_internal {
+
+template<typename SamplerT,
+         typename AccessorT,
+         typename TargetValueT,
+         typename PositionT = Vec3f,
```
danrbailey: PositionT cannot currently be anything other than Vec3f so this template argument should be removed, especially as it's only part of an internal API.

openvdb/points/PointSample.h
```diff
@@ -0,0 +1,379 @@
...
+         typename PointFilterT = points::NullFilter,
+         typename CodecT = points::UnknownCodec,
```
danrbailey: As the CodecT argument isn't used for anything other than UnknownCodec in the public API, I would suggest removing it.

openvdb/points/PointSample.h
```diff
@@ -0,0 +1,379 @@
...
+                    // voxels when doing a staggered offset by 0.5
+
+                    Vec3d position = Vec3d(positionHandle->get(index)) + coordVec3d;
+
+                    // transform it to source index space if transforms differ
+
+                    if (Transform) {
+                        Vec3d worldPosition = mPointDataGridTransform->indexToWorld(position);
+                        position = mSourceGridTransform->worldToIndex(worldPosition);
+                    }
```
danrbailey: I think I'd suggest wrapping up all the position access logic into a new struct so this code can be cleaned up a bit.

openvdb/points/PointCount.h
```diff
@@ -117,6 +117,36 @@ Index64 inactiveGroupPointCount(const PointDataTreeT& tree, const Name& name,
     const bool inCoreOnly = false);
 
 
+/// @brief Total points in a tree which are included by a filter
+/// @param tree PointDataTree.
+/// @param filter Filter of the specified template type
+/// @param inCoreOnly if true, points in out-of-core leaf nodes are not counted
+template <typename PointDataTreeT, typename FilterT>
+Index64 filterPointCount(const PointDataTreeT& tree,
```
danrbailey: A lot of this functionality was written before we went over to C++11, so features like default function template arguments weren't available then! Having access to this kind of functionality could improve this public API, I think perhaps something like this would be cleaner than having separate pointCount() and filterPointCount() functions:

```
template <typename PointDataTreeT, typename FilterT = NullFilter>
Index64 pointCount(...)
```

What do you think?

openvdb/points/PointCount.h
```diff
@@ -117,6 +117,36 @@ Index64 inactiveGroupPointCount(const PointDataTreeT& tree, const Name& name,
...
+    const FilterT& filter,
+    const bool inCoreOnly = false);
+
+
+/// @brief Total active points in a tree which are included by a filter
+/// @param tree PointDataTree.
+/// @param filter Filter of the specified template type
+/// @param inCoreOnly if true, points in out-of-core leaf nodes are not counted
+template <typename PointDataTreeT, typename FilterT>
+Index64 filterActivePointCount(const PointDataTreeT& tree,
```
danrbailey: Probably out of the scope of this feature, but just to mention an inconsistency - in the PointDataGrid class there's an "onPointCount()" instead of "activePointCount()". Perhaps it would make sense to attempt to unify these.

openvdb_houdini/houdini/SOP_OpenVDB_Sample_Points.cc
```diff
@@ -265,6 +302,24 @@ newSopOperator(OP_OperatorTable* table)
...
             " (see [specifying volumes|/model/volumes#group])"));
 
+    // vdb points groups
+    parms.add(hutil::ParmFactory(PRM_STRING, "vdbpointsgroups", "VDB Points Groups")
+        .setHelpText("Specify subset of VDB points to sample."
+                     " This parameter has no effect if there are no input VDB point grids")
+        .setChoiceList(&hvdb::VDBPointsGroupMenuInput1));
+
+    // rename "vel" attribute to "v"
+    parms.add(hutil::ParmFactory(PRM_TOGGLE, "renamevel", "Rename Vel To V")
```
danrbailey: It would probably be a fairly minor change to support this for Houdini points as well? I think for consistency that would be good, because the comment about it being VDB Points only in the help text could be easily missed, causing some confusion. Admittedly that could break existing scenes, so perhaps making this default to off would be preferred to having functionality apply just to VDB points.

openvdb_houdini/houdini/SOP_OpenVDB_Sample_Points.cc
```diff
@@ -265,6 +302,24 @@ newSopOperator(OP_OperatorTable* table)
...
+        .setDefault(PRMoneDefaults)
+        .setHelpText("When toggled on, if the input VDB's name is vel, "
+            "rename the VDB points attribute to v"));
+
+    // warn if attempting to overwrite existing attribute
+    parms.add(hutil::ParmFactory(PRM_TOGGLE, "attributeexists", "Attribute Exists Warning")
+        .setDefault(PRMoneDefaults)
```
danrbailey: In my opinion, having this default to off would be preferable. To sample into an attribute with a compression codec, you would have to add it before sampling, so that workflow would always create a warning on this node by default. Again, this one would be good if it's possible to support for Houdini points in a trivial way (if it doesn't already).

openvdb_houdini/houdini/SOP_OpenVDB_Sample_Points.cc
```diff
@@ -424,24 +556,35 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Sample_Points)::cookVDBSop(OP_
                         grid, threaded, aGdp, attribHandle, &scalarInterrupt);
                     theSampler.sample();
 
+                    // loop over vdb point grids
+                    sampleToVdbPointsGrids<cvdb::FloatGrid>(vdb, attributeName,
```
danrbailey: It might be worth supporting threaded and interrupt parameters to be consistent with the Houdini point sampling? (As an aside, much of the existing point functionality is currently missing the threaded option, which I think is probably a worthwhile addition).

openvdb_houdini/houdini/SOP_OpenVDB_Sample_Points.cc
```diff
@@ -504,20 +662,24 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Sample_Points)::cookVDBSop(OP_
                             << grid.valueType() << std::endl;
                     }
 
-                    UT_AutoInterrupt vectorInterrupt{
-                        "Sampling from VDB vector-type staggered grids"};
-
+                    UT_AutoInterrupt vectorInterrupt("Sampling from VDB vector-type staggered grids");
```
danrbailey: Should this have changed?

openvdb_houdini/houdini/SOP_OpenVDB_Sample_Points.cc
```diff
@@ -346,20 +402,86 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Sample_Points)::cookVDBSop(OP_
...
+                        leaf->attributeSet());
+
+                    nVDBPoints +=
+                        cvdb::points::filterPointCount<cvdb::points::PointDataTree,
+                            cvdb::points::MultiGroupFilter>(pointDataGrid->tree(), filter);
+                }
+            }
+
+            pointGrids.push_back(pointDataGrid);
+        }
```
danrbailey: For the sake of simplifying the code, it might be cleaner to just change the return value of points::sampleGrid() to return the number of points sampled so that this logic could be removed entirely. That could have the added benefit of extracting the point count changes into a new PR to be done independently.

openvdb/points/PointSample.h
```diff
@@ -0,0 +1,359 @@
...
+///                          attribute will be created with the source grid type if it does
+///                          not exist
+///  @param includeGroups    names of groups to include
+///  @param excludeGroups    names of groups to exclude
+template <typename SourceGridT, typename PointDataGridT, size_t Order = 1>
+inline void sampleGrid(PointDataGridT& points,
+                       const SourceGridT& sourceGrid,
+                       const Name& targetAttribute,
+                       const std::vector<Name>& includeGroups = std::vector<Name>(),
+                       const std::vector<Name>& excludeGroups = std::vector<Name>());
```
danrbailey: Although there are some exceptions, in most of the public API, the template arguments for functions are automatically deduced, so having an interface where you're asking the user to change a template argument is perhaps not ideal, even if that argument has a sensible default value.

One solution could be doing what tools/Interpolation.h does and introduce convenience functionality that wrap this - pointSample(PointDataGridT&, ...), boxSample(PointDataGridT&, ...), etc. I appreciate it's a bit more code to handle though.

openvdb/points/PointSample.h
```diff
@@ -0,0 +1,359 @@
...
+///  @param sourceGrid       VDB grid which will be sampled
+///  @param targetAttribute  a target attribute on the points which will hold samples. This
+///                          attribute will be created with the source grid type if it does
+///                          not exist
+///  @param includeGroups    names of groups to include
+///  @param excludeGroups    names of groups to exclude
+template <typename SourceGridT, typename PointDataGridT, size_t Order = 1>
+inline void sampleGrid(PointDataGridT& points,
+                       const SourceGridT& sourceGrid,
+                       const Name& targetAttribute,
```
danrbailey: Could targetAttribute take a default argument of an empty string? That way, if the name was empty, the attribute name would simply be the name of the source grid?

PR #232, opened 2018-04-24
VDB Points Advection

danrbailey:
Ability to advect VDB Points through a velocity field and a point move API that relies on defining a custom deformer. In addition, some minor changes to AttributeHandle and the addition of methods to IndexFilters that can indicate whether all or none of the points are evaluated. Support for VDB Points advection has been added to OpenVDB Advect Points SOP.

### DIFF
--- a/openvdb/doc/examplecode.txt
+++ b/openvdb/doc/examplecode.txt
@@ -45,7 +45,9 @@ illustrate how to use OpenVDB and how to perform common tasks.
   - @ref sPointCustomFiltering
 - @ref sPointStride
   - @ref sConstantStride
-
+- @ref sPointMove
+  - @ref sPointAdvect
+  - @ref sPointCustomDeformer
 
 @section sHelloWorld &ldquo;Hello, World&rdquo; for OpenVDB
 This is a very simple example showing how to create a grid and access
@@ -1551,8 +1553,6 @@ Leaf[0, 0, 0]
 This example demonstrates how to create a new point grid and to populate it with
 random point positions initialized inside a level set sphere.
 
-Note that multi-threaded scattering tools are currently in development for
-introduction to the library to handle this use-case.
 @code
 #include <iostream>
 #include <openvdb/openvdb.h>
@@ -2058,6 +2058,59 @@ for (auto leafIter = points->tree().beginLeaf(); leafIter; ++leafIter) {
         }
     }
 }
+@endcode
+
+@section sPointMove Moving Points in Space
+
+As points are stored within voxels in an implicit spatially organised data structure, moving points in space requires re-sorting the data.
+
+@subsection sPointAdvect Advecting Points
+
+Advection uses a specified integration order (4 = runge-kutta 4th) as well as delta time and time-step parameters
+to advect the points in-place using the supplied velocity grid.
+
+@code
+// Create an empty velocity grid with gravity as background value
+auto gravity = openvdb::Vec3SGrid::create(openvdb::Vec3s(0, -9.81, 0));
+
+// Advect points in-place using gravity velocity grid
+openvdb::points::advectPoints(*points, *gravity,
+    /*integrationOrder=*/4, /*dt=*/1.0/24.0, /*timeSteps=*/1);
+@endcode
+
+@subsection sPointCustomDeformer Moving Points with a Custom Deformer
+
+A custom deformer generates the new position of each existing point in a point set. This can use any number of
+mechanisms to achieve this such as a static value, a hard-coded list of positions, a function that uses the
+existing position to compute the new one or a function that uses the index of the point within the leaf array in
+some other way. This example simply takes the input position and adds a Y offset. Note that it is also possible
+to configure a custom deformer to operate in index-space.
+
+@code
+// This custom deformer is also used in the TestPointMove unit tests.
+struct OffsetDeformer
+{
+    OffsetDeformer(const openvdb::Vec3d& _offset)
+        : offset(_offset){ }
+
+    template <typename LeafIterT>
+    void reset(const LeafIterT&) { }
+
+    template <typename IndexIterT>
+    void apply(openvdb::Vec3d& position, const IndexIterT&) const
+    {
+        position += offset;
+    }
+
+    openvdb::Vec3d offset;
+};
+
+// Create an OffsetDeformer that moves the points downwards in Y by 10 world-space units.
+openvdb::Vec3d offset(0, -10, 0);
+OffsetDeformer deformer(offset);
+
+// Move the points using this deformer
+openvdb::points::movePoints(*points, deformer);
 @endcode
 
 */
--- a/openvdb/doc/examplecode.txt
+++ b/openvdb/doc/examplecode.txt
@@ -2062,7 +2062,7 @@ for (auto leafIter = points->tree().beginLeaf(); leafIter; ++leafIter) {
 
 @section sPointMove Moving Points in Space
 
-As points are stored within voxels in an implicit spatially organised data structure, moving points in space requires re-sorting the data.
+As points are stored within voxels in an implicit spatially organised data structure, moving points in space requires re-bucketing the data.
 
 @subsection sPointAdvect Advecting Points
 
--- a/openvdb/points/PointAdvect.h
+++ b/openvdb/points/PointAdvect.h
@@ -1,0 +1,276 @@
+///////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2012-2018 DreamWorks Animation LLC
+//
+// All rights reserved. This software is distributed under the
+// Mozilla Public License 2.0 ( http://www.mozilla.org/MPL/2.0/ )
+//
+// Redistributions of source code must retain the above copyright
+// and license notice and the following restrictions and disclaimer.
+//
+// *     Neither the name of DreamWorks Animation nor the names of
+// its contributors may be used to endorse or promote products derived
+// from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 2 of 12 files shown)


openvdb/points/PointMove.h
```diff
@@ -0,0 +1,1009 @@
...
+            for (int i = 1; i < leaf->buffer().size(); i++) {
+                buffer[i] = buffer[i-1] + buffer[i];
+            }
+        }
+    };
+    if (threaded) {
+        tbb::parallel_for(targetLeafManager.leafRange(), histogramLambda);
+    } else {
+        histogramLambda(targetLeafManager.leafRange());
+    }
```
richhones: should probably use the leaf manager foreach method here to avoid branching

danrbailey: fair comment, I'll improve that

openvdb/doc/examplecode.txt
```diff
@@ -2060,4 +2060,57 @@ for (auto leafIter = points->tree().beginLeaf(); leafIter; ++leafIter) {
 }
 @endcode
 
+@section sPointMove Moving Points in Space
+
+As points are stored within voxels in an implicit spatially organised data structure, moving points in space requires re-sorting the data.
```
fgochez: (minor) as suggested by @toddkeeler , the term "re-bucketing" might be more appropriate than "re-sorting".

openvdb/points/PointAdvect.h
```diff
@@ -0,0 +1,288 @@
...
+    {
+        mFilter.reset(*leaf);
+    }
+
+    template <typename IndexIterT>
+    void apply(Vec3d& position, const IndexIterT& iter) const
+    {
+        if (mFilter.valid(iter)) {
+            for (int n = 0; n < mSteps; ++n) {
+                mIntegrator.template rungeKutta<IntegrationOrder, openvdb::Vec3d>(mTimeStep, position);
```
dnegmw: Looks like mTimeStep never gets divided by mSteps prior to this call so is it currently stepping too far if there are multiple advection substeps?

danrbailey: @dnegmw I agree this feels a little counter-intuitive, but in fact the timestep parameter here is already per-step. Here's where the division you're referring to happens in the SOP:

https://github.com/dreamworksanimation/openvdb/blob/master/openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc#L1021

If you pass in Houdini points and VDB points while varying these parameters, you should see identical behaviour.

openvdb/doc/examplecode.txt
danrbailey: @fgochez Yes, re-bucketing is probably more correct, but I think I might just add both terms here for additional clarity. :)

Idclip:
@danrbailey is this the latest version of this work?

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,952 @@
...
+            // transform to world-space position and apply world-space deformation (if applicable)
+
+            Vec3d positionWS = mSourceTransform.indexToWorld(positionIS);
+            if (!useIndexSpace) {
+                deformer.apply(positionWS, iter);
+            }
+
+            // transform to index-space position of target grid
+
+            positionIS = mTargetTransform.worldToIndex(positionWS);
```
richhones: cant you put the above lines into the !indexSpace scope? i.e. 
```
if(!useIndexSpace) {
    Vec3d positionWS = mSourceTerm.indexToWorld(positionIS);
    deformer.apply(positionWS, iter);
    positionIS = mTargetTransform.worldToIndex(positionWS);
}
```

danrbailey: @richhones - you still need to apply the world->index->world transformation if the source and target transforms don't match, so as to put these positions into the index-space of the target. Note that there are other optimisations that could be introduced here, ignoring this world->index->world transformation if the transforms are the same as well as collapsing these two transformations down into a single transform or matrix if they're not. Neither of which I introduced because the world-space deformer is by far the most common, the index-space deformation is currently just an option and is only used in the unit tests. If that were to change, this can be optimised later without affecting ABI or the public API.

openvdb/points/PointMask.h
```diff
@@ -80,14 +83,51 @@ inline typename std::enable_if<std::is_same<typename MaskT::ValueType, bool>::va
...
+    void apply(Vec3d&, IterT&) const { }
+};
+
+/// @brief Deformer Traits for optionally configuring deformers to be applied
+/// in index-space. The default is world-space.
+template <typename DeformerT>
+struct DeformerTraits
+{
+    static const bool IndexSpace = false;
+};
```
richhones: Only bugbear with this is that you have to inject new specialisations into the openvdb::points namespace (supposing you are creating a new Deformer elsewhere),  whereas if this were a required member on the Deformer interface it wouldn't matter. 
For clarity suppose I am creating a new deformer FooDeformer: 
```
namespace foo {
struct FooDeformer(){
    template<LeafT>
    void reset(LeafT&, size_t) {}
    template<IterT>
    void apply(Vec3d&, IterT&) {}
}
} // namespace foo
```
then I also need to somewhere have 
```
namespace openvdb {
namespace points {
struct DeformerTraits <foo::FooDeformer>
{
    static const bool IndexSpace = true;
}
```
I'm not too well versed in whether this is a common approach and it probably doesn't matter that much but just pointing it out. Don't know whether there would be any better way to avoid this without requiring some inheritance or adding another required member into the Deformer?

danrbailey: I agree it's not ideal. I thought quite carefully about this and my logic is as follows. World-space deformation is going to be by far the most common and should not require the user to have to define that they want a deformer to be world-space, it should be the default behaviour. To achieve maximum performance, this needs to be compile-time and cannot use virtual functions, so I didn't see any other way of exposing this that would allow an optional (but very rare) override to be able to make a deformer evaluate in index-space. Ultimately I kind of see this as a hook that I don't expect anyone to actually use (aside from a few unit tests), but it's there just in case someone wants to use it and means they don't have to patch the library in order to do this, which is far worse than injecting new specialisations into the openvdb::points namespace.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+/// @param points           the PointDataGrid containing the points to be moved.
+/// @param deformer         a custom deformer that defines how to move the points.
+/// @param filter           an optional index filter
+/// @param objectNotInUse   for future use, this object is currently ignored
+/// @param threaded         enable or disable threading  (threading is enabled by default)
+template <typename PointDataGridT, typename DeformerT, typename FilterT = NullFilter>
+inline void movePoints(PointDataGridT& points,
+                       DeformerT& deformer,
+                       const FilterT& filter = NullFilter(),
+                       future::Advect* objectNotInUse = nullptr,
```
richhones: can you give some insight into why this argument is here?

danrbailey: @richhones It is a future hook to allow injecting points from other VDB points grids into an existing one while moving the points. This allows for a single re-bucketing operation instead of being required to perform re-bucketing in two separate steps for moving and merging. A convenience API will also be offered to perform merging without requiring a deformer. The main reason this is a generic object pointer and not simply a vector of PointDataGrid pointers is to be able to optionally filter these secondary point grids while merging. Clearly, the future hook eliminates the need to make a breaking change to the API and deprecate and support the old function for a time, however as merging is a new and quite distinct behaviour, I didn't want it to hold up the introduction of point moving.

Idclip:
@danrbailey I had written a bunch of comments here which I annoyingly forgot to submit. Just realised this upon merge notification!

dokipen3d:
great that this is merged. looking forward to using it.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+///
+/// @brief Ability to move VDB Points using a custom deformer.
+///
+/// Deformers used when moving points are in world space by default and must adhere
+/// to the interface described in the example below:
+/// @code
+/// struct MyDeformer
+/// {
+///     // A reset is performed on each leaf in turn before the points in that leaf are
+///     // deformed. A leaf and depth-first index within the tree are supplied as the
```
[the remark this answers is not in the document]
danrbailey: Yes, that's fair, I can improve the language a little here.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+    // early exit if no LeafNodes
+
+    PointDataTree::LeafCIter iter = tree.cbeginLeaf();
+
+    if (!iter)      return;
+
+    // build voxel topology taking into account any point group deletion
+
+    auto newPoints = point_mask_internal::convertPointsToScalar<PointDataGrid>(
+        points, transform, filter, deformer, threaded);
```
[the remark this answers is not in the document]
danrbailey: While it's true that the pdg being generated isn't valid (yet), there's nothing in the API to prevent partial creation of a pdg and I feel it should be encouraged in cases like there where there's some efficiency to be gained in doing so. After all, this is an "internal" method calling another "internal" method in a way that isn't explicitly documented as part of the public API.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+            assert(mTargetLeafMap.find(targetLeafOrigin) != mTargetLeafMap.end());
+            const Index targetLeafOffset(mTargetLeafMap.at(targetLeafOrigin));
+
+            // insert into move map based on whether point ends up in a new leaf node or not
+
+            if (targetLeafOrigin == sourceLeafOrigin) {
+                mLocalMoveLeafMap[targetLeafOffset].emplace_back(targetOffset, *iter);
+            }
+            else {
+                mGlobalMoveLeafMap[targetLeafOffset].push_back(IndexTriple(idx,
```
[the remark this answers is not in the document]
danrbailey: I'll have to check on the exact details, but I believe TBB's support for emplace_back in concurrent vectors was added relatively recently, so this was left like this to be fully compatible with the VFX Reference Platform rather than branching based on TBB version.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+            // set new local position in source transform space (if point has been deformed)
+
+            Vec3d voxelPosition(positionIS - targetVoxel.asVec3d());
+            sourceHandle->set(*iter, voxelPosition);
+
+            // determine target leaf node origin and offset in the target leaf vector
+
+            Coord targetLeafOrigin = targetVoxel & ~(LeafT::DIM - 1);
+            assert(mTargetLeafMap.find(targetLeafOrigin) != mTargetLeafMap.end());
+            const Index targetLeafOffset(mTargetLeafMap.at(targetLeafOrigin));
```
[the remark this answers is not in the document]
danrbailey: You're suggesting this so as to be able to use switch the std::map into a std::unordered_map? If so that's a modest but nonetheless worthwhile improvement. You do however still need to compute the origin for target <-> source leaf comparison.

openvdb/points/PointMove.h
```diff
@@ -0,0 +1,997 @@
...
+        CacheHandleOp op(mHandles);
+
+        for (auto leaf = range.begin(); leaf; ++leaf) {
+            auto& array = leaf->attributeArray(attributeIndex);
+            processTypedArray(array, op);
+        }
+    }
+
+private:
+    HandleArray mHandles;
```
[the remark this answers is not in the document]
danrbailey: Absolutely, however it will require an ABI change to be able do this properly. See the comment discussing this in detail here - https://github.com/dreamworksanimation/openvdb/blob/master/openvdb/points/PointMove.h#L197

I intend to address this deficiency in the AttributeArray API for ABI=6 at which stage lots of this framework can be deprecated.

openvdb/points/PointMask.h
```diff
@@ -326,10 +389,15 @@ inline typename std::enable_if<std::is_same<typename MaskT::ValueType, bool>::va
...
     const PointDataGridT& points,
-    const FilterT& filter)
+    const FilterT& filter,
+    bool threaded)
 {
-    return point_mask_internal::convertPointsToScalar<PointDataGridT, MaskT>(
-        points, filter);
+    // This is safe because the PointDataGrid can only be modified by the deformer
+    using AdapterT = TreeAdapter<typename PointDataGridT::TreeType>;
+    auto& nonConstPoints = const_cast<typename AdapterT::NonConstGridType&>(points);
```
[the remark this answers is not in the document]
danrbailey: I think you're correct, I'll verify this and change it if so.

PR #233, opened 2018-05-03
Improved Point Filtering and Count Optimizations

danrbailey:
* Adding extra state() methods to index filter interface to improve optimization opportunities
* Adding active and inactive value mask index filters
* Switching include/exclude groups API to instead use explicit filtering
* Refactoring and simplifying of point count API
* Computing cumulative per-leaf point offsets is now parallelized

### DIFF
--- a/openvdb/points/IndexFilter.h
+++ b/openvdb/points/IndexFilter.h
@@ -126,6 +126,7 @@ generateRandomSubset(const unsigned int seed, const IntType n, const IntType m)
 } // namespace index_filter_internal
 
 
+/// Index filtering on active / inactive state of host voxel
 template <bool On>
 class ValueMaskFilter
 {
@@ -146,7 +147,7 @@ public:
     template <typename IterT>
     bool valid(const IterT& iter) const
     {
-        const bool valueOn = iter.valueIter().isValueOn();
+        const bool valueOn = iter.isValueOn();
         return On ? valueOn : !valueOn;
     }
 };
--- a/openvdb/points/IndexFilter.h
+++ b/openvdb/points/IndexFilter.h
@@ -33,6 +33,33 @@
 /// @author Dan Bailey
 ///
 /// @brief  Index filters primarily designed to be used with a FilterIndexIter.
+///
+/// Filters must adhere to the interface described in the example below:
+/// @code
+/// struct MyFilter
+/// {
+///     // Return true when the filter has been initialized for first use
+///     bool initialized() { return true; }
+///
+///     // Return index::ALL if all points are valid, index::NONE if no points are valid
+///     // and index::PARTIAL if some points are valid
+///     index::State state() { return index::PARTIAL; }
+///
+///     // Return index::ALL if all points in this leaf are valid, index::NONE if no points
+///     // in this leaf are valid and index::PARTIAL if some points in this leaf are valid
+///     template <typename LeafT>
+///     index::State state(const LeafT&) { return index::PARTIAL; }
+///
+///     // Resets the filter to refer to the specified leaf, all subsequent valid() calls
+///     // will be relative to this leaf until reset() is called with a different leaf.
+///     // Although a required method, many filters will provide an empty implementation if
+///     // there is no leaf-specific logic needed.
+///     template <typename LeafT> void reset(const LeafT&) { }
+///
+///     // Returns true if the filter is valid for the supplied iterator
+///     template <typename IterT> bool valid(const IterT&) { return true; }
+/// };
+/// @endcode
 
 #ifndef OPENVDB_POINTS_INDEX_FILTER_HAS_BEEN_INCLUDED
 #define OPENVDB_POINTS_INDEX_FILTER_HAS_BEEN_INCLUDED
@@ -63,6 +90,7 @@ namespace points {
 
 
 ////////////////////////////////////////
+
 
 
 namespace index_filter_internal {
@@ -96,6 +124,36 @@ generateRandomSubset(const unsigned int seed, const IntType n, const IntType m)
 
 
 } // namespace index_filter_internal
+
+
+template <bool On>
+class ValueMaskFilter
+{
+public:
+    static bool initialized() { return true; }
+    static index::State state() { return index::PARTIAL; }
+    template <typename LeafT>
+    static index::State state(const LeafT& leaf)
+    {
+        if (leaf.isDense())         return On ? index::ALL : index::NONE;
+        else if (leaf.isEmpty())    return On ? index::NONE : index::ALL;
+        return index::PARTIAL;
+    }
+
+    template <typename LeafT>
+    void reset(const LeafT& leaf) { }
+
+    template <typename IterT>
+    bool valid(const IterT& iter) const
+    {
+        const bool valueOn = iter.valueIter().isValueOn();
+        return On ? valueOn : !valueOn;
+    }
+};
+
+
+using ActiveFilter = ValueMaskFilter<true>;
+using InActiveFilter = ValueMaskFilter<false>;
 
 
 /// Index filtering on multiple group membership for inclusion and exclusion
@@ -142,15 +200,23 @@ public:
 
     inline bool initialized() const { return mInitialized; }
 
+    inline index::State state() const
+    {
+        return (mInclude.empty() && mExclude.empty()) ? index::ALL : index::PARTIAL;
+    }
+
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
+
     template <typename LeafT>
     void reset(const LeafT& leaf) {
         mIncludeHandles.clear();
         mExcludeHandles.clear();
-        for (const AttributeSet::Descriptor::GroupIndex& index : mInclude) {
-            mIncludeHandles.emplace_back(leaf.groupHandle(index));
+        for (const auto& i : mInclude) {
+            mIncludeHandles.emplace_back(leaf.groupHandle(i));
         }
-        for (const AttributeSet::Descriptor::GroupIndex& index : mExclude) {
-            mExcludeHandles.emplace_back(leaf.groupHandle(index));
+        for (const auto& i : mExclude) {
+            mExcludeHandles.emplace_back(leaf.groupHandle(i));
         }
         mInitialized = true;
     }
@@ -226,6 +292,10 @@ public:
 
     inline bool initialized() const { return mNextIndex == -1; }
 
+    static index::State state() { return index::PARTIAL; }
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
+
     template <typename LeafT>
     void reset(const LeafT& leaf) {
         using index_filter_internal::generateRandomSubset;
@@ -297,6 +367,10 @@ public:
 
     inline bool initialized() const { return bool(mIdHandle); }
 
+    static index::State state() { return index::PARTIAL; }
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
+
     template <typename LeafT>
     void reset(const LeafT& leaf) {
         assert(leaf.hasAttribute(mIndex));
@@ -349,6 +423,10 @@ public:
     }
 
     inline bool initialized() const { return bool(mPositionHandle); }
+
+    static index::State state() { return index::PARTIAL; }
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
 
     template <typename LeafT>
     void reset(const LeafT& leaf) {
@@ -410,6 +488,13 @@ public:
 
     inline bool initialized() const { return bool(mPositionHandle); }
 
+    inline index::State state() const
+    {
+        return mBbox.empty() ? index::NONE : index::PARTIAL;
+    }
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
+
     template <typename LeafT>
     void reset(const LeafT& leaf) {
         mPositionHandle.reset(new Handle(leaf.constAttributeArray("P")));
@@ -450,6 +535,16 @@ public:
 
     inline bool initialized() const { return mFilter1.initialized() && mFilter2.initialized(); }
 
+    inline index::State state() const
+    {
+        return this->computeState(mFilter1.state(), mFilter2.state());
+    }
+    template <typename LeafT>
+    inline index::State state(const LeafT& leaf) const
+    {
+        return this->computeState(mFilter1.state(leaf), mFilter2.state(leaf));
+    }
+
     template <typename LeafT>
     void reset(const LeafT& leaf) {
         mFilter1.reset(leaf);
@@ -463,6 +558,19 @@ public:
     }
 
 private:
+    inline index::State computeState(   index::State state1,
+                                        index::State state2) const
+    {
+        if (And) {
+            if (state1 == index::NONE || state2 == index::NONE)       return index::NONE;
+            else if (state1 == index::ALL && state2 == index::ALL)    return index::ALL;
+        } else {
+            if (state1 == index::NONE && state2 == index::NONE)       return index::NONE;
+            else if (state1 == index::ALL && state2 == index::ALL)    return index::ALL;
+        }
+        return index::PARTIAL;
+    }
+
     T1 mFilter1;
     T2 mFilter2;
 }; // class BinaryFilter
--- a/openvdb/points/IndexFilter.h
+++ b/openvdb/points/IndexFilter.h
@@ -154,7 +154,7 @@ public:
 
 
 using ActiveFilter = ValueMaskFilter<true>;
-using InActiveFilter = ValueMaskFilter<false>;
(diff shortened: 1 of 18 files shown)


openvdb/points/IndexFilter.h
```diff
@@ -98,6 +126,36 @@ generateRandomSubset(const unsigned int seed, const IntType n, const IntType m)
...
+{
+public:
+    static bool initialized() { return true; }
+    static index::State state() { return index::PARTIAL; }
+    template <typename LeafT>
+    static index::State state(const LeafT& leaf)
+    {
+        if (leaf.isDense())         return On ? index::ALL : index::NONE;
+        else if (leaf.isEmpty())    return On ? index::NONE : index::ALL;
+        return index::PARTIAL;
```
Idclip: Shouldn't this be returning index::PARTIAL if On and at least one voxel is active and vice-versa? i.e:

if (On && leaf.isDense()) return index::ALL;
else if (On && leaf.onVoxelCount()) return index::PARTIAL;
else if (!On && leaf.isEmpty()) return index::ALL;
else if (!On && leaf.offVoxelCount()) return index::PARTIAL;
return index::PARTIAL;

openvdb/points/PointConversion.h
```diff
@@ -490,27 +501,15 @@ struct ConvertPointDataGridPositionOp {
...
             else {
-                auto iter = leaf->beginIndexOn();
-
-                for (; iter; ++iter) {
-                    const Vec3d xyz = iter.getCoord().asVec3d();
-                    const Vec3d pos = handle->get(*iter);
-                    pHandle.set(static_cast<Index>(offset++), /*stride=*/0,
-                        mTransform.indexToWorld(pos + xyz));
-                }
+                auto iter = leaf->beginIndexOn(newFilter);
```
Idclip: beginIndexOn will again copy the filter... tbh i think we should remove this internal copying of the filter in beginIndex() but if not, move the copying into the required branches

openvdb/points/IndexFilter.h
Idclip: Also if this is the case, we need unit tests that catch this as currently everything passes

openvdb/points/PointConversion.h
Idclip: This needs to be updated throughout this file

openvdb/points/PointConversion.h
```diff
@@ -117,29 +119,30 @@ createPointDataGrid(const std::vector<ValueT>& positions, const math::Transform&
...
-template <typename PositionAttribute, typename PointDataGridT>
+template <typename PositionAttribute, typename PointDataGridT, typename FilterT = NullFilter>
 inline void
 convertPointDataGridPosition(   PositionAttribute& positionAttribute,
                                 const PointDataGridT& grid,
                                 const std::vector<Index64>& pointOffsets,
                                 const Index64 startOffset,
-                                const std::vector<Name>& includeGroups = std::vector<Name>(),
-                                const std::vector<Name>& excludeGroups = std::vector<Name>(),
+                                const FilterT& filter = NullFilter(),
```
Idclip: If you're updating this signatures are you planning to add a threaded argument?

openvdb/points/PointCount.h
```diff
@@ -55,328 +54,287 @@ namespace OPENVDB_VERSION_NAME {
...
+    leafManager.foreach(
+        [&pointOffsets, &filter, &inCoreOnly](const LeafT& leaf, size_t pos) {
+            FilterT newFilter(filter);
+            if (inCoreOnly && leaf.buffer().isOutOfCore())  return;
+            auto state = filter.state(leaf);
+            if (state == index::ALL) {
+                pointOffsets[pos] = leaf.pointCount();
+            } else if (state != index::NONE) {
+                newFilter.reset(leaf);
+                pointOffsets[pos] = iterCount(leaf.beginIndexAll(newFilter));
```
Idclip: double copy

openvdb/points/PointDataGrid.h
```diff
@@ -999,13 +999,20 @@ template<typename ValueIterT, typename FilterT>
 inline IndexIter<ValueIterT, FilterT>
 PointDataLeafNode<T, Log2Dim>::beginIndex(const FilterT& filter) const
 {
+    FilterT newFilter(filter);
+    newFilter.reset(*this);
+
+    // generate no-op iterator if filter evaluates no indices
+
+    if (filter.state() == index::NONE) {
```
Idclip: Can move this up before the filter copy, though this still dosen't solve the double copy problem

openvdb/points/PointDataGrid.h
```diff
@@ -1092,7 +1099,7 @@ template<typename T, Index Log2Dim>
 inline Index64
 PointDataLeafNode<T, Log2Dim>::pointCount() const
 {
-    return iterCount(this->beginIndexAll());
+    return this->getLastValue();
```
Idclip: :thumbsup:

openvdb/points/PointMask.h
```diff
@@ -56,33 +56,31 @@ namespace points {
...
 template <typename PointDataGridT,
-          typename MaskT = typename PointDataGridT::template ValueConverter<bool>::Type>
+          typename MaskT = typename PointDataGridT::template ValueConverter<bool>::Type,
+          typename FilterT = NullFilter>
 inline typename std::enable_if<std::is_same<typename MaskT::ValueType, bool>::value,
     typename MaskT::Ptr>::type
 convertPointsToMask(const PointDataGridT& grid,
-                    const std::vector<Name>& includeGroups = std::vector<Name>(),
-                    const std::vector<Name>& excludeGroups = std::vector<Name>());
+                    const FilterT& filter = NullFilter());
```
Idclip: threaded arguments?

openvdb/points/IndexFilter.h
```diff
@@ -142,15 +200,23 @@ class MultiGroupFilter
 
     inline bool initialized() const { return mInitialized; }
 
+    inline index::State state() const
+    {
+        return (mInclude.empty() && mExclude.empty()) ? index::ALL : index::PARTIAL;
```
Idclip: I think it's weird that this filter accepts no include groups as valid. imo it should be:

mInclude.empty() && mExclude.empty() = None
!mInclude.empty() && !mExclude.empty() = Partial
!mInclude.empty() && mExclude.empty() = Partial
mInclude.empty() && !mExclude.empty() = Partial

If the descriptor is known then you can optimise even further. This is a potentially destructive change, though I think it makes far more sense. Here is some custom logic I'm using to improve this behaviour. Assuming that a group does not exist in both the include and exclude groups, then:

```
        // The following assumes that a group does not exist inside both
        // the includeGroups and excludeGroups containers

        const bool hasExcludes(!excludeGroups.empty());
        for (auto iter = excludeGroups.begin(); iter != excludeGroups.end();) {
            if (!descriptor->hasGroup(*iter)) {
               std::swap(*iter, excludeGroups.back());
               excludeGroups.pop_back();
            }
            else {
                ++iter;
            }
        }

        // if exclude groups were specified but are now empty,
        // regardless of which include groups were specified we
        // process all points

        if (hasExcludes && excludeGroups.empty()) {
            includeGroups.clear();
            return ALL;
        }

        const bool hasIncludes(!includeGroups.empty());
        for (auto iter = includeGroups.begin(); iter != includeGroups.end();) {
            if (!descriptor->hasGroup(*iter)) {
               std::swap(*iter, includeGroups.back());
               includeGroups.pop_back();
            }
            else {
                ++iter;
            }
        }

        // if there were specified include groups but have all been removed
        // as they don't exist, check if any exclude groups exist. If not then
        // we shouldn't perform any execution

        if (hasIncludes && includeGroups.empty()) {
            if (!excludeGroups.empty()) return EXCLUDE;
            else return INVALID;
        }
```

openvdb/points/IndexFilter.h
```diff
@@ -142,15 +200,23 @@ class MultiGroupFilter
...
+    }
+
+    template <typename LeafT>
+    static index::State state(const LeafT&) { return index::PARTIAL; }
```
Idclip: Why not use the leaf here to optimise further? See above comment

openvdb/points/PointMask.h
```diff
@@ -240,41 +237,38 @@ inline typename GridT::Ptr convertPointsToScalar(
...
+        NullFilter nullFilter;
         PointsToScalarOp<GridT, PointDataGridT, NullFilter> pointsToScalarOp(
-            points, filter);
+            points, nullFilter);
+        tbb::parallel_for(leafManager.leafRange(), pointsToScalarOp);
+    } else {
+        // build mask from points in parallel only where filter evaluates to true
+        FilterT newFilter(filter);
+        PointsToScalarOp<GridT, PointDataGridT, FilterT> pointsToScalarOp(
+            points, newFilter);
```
Idclip: threaded argument with foreach?

openvdb/points/IndexFilter.h
danrbailey: For a start, if you're evaluating *only* active voxels and you know all voxels are inactive, then the result should be index::NONE. Perhaps you meant the return statement you listed to be 'return index::NONE'? If so, I believe that's actually the same logic written differently.

I'll look at the unit tests and make to get this case covered (I should really have included a test case for this filter in TestIndexFilter).

openvdb/points/IndexFilter.h
danrbailey: An empty includeGroups vector is simply treated as including "*" for convenience.

As for your logic here, this looks great. I consider this a further optimisation though, index::PARTIAL does produce the correct result, it's just not as efficient as it could be, as you highlight. There are other potential optimisations I didn't include too - BBoxFilter for example. Provided the results are correct and it doesn't introduce a regression, this can be introduced later I feel, or of course I'm more than happy if you'd like to add this to my PR with all the relevant unit tests. :)

openvdb/points/IndexFilter.h
danrbailey: Again, happy to accept a contribution here. :)

openvdb/points/PointMask.h
danrbailey: These threaded arguments are actually added in the advection PR - https://github.com/dreamworksanimation/openvdb/pull/232. I was just trying to restrict this PR to addressing filtering and counting here.

openvdb/points/PointMask.h
danrbailey: Again this is part of the changes proposed in https://github.com/dreamworksanimation/openvdb/pull/232. If we can get this PR accepted first, I'll then make the relevant changes to update that one.

openvdb/points/PointConversion.h
danrbailey: You're right, I'll address this.

openvdb/points/IndexFilter.h
Idclip: This isn't just an optimisation, it would change how the Houdini SOPs parse groups + anything which tries to use this filter with groups (names) which do no exist. Currently unavailable groups are dropped by the namesToIndices() convenience method indirectly through the first MultiGroupFilter constructor. This means that groups which don't exist are not handled "correctly". With the above:

- If exclude groups have been provided and there are no valid exclude groups, all points are considered regardless of include groups
- If include groups have been provided and there are no valid include groups, no points are considered regardless of exclude groups
- If include groups have been provided and there are no valid include groups BUT there are valid exclude groups, only the exclude groups are considered. This is not inversely true
- Otherwise use both include and exclude groups

Currently:

- If exclude groups have been provided and there are no valid exclude groups, the include groups are used (if these are also invalid/are not provided, everything is parsed)
- If include groups have been provided and there are no valid include groups, the exclude groups are used (if these are also invalid/are not provided, everything is parsed)

i.e: 

"!a b" - if a doesn't exist, everything should be used. Currently, only b will be used
"a b"  - if a and b don't exist, nothing should be used. Currently, everything will be used. Note this is the same for 'if "a" doesn't exist'. You should be using a GroupFilter for this, but a lot of algorithms don't.

You can try this stuff out with the current implementation from the SOPs. I hope I've explained this correctly. I've mentioned this before, (I think ^ behaviour also wrong but that's Houdini specific so slightly irrelevant) but as the filters are to be used directly and a lot of this is changing, it would be worthwhile to consider changing this.

openvdb/points/IndexFilter.h
Idclip: Potentially something that should be done as part of the PR?

openvdb/points/IndexFilter.h
danrbailey: Hi @Idclip, thanks, your explanation makes perfect sense. The idea behind the state() method returning index::PARTIAL is that in this case, you cannot *guarantee* that all or no points will be evaluated. As such this maintains (status quo) correctness and performs *no slower* than the current mechanism. 

What you're talking about here is a worthwhile change, but I suggest could be tackled completely tangentially in a new PR (by modifying namesToIndices) to correct the behaviour and *then* introducing extra optimisations in the state() method once this extra optimisation framework in this PR has been merged in. Until then it just relies on returning index::PARTIAL which ultimately falls back to the implementation of the valid() index filter method. How does that sound?

OpenVDB-DevTeam:
Some minor issues:
- Some deprecated functions, such as `points::groupPointCount`, are still used in a couple of places, including in `PointUtils.cc` and `GR_PrimVDBPoints.cc`.
- `InActiveFilter` should probably be `InactiveFilter`.
- Doxygen doesn't handle complex return type expressions gracefully, and that renders the documentation for a file like `PointCount.h` nearly unusable.  Can you maybe use `static_assert` instead of `std::enable_if` in that file?

openvdb/points/PointMask.h
```diff
@@ -304,18 +295,13 @@ inline typename GridT::Ptr convertPointsToScalar(
...
-    if (useGroup) {
-        const auto leaf = points.constTree().cbeginLeaf();
-        MultiGroupFilter filter(includeGroups, excludeGroups, leaf->attributeSet());
-        PointsToTransformedScalarOp<GridT, PointDataGridT, MultiGroupFilter> pointsToScalarOp(
-            transform, pointsTransform, filter, combiner);
-        tbb::parallel_for(leafManager.leafRange(), pointsToScalarOp);
-    }
-    else {
-        NullFilter filter;
+    if (filter.state() == index::ALL) {
```
Idclip: Is this static state method really needed? It seems redundant as the only filter capable of returning ALL is a null filter anyway? I would have expected callers of function requiring a filter to perform some filtering and use a NullFilter otherwise (which should be the default argument in the majority of cases). Imo that method can be removed, simplifying the filter object and branching patterns within all methods.

openvdb/points/IndexFilter.h
danrbailey: I appreciate what you're saying and I think the most prominent point to consider here is regarding backwards-compatibility. If we were to introduce behaviour-changing logic to the MultiGroupFilter, the only way to effectively retain backwards compatibility would be to have two MultiGroupFilter implementations, keep the old one for the deprecated includeGroups/excludeGroups implementation and introduce a new one when used with the explicit MultiGroupFilter parameter method. Essentially, if we were to rigorously uphold the existing behaviour, we would also then be forced to branch the SOPs that use it so that a newly-placed SOP would get the new behaviour and an existing one keep the old behaviour. This is a lot of work and perhaps more an argument to keep the existing (incorrect) behaviour. 

Ultimately, for the relatively minor use cases (in my opinion) that your changes will tidy up, I'd still rather that we include behaviour-changing logic as a separate PR, because there's also an impact in delaying an imminent change to the API.

openvdb/points/PointMask.h
danrbailey: Actually, the MultiGroupFilter can also return ALL. One of the key ideas here is to eliminate some of the complexity for the caller - in adopting this the caller no longer needs to check their own include / exclude arguments and determine whether to branch their code to use a NullFilter or a MultiGroupFilter, they can simply call with a MultiGroupFilter and default empty include / exclude arguments and the underlying logic will branch accordingly to pick up the faster implementation.

danrbailey: I don't agree it would be clearer - from the user perspective, the API is simpler because they can use MultiGroupFilter and not worry about the filtering performance. You're right that it trades a little compile-time cost and a little extra IndexFilter complexity, but in my opinion I think that's worthwhile.

There is a filter.state() == index::NONE check, but this is in the creation of the iterator. Not as useful, but I included it for consistency. Also, regardless of implementation, there should always be an input to the MultiGroupFilter that evaluates all points, so whatever that input is should see this branching optimization (in reference to your last comment).

Idclip: They can use MultiGroupFilter and not worry about filtering performance: would this not also be true if the state function that took a leaf was implemented (perhaps a very minor penalty but likely negligible). I thought this is what this method was for and was my main point in my previous comment.

trades a little compile-time cost: didn't mention this, don't think this is a factor

I think this confuses things slightly and not sure I see its worth but oh well!

danrbailey:
@OpenVDB-DevTeam - it's ready to go, let's merge it. Thanks!

PR #250, opened 2018-07-26
Move license to LICENSE file

zxiiro:
Move to LICENSE file so that the Markdown file can be properly
formatted as '#' are header sections in Markdown formatting.


OpenVDB-DevTeam:
Thanks!

PR #251, opened 2018-07-27
Add common file system locations to search paths

zxiiro:
Rather than having the dev specifcy PACKAGE_ROOT variables for
every dependency set a sane set of default search paths where
most distros typically install to so that CMake can fallback on if
PACKAGE_ROOT is not explicitly specified.


cmake/FindBlosc.cmake
```diff
@@ -53,7 +65,7 @@ IF ( BLOSC_FOUND )
     CACHE STRING "Blosc library directories")
 
   SET ( _blosc_library_name "blosc" )
-  
+
```
danrbailey: I'd rather the trailing space changes were in a separate PR.

cmake/FindBlosc.cmake
```diff
@@ -35,12 +35,24 @@
 # BLOSC_LIBRARYDIR      Blosc's library directory
 # BLOSC_LIBRARIES        all Blosc libraries
 
+SET ( BLOSC_SEARCH_PATHS
```
danrbailey: I don't like all the duplication here, adding a new search path would require modifying all of these files, could this be moved somewhere where it can be shared?

cmake/FindBlosc.cmake
zxiiro: I'll see if I can break the whitespace changes out (my editor automatically removes them so it's not something I usually think about).

cmake/FindBlosc.cmake
zxiiro: Indeed, I'll see if I can figure out a better way to handle this.

zxiiro:
@danrbailey  Sorry I had not realized there was comments on this patch until now. I'll try to address them and find a way to add the switch even if it's just for CI builds.

Idclip:
This looks good to me. @danrbailey do you have any more comments? This will help massively with the travis cmake ci

danrbailey:
Thanks for this!

PR #259, opened 2018-08-10
openvdb/python/CMakeLists.txt: Minor fix and CMake v3.11 compatibility

dracwyrm:
Adding the python version string to finding the python module of
Boost fixes this error: CMake Error at /.../FindBoost.cmake:2044
More on the error here: https://bugs.gentoo.org/652266

Remove unneeded python version variables as PythonInterp already
defines them.

PYOPENVDB_INSTALL_DIRECTORY variable cannot be set during configure
phase because it is hard set. Test if it is not defined before
setting it to a default.

Signed-off by: Jonathan Scruggs <dracwyrm@gentoo.org>


dracwyrm:
How come all tests passed except this one?
https://travis-ci.org/dreamworksanimation/openvdb/jobs/414410409

OpenVDB-DevTeam:
> How come all tests passed except this one?

Just some random network timeout.  It reran successfully.

PR #262, opened 2018-09-07
GR_Prim VDB Points fixes and improvements

Idclip:


Idclip:
@danrbailey @jmlait tagging for reference

openvdb_houdini/houdini/GR_PrimVDBPoints.cc
```diff
@@ -1135,90 +840,85 @@ GR_PrimVDBPoints::inViewFrustum(const UT_Matrix4D& objviewproj)
...
+        if (useGroup) includeGroups.emplace_back(groupName);
 
         MultiGroupFilter filter(includeGroups, excludeGroups, iter->attributeSet());
 
         std::vector<Index64> offsets;
         pointOffsets(offsets, grid.tree(), filter, /*inCoreOnly=*/true);
 
         if (type == "vec3s") {
+            std::vector<Index64> pointOffsets;
+            getPointOffsets(pointOffsets, grid.tree(), includeGroups, excludeGroups, /*inCoreOnly=*/true);
```
OpenVDB-DevTeam: `getPointOffsets` is a deprecated function, but the vector shadows its replacement, and the vector doesn't seem to get used once it's been populated.  It's not obvious what the intent is here; can you clarify?

danrbailey: This is true, good catch. This looks like a possible merge error to me, @Idclip - could the below conversion method not use the offsets built in the call to pointOffsets() above?

Idclip: Yeah must have been a merge error, will fix

PR #267, opened 2018-10-24
Add methods to StringAttributeArray and StringAttributeWriteHandle

dnegmw:
Added missing array() method for StringAttributeHandle and StringAttributeWriteHandle. Also added convenience hasIndex() method to StringAttributeWriteHandle for checking if a string exists in the metadata cache already.

Updated StringAttributeWriteHandle unit tests to use hasIndex to verify it is working as intended.

### DIFF
--- a/openvdb/points/AttributeArrayString.h
+++ b/openvdb/points/AttributeArrayString.h
@@ -148,6 +148,8 @@ public:
     Name get(Index n, Index m = 0) const;
     void get(Name& name, Index n, Index m = 0) const;
 
+    const AttributeArray& array() const;
+
 protected:
     AttributeHandle<StringIndexType, StringCodec<false>>    mHandle;
     const MetaMap&                                          mMetadata;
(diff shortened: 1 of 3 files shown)


openvdb/points/AttributeArrayString.h
```diff
@@ -192,6 +194,10 @@ class OPENVDB_API StringAttributeWriteHandle : public StringAttributeHandle
     /// Reset the value cache from the metadata
     void resetCache();
 
+    AttributeArray& array();
+
+    bool hasIndex(const Name& name) const;
```
danrbailey: Perhaps this could follow a stdlib convention here and use the upcoming "contains()" method for checking the key:

```suggestion
    bool contains(const Name& name) const;
```

https://en.cppreference.com/w/cpp/container/unordered_map/contains

Also, these two methods need comment strings.

dnegmw: Apologies for the delay in responding to this @danrbailey. I've made the changes you've suggested now. If you have another moment to cast your eye over them it'd be much appreciated.

danrbailey: Looks good to me, thanks for making those changes Matt!

danrbailey:
Hi @dnegmw - "This branch is out-of-date with the base branch".

If you can update this branch to master, it will also trigger a build that should mean all 3/3 checks will go through successfully so this can be considered for merging. As the author, only you are able to do that unfortunately.

PR #276, opened 2018-11-01
Use the inputGeo() for the group generation and default Keep Input Off

jmlait:
When merging into H17, I found the building of the group from gdp rather than inputGeo(1) causes problems when not in keep input mode.   I also changed the default to not keep the input as I believe that matched the old tools (But it could be an old default on our part?)


danrbailey:
Previously the old tool didn't have a "keep original geometry" mode and always deleted everything else, so yes this is more correct to match the previous behaviour. 

It's still not ideal in my opinion, but then at least it's consistent with the behaviour in an Unpack SOP where all the non-packed geometries are deleted by default.

PR #277, opened 2018-11-01
Fix for Strict Type builds

jmlait:
In strict type builds 0 cannot be assigned to a GA_Offset.  Since the initialization was redundant, it was removed in this case.  If we prefer to always initialize, and explicit cast of (GA_Offset)0 may be used instead.


danrbailey:
Looks good to me.

pcucka:
How about "`auto start = GA_Offset(), end = GA_Offset();`"?

danrbailey:
@jmlait - I think this one can be rejected now that https://github.com/AcademySoftwareFoundation/openvdb/pull/303 has been merged?

PR #278, opened 2018-11-01
Advect Points will either advect VDB points or points, not both

jmlait:
The previous code attempted to advect both, but ended up re-using the same group field for two different meanings, causing various problems.   This moves the control parameter to the top to reflect its new importance.

### DIFF
--- a/openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
+++ b/openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
@@ -705,6 +705,18 @@ newSopOperator(OP_OperatorTable* table)
 
     hutil::ParmList parms;
 
+    // VDB Points advection
+    parms.add(hutil::ParmFactory(PRM_TOGGLE, "advectvdbpoints", "Advect VDB Points")
+        .setDefault(PRMzeroDefaults)
+        .setTooltip("Enable/disable advection of VDB Points.")
+        .setDocumentation(
+            "If enabled, advect the points in a VDB Points grid, otherwise apply advection"
+            " only to the Houdini point associated with the VDB primitive.\n\n"
+            "The latter is faster to compute but updates the VDB transform only"
+            " and not the relative positions of the points within the grid."
+            " It is useful primarily when instancing multiple static VDB point sets"
+            " onto a dynamically advected Houdini point set."));
+
     // Points to process
     parms.add(hutil::ParmFactory(PRM_STRING, "group", "Point Group")
         .setChoiceList(&hutil::PrimGroupMenuInput1)
@@ -822,17 +834,6 @@ newSopOperator(OP_OperatorTable* table)
             " It may also be useful for special effects (see also the"
             " [Trail SOP|Node:sop/trail])."));
 
-    // VDB Points advection
-    parms.add(hutil::ParmFactory(PRM_TOGGLE, "advectvdbpoints", "Advect VDB Points")
-        .setDefault(PRMoneDefaults)
-        .setTooltip("Enable/disable advection of VDB Points.")
-        .setDocumentation(
-            "If enabled, advect the points in a VDB Points grid, otherwise apply advection"
-            " only to the Houdini point associated with the VDB primitive.\n\n"
-            "The latter is faster to compute but updates the VDB transform only"
-            " and not the relative positions of the points within the grid."
-            " It is useful primarily when instancing multiple static VDB point sets"
-            " onto a dynamically advected Houdini point set."));
 
     // Obsolete parameters
     hutil::ParmList obsoleteParms;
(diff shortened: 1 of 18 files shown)


danrbailey:
> I'm not sure... The problem is any VDB Points primitive has some normal points, so a default blank point group would imply those need to be moved as well. I do agree that this isn't a nice answer. As an artist I don't want to think about it.

What I was going for is that by default (ie "advectvdbpoints" enabled), all VDB points and all Houdini points that **don't** have a VDB Points primitive attached should be advected. Even if you include the point that a VDB Points primitive is attached to in a point group, it would still be excluded from advection. Is that acceptable?

Regarding re-using the "group" field to mean "point groups" for Houdini points and "primitive groups" for VDB points, I would concede to dropping the "primitive groups" meaning over having to manually toggle between which input type you're expecting.

openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
```diff
@@ -1102,18 +1107,26 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Advect_Points)::evalAdvectionP
     UT_String ptGroupStr;
     evalString(ptGroupStr, "group", 0, now);
 
-    parms.mPointGroup = parsePointGroups(ptGroupStr, GroupCreator(gdp));
-    parms.mGroup = matchGroup(*parms.mPointGeo, evalStdString("group", now));
+    if (advectVdbPoints)
```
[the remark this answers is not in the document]
danrbailey: You're right that the assignment of mGroup and mPointGroup is correct, it's the rest of the logic that's flipped, I'll highlight the exact lines.

openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
```diff
@@ -1102,18 +1107,26 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Advect_Points)::evalAdvectionP
...
+    {
+	parms.mPointGroup = 0;
+	parms.mGroup = matchGroup(*parms.mPointGeo, ptGroupStr);
 
-    const std::string groups = evalStdString("vdbpointsgroups", now);
+	if (!parms.mPointGroup && ptGroupStr.length() > 0) {
```
danrbailey: mPointGroup is explicitly set to 0 above, so the first part of this conditional will always evaluate to true (!parms.mPointGroup)

openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
```diff
@@ -1102,18 +1107,26 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Advect_Points)::evalAdvectionP
...
-    openvdb::points::AttributeSet::Descriptor::parseNames(
-        parms.mIncludeGroups, parms.mExcludeGroups, evalStdString("vdbpointsgroups", now));
+	const std::string groups = evalStdString("vdbpointsgroups", now);
 
-    if (!parms.mPointGroup && ptGroupStr.length() > 0) {
-        addWarning(SOP_MESSAGE, "Point group not found");
-        return false;
+	// Get and parse the vdb points groups
+	openvdb::points::AttributeSet::Descriptor::parseNames(
+	    parms.mIncludeGroups, parms.mExcludeGroups, evalStdString("vdbpointsgroups", now));
```
danrbailey: parms.mIncludeGroups and parms.mExcludeGroups are required to have been set in the other branch (when VDB Points are being advected)

jmlait: Yes, this does look backwards.   I'll have to make some new tests to ensure this code is exercised & I'll resubmit when I've verified one way or another.

jmlait:
On Thu, Nov 1, 2018 at 8:59 PM Dan Bailey <notifications@github.com> wrote:
>
> @danrbailey requested changes on this pull request.
>
> You're also introducing a mixture of tabs into this spaces-only file. Perhaps we can talk about changing the conventions for the OpenVDB Houdini sub-project to align more with SideFX's own in the TSC meetings, but for now perhaps best to stick to spaces.

I don't think we should let SideFX's tab conventions spread anywhere
else.  Keep this space-only.  Maybe we can add a validation hook for
tab vs space to ensure we don't have to watch as closely?

jmlait:
Thank you for finding that problem.  I set up a vdb point group reg test and sure enough the group wasn't being used properly before.  17.0 has also been patched.  I've also retabbed so hopefully we are tab clean now.

As for the question of auto-selecting: I'd recommend make that a separate change once we've figured out what would work.

danrbailey:
> Maybe we can add a validation hook for tab vs space to ensure we don't have to watch as closely?
Yes, that would be great to have! :)

danrbailey:
> Thank you for finding that problem. I set up a vdb point group reg test and sure enough the group wasn't being used properly before. 17.0 has also been patched. I've also retabbed so hopefully we are tab clean now.
> 
> As for the question of auto-selecting: I'd recommend make that a separate change once we've figured out what would work.

Great, thanks. To clarify, the switching behavior for VDB points vs Houdini points is what's currently in H17? So this PR is merely to align ourselves with what Houdini is doing? 

Would you mind quickly going over the issue with the current auto-selecting behaviour again for me just to make sure I've properly understood the problems you're referring to?

jmlait:
On Fri, Nov 2, 2018 at 7:18 PM Dan Bailey <notifications@github.com> wrote:
> Great, thanks. To clarify, the switching behavior for VDB points vs Houdini points is what's currently in H17? So this PR is merely to align ourselves with what Houdini is doing?
>
> Would you mind quickly going over the issue with the current auto-selecting behaviour again for me just to make sure I've properly understood the problems you're referring to?

Correct.  17.0 already has the split-behaviour that I've submitted as
a PR.  Next daily build will have the fix you found as well.

Without that fix, if you specify a normal point group like "0-20" the
node will red-flag as it will fail to find a 0-20 primitive group.  I
wasn't sure if the primitive group selection was important, but it
certainly seemed a reasonable feature.  But if a group field is to be
overridden, it needs to be one or the other in any particular cook.

If we want dual-advection, we could add an extra parameter, or, as you
suggest, cut the primitive group feature entirely.  Disabling the
advection of the vdb primitive's point can be done by using
gdp->containsPrimitiveType() to cheaply verify if there are any vdbs
before going over the primitive list and removing points from the
group.

This is a new feature of VDB Advect Points, so we probably can still
change with a Compatibility note in 17.0 if a patch is made soon
enough.

danrbailey:
> Correct. 17.0 already has the split-behaviour that I've submitted as a PR. Next daily build will have the fix you found as well. Without that fix, if you specify a normal point group like "0-20" the node will red-flag as it will fail to find a 0-20 primitive group. I wasn't sure if the primitive group selection was important, but it certainly seemed a reasonable feature. But if a group field is to be overridden, it needs to be one or the other in any particular cook. If we want dual-advection, we could add an extra parameter, or, as you suggest, cut the primitive group feature entirely. Disabling the advection of the vdb primitive's point can be done by using gdp->containsPrimitiveType() to cheaply verify if there are any vdbs before going over the primitive list and removing points from the group. This is a new feature of VDB Advect Points, so we probably can still change with a Compatibility note in 17.0 if a patch is made soon enough.

Right, understood. Let's accept this then and I'll make sure to spend some time looking into how best to address this soon.

PR #281, opened 2018-11-06
ABI=6 and Fast AttributeArray Copying

danrbailey:
* Added ABI=6, bumped Travis to build for ABI=6 and deprecated some older ABI builds
* Add various new methods to AttributeArray for querying value and storage type and size metadata
* In-memory compression for AttributeArray has now been deprecated
* AttributeArray member variables have been unioned, cleaned up and packed to reduce memory from 64 bytes to 40 bytes (for an empty array)
* New AttributeArray copyValues() method that can copy batches of attributes using a custom iterator to avoid the cost of doing a virtual function call per-point
* Add copyValues() implementation to deletion of points to accelerate this method for ABI=6
* Add extra AttributeArray read checks to error if attempting to access or write partially-read arrays

### DIFF
--- a/openvdb/points/AttributeArray.h
+++ b/openvdb/points/AttributeArray.h
@@ -285,11 +285,13 @@ public:
     /// both value types are floating-point or both integer
     /// @param rangeChecking        if true, checks for out-of-range errors on source or target and
     ///                             checks that this array is not uniform
+    /// @note it is possible to use this method to write to a uniform target array if the iterator
+    /// does not have non-zero target indices, simply disable range-checking for this case
     template <typename IterT>
     void copyValuesUnsafe(const AttributeArray& sourceArray, const IterT& iter, bool rangeChecking = false);
-    /// @param preserveUniformity   if true, attempts to collapse this array if the array started as uniform
+    /// @param compact   if true, attempts to collapse this array
     template <typename IterT>
-    void copyValues(const AttributeArray& sourceArray, const IterT& iter, bool preserveUniformity = true);
+    void copyValues(const AttributeArray& sourceArray, const IterT& iter, bool compact = true);
     //@}
 #endif
 
@@ -303,11 +305,12 @@ public:
     /// Compact the existing array to become uniform if all values are identical
     virtual bool compact() = 0;
 
-    /// Return @c true if this array is compressed.
+    /// Deprecated method that previously returned true if this array is compressed,
+    /// now always returns false.
     OPENVDB_DEPRECATED bool isCompressed() const { return false; }
-    /// Compress the attribute array.
+    /// Deprecated method that previously compressed the attribute array, now it does nothing.
     OPENVDB_DEPRECATED virtual bool compress() = 0;
-    /// Uncompress the attribute array.
+    /// Deprecated method that previously uncompressed the attribute array, now it does nothing.
     OPENVDB_DEPRECATED virtual bool decompress() = 0;
 
     /// @brief   Specify whether this attribute should be hidden (e.g., from UI or iterators).
@@ -1029,7 +1032,7 @@ void AttributeArray::copyValuesUnsafe(const AttributeArray& sourceArray, const I
     bool rangeChecking/* = false*/)
 {
     // ensure both arrays have float-float or integer-integer value types
-    assert(sourceArray.valueTypeIsFloatingPoint() != this->valueTypeIsFloatingPoint());
+    assert(sourceArray.valueTypeIsFloatingPoint() == this->valueTypeIsFloatingPoint());
     // ensure both arrays have been loaded from disk (if delay-loaded)
     assert(sourceArray.isDataLoaded() && this->isDataLoaded());
     // ensure storage size * stride matches on both arrays
@@ -1066,6 +1069,7 @@ void AttributeArray::copyValuesUnsafe(const AttributeArray& sourceArray, const I
             // range-checking asserts
             assert(sourceIndex < sourceArray.dataSize());
             assert(targetIndex < this->dataSize());
+            if (this->isUniform())  assert(targetIndex == Index(0));
         }
 
         const size_t targetOffset(targetIndex * bytes);
@@ -1077,7 +1081,7 @@ void AttributeArray::copyValuesUnsafe(const AttributeArray& sourceArray, const I
 
 template <typename IterT>
 void AttributeArray::copyValues(const AttributeArray& sourceArray, const IterT& iter,
-    bool preserveUniformity/* = true*/)
+    bool compact/* = true*/)
 {
     const Index bytes = sourceArray.storageTypeSize();
     if (bytes != this->storageTypeSize()) {
@@ -1088,17 +1092,13 @@ void AttributeArray::copyValues(const AttributeArray& sourceArray, const IterT& 
     sourceArray.loadData();
     this->loadData();
 
-    const bool uniform = this->isUniform();
-
     // if the target array is uniform, expand it first
-    if (uniform) {
-        this->expand();
-    }
+    this->expand();
 
     this->copyValuesUnsafe(sourceArray, iter, /*rangeChecking = */true);
 
-    // attempt to compact target array if previously uniform
-    if (preserveUniformity && uniform) {
+    // attempt to compact target array
+    if (compact) {
         this->compact();
     }
 }
@@ -1150,7 +1150,7 @@ TypedAttributeArray<ValueType_, Codec_>::TypedAttributeArray(const TypedAttribut
     , mIsUniform(rhs.mIsUniform)
 #endif
 {
-    if (!this->isOutOfCore()) {
+    if (this->validData()) {
         this->allocate();
         std::memcpy(this->data(), rhs.data(), this->arrayMemUsage());
     }
@@ -1172,7 +1172,7 @@ TypedAttributeArray<ValueType_, Codec_>::operator=(const TypedAttributeArray& rh
         mStrideOrTotalSize = rhs.mStrideOrTotalSize;
         mIsUniform = rhs.mIsUniform;
 
-        if (!rhs.isOutOfCore()) {
+        if (this->validData()) {
             this->allocate();
             std::memcpy(this->newDataAsByteArray(), rhs.newDataAsByteArray(), this->arrayMemUsage());
         }
@@ -1304,6 +1304,10 @@ bool
 TypedAttributeArray<ValueType_, Codec_>::valueTypeIsFloatingPoint() const
 {
     // std::is_floating_point evaluates to false for half so use std::is_integral
+    // Note that as implemented, this only works as expected for vectors, not matrices
+    // and quaternions. However, because we don't use integer matrices or quaternions,
+    // this still performs correctly for all the types we care about.
+    // TODO: Update to use Traits that correctly handle matrices and quaternions.
     return !std::is_integral<typename VecTraits<ValueType>::ElementType>::value;
 }
 
--- a/openvdb/points/AttributeArray.h
+++ b/openvdb/points/AttributeArray.h
@@ -148,8 +148,32 @@ public:
     template <typename ValueType, typename CodecType> friend class AttributeHandle;
 
     AttributeArray() = default;
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+    AttributeArray(const AttributeArray& rhs)
+        : mCompressedBytes(rhs.mCompressedBytes)
+        , mFlags(rhs.mFlags)
+        , mSerializationFlags(rhs.mSerializationFlags)
+        , mOutOfCore(rhs.mOutOfCore)
+    {
+        if (rhs.mPageHandle)    mPageHandle = rhs.mPageHandle->copy();
+        else                    mPageHandle = compression::PageHandle::Ptr();
+    }
+    AttributeArray& operator=(const AttributeArray& rhs)
+    {
+        mCompressedBytes = rhs.mCompressedBytes;
+        mFlags = rhs.mFlags;
+        mSerializationFlags = rhs.mSerializationFlags;
+        mOutOfCore = rhs.mOutOfCore;
+        if (rhs.mPageHandle)    mPageHandle = rhs.mPageHandle->copy();
+        else                    mPageHandle = compression::PageHandle::Ptr();
+        return *this;
+    }
+#else
     AttributeArray(const AttributeArray&) = default;
     AttributeArray& operator=(const AttributeArray&) = default;
+#endif
+    AttributeArray(AttributeArray&&) = default;
+    AttributeArray& operator=(AttributeArray&&) = default;
     virtual ~AttributeArray() = default;
 
     /// Return a copy of this attribute.
--- a/openvdb/points/AttributeArray.h
+++ b/openvdb/points/AttributeArray.h
@@ -768,7 +768,16 @@ public:
 protected:
     AccessorBasePtr getAccessor() const override;
 
+    /// Return the raw data buffer
+    inline StorageType* data() { assert(validData()); return mData.get(); }
+    inline const StorageType* data() const { assert(validData()); return mData.get(); }
+
+    /// Verify that data is not out-of-core or in a partially-read state
+    inline bool validData() const { return !(isOutOfCore() || (flags() & PARTIALREAD)); }
+
 private:
+    friend class ::TestAttributeArray;
+
     /// Load data from memory-mapped file.
     inline void doLoad() const;
     /// Load data from memory-mapped file (unsafe as this function is not protected by a mutex).
@@ -1128,7 +1137,7 @@ TypedAttributeArray<ValueType_, Codec_>::TypedAttributeArray(
     }
     mSize = std::max(Index(1), mSize);
     mStrideOrTotalSize = std::max(Index(1), mStrideOrTotalSize);
-    Codec::encode(uniformValue, mData.get()[0]);
+    Codec::encode(uniformValue, this->data()[0]);
 }
 
 
@@ -1143,7 +1152,7 @@ TypedAttributeArray<ValueType_, Codec_>::TypedAttributeArray(const TypedAttribut
 {
     if (!this->isOutOfCore()) {
         this->allocate();
-        std::memcpy(mData.get(), rhs.mData.get(), this->arrayMemUsage());
+        std::memcpy(this->data(), rhs.data(), this->arrayMemUsage());
     }
 }
 
@@ -1349,10 +1358,9 @@ typename TypedAttributeArray<ValueType_, Codec_>::ValueType
 TypedAttributeArray<ValueType_, Codec_>::getUnsafe(Index n) const
 {
     assert(n < this->dataSize());
-    assert(!this->isOutOfCore());
 
     ValueType val;
-    Codec::decode(/*in=*/mData.get()[mIsUniform ? 0 : n], /*out=*/val);
+    Codec::decode(/*in=*/this->data()[mIsUniform ? 0 : n], /*out=*/val);
     return val;
 }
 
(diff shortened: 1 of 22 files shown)


openvdb/points/AttributeArray.h
```diff
@@ -836,6 +1019,92 @@ UnitVecCodec::encode(const math::Vec3<T>& val, StorageType& data)
...
+
+// AttributeArray implementation
+
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+template <typename IterT>
+void AttributeArray::copyValuesUnsafe(const AttributeArray& sourceArray, const IterT& iter,
+    bool rangeChecking/* = false*/)
+{
+    // ensure both arrays have float-float or integer-integer value types
+    assert(sourceArray.valueTypeIsFloatingPoint() != this->valueTypeIsFloatingPoint());
```
Idclip: Is this assert correct?

openvdb/points/AttributeArray.h
```diff
@@ -1063,6 +1298,53 @@ TypedAttributeArray<ValueType_, Codec_>::deallocate()
...
+
+
+template<typename ValueType_, typename Codec_>
+bool
+TypedAttributeArray<ValueType_, Codec_>::valueTypeIsScalar() const
+{
+    return  !this->valueTypeIsVector() &&
+            !this->valueTypeIsQuaternion() &&
+            !this->valueTypeIsMatrix();
+}
```
Idclip: If we introduce a any new array type this will need to be tracked. Maybe just use is_arithmetic<> (and check is_same<char>)?

openvdb/points/AttributeArray.h
```diff
@@ -836,6 +1019,92 @@ UnitVecCodec::encode(const math::Vec3<T>& val, StorageType& data)
...
+    assert(this->storageTypeSize()*this->stride() == sourceArray.storageTypeSize()*sourceArray.stride());
+
+    const size_t bytes(sourceArray.storageTypeSize()*sourceArray.stride());
+    const char* const sourceBuffer = sourceArray.dataAsByteArray();
+    char* const targetBuffer = this->dataAsByteArray();
+    assert(sourceBuffer && targetBuffer);
+
+    if (rangeChecking && this->isUniform()) {
+        OPENVDB_THROW(IndexError, "Cannot copy array data as target array is uniform.");
+    }
```
Idclip: can we move ```this->expand();``` from copyValues into here so this is never a concern?

openvdb/points/AttributeArray.h
```diff
@@ -836,6 +1019,92 @@ UnitVecCodec::encode(const math::Vec3<T>& val, StorageType& data)
...
+
+        if (rangeChecking) {
+            if (sourceIndex >= sourceDataSize) {
+                OPENVDB_THROW(IndexError,
+                    "Cannot copy array data as source index exceeds size of source array.");
+            }
+            if (targetIndex >= targetDataSize) {
+                OPENVDB_THROW(IndexError,
+                    "Cannot copy array data as target index exceeds size of target array.");
+            }
```
Idclip: This feels a bit clunky, especially in an "unsafe" method. Is there no way we can promote this rangeChecking flag to copyValues() which would check the iterator extrema?

openvdb/points/AttributeArray.h
```diff
@@ -836,6 +1019,92 @@ UnitVecCodec::encode(const math::Vec3<T>& val, StorageType& data)
...
+    if (uniform) {
+        this->expand();
+    }
+
+    this->copyValuesUnsafe(sourceArray, iter, /*rangeChecking = */true);
+
+    // attempt to compact target array if previously uniform
+    if (preserveUniformity && uniform) {
+        this->compact();
+    }
```
Idclip: How about simply a compact argument instead of preserveUniformity which tries to do this regardless. You lose the automatic ability to compact if originally uniform, but you just pass ```target.isUniform()``` as the argument instead

openvdb/points/AttributeArray.h
```diff
@@ -1861,18 +2072,18 @@ struct AccessorEval<UnknownCodec, ValueType>
 
 template <typename ValueType, typename CodecType>
 typename AttributeHandle<ValueType, CodecType>::Ptr
-AttributeHandle<ValueType, CodecType>::create(const AttributeArray& array, const bool preserveCompression)
+AttributeHandle<ValueType, CodecType>::create(const AttributeArray& array, const bool collapseOnDestruction)
```
Idclip: Is there any impact with re-using the preserveCompression argument for this?

openvdb/points/AttributeArray.h
```diff
@@ -203,11 +304,11 @@ class OPENVDB_API AttributeArray
...
 
     /// Return @c true if this array is compressed.
-    bool isCompressed() const { return mCompressedBytes != 0; }
+    OPENVDB_DEPRECATED bool isCompressed() const { return false; }
     /// Compress the attribute array.
-    virtual bool compress() = 0;
+    OPENVDB_DEPRECATED virtual bool compress() = 0;
     /// Uncompress the attribute array.
-    virtual bool decompress() = 0;
+    OPENVDB_DEPRECATED virtual bool decompress() = 0;
```
Idclip: I'd improve these comments to signify these are noops now

openvdb/points/AttributeArray.h
```diff
@@ -437,8 +563,13 @@ struct UnitVecCodec
 
 
 /// Typed class for storing attribute data
+
 template<typename ValueType_, typename Codec_ = NullCodec>
+#if OPENVDB_ABI_VERSION_NUMBER >= 6 // for ABI=6, class is final to allow for de-virtualization
+class TypedAttributeArray final: public AttributeArray
```
Idclip: be good to hear if you've confirmed this breaks ABI or not as I'm curious too

openvdb/points/AttributeArray.h
```diff
@@ -169,6 +205,37 @@ class OPENVDB_API AttributeArray
...
+    /// Return the size in bytes of the value type of a single element in this array.
+    /// (e.g. "float" -> 4 bytes, "vec3d" -> 24 bytes").
+    virtual Index valueTypeSize() const = 0;
+
+    /// Return the size in bytes of the storage type of a single element of this array.
+    /// @note If the Codec is a NullCodec, valueSize() == storageSize()
+    virtual Index storageTypeSize() const = 0;
+
+    /// Return @c true if the value type is floating point
+    virtual bool valueTypeIsFloatingPoint() const = 0;
```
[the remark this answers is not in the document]
danrbailey: No, this PR doesn't rely on that change. I made these two changes orthogonal, so they wouldn't have to be merged together. If there's no consensus for accepting the other PR, I can always make a new change to introduce those traits into a local namespace of this class to speed up this slow runtime string comparison.

openvdb/points/AttributeArray.h
danrbailey: No, that doesn't look correct, I'll fix that. Thanks.

openvdb/points/AttributeArray.h
danrbailey: The only reason I didn't do that is to be able to support half, which isn't included in std::is_floating_point. Question is whether an unknown type should be scalar or non-scalar - perhaps non-scalar makes more sense, I'm not sure?

openvdb/points/AttributeArray.h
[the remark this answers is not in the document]
Idclip: I'm referring explicitly to the functionality of this particular function - ```VecTraits<ValueType>::ElementType``` will work correctly for scalars and vectors but quats and matrices will behave unexpectedl i.e. float = true, vec3f = true, mat3f = false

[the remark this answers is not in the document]
danrbailey: Actually, that's not true and this specific case is tested in the unit tests - https://github.com/AcademySoftwareFoundation/openvdb/pull/281/files#diff-09ba2ff10db2a9aef3551d5ca873020dR537.

What you're saying is almost correct though in that a mat3i type would fail because the std::is_integral test here would resolve to using the struct not the element type. I'm ignoring this particular case because we don't use integer matrices or quaternions. However were the traits PR to go in, I'd update this method to use that instead so it's more correct in handling this edge case.

openvdb/points/AttributeArray.h
danrbailey: We compile a debug build, however we don't execute it because the time to do so is prohibitively slow. Ideally we should have an optimized build with asserts switched on which does get executed to catch these cases, but that's probably something to factor into the migration to the new Jenkins CI at this stage.

openvdb/points/AttributeArray.h
danrbailey: I'm still not entirely sure, my intuition is that this is implemented at the compiler optimization level so why would it affect ABI? For safety and until I learn otherwise though, I'm leaving this behind an ABI=6 macro.

openvdb/points/AttributeArray.h
danrbailey: On reflection, I'm not going to make this change as without it, it becomes possible to call this method with a uniform target array and an iterator that only generates zero indices to avoid having to go through type instantiation.

I have however documented this use case and added an additional unit test to ensure this behavior is verified.

openvdb/points/AttributeArray.h
danrbailey: I've changed this to match your suggestion.

openvdb/points/AttributeArray.h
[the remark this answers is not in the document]
danrbailey: Added a comment to this effect in the code.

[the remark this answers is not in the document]
Idclip: Ah I assumed you were using is_floating_point not is_integral. I think you should be using is_floating_point due to the nature of this function and the expected result. With all these methods, it's better practise to implement what they are actually defining instead of what they aren't e.g. what If i want a pointer array (float*) or string array (I realise strings are handled differently but there's not reason why you shouldn't be able to create one). These would return true!

If you're worried about supporting half, just add as is_same\<half>. For the rest, I'd advise adding an is_same<> for mat and quat fp types until your value traits PR is in (with a comment).

openvdb/points/AttributeArray.h
Idclip: Okay well two suggestions then:

1) Let the user deal with this in their iterator. Update the example iterator to store two size_t sizes for the source and target arrays and show range checking in the iterator. Completely remove everything bar the asserts.
2) Define functionality for out_of_range values which doesn't involve a throw. Maybe just ```if (sourceIndex >= sourceDataSize) continue;```

I think the first is the best solution

openvdb/points/AttributeArray.h
Idclip: I mean ideally you'd use is_scalar here to reflect the function name but it depends on your use case. If really all you want is to return true for ints and floats then I'd use is_arithmetic with char and half with an is_same

danrbailey: I think really the information I'd like to know is whether the value type is a POD type like int, half, double, etc or a struct-based type like vec3s, mat4d, etc. I agree with you that it's better if the implementation here aligns more with the C++ traits, so I've changed this to valueTypeIsClass() which overlays std::is_class.

Note that half has to be explicitly excluded, though it has been raised that the type traits there should be better specialized - https://github.com/openexr/openexr/issues/253.

openvdb/points/AttributeArray.h
danrbailey: I'm not sure I agree with your proposed solution as I don't like the idea that the user has to provide the minimum and maximum values. If the indices are not predictable and not in order that means either a sort or a second pass through the data to figure that out, just in order to do range checking. The range checking _is_ done automatically, but is done on every index. If you _really_ care about this performance the unsafe method is offered to eliminate that cost, provided you know what you're doing.

I do however agree that the range-checking parameter on the unsafe method is redundant, which bloats the public API slightly. As a result, I've introduced a new private member function which performs the copying and the range-checking parameter and both copyValues() and copyValuesUnsafe() call that method with the appropriate value.

openvdb/points/AttributeArray.h
Idclip: @danrbailey I've made an additional comment on this but you've marked it as resolved

openvdb/points/AttributeArray.h
Idclip: The user only needs to provide at most the two sizes of the arrays from dataSize(), not the minimum and maximum values. What I'm proposing is not performing any additional pass to do range checking. Its not necessarily about performance but also about what an "unsafe" method should represent. imo it shouldn't throw. I'm suggesting that a user can provide something like this (taken from your unit tests):

```
struct MyIterator
{
    using T = std::vector<std::pair<Index, Index>>;

    VectorWrapper(const T& _data, Index max) : data(_data), mMax(max) { }
    operator bool() const { return index < data.size(); }
    VectorWrapper& operator++() { index++; return *this; }
    Index sourceIndex() const { assert(*this); 
        if (data[index].first >= mMax) throw std::runtime_error("out_of_range"); // throw
        return data[index].first; 
    }
    Index targetIndex() const { assert(*this); 
        if (data[index].second >= mMax) return 0; // do some custom thing
        return data[index].second; 
    }

private:
    const T& data;
    const Index mMax;
    T::size_type index = 0;
};
```
simplifies the API, no additional pass, defers the control to the user as to how to handle out of range values in all cases, allows for optimal performance in the general case.

Idclip:
To clarify the outstanding conversations here:
- Implementation of valueTypeIsFloatingPoint - https://github.com/AcademySoftwareFoundation/openvdb/pull/281#discussion_r234567761
- Implementation and API of copyValues - https://github.com/AcademySoftwareFoundation/openvdb/pull/281#discussion_r234986656

I'll ask the dneg folk to voice their opinions too

openvdb/points/AttributeArray.h
danrbailey: The point is **not** to defer to the user to implement their own bounds-checking and hope they do it correctly, this _safe_ method should guarantee to **never** result in a seg-fault regardless of what array data or input you call it with. Consider std::vector::at() which does bounds-checking vs std::vector[] which doesn't for example. 

Since I've removed the range-checking parameter from the _unsafe_ method, this method does **not** throw.

We could provide just one implementation of this method with various different toggles for different permutations of checking ranges, strided sizes, value type checking, auto-expanding, throwing or asserting on errors, etc, etc. I think this overly complicates the API, hence the preferred balance between flexibility, simplicity, safety and performance in my opinion is a _safe_ method which provides all guarantees and checks and throws exceptions on errors and an _unsafe_ method which has none of these guarantees but offers bare-metal performance with asserts to help track down seg-faults.

openvdb/points/AttributeArray.h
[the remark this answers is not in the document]
danrbailey: I've tweaked the logic here to match what you're suggesting. Bear in mind that the most important aspect of this PR is adding the virtual function, not the actual implementation itself which we can continue to improve later. I considered simply throwing NotImplementedErrors as the 6.0.0 change and returning to do the implementation later, but felt it was simple enough to at least get a first version implementation in place.

danrbailey: Unresolved until you've approved (will resolve this comment).

openvdb/points/AttributeArray.h
Idclip: Right I didn't realise your change removed the issue that the unsafe method could throw, that's great! In which case we're left with the unnecessary branch complexity (the difference with your at vs vec analogy...). I am not suggesting providing a method for each permutation :\ yes this would definitely over-complicate the API.... I'm making the case that its better overall to defer this logic to the user (where they could define this). The 'hope they do it correctly' statement is a bit weird - you're designing an API to be used in a certain way, you're allowed to define that - there's a number of weird ways I could call copyValuesUnsafe with a custom object to cause problems (what if I don't increment, should we catch that? clearly not!). Obviously there's a fine line but I guess it's personal preference.

openvdb/points/AttributeArray.h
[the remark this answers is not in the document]
Idclip: Yeah sure I get that, that would have been fine also imo. This looks good now!

Idclip:
Approved

openvdb/points/AttributeArray.h
danrbailey: True, you can't protect for everything. I suppose I see it as most important to protect the user against accidentally introducing undefined behavior when using the safe implementation than introducing an infinite loop.

openvdb/points/AttributeArray.h
```diff
@@ -503,6 +640,36 @@ class TypedAttributeArray: public AttributeArray
...
+    Index valueTypeSize() const override { return sizeof(ValueType); }
+
+    /// Return the size in bytes of the storage type of a single element of this array.
+    /// @note If the Codec is a NullCodec, valueSize() == storageSize()
+    Index storageTypeSize() const override { return sizeof(StorageType); }
+
+    /// Return @c true if the value type is floating point
+    bool valueTypeIsFloatingPoint() const override;
+
+    /// Return @c true if the value type is scalar (not vector, quaternion or matrix)
```
richhones: Think this comment is the wrong way round?

openvdb/points/AttributeArray.h
```diff
@@ -169,6 +205,37 @@ class OPENVDB_API AttributeArray
...
+    virtual Index valueTypeSize() const = 0;
+
+    /// Return the size in bytes of the storage type of a single element of this array.
+    /// @note If the Codec is a NullCodec, valueSize() == storageSize()
+    virtual Index storageTypeSize() const = 0;
+
+    /// Return @c true if the value type is floating point
+    virtual bool valueTypeIsFloatingPoint() const = 0;
+
+    /// Return @c true if the value type is scalar (not vector, quaternion or matrix)
```
richhones: The comment is the wrong way round here too

openvdb/points/AttributeArray.h
danrbailey: Thanks, just resolved that.

openvdb/points/AttributeArray.h
danrbailey: Thanks, just resolved that.

zxiiro:
recheck

jmlait:
openvdb_houdini/houdini/PointUtils.cc still calls bloscCompressAttribute, despite being marked as deprecated.   This breaks our build as we treat deprecation use as error.

Since it is a no-op now, I think the right thing to do is just remove that loop entirely from the PointUtils.cc

danrbailey:
> openvdb_houdini/houdini/PointUtils.cc still calls bloscCompressAttribute, despite being marked as deprecated. This breaks our build as we treat deprecation use as error.
> 
> Since it is a no-op now, I think the right thing to do is just remove that loop entirely from the PointUtils.cc

@jmlait Thanks for highlighting this, I hadn't realized we'd exposed this within the 'Specific Attributes' branch. I've also made the 'bloscCompression' toggle invisible as it's now a no-op (I believe you can't mark a multi-parm parameter obsolete) and added a change to the ParmFactory to expose PRM_Template::setInvisible().

openvdb_houdini/houdini/ParmFactory.h
```diff
@@ -338,6 +338,12 @@ class OPENVDB_HOUDINI_API ParmFactory
     /// @details (The default vector size is one element.)
     ParmFactory& setVectorSize(int);
 
+    /// @brief Mark this parameter as hidden from the UI.
+    /// @note Marking parameters as obsolete is preferable to making them invisible as changing
+    /// invisible parameter values will still trigger a re-cook, however this is not possible
+    /// when using multi-parms.
+    ParmFactory& setInvisible(bool);
```
pcucka: Drop the boolean argument.  No one is going to need to `setInvisible(false)`, since that's the default setting.  And double negatives like that are awkward.

danrbailey: Agreed, just made that change, thanks.

PR #288, opened 2018-11-21
Crash fix in extractActiveVoxelSegmentMasks()

stasradz:
Fixed a crash in extractActiveVoxelSegmentMasks() when the first member of nodeSegmentArray is of size 0. Also, made the segmentActiveVoxels() to not to push any segments, when numSegments is 0.

### DIFF
--- a/openvdb/tools/LevelSetUtil.h
+++ b/openvdb/tools/LevelSetUtil.h
@@ -2416,8 +2416,15 @@ extractActiveVoxelSegmentMasks(const GridOrTreeType& volume,
     // 3. Group connected segments
 
     std::deque<NodeMaskSegmentRawPtrVector> nodeSegmentGroups;
+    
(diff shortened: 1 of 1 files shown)


danrbailey:
Hi @stasradz, I've just submitted a new PR that addresses the core issue you're fixing here - https://github.com/AcademySoftwareFoundation/openvdb/pull/336, so I'm going to close this one. 

The main difference is a slightly simpler approach using tools::pruneInactive() on the mask to drop any inactive leaf nodes instead of iterating over the leaf nodes to check their size and a unit test to verify the behaviour. In addition, regarding your second fix, I think we should still retain the behaviour where a grid with no inactive leaf nodes is output. 

Happy to have your feedback on my PR, thanks for the contribution! We discovered this issue at ILM about a year ago, but hadn't got around to submitting it. :)

PR #292, opened 2018-12-07
Fixed a backwards compatibility bug with SOP_OpenVDB_From_Particles

Idclip:
Fixed an issue which would stop attribute transfers if no primary output was selected

### DIFF
--- a/openvdb_houdini/houdini/SOP_OpenVDB_From_Particles.cc
+++ b/openvdb_houdini/houdini/SOP_OpenVDB_From_Particles.cc
@@ -1118,14 +1118,24 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_From_Particles)::cookVDBSop(OP
             outputInteriorMaskGrid = (0 != evalInt("buildinteriormask", 0, time)),
             outputBoundingMaskGrid = (0 != evalInt("buildmask", 0, time)),
             outputAttributeGrid =
-                ((0 != evalInt("buildattrs", 0, time)) && (evalInt("attrList", 0, time) > 0)),
-            needLeveLSet = (outputLevelSetGrid || outputFogVolumeGrid || outputBoundingMaskGrid);
+                ((0 != evalInt("buildattrs", 0, time)) && (evalInt("attrList", 0, time) > 0));
 
         if (!outputFogVolumeGrid && !outputLevelSetGrid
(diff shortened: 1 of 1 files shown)


Idclip:
@pcucka This is a small fix from some of the changes to this SOP - previously you used to be able to transfer attributes with no primary output (and the sop would build a surface behind the scenes to do this). This behaviour seems to have been lost with some recent changes.

openvdb_houdini/houdini/SOP_OpenVDB_From_Particles.cc
```diff
@@ -1128,6 +1127,17 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_From_Particles)::cookVDBSop(OP
...
+        bool needLeveLSet = (outputLevelSetGrid || outputFogVolumeGrid || outputBoundingMaskGrid);
+
+        // Backwards compatibility for when attribute grids are being requested
+        // but no level set output is desired. outputInteriorMaskGrid will always
+        // be false for older nodes. We need to provide convertWithAttributes()
+        // a valid level set in this case.
+
+        if (outputAttributeGrid && !outputLevelSetGrid && !outputInteriorMaskGrid) {
+            needLeveLSet = true;
+        }
```
Idclip: @pcucka Feel free to move this back up to be part of the assignment to needLeveLSet - split it out to be clear but perhaps this is always the intended behaviour anyway.

pcucka: Changing line 1122 to
```
needLevelSet = (outputLevelSetGrid || outputFogVolumeGrid
    || outputBoundingMaskGrid || outputAttributeGrid);
```
would be preferable.  Is that what you meant?

Idclip: Yeah exactly - though you can probably do:

```
needLevelSet = (outputLevelSetGrid || outputFogVolumeGrid
    || outputBoundingMaskGrid || (outputAttributeGrid && !outputInteriorMaskGrid));
```
to avoid building the surface if the mask is selected.

pcucka: That sounds good.  Do you want to update your PR?

Idclip: sure, done

PR #293, opened 2018-12-07
Minutes for last two TSC meetings.

jmlait:


tsc/meetings/2018-12-06.md
```diff
@@ -0,0 +1,74 @@
...
+A quorum was confirmed.
+
+Secretary - Jeff Lait
+
+3) Guidelines on Code Reviews
+
+We will attempt the CODEOWNER files to automate assignment of initial
+code reviews of pull requests.   Squashing is discouraged unless
+required for bisection; but will be reviewed in a few months.  We
+probably want documentation explaining how third parties can submit
```
pcucka: Did we decide that documentation for third parties should go in the CONTRIBUTING file and documentation for the TSC (code review policies, etc.) should go in the `process/` subdirectory?

jmlait: I could have missed a conclusion there as that was when my connection dropped :>  I seem to recall it was agreed that we want to split off a process directory, but I seem to recall the goal was the process directory would have the documentation for 3rd parties.  Thinking of it, that does seem backwards, so I probably heard backwards.  I think the immediate plan was to go forward with the single file focusing only on the TSC process, leaving as a future step providing 3rd party information.

PR #296, opened 2018-12-08
Adding TSC code review policy and best practices

danrbailey:


danrbailey:
@dokipen3d - I agree with you. I actually had this point as part of the initial draft, however as far as I'm aware there's no way to restrict commenting on a public pull request for GitHub (aside from actually blocking users), so I figured it's probably not useful to include it. If you think it's still worth explicitly stating that we encourage comments from anyone, we could always add it back in.

danrbailey:
Added a policy item about community feedback.

This is up-to-date for discussion in today's TSC meeting (@kmuseth @pcucka @Idclip @jmlait)

danrbailey:
I agree @kmuseth, I've made that change now. I've also introduced a CODEOWNERS file to this PR.

Based on my understanding, there are some subtle implications about the use of this CODEOWNERS that I'd like to try and summarize under this proposed code review policy: 

Before merging a PR which changes...

* a single file with just one owner and the author is _not_ the owner, that owner must approve
* a single file with just one owner and the author _is_ the owner, one of the other TSC members must approve
* a single file with two owners and the author is _not_ one of the owners, either of the owners must approve
* a single file with two owners and the author _is_ one of the owners, the other owner must approve
* two files with different owners, **both** owners must approve
* files that have no explicit owner (ie docs, unit tests), any of the TSC members must approve
* two files, one which has two owners, one which has just one owner, one of the two-owners must approve **and** the other owner (unless they are also one of the two-owners)

PR #299, opened 2018-12-15
Adding TSC meeting notes from 13th Dec

danrbailey:


kmuseth:
Looks good - only comment is please replace

"OpenVDB won and award" 

with

"OpenVDB won a Luminaire Tech award from the Advanced Imaging Society".

danrbailey:
Done, thanks @kmuseth, I didn't note down the details at the time.

PR #301, opened 2018-12-18
Starting OpenVDB 6.1.0 development

danrbailey:


pcucka:
Do we already know that there will be API-breaking changes?  If not, then we should bump the version to 6.0.1, not 6.1.0.

PR #303, opened 2019-01-02
Prim attr xfer

pcucka:
Refactored `openvdb_houdini::TransferPrimitiveAttributesOp` and added logic to iterate over polygon soups.  (The old code treated a polygon soup as a single primitive and therefore failed to find a corresponding reference polygon from which to transfer attributes.)


openvdb_houdini/houdini/SOP_OpenVDB_Scatter.cc
```diff
@@ -448,7 +448,7 @@ class SnapPointsOp
             // Translate Houdini points toward the isosurface.
             UTparallelForLightItems(GA_SplittableRange(mRange), [&](const GA_SplittableRange& r) {
                 const auto cptAcc = cpt->getConstAccessor();
-                GA_Offset start = 0, end = 0;
+                auto start = GA_Offset(), end = GA_Offset();
```
pcucka: This is actually an unrelated fix (for PR #277).

jmlait: Yay!  Then I can cancel that PR if this gets committed.

pcucka: I'm not sure I follow you.  Wouldn't `auto x = GA_Offset()` invoke the same special default constructor in strict builds that `GA_Offset x` does?

The problem with leaving these uninitialized is that in normal builds one gets (spurious, in this case) "variable may be used uninitialized" compiler warnings.

jmlait: Disclaimer: This is some confusing parts of the standard, so it is likely I've got confused.

In normal code, GA_Offset typedefs to exint, which typedefs to whatever the int64 intrinsic type is.  As such, it has no constructor.  The statement
```auto foo = GA_Offset();```
is similar to
```auto foo = int64();```
thus does value-initialization (previously called default-initialization), which translates to zero-ifying the contents.

In GA_STRICT_TYPE builds, however, GA_Offset isn't a typedef or we can't prohibit mixing of GA_Offset and GA_Index.  

The relevant code:

```class GA_API_TMPL GA_Offset
{
    typedef     GA_Offset           ThisType;

public:
                GA_Offset()
#if (UT_ASSERT_LEVEL >= UT_ASSERT_LEVEL_NORMAL)
                    : myVal(GA_MAGIC_BAD_VALUE)
#endif
                {}
    explicit    GA_Offset(GA_Size val) : myVal(val) { }
```

This defines a default constructor that leaves the contents undefined, except in debug builds where we set it to GA_MAGIC_BAD_VALUE.   In this case
```auto foo = GA_Offset();```
will invoke that constructor and leave foo either BAD_VALUE or uninitialized....  But we want those behaviours to catch the more common traditional error:
```GA_Offset foo;```

If you want to silence those warnings, I'd say instead to explicitly initialize them.  Zero is a pretty bad default value for GA_Offsets, and I shouldn't have suggested it in the first PR.  GA_INVALID_OFFSET would be better.

```auto start = GA_Offset(GA_INVALID_OFFSET), end = GA_Offset(GA_INVALID_OFFSET);```

(The GA_Offset I think is redundant here, but it might be easier to read)

pcucka: OK, I get it now: `GA_Offset()` behaves differently in debug vs. release builds.

I made your suggested change, to initialize with `GA_INVALID_OFFSET`.

PR #304, opened 2019-01-04
Downgrade library version to 6.0.1

pcucka:
Downgraded library version to 6.0.1 from 6.1.0 pending any API-breaking changes and removed a broken Doxygen link.

Reviewers: @kmuseth  @danrbailey

### DIFF
--- a/openvdb/version.h
+++ b/openvdb/version.h
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////
 //
-// Copyright (c) 2012-2018 DreamWorks Animation LLC
+// Copyright (c) 2012-2019 DreamWorks Animation LLC
 //
 // All rights reserved. This software is distributed under the
 // Mozilla Public License 2.0 ( http://www.mozilla.org/MPL/2.0/ )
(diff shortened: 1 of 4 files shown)


openvdb/version.h
```diff
@@ -174,7 +174,7 @@
 
 /// By default, the @b OPENVDB_REQUIRE_VERSION_NAME macro is undefined, and
 /// symbols from the version namespace are promoted to the top-level namespace
-/// so that, for example, @b openvdb::v6_0::io::File can be referred to
+/// so that, for example, @b openvdb::v5_0::io::File can be referred to
```
danrbailey: This is to revert this back to the previous version number with the intention that it isn't updated every version change right?

openvdb/version.h
```diff
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////
 //
-// Copyright (c) 2012-2018 DreamWorks Animation LLC
+// Copyright (c) 2012-2019 DreamWorks Animation LLC
```
danrbailey: @pcucka - I know in previous years you've done a bulk update of these copyright version numbers when the new year rolls around. Is the idea in just changing this one because the copyright should always be up-to-date for any files edited but it's not worth doing a bulk update because we're considering a move to SPDX soon?

openvdb/version.h
pcucka: > I can see why a patch version might be preferable though because it won't (potentially) require downgrading the version number.

Changing only the patch number also preserves the library namespace, FWIW.  And note that introducing new features, even substantial ones, doesn't necessitate a minor version bump, unless those changes break existing APIs.

> This is to revert this back to the previous version number with the intention that it isn't updated every version change right?

Right, there's no need to keep that comment in sync with the library version.

On the other hand, it _is_ necessary to keep the `vdbnamespace` and `OPENVDB_VERSION_NAME` definitions in sync (in [openvdb/doxygen-config](https://github.com/AcademySoftwareFoundation/openvdb/blob/d7f6c3d9afd74c1a3f585ac4ee4dcfd526dab809/openvdb/doxygen-config#L44)).

openvdb/version.h
pcucka: That was my thinking, yes.  Hopefully we can make a final decision soon on whether to remove the copyright dates, but maybe in the meantime I should just submit a PR with a bulk update.

PR #309, opened 2019-01-09
Add OPENVDB_NO_DEPRECATION_WARNING_BEGIN/END macros

pcucka:
These macros disable deprecation warnings from Clang, GCC and ICC.
With GCC and ICC, the macros also output a `#pragma message`, so as to discourage their use long-term.  (With Clang, there doesn't appear to be a way to output a message without generating a warning.)

Reviewers: @kmuseth @danrbailey

### DIFF
--- a/openvdb/Platform.h
+++ b/openvdb/Platform.h
@@ -186,6 +186,9 @@
         _Pragma("message(\"NOTE: ignoring deprecation warning\")")
     #define OPENVDB_NO_DEPRECATION_WARNING_END \
         _Pragma("GCC diagnostic pop")
+#else
+    #define OPENVDB_NO_DEPRECATION_WARNING_BEGIN
+    #define OPENVDB_NO_DEPRECATION_WARNING_END
 #endif
 
 
--- a/openvdb/Platform.h
+++ b/openvdb/Platform.h
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////
 //
-// Copyright (c) 2012-2018 DreamWorks Animation LLC
+// Copyright (c) 2012-2019 DreamWorks Animation LLC
 //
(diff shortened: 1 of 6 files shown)


Idclip:
This looks good. One minor thing is that I've found pragma messages a nuisance in the past as they're impossible (?) to disable without modifying the source - especially annoying if you're locked down to a particular version. For example some boost 1.61.0 trait headers had this:

```
In file included from In file included from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/detail/is_dereferenceable.hpp:12:0,
                 from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/detail/resolve.hpp:26,
                 from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/detail/push.hpp:24,
                 from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/detail/streambuf/indirect_streambuf.hpp:31,
                 from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/stream_buffer.hpp:22,
                 from /builds/boost/1.61.0/bd20733b58/include/boost/iostreams/stream.hpp:21,
...
 # pragma message("NOTE: Use of this header (bool_trait_def.hpp) is deprecated")
```

pcucka:
In this case, the annoyance is intentional.  When code is deprecated, references to it should be removed as expeditiously as possible, if not immediately.

Idclip:
I agree, but not being able to control output like this at all isn't always ideal. I think that anyone that includes PointMove.h will get deprecation output regardless of the methods they use? I guess you can include Platform and redefine the macros so at least there is a workaround unlike the case above. Not too concerned, thought I'd mention it anyway. Maybe we should consider building with -Wall by default

PR #314, opened 2019-01-11
OVDB-13: Optimize Travis build

danrbailey:
* Split out header tests into a separate concurrent job
* Removing the building of Python module and binaries from Houdini builds

Results in ~44m being the longest build when caching is not being used.


danrbailey:
Merging this to resolve the timeout issues that many of the Travis jobs are encountering, as mentioned in today's TSC meeting.

PR #315, opened 2019-01-12
Add missing override keyword to TypedAttributeArray deconstructor

danrbailey:


Idclip:
Weird travis timeouts but otherwise looks good. Would be good to fix the travis issues first even if they are unrelated.

danrbailey:
Not sure about the timeouts, but pretty sure it’s unrelated.

PR #316, opened 2019-01-14
OVDB-10: Introduce stricter Clang and GCC warnings for Travis builds

danrbailey:


Idclip:
Could we just bundled these into the current make/cmake setup with an optional flag such that we have an offline test environment which mimics what travis is doing to some degree?

danrbailey:
Thanks! I created a ticket to keep track of that change to Make/CMake - https://jira.aswf.io/browse/OVDB-23

PR #317, opened 2019-01-15
Advect Points can advect *both* VDB points and points

danrbailey:
Introduced a new primitive group field that applies to just VDB points primitives.

@jmlait - I've layered a commit on top of your previous commit (rather than appending to your https://github.com/AcademySoftwareFoundation/openvdb/pull/278). I tried a few different approaches to solve the filtering issue here, this seemed the best option to me. There are now two native filtering fields - a points one for Houdini points and a primitive one for VDB points. In addition there is a VDB points group field too. Although this does slightly confuse the UI having three filtering fields, I've tried to name and arrange the options so that it's as clear as possible. Obviously the main goal here is that the user can feed in any data, points or VDB points and by default (no filtering) everything will get advected. Hopefully you agree with these design decisions, but happy to hear your feedback.

### DIFF
--- a/openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
+++ b/openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
@@ -211,7 +211,6 @@ struct AdvectionParms
         : mOutputGeo(outputGeo)
         , mPointGeo(nullptr)
         , mPointGroup(nullptr)
-        , mVDBGroup(nullptr)
         , mOffsetsToSkip()
         , mIncludeGroups()
         , mExcludeGroups()
@@ -230,7 +229,6 @@ struct AdvectionParms
     GU_Detail* mOutputGeo;
     const GU_Detail* mPointGeo;
     const GA_PointGroup* mPointGroup;
-    const GA_PrimitiveGroup* mVDBGroup;
     std::vector<GA_Offset> mOffsetsToSkip;
     std::vector<std::string> mIncludeGroups;
     std::vector<std::string> mExcludeGroups;
@@ -1013,10 +1011,12 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Advect_Points)::cookVDBSop(OP_
 #endif
 
         // Evaluate UI parameters
+        const fpreal now = context.getTime();
+
         AdvectionParms parms(gdp);
         if (!evalAdvectionParms(context, parms)) return error();
 
-        const bool advectVdbPoints = (0 != evalInt("advectvdbpoints", 0, context.getTime()));
+        const bool advectVdbPoints = (0 != evalInt("advectvdbpoints", 0, now));
 
         hvdb::Interrupter boss("Processing points");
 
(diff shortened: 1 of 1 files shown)


danrbailey:
Rebased this PR on top of master to resolve conflicts as a result of https://github.com/AcademySoftwareFoundation/openvdb/pull/278 having been merged.

openvdb_houdini/houdini/SOP_OpenVDB_Advect_Points.cc
```diff
@@ -1107,25 +1111,18 @@ VDB_NODE_OR_CACHE(VDB_COMPILABLE_SOP, SOP_OpenVDB_Advect_Points)::evalAdvectionP
     UT_String ptGroupStr;
     evalString(ptGroupStr, "group", 0, now);
 
-    if (advectVdbPoints)
-    {
-        parms.mPointGroup = 0;
-        parms.mGroup = matchGroup(*parms.mPointGeo, ptGroupStr);
+    parms.mPointGroup = parsePointGroups(ptGroupStr, GroupCreator(gdp));
+    parms.mVDBGroup = matchGroup(*parms.mPointGeo, evalStdString("vdbgroup", now));
```
jmlait: I think we still need to guard mVDBGroup with an advectVdbPoints test.  We are not using this group if the advectVdbPoints is false, but matchGroups will throw an exception that will red flag the node if it is invalid.  So there is not just the trivial cost of computing the group; but the potential a node may flag red due to an invisible parameter value.  (As an aside, if these were using the .ds-style verbs disabled parameters would always evaluate to defaults so this would be safe; but I believe the current VDB shim will fully evaluate the parameters regardless of the disabled state)

danrbailey: Yes, good point, you're right, I can fix that. 

I'm still a bit wary about using the ds stuff as I've worked with other autogen c++ tools in the past and always found them quite fiddly to work with, but perhaps I just need to stick with it. I'm sure proper Make/CMake support for this in VDB would help. Certainly something worth discussing in our TSC meetings at some point though if you think there's a significant enough benefit to make it worthwhile making the transition or if it simplifies the integration of VDB into Houdini.

jmlait: I certainly agree with your suspicious on autogen c++ tools.

The first benefit of ds-style is that it exactly matches how HDAs store parameters.  This ensures there is always compatibility between what can be exported and what can be created.  The second is that it uses the disablewhen syntax.  This allows the parameter evaluation to be restricted to enabled parameters, avoiding any false dependencies and returning defaults.

But since VDB nodes themselves tend to be heavyweight, there isn't much benefit in converting them all to .ds.  It would be a big sweep.

danrbailey: Good to know, thanks.

Idclip: @jmlait @danrbailey just to mention here that the biggest benefit of converting to .ds is to resolve disabled time dependant parameters. Whilst this luckily doesn't seem to be common on many VDB nodes it still is significant and would be good to resolve, otherwise you end up cooking every frame!

PR #319, opened 2019-01-17
OVDB-18: Remove deprecated use of AttributeArray::set() in points/PointMove.h

danrbailey:


openvdb/points/PointMove.h
```diff
@@ -509,13 +535,41 @@ struct GlobalMovePointsOp
...
+
+            const Index sourceOffset = std::get<2>(it);
+            const Index targetOffset = indexOffsetFromVoxel(std::get<1>(it), targetLeaf, offsets);
+
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+            sourceArray.loadData();
+            SingleCopyIterator singleCopyIterator;
+            singleCopyIterator.setIndices(sourceOffset, targetOffset);
+            targetArray.copyValues(sourceArray, singleCopyIterator);
+#else
```
Idclip: There's a lot of unnecessary virtual methods being called here. copyValues calls loadData on both source and targte, and expand and compact on the target. You can alleviate the call to compact with the last argument but that still leaves the others. It seems that what you really want is to loop over all source arrays and load/expand the data once, then call copyValuesUnsafe() with range checking as true. I think I mentioned this in the original PR for this implementation.

openvdb/points/PointMove.h
```diff
@@ -509,13 +535,41 @@ struct GlobalMovePointsOp
...
-            OPENVDB_NO_DEPRECATION_WARNING_BEGIN
-            targetArray.set(tgtOffset, sourceArray, std::get<2>(it));
-            OPENVDB_NO_DEPRECATION_WARNING_END
+
+            const Index sourceOffset = std::get<2>(it);
+            const Index targetOffset = indexOffsetFromVoxel(std::get<1>(it), targetLeaf, offsets);
+
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+            sourceArray.loadData();
+            SingleCopyIterator singleCopyIterator;
```
Idclip: could move the construction of SingleCopyIterator out of this loop

danrbailey: I'd be very surprised if the compiler doesn't do this optimisation. Reason for putting it in the loop is to reduce the number of #if statements littering the code without incurring unused variable warnings.

openvdb/points/PointMove.h
danrbailey: I'll look at doing what you're suggesting here. Bear in mind, the main purpose of this PR is to eliminate the use of a deprecated function in this (very rare) code branch for ABI=6+. The current ABI=5 implementation is a lot of code but essentially provides similar performance to the newer method. 

Ultimately, what I'd really like to do is wait until ABI=5 is mostly no longer being used then do a big refactor of this whole logic to simplify the code dramatically and then leave ABI=5 using the slow virtual function copying method for backwards-compatibility only.

Idclip: Sure, though the removal of sourceArray.loadData() and passing false to copyValues for compact should definitely be done imo. Why is this code branch a very rare case? I just happened to choose this change set to comment on, but it applies to the other one in this PR too. I can see a case of wanting to copy arbitrary (potentially single) indices from multiple source arrays to a single target array being common (e.g. point merge by groups) and think it's good that this is highlighted now as it does seem a fairly obvious limitation.

openvdb/points/PointMove.h
Idclip: Sure, this is minor, I think it's clearer regardless of what we assume the compilers doing (at the cost of more #if statements) but I'll leave it up to you.

openvdb/points/PointMove.h
danrbailey: It's a rare case, because for all registered types (which includes all types available in Houdini), this logic doesn't get invoked at all. It purely exists for the cases where you're registering and using an attribute type which isn't registered by default.

I'm going to do the aforementioned refactor now to produce a more tailored ABI=6 version. It's a bit of code duplication, but the ABI=6 one can be much simpler as a result of the introduction of AttributeArray::copyValues().

Idclip: Thanks for clarifying, that sounds good to me. Though is there any reason why rangeChecking couldn't be exposed as an argument on copyValuesUnsafe? Seems that that would resolve this too

danrbailey:
@Idclip - updated this PR with two major changes:

* branched the global and local point moving logic into ABI=5 and ABI=6 logic, the latter of which has been re-implemented using custom copy iterators and the new AttributeArray::copyValues() methods (as discussed) - the number of lines of code in the ABI=6 logic is around 2/3 of the ABI=5 logic as a result
* abstracted out the sorting of the point orders into deterministic lists per-leaf so that this is now computed just once for the whole point set instead of once per attribute for improved performance

My intention is to leave this code as-is until ABI=6+ usage is significantly more common than ABI=5, then I'll drastically simplify the ABI=5 logic to remove the code that attempts to work-around the performance issues related to the use of the virtual functions (until we can finally deprecate ABI=5 in years to come)

Idclip:
Sounds great @danrbailey thanks for doing this now. I'll take a look early next week

openvdb/points/PointMove.h
```diff
@@ -186,26 +186,377 @@ class CachedDeformer
 
 namespace point_move_internal {
 
+using IndexArray = std::vector<Index>;
 
 using IndexTriple = std::tuple<LeafIndex, Index, Index>;
 using IndexTripleArray = tbb::concurrent_vector<IndexTriple>;
 using GlobalPointIndexMap = std::vector<IndexTripleArray>;
+using GlobalPointIndexIndices = std::vector<IndexArray>;
```
Idclip: All these typedefs are becoming confusing, we should really implement a class which holds these values with a sensible API to help people follow this code (specifically for GlobalPointIndexMap, GlobalPointIndexIndices and LocalPointIndexMap). Perhaps not required for this PR

Idclip: > Ultimately, it's all internal logic and not exposed via the public API, so I didn't spend too much time on it
No worries, but will help with future reviews too :)

PR #322, opened 2019-01-17
OVDB-19: Add support for 8-bit integer point attribute types to match Houdini

danrbailey:


danrbailey:
> Could you update PointSample.h to support int8_t too?

Well spotted, done.

PR #325, opened 2019-01-18
Additional fixes for PR 263

Idclip:


openvdb/doc/changes.txt
```diff
@@ -10,9 +10,14 @@
 Bug fixes:
 - Fixed a bug in the @link tools/Clip.h clip@endlink tool that caused
   some grid metadata to be discarded.
+- Added a check to @vdblink::points::setGroup setGroup@endlink to compare
```
pcucka: What bug did this change address, specifically (data corruption, a crash, etc.)?

openvdb_houdini/houdini/PointUtils.h
```diff
@@ -117,6 +117,7 @@ computeVoxelSizeFromHoudini(
 /// @param  attributes     a vector of VDB Points attributes to be included
 ///                        (empty vector defaults to all)
 /// @param  transform      transform to use for the new point grid
+/// @param  warnings       list of warnings to be added to the SOP
```
pcucka: Is this warning API really necessary?  Can you not use the existing logging mechanism?

Idclip: If I've understood correctly, the existing mechanism requires using OPENVDB_LOG macros with log4cplus? We don't currently use that here though it should be easy enough for us to change. My main concern is anyone building without log4cplus won't get these messages on the nodes, is that correct? It looks like the macros aren't used much in the houdini lib, so it would be a bit strange to have all other messages propagate to SOPs correctly except these ones when building without l4c, and I imagine we don't want to make that a required dependency. If generally everyone is building with this then I'm happy to change it

openvdb/doc/changes.txt
Idclip: perhaps an improvement is more relevant here as the unexpected behaviour only manifested as an incorrect runtime error:

    - Fixed point group initialization in points::setGroup() when a PointDataIndex
      tree is used but has been constructed from NaN positional values.

openvdb_houdini/houdini/PointUtils.h
pcucka: The core library uses logging exclusively for warnings, and the SOPs are mostly wrappers around core library functions, so while it is true that without log4cplus warnings won't appear on nodes, that is already largely the case for OpenVDB SOPs.

If building log4cplus is a concern for you, perhaps we should consider switching from log4cplus to Boost.Log, given that we already depend on Boost.

Idclip: Most of the SOPs seem to add their "main" user facing warnings directly using addWarning() and there's not too many usages of OPENVDB_LOG_WARN in the core library by my count. That being said it shouldn't be a problem for us to start using this, I'll just need a bit more time to transition here and test these changes. 

We should probably use the same system everywhere within the Houdini lib. The implementation of the error manager with log4cplus looks good, but I suspect sidefx don't build with this either @jmlait? This would be an issue as we'd definitely want them visible on the nodes bundled with Houdini.

jmlait: Houdini doesn't build with log4cplus currently.  And I don't think it should.

I don't think any core library warnings should ever show up as yellow warnings on Nodes.  It is practically impossible to generate a useful warning in the core library for the end artist as you have no idea of the context.  We already have had problems trying to get error messages from exceptions properly re-contextualized.  This is why any required pre-condtions should be tested at the wrapper level where you can present errors/warnings that make as clear as possible references to which parameter/value is invalid.

In any case, PointUtils.h isn't in the core library, but is in the houdini library.  As such, I think it should explicitly use addWarning for its warnings.  Using a std::vector to accumulate text, or passing in a UT_ErrorManager to add them to, are both reasonable approaches.

pcucka: > It is practically impossible to generate a useful warning in the core library for the end artist as you have no idea of the context.

That is one reason why the core library issues warnings only sparingly.  But surely it is better to know that an operation failed due to a singular matrix somewhere or because some substep is not supported for operands of certain types than to know nothing at all.  It is similarly practically impossible to anticipate all such errors simply by examining parameter values, though I don't dispute that that is the preferred approach.

jmlait: If I have no idea how my parameters caused a singular matrix (or what a singular matrix is) then it is not very helpful to have that warning.  As an artist, failure of the operation is determined by looking at the output.  Sometimes these warnings can be generated by side-effects that weren't contributing to the final result; so my output is fine and they just causing confusion.

Some warning-like things can only be found at run time.  PCG failing to converge, for example.  But, in addition to any logging output, PCG should provide that information to its caller, and not rely on a logging message to send it up.  The caller may be able to recover gracefully from that failure, so have no need to send the result to the top level.  Sending it to a log makes it infeasible to suppress.   

I prefer it if things like matrix invert() return a status code instead of having to first pre-determine if they are singular to avoid triggering exceptions/logs.  

To be clear, logging is very useful for those using the VDB API as programmers.  If log output always went to node info, we'd have to be more sparing with logging, which I don't want either.

Idclip: Okay, based on this discussion I think I'm going to leave this API as it is for now, as long as that's okay with you @pcucka ? I'll update the change files in a new PR

PR #330, opened 2019-01-25
Pre-defined compiler warnings for gcc and clang

Idclip:
- Introduced a set of pre-defined compiler warnings for gcc and clang for all projects. 
 - Updated travis to use warnings flag

### DIFF
--- a/openvdb/Makefile
+++ b/openvdb/Makefile
@@ -290,7 +290,9 @@ ifeq (yes,$(strip $(warnings)))
     ifneq (,$(USING_CLANG))
         CXXFLAGS += \
             -Qunused-arguments \
-            -Weverything \
+            -Wall \
+            -Wextra \
+            -Wconversion \
             -Wc++0x-compat \
             -Wno-c++98-compat-pedantic \
             -Wno-covered-switch-default \
--- a/openvdb/Makefile
+++ b/openvdb/Makefile
@@ -62,6 +62,8 @@
 #                       (default: link against shared libraries)
 #   debug=yes           build with debugging symbols and without optimization
 #   verbose=yes         run commands (e.g., doxygen) in verbose mode
+#   warnings=yes        Enable a collection of pre defined compiler warnings
+#                       for GCC and clang
 
 
 #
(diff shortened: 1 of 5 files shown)


openvdb/Makefile
```diff
@@ -282,6 +284,41 @@ ifneq (2,$(strip $(GLFW_MAJOR_VERSION)))
     CXXFLAGS += -DOPENVDB_USE_GLFW_3
 endif
 
+ifeq (yes,$(strip $(warnings)))
+    USING_CLANG=$(shell ${CXX} -dM -E -x c++ - < /dev/null | grep __clang__)
+    USING_GCC=$(shell ${CXX} -dM -E -x c++ - < /dev/null | grep __GNUC__)
```
Idclip: The most reliable way I've been able to come up with for detecting gcc vs clang irrespective of any wrappers around CXX, assuming that either gcc or clang has indeed been called. clang also defines __GNUC__.

pcucka: ICC also defines `__GNUC__`, but its diagnostic flags are incompatible.

How about "`${CXX} --version | grep clang`", "`${CXX} --version | grep GCC`", etc.?

openvdb/Makefile
```diff
@@ -282,6 +284,41 @@ ifneq (2,$(strip $(GLFW_MAJOR_VERSION)))
...
+    ifneq (,$(USING_CLANG))
+        CXXFLAGS += \
+            -Qunused-arguments \
+            -Weverything \
```
pcucka: Evidently, `-Weverything` is [not recommended](https://softwareengineering.stackexchange.com/questions/122608/clang-warning-flags-for-objective-c-development/124574#124574).  I think we've agreed to go with "`-Wall -Wextra -Wconversion`" instead ([OVDB-10](https://jira.aswf.io/browse/OVDB-10)).

openvdb/Makefile
```diff
@@ -62,6 +62,8 @@
 #                       (default: link against shared libraries)
 #   debug=yes           build with debugging symbols and without optimization
 #   verbose=yes         run commands (e.g., doxygen) in verbose mode
+#   warnings=yes        Enable a collection of pre defined compiler warnings
```
pcucka: "`strict`" would be a better name for this option, since all compilers generate some warnings even without any flags.

openvdb/Makefile
Idclip: Yeah I considered this but for some reason thought querying the defines would be more 'correct'. Not sure why, I'll change this

openvdb/Makefile
Idclip: Sure will do, didn't see that conversation and was going to ask about -Werror!

openvdb/Makefile
Idclip: done

danrbailey:
Looks like it's heading in the right direction, thanks a lot for doing this @Idclip. 

To let you know, I'm currently narrowing in on a sane set of compiler flags for Clang ("-Wall -Wextra -Wconversion" currently produces so many warnings, it exceeds the maximum log size for Travis).

Idclip:
Ha nice, okay well feel free to add to these changes

danrbailey:
The correct flags to use for Clang are (see that JIRA):

-Wall -Wextra -Wconversion -Wno-sign-conversion

PR #332, opened 2019-01-25
Release process documentation

pcucka:

### DIFF
--- a/tsc/process/release.md
+++ b/tsc/process/release.md
@@ -3,7 +3,7 @@
 The following assumes that the current OpenVDB library version number is 6.0.0 and the new version number is 6.1.0. Adjust for the actual version numbers as appropriate.
 
 - [ ] Open a Jira "Release OpenVDB 6.1.0" ticket with "OpenVDB_6.1.0" as the Fix Version.
-- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.
+- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.  [_Specifics TBD, pending a review of release note management tools._]
 - [ ] Update `openvdb/INSTALL` with the library version number and any directory hierarchy changes.
 - [ ] Open a pull request to merge the above changes into `openvdb/master`.  Associate the pull request with the Jira ticket created earlier, and verify that the Travis CI build runs successfully.
 - [ ] Draft a new [GitHub release](https://github.com/AcademySoftwareFoundation/openvdb/releases). Title it "OpenVDB 6.1.0" and tag it as `v6.1.0`.
--- a/tsc/process/release.md
+++ b/tsc/process/release.md
@@ -1,0 +1,20 @@
+**Release Process for OpenVDB**
+
+The following assumes that the current OpenVDB library version number is 6.0.0 and the new version number is 6.1.0. Adjust for the actual version numbers as appropriate.
+
+- [ ] Open a Jira "Release OpenVDB 6.1.0" ticket with "OpenVDB_6.1.0" as the Fix Version.
+- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.
+- [ ] Update `openvdb/INSTALL` with the library version number and any directory hierarchy changes.
+- [ ] Push the above changes to `openvdb/master`, associating them with the Jira ticket created earlier, and verify that the Travis CI build runs successfully.
+- [ ] Draft a new [GitHub release](https://github.com/AcademySoftwareFoundation/openvdb/releases). Title it "OpenVDB 6.1.0" and tag it as `v6.1.0`.
+
+- [ ] Update `openvdb-website/contents/download/index.html` with the new version number (6.1.0) and release date, and `openvdb-website/contents/index.html` with a news item. Open the latter in a browser and check that the website renders correctly and that there are no broken links.
+- [ ] Build the documentation (both the `doc` and `pydoc` targets) and replace the contents of `openvdb-website/contents/documentation/doxygen/` with the output.  [_This step should be automated, and the thousands of files it generates should preferably not be committed to the repository._]
+- [ ] Push the above changes to `openvdb-website/master`, associating them with the Jira ticket created earlier.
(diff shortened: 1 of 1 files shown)


tsc/process/release.md
```diff
@@ -0,0 +1,20 @@
+**Release Process for OpenVDB**
+
+The following assumes that the current OpenVDB library version number is 6.0.0 and the new version number is 6.1.0. Adjust for the actual version numbers as appropriate.
+
+- [ ] Open a Jira "Release OpenVDB 6.1.0" ticket with "OpenVDB_6.1.0" as the Fix Version.
```
danrbailey: Note that if we use the JIRA ASWF for this, we'll need to adjust the configuration to add "Fix Version" (among other currently disabled features referenced here). I'm fine with leaving this as-is for the time being though.

tsc/process/release.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.
```
danrbailey: Can you clarify - assuming that the release notes have been maintained throughout the commit history since the last release, the only task needed here would be to adjust the release numbers in these changes files (ie "In Development" -> "6.1.0")?

tsc/process/release.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] Build the documentation (both the `doc` and `pydoc` targets) and replace the contents of `openvdb-website/contents/documentation/doxygen/` with the output.  [_This step should be automated, and the thousands of files it generates should preferably not be committed to the repository._]
+- [ ] Push the above changes to `openvdb-website/master`, associating them with the Jira ticket created earlier.
+- [ ] Post a release announcement to the [OpenVDB forum](https://groups.google.com/forum/#!forum/openvdb-forum).
+
+- [ ] In preparation for the next release, change one or more of `OPENVDB_LIBRARY_PATCH_VERSION_NUMBER`, `OPENVDB_LIBRARY_MINOR_VERSION_NUMBER` and `OPENVDB_LIBRARY_MAJOR_VERSION_NUMBER` in `openvdb/version.h`.  Unless it is known that the next release will include API- or ABI-breaking changes, increment only the patch number to begin with (in this case, from 6.1.0 to 6.1.1).  In `openvdb/doxygen-config` update `PROJECT_NUMBER`, `OPENVDB_VERSION_NAME`, `OPENVDB_ABI_VERSION_NUMBER` and the `@vdbnamespace` alias to match `version.h`.  Push these changes to `openvdb/master`.
```
danrbailey: Perhaps a missing step to add the "In Development" for the next version in openvdb/changes?

(As above, no direct pushes to master, so this would still need to be a PR that's approved.)

tsc/process/release.md
```diff
@@ -0,0 +1,20 @@
...
+
+- [ ] Open a Jira "Release OpenVDB 6.1.0" ticket with "OpenVDB_6.1.0" as the Fix Version.
+- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.
+- [ ] Update `openvdb/INSTALL` with the library version number and any directory hierarchy changes.
+- [ ] Push the above changes to `openvdb/master`, associating them with the Jira ticket created earlier, and verify that the Travis CI build runs successfully.
+- [ ] Draft a new [GitHub release](https://github.com/AcademySoftwareFoundation/openvdb/releases). Title it "OpenVDB 6.1.0" and tag it as `v6.1.0`.
+
+- [ ] Update `openvdb-website/contents/download/index.html` with the new version number (6.1.0) and release date, and `openvdb-website/contents/index.html` with a news item. Open the latter in a browser and check that the website renders correctly and that there are no broken links.
+- [ ] Build the documentation (both the `doc` and `pydoc` targets) and replace the contents of `openvdb-website/contents/documentation/doxygen/` with the output.  [_This step should be automated, and the thousands of files it generates should preferably not be committed to the repository._]
```
danrbailey: I note that this doxygen documentation is for the last release. As part of the move to automate this, would it be desirable if we were to instead make this the doxygen for the current master branch? 

We're building all doc targets as part of our CI. If we implement the policy where we ensure that these are successful before a PR is merged, that means the doxygen would be always correct and in-sync with a fresh clone of the repo. I still think it could be useful to maintain doxygen for previous releases too though.

tsc/process/release.md
```diff
@@ -0,0 +1,20 @@
+**Release Process for OpenVDB**
+
+The following assumes that the current OpenVDB library version number is 6.0.0 and the new version number is 6.1.0. Adjust for the actual version numbers as appropriate.
+
+- [ ] Open a Jira "Release OpenVDB 6.1.0" ticket with "OpenVDB_6.1.0" as the Fix Version.
+- [ ] Update `openvdb/CHANGES` and `openvdb/doc/changes.txt` with release notes.
+- [ ] Update `openvdb/INSTALL` with the library version number and any directory hierarchy changes.
+- [ ] Push the above changes to `openvdb/master`, associating them with the Jira ticket created earlier, and verify that the Travis CI build runs successfully.
```
danrbailey: Our policy is no direct pushes to master, so I think this should still be a PR which is then approved before being merged in.

tsc/process/release.md
pcucka: Right, and at least one of us would have to have the necessary permissions to create new versions.

tsc/process/release.md
pcucka: If the release notes have been maintained throughout, there should be nothing to do here other than to replace "In Development" with the target release date.  But if we're going to consider alternatives like reno or temporary change logs, then yes, this step should specify precisely what needs to be done to bring the release notes up to date.

tsc/process/release.md
pcucka: I think we should make docs for all official releases available on the website (I do have archives of all of them), but certainly it wouldn't hurt to include an "in development" version as well.

tsc/process/release.md
pcucka: Agreed—I've added that clarification.

tsc/process/release.md
pcucka: I've added that to the "In preparation" step.

PR #342, opened 2019-02-05
OVDB-26: Fix Clang warning in ParticlesToLevelset

danrbailey:


openvdb/tools/ParticlesToLevelSet.h
```diff
@@ -952,6 +952,14 @@ inline BlindData<VisibleT, BlindT> Abs(const BlindData<VisibleT, BlindT>& x)
     return BlindData<VisibleT, BlindT>(math::Abs(x.visible()), x.blind());
 }
 
+/// @private
+// Required to support the (zeroVal<ValueType>() + val) idiom used in Grid::pruneGrid()
+template<typename VisibleT, typename BlindT>
+inline BlindData<VisibleT, BlindT> operator+(const BlindData<VisibleT, BlindT>& x, float)
+{
+    return x;
```
pcucka: This almost certainly breaks addition of `BlindData` with `float` when `BlindData::VisibleT` is any type other than `float`.  Better to just add a `static_cast` to `BlindData::operator+(const VisibleT&)`.

danrbailey: You're right, I just had another go at this one which is slighly different to what you suggested. There's now also a unit test which I believe should produce warnings on all compilers. 

I've slightly changed (fixed a bug in) the implementation too, as previously:

operator+(const BlindData<VisibleT, BlindT>& rhs, float x) would drop the blind data from rhs on addition of the visible portion, which I don't believe it should do.

pcucka: `operator+(const BlindData&)` and `operator-(const BlindData&)` both also drop `mBlind`, so maybe that's the intended behavior.  Seems wrong to me too, though.

danrbailey: I think the idea is that there isn't a strict requirement for operator addition to be defined for the blind portion of blind data?

Negation and Abs() do however retain the blind data, so I think this is _more_ correct than before.

danrbailey: Any further thoughts on this @pcucka? Do you think we should drop the blind data to be consistent?

pcucka: I guess that's a question for @kmuseth.  It doesn't make sense to me that the blind data component is discarded when adding or subtracting but preserved when negating or taking the absolute value.  It seems to me that either all of the arithmetic operations should preserve the blind data or none of them should.

danrbailey: Ok, thanks, I'll put a comment on the ticket.

kmuseth: Most of the methods in the BiindData class are only there to make the compiler happy and should never actually be invoked. Remember this class is in a private namespace and should never be used outside of this file. If I remember correctly the only important methods are operator < and > which are used during the rasterization. So in short it doesn't really matter if the blind data member is left untouched for operator + and - since none of them strictly makes sense and there is no meaningful way to define them. Note the blind data member is a particle index so adding or subtracting them could produce an invalid index. Just leave it as is - it's been working find for more than five years so why change it by introducing operations that are ultimately meaningless. Abs and Negation do however make sense since it changes the distance value (as you'd expect) but leaves the blind data member (the point index) unchanged.

kmuseth: and to Peters original comment about possibly failing if BlindData::VisibleT is a not a float - correct, but due to multiple static_assert(std::is_floating_point<typename GridT::ValueType>::value,... it's exactly guarantees that  BlindData::VisibleT is always a floating point number!

pcucka: `std::is_floating_point` guarantees that `VisibleT` is a floating-point type, not that it is `float`.

PR #345, opened 2019-02-06
Non constant stride value fix on OpenVDB Points Attribute Handles

Idclip:
Required for #341

@danrbailey This caught me out whilst writing new unit tests for the string array/handle improvements. Non constant strides should return 0, not 1

### DIFF
--- a/openvdb/points/AttributeArray.h
+++ b/openvdb/points/AttributeArray.h
@@ -2101,7 +2101,7 @@ AttributeHandle<ValueType, CodecType>::create(const AttributeArray& array, const
 template <typename ValueType, typename CodecType>
 AttributeHandle<ValueType, CodecType>::AttributeHandle(const AttributeArray& array, const bool collapseOnDestruction)
     : mArray(&array)
-    , mStrideOrTotalSize(array.hasConstantStride() ? array.stride() : 1)
+    , mStrideOrTotalSize(array.stride())
     , mSize(array.hasConstantStride() ? array.size() : array.dataSize())
     , mCollapseOnDestruction(collapseOnDestruction && array.isStreaming())
 {
--- a/openvdb/unittest/TestAttributeArray.cc
+++ b/openvdb/unittest/TestAttributeArray.cc
@@ -1413,10 +1413,10 @@ TestAttributeArray::testStrided()
         CPPUNIT_ASSERT_NO_THROW(StridedHandle::create(*array));
         CPPUNIT_ASSERT_NO_THROW(StridedWriteHandle::create(*array));
 
-        // handle is bound as if a linear array with stride 1
+        // handle is bound as if a linear array with stride 0
         StridedHandle handle(*array);
         CPPUNIT_ASSERT(!handle.hasConstantStride());
-        CPPUNIT_ASSERT_EQUAL(Index(1), handle.stride());
+        CPPUNIT_ASSERT_EQUAL(Index(0), handle.stride());
         CPPUNIT_ASSERT_EQUAL(array->dataSize(), handle.size());
     }
 }


openvdb/points/AttributeArray.h
```diff
@@ -2101,7 +2101,7 @@ AttributeHandle<ValueType, CodecType>::create(const AttributeArray& array, const
 template <typename ValueType, typename CodecType>
 AttributeHandle<ValueType, CodecType>::AttributeHandle(const AttributeArray& array, const bool collapseOnDestruction)
     : mArray(&array)
-    , mStrideOrTotalSize(array.hasConstantStride() ? array.stride() : 1)
+    , mStrideOrTotalSize(array.stride())
```
danrbailey: This isn't as simple as it looks to change. One implication of this change is that handle.get(5) returns the same offset into the array as handle.get(0) due to the fact that this calls index(n, m=0) { return n * stride + m; }

We need to think about how we index into a dynamic array. I'd be more confident making a change like this if it were included with full support for dynamic arrays (including resizing) and with a bunch of unit tests to verify the expected behaviour.

Idclip: But in the case you've given, that's also wrong for non constant strides :| I could alternative change stride() to return mArray->stride(), so at least the stride values match the attribute array functionality and leave mStrideOrTotalSize set to 1.

danrbailey: I'm confused. First clarification - dynamic arrays and non constant strides are the same thing.

For dynamic arrays, stride() is kind of undefined, there's really no useful value it can return. Though we used to have this behaviour, you should not be relying on it being zero to indicate that this is a dynamic array any more, you should be using hasConstantStride() to tell you that. However, it _is_ defined in certain contexts for convenience and performance - indexing being one of them, so for that reason I think the best value for it to return is 1 for dynamic arrays.

I do agree with you that AttributeHandle::stride() should return the same value as AttributeArray::stride(), however where possible we also should be trying not to call these virtual functions each time, as we would prefer that value to be cached on the handle when first created. (I'm aware there are other functions in the handle that do this which we should also fix).

Idclip: Okay I thought I understood this but this comment has confused me slightly. You've mentioned that dynamic  arrays and non constant strides are the same, but then also mention that we should not be using hasConstantStride() to indicate a dynamic array? 

My original point was that although changing stored value to zero changes the result of get, get already returns non sensical values for non constant strides anyway? 

Sorry if I've misunderstood anything here, though the documentation and methods for these cases are confusing. 

Regarding the stride virtual call, can we just static cast the array to the expected typed array type when called?

danrbailey: Yes, use hasConstantStride() to determine dynamic array or not.

There's definitely more work to be done on these dynamic arrays. I have some mostly finished changes on a branch I can share, but it's not ready to be made into a PR yet. 

Another way of thinking about dynamic arrays is externally indexed arrays. You're essentially giving up ownership of the indexing strategy - these are just blind arrays that simply store length, the ability to get and set them and I/O. However, which point associates with which value is not necessarily known or stored by the data structure. With that in mind, you can access the data by doing handle->get(i) and that will index into it like it's a simple linear array, which of course it is under-the-hood. I believe that your proposal will break that functionality, requiring you to do something weird like handle->get(0, i) to access the linear array.

Going forward with this, we should properly distinguish two specific use-cases for dynamic arrays:

* blind data - a really convenient way to put one or more buffers of random data into a leaf that you wish to be serialized and deserialized with the leaf without requiring any extra registration of new types and any consistency in their lengths and in relation to the number of points.
* array of array attributes - the more conventional use similar to Houdini's implementation where you can store an arbitrary number of values per-point. Currently, this indexing strategy needs to be self-managed somehow and it's this case where we should extend the API to enforce a convention for how to represent this indexing so it can be passed around with the data structure. The key design decision that differs this implementation from Houdini's is that Houdini physically stores arrays of arrays whereas we map them into a single array with all the marshalling needed to convert between the two, but there are pros and cons for each style of storage.

Hope that clarifies, I'll try and put in a JIRA (or two) and put some information on that.

danrbailey: I realize I didn't answer your casting question - no you can't do that because AttributeHandle offers the option of using UnknownCodec so you don't need to specify it for convenience and you can't static_cast without knowing all the template arguments.

Idclip: Ah yeah, forgot about UnknownCodecs, okay. Yeah is all pretty much how I understood it. One thing:

> you should not be relying on it being zero to indicate that this is a dynamic array any more

Feel like this is only the case because of the limitation with the attribute handles rather than an enforced workflow? The comment on the attribute array for stride() is:

`/// @note a return value of zero means a non-constant stride`

Which makes sense, I like the idea of stride returning 0 for a dynamic array if the arrays themselves are never going to have a known implementation of their offsets. So what it boils down to is that we have an API for the attribute handles which isn't ideal for dynamic arrays and has unfortunately already established a type of implementation for get(), correct? It felt a bit odd writing unit tests to test that the stride form the attribute handles is 1 for dynamic arrays, but it's good to hear you have solutions in the works. I'll close this for now, look forward to seeing them!

danrbailey: That note needs updating, it has obviously been missed. 

I guess one way of thinking about this is the stride is not necessarily how the data is stored (as I mentioned for dynamic arrays, it isn't well-defined), but how you wish to iterate through the data and then arguably a stride of zero doesn't mean anything.

PR #350, opened 2019-02-08
OVDB-30: Fix bug where AttributeArray point moving was not read-only 

danrbailey:


openvdb/points/PointMove.h
```diff
@@ -442,8 +442,8 @@ struct GlobalMovePointsOp
             if (newSourceLeafIndex > sourceLeafIndex) {
                 copyIterator.reset(startIndex, endIndex);
 
-                LeafT& sourceLeaf = mSourceLeafManager.leaf(sourceLeafIndex);
-                const auto& sourceArray = sourceLeaf.attributeArray(mAttributeIndex);
+                const LeafT& sourceLeaf = mSourceLeafManager.leaf(sourceLeafIndex);
+                const auto& sourceArray = sourceLeaf.constAttributeArray(mAttributeIndex);
```
Idclip: Would you mind adding a comment to constAttributeArray() and attributeArray() which clarify the copying involved if shared? Not sure if the arrays were shared in this scenario or if this is just a bit of const correctness, but I've ran into this issues a few times too!

danrbailey: Sure, I'll put this change on the other PR:

"Attribute arrays use copy-on-write. When deep-copying a leaf node, the arrays will be shared across both leafs. On non-const access to an array, the data will then be deep-copied to ensure it is unique. Always use const arrays where possible to eliminate redundant copying."

Idclip: Great, thanks a lot

PR #352, opened 2019-02-08
OVDB-27: MeshToVolume uses infinite memory when given NANs

jmlait:
Check for NANs when computing distances to avoid infinite loops
with bad input geometry.

Rename the computeDistance to updateDistance to better reflect
its behaviour.

### DIFF
--- a/openvdb/tools/MeshToVolume.h
+++ b/openvdb/tools/MeshToVolume.h
@@ -2144,7 +2144,10 @@ private:
         ijk = Coord::floor(prim.a);
         coordList.push_back(ijk);
 
-        computeDistance(ijk, prim, data);
+        // The first point may not be quite in bounds, and rely
+        // on one of the neighbours to have the first valid seed,
+        // so we cannot early-exit here.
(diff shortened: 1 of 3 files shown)


openvdb/tools/MeshToVolume.h
```diff
@@ -2172,6 +2175,11 @@ class VoxelizePolygons
         const ValueType dist = ValueType((voxelCenter -
             closestPointOnTriangleToPoint(prim.a, prim.c, prim.b, voxelCenter, uvw)).lengthSqr());
 
+        // Either the points may be NAN, or they could be far enough from
+        // the origin that computing distance fails.
+        if (std::isnan(dist))
```
pcucka: Would it not be preferable to use "`!std::isfinite(dist)`", to handle the case of negative infinity?

jmlait:
> LGTM, but please add an entry to the change logs.

Added entry to change logs.  Immediately regretted that we have a single conflict-prone change log :>

PR #353, opened 2019-02-09
Synchronize with Houdini 17.5's version.

jmlait:
Improved topology merging.  Expansion to bool grids for topology
operations.

Note that there are no longer non-AMD64 platforms, and that is placeholder
for 32bit.  So the atomic counter is effectively 64bit.`

### DIFF
--- a/openvdb_houdini/houdini/GEO_PrimVDB.cc
+++ b/openvdb_houdini/houdini/GEO_PrimVDB.cc
@@ -1928,6 +1928,30 @@ geoIntersect(GridTypeA& grid_a, const GridTypeB &grid_b)
     }
 }
 
+/// This class is used as a functor to set inactive voxels to the background
+/// value.
+template<typename GridType>
+class geoInactiveToBackground
+{
+public:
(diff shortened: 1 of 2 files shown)


openvdb_houdini/houdini/GEO_PrimVDB.cc
```diff
@@ -1928,6 +1928,30 @@ geoIntersect(GridTypeA& grid_a, const GridTypeB &grid_b)
...
+
+    geoInactiveToBackground(const GridType& grid)
+    {
+        background = grid.background();
+    }
+
+    inline void operator()(const Iterator& iter) const
+    {
+        iter.setValue(background);
+        iter.setValueOff();
```
pcucka: This line isn't needed given that you're iterating over inactive values only.  It's also not strictly thread-safe, although that's probably OK in this case since you're not actually changing the values of any active masks.

jmlait: Thank you!  Not sure how that slipped in there. There was some concern that setValue() was implicitly activating; but it isn't.

PR #355, opened 2019-02-11
OVDB-31: Add ABI version to cmd -version output.

pcucka:


openvdb/CHANGES
```diff
@@ -12,6 +12,9 @@ Version 6.0.1 - In development
       attribute generation.
     - AttributeArray::copy() and the copy assignment operator for
       AttributeArrays are now thread-safe.
+    - The command-line tools (vdb_print, etc.) now include the library
+      ABI version (when it differs from the library major version)
+      in their -version output.
```
Idclip: It seems that OPENVDB_LIBRARY_ABI_VERSION_STRING is always going to contain the abi number even if it's the same as the lib major version? This comment suggests that it won't. If I've missed something, it would be good if --version always included the abi regardless of if it's the same as the library version so that dependant software has a reliable way of getting this value. Though perhaps a --abi flag would be more useful in the future to avoid having to cut the return of --version.

pcucka: I like the idea of adding a `vdb_config` command with options like `--cxxflags`, `--ldflags` and `--abi`.  I don't really want to add those options to any of the existing commands, though.

Idclip: Yeah that's not a problem, this change is a good one in the interim. Should "(when it differs from the library major version)" be removed?

pcucka: You're right, it should be removed.  Thanks for catching that!

PR #358, opened 2019-02-15
OVDB-46: Collapse openvdb/cmd/ subdirectories

pcucka:
- Renamed cmd source files and merged them into `openvdb/cmd/`.
- Updated makefile and CMake script.
- Added missing `vdb_lod` target to CMake script.


openvdb/Makefile
```diff
@@ -822,21 +823,21 @@ pdfdoc:
 	@echo "$@"': $$DOXYGEN is undefined'
 endif
 
-vdb_lod: $(LIBOPENVDB) cmd/openvdb_lod/main.cc
+vdb_lod: $(LIBOPENVDB) cmd/openvdb_lod.cc
```
danrbailey: I find it slightly obscure that "openvdb_lod.cc" produces "vdb_lod". Could we perhaps rename to vdb_lod.cc so at least the names are consistent (and likewise for all the other binaries)?

pcucka: I guess, though "openvdb" is the prefix used for everything else.  The reason I didn't name the commands `openvdb_print`, `openvdb_view`, etc. is that I figured users would rebel if they had to type those extra four characters all the time.

Idclip: We could build openvdb_* binaries and also install symlinks to their vdb_* aliases

openvdb/CMakeLists.txt
```diff
@@ -215,7 +215,25 @@ SET_TARGET_PROPERTIES(
   VERSION ${OPENVDB_MAJOR_VERSION_NUMBER}.${OPENVDB_MINOR_VERSION_NUMBER}.${OPENVDB_PATCH_VERSION_NUMBER}
   )
 
-SET ( VDB_PRINT_SOURCE_FILES  cmd/openvdb_print/main.cc )
+SET ( VDB_LOD_SOURCE_FILES  cmd/openvdb_lod.cc )
+SET_SOURCE_FILES_PROPERTIES ( ${VDB_LOD_SOURCE_FILES}
+  PROPERTIES
+  COMPILE_FLAGS "-DOPENVDB_USE_BLOSC"
+  )
```
Idclip: Correct me if I'm wrong, but OPENVDB_USE_BLOSC is purely for the objects built by the core library and isn't required anymore for dependant builds (except for the write SOP which uses it to introduce a blosc mode)? Or is that just happenstance and we want to continue to make sure people are defining this in-case it makes its way back into a header

pcucka: You're not wrong, but all of the other executables get that flag, so it's possible that CMake requires it to be set at that level.  Then again, all of the other executables get `OPENVDB_USE_GLFW_FLAG`, which is needed only when building the viewer, so maybe not.

Idclip: I'm fairly certain it's not needed, but I don't mind it staying for now to match the rest of the CMake until it's cleaned up

openvdb/Makefile
pcucka: I'm going to punt on the file naming issue for now.  Let's revisit it after the CMake switchover.

PR #359, opened 2019-02-16
improved util::CpuTimer and added til::printTime

kmuseth:
1) Added util::printTime that outputs nicely-formatted time information. 

2) The API of util::CpuTimer has been improved and is now making use
     a new util::printTime function for nicer output of time information.


openvdb/util/CpuTimer.h
```diff
@@ -56,65 +57,130 @@ namespace util {
...
-        std::cerr << ostr.str();
+        const double msec = this->milliseconds();
+        printTime(mOutStream, msec, " completed in ", "\n", true, 4, 1);
+        return msec;
     }
 
-private:
+    /// @brief Returns and prints time in milliseconds since construction or start was called.
+    ///
+    /// @note Combine this method with start() to delay output to will completion of task being timed.
```
danrbailey: typo (though I'm not sure what this is intended to be) - "delay output to will completion of task being timed"

openvdb/util/Formats.cc
```diff
@@ -112,10 +112,53 @@ printNumber(std::ostream& os, uint64_t number,
...
+    int group = 0;
+
+    // Write to a string stream so that I/O manipulators like
+    // std::setprecision() don't alter the output stream.
+    std::ostringstream ostr;
+    ostr << head;
+    ostr << std::setprecision(precision) << std::setiosflags(std::ios::fixed);
+
+    double msec = milliseconds;
+    if (msec >= 1000.0) {// avoid unnecessary coverheads
```
danrbailey: coverheads -> overheads

openvdb/util/CpuTimer.h
```diff
@@ -56,65 +57,130 @@ namespace util {
...
     {
-        this->stop();
-        this->start(msg);
+        tbb::tick_count::interval_t dt = tbb::tick_count::now() - mT0;
+        return 1000.0*dt.seconds();
     }
 
-    /// Return Time diference in milliseconds since construction or start was called.
-    inline double delta() const
+    /// @brief This method is identical to milliseconds() - will soon be deprecated
```
danrbailey: Could we perhaps mark it deprecated now?

openvdb/util/CpuTimer.h
```diff
@@ -56,65 +57,130 @@ namespace util {
...
-    /// @note Should normally be followed by a call to stop()
-    inline void restart(const std::string& msg)
+    /// @note Combine this method with start() to get timing without any outputs.
+    inline double milliseconds() const
     {
-        this->stop();
-        this->start(msg);
+        tbb::tick_count::interval_t dt = tbb::tick_count::now() - mT0;
+        return 1000.0*dt.seconds();
     }
```
Idclip: Any chance we could add a seconds() method to this class too? I'm using this in a few places where the time scale is much larger so end up having to multiple delta() by 1000 everywhere. Could go even further and introduce something like an enum which defines the return scale:

```
inline double detla(const Precision p = MILLISECONDES) {
  tbb::tick_count::interval_t dt = tbb::tick_count::now() - mT0;
  double s = dt.seconds();
  switch (p) {
    case MILLISECONDES : return s*1000.0;
    // etc
   }
}
```

But if not, something which just returns dt.seconds() would be great

openvdb/util/Formats.cc
```diff
@@ -112,10 +112,53 @@ printNumber(std::ostream& os, uint64_t number,
...
+        group = 3;
+      } else if (minutes) {
+        ostr << minutes << " minutes, " << seconds << " seconds, and ";
+        group = 2;
+      } else if (seconds) {
+        ostr << seconds << " seconds and ";
+        group = 1;
+      }
+    }
+    ostr << std::setw(width) << msec << " milliseconds";
```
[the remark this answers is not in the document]
pcucka: I suppose there are situations where output as verbose as "1 hour, 10 minutes, and 20.5 seconds (4220500.0)" (48 characters in all) is welcome, but I would prefer to at least have the option of the much more compact "1h 10m 20.5s".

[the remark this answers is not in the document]
kmuseth: adding an option for both a long and compact output format

openvdb/util/Formats.cc
```diff
@@ -112,10 +112,53 @@ printNumber(std::ostream& os, uint64_t number,
...
+      const uint32_t days    = static_cast<uint32_t>(msec / (1000.0*60*60*24));
+      msec -= (seconds + (minutes + (hours + days * 24) * 60) * 60) * 1000.0;
+      if (days) {
+        ostr << days << " days, " << hours << " hours, " << minutes << " minutes, " << seconds << " seconds, and ";
+        group = 4;
+      } else if (hours) {
+        ostr << hours << " hours, " << minutes << " minutes, " << seconds << " seconds, and ";
+        group = 3;
+      } else if (minutes) {
+        ostr << minutes << " minutes, " << seconds << " seconds, and ";
```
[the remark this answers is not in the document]
kmuseth: fixed

openvdb/util/CpuTimer.h
kmuseth: will do

openvdb/util/CpuTimer.h
kmuseth: added

kmuseth: I'm deprecating the delta() method in favor of a seconds() and miliseconds() method

PR #354, opened 2019-02-10
Remove trailing whitespace from cmake files

zxiiro:

PR #240, opened 2018-06-15
Windows Updates & Appveyor Config

chrisheckey:
Updates to make building on windows a bit less painful. 

- Assumes vcpkg as the source of dependencies, as it seems to be the easiest way to build openexr, ilmbase, blosc, glew, glfw, and zlib on windows. 
- vdb_viewer now uses GLEW, along with the maya plugin. 
- An appveyor.yml is also included for windows ci

I'm no expert with cmake or windows native development, so any feedback on these changes is appreciated.

Example of the Appveyor in action is here: https://ci.appveyor.com/project/chrisheckey/openvdb

Signed-off-by: Christopher Heckey <christopher.heckey@dreamworks.com>

PR #362, opened 2019-02-20
CMake Improvements

Idclip:
This is a whole batch of CMake changes that we've made at DNEG

All
 - Formatting, functions upper case, white space (#354), moving various code to more appropriate locations
 - Copyright dates

Root CMake
 - Specifically disable in source builds
 - More controls for default behaviour for OPENVDB_ABI_VERSION_NUMBER
 - Moved OpenVDB version detection from core CMake to root CMake
 - Fixed some CMake warnings

Core CMake
 - Removed Boost Python as a requirements of the Core lib
 - Removed directory install command in favour of specific header file installations (resolving recursive directory issues)
 - Removed unnecessary binary compile flags for view, render and print
 - Fixed vdb_view linking issues
 
Python CMake

 - Changed python version requirement from boost (this may not be correct)
 - Added missing pyPointGrid.cc

FindHDK CMake

 - Fixed $HFS not being respected
 - Changed DMAKING_DSO to a target definition, not a global definition

FindOpenEXR CMake

 - Fixed OPENEXR_ROOT not being respected 

Houdini CMake

 - Moved OPENVDB_HFS_INSTALL_BASE_DIR to top of file
 - Added missing houdini/GEO_VDBTranslator.cc to openvdb_houdini lib
 - Improved SOP list formatting
 - Added VRAY_OpenVDB_Points target
 - Added install rules for icons
 - Changed the behaviour of OPENVDB_HOUDINI_INSTALL_LIBRARY to install to CMAKE_INSTALL_PREFIX


Idclip:
This is more for discussion vs being a ready/viable merge candidate. The main things that I need feedback on are the boost python changes and the Houdini CMake changes. We still don't use the Houdini CMake in production, but this should work. We run the CMake twice, once for the core lib:

```
"CMAKE_BUILD_TYPE": Release
"OPENVDB_ABI_VERSION_NUMBER": <>,
"MINIMUM_BOOST_VERSION": "1.53",
"ILMBASE_NAMESPACE_VERSIONING": "OFF",
"OPENEXR_NAMESPACE_VERSIONING": "OFF",
"OPENVDB_BUILD_DOCS": "ON",
"OPENVDB_BUILD_PYTHON_MODULE": "ON",
"PYOPENVDB_INSTALL_DIRECTORY": "lib/python",
"OPENVDB_BUILD_HOUDINI_SOPS": "OFF",
"OPENVDB_BUILD_MAYA_PLUGIN": "OFF",
"OPENVDB_BUILD_UNITTESTS": "ON",
"OPENVDB_BUILD_CORE": "ON",
"USE_GLFW3": "ON",
"GLFW3_USE_STATIC_LIBS": "ON"
```

and once for the houdini lib

```
"OPENVDB_ABI_VERSION_NUMBER": <>,
"MINIMUM_BOOST_VERSION": "1.53",
"ILMBASE_NAMESPACE_VERSIONING": "OFF",
"OPENEXR_NAMESPACE_VERSIONING": "OFF",
"OPENVDB_BUILD_HOUDINI_SOPS": "ON",
"OPENVDB_BUILD_MAYA_PLUGIN": "OFF",
"OPENVDB_BUILD_UNITTESTS": "OFF",
"OPENVDB_BUILD_CORE": "OFF",
"HDK_LOCATION": <>
```

Some questions about the Houdini installation:

What are the desired range of controls for the houdini install path. Currently there are three CMake options for this, but we only use OPENVDB_HOUDINI_INSTALL_LIBRARY which we set to ON. Is there ever a case where we'd want OPENVDB_HOUDINI_INSTALL_LIBRARY to be off? Is it important to be able to automatically add the Houdini version to the Houdini install path? Our install dir tree looks like:

```
├── include
│   ├── houdini_utils
│   └── openvdb_houdini
└── lib
    └── houdini
        ├── config
        │   └── Icons
        ├── dso
        │   └── mantra
        ├── help
        │   └── nodes
        │       └── sop
        └── scripts
            └── sop
```

Regarding the "houdini_utils" folder, does this have to be generated in the CMake? The folder structure in the houdini plugin doesn't reflect the install header locations.

CMakeLists.txt
```diff
@@ -26,58 +26,99 @@
...
+FIND_PROGRAM ( CCACHE_FOUND ccache )
+IF ( CCACHE_FOUND )
+    SET_PROPERTY ( GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache )
+    SET_PROPERTY ( GLOBAL PROPERTY RULE_LAUNCH_LINK ccache )
+ENDIF ( CCACHE_FOUND )
 
 OPTION ( OPENVDB_BUILD_UNITTESTS "Build the OpenVDB unit tests" ON )
 OPTION ( OPENVDB_BUILD_DOCS "Build the OpenVDB documentation" OFF )
-OPTION ( OPENVDB_BUILD_PYTHON_MODULE "Build the pyopenvdb Python module" ON )
+OPTION ( OPENVDB_BUILD_PYTHON_MODULE "Build the pyopenvdb Python module" OFF )
```
Idclip: I'll revert this

CMakeLists.txt
```diff
@@ -26,58 +26,99 @@
...
 
-SET ( OPENVDB_ABI_VERSION_NUMBER "" CACHE STRING "build for compatibility
-with version N of the OpenVDB Grid ABI, where N is 2, 3, 4, etc.
-(some newer features will be disabled)")
+SET ( OPENVDB_ABI_VERSION_NUMBER "" CACHE STRING "Build for compatibility with version N of the
+    OpenVDB Grid ABI, where N is 3, 4, 5 etc. (some newer features will be disabled)")
 
-IF (OPENVDB_BUILD_HOUDINI_SOPS)
+IF ( OPENVDB_BUILD_HOUDINI_SOPS )
+  # @todo why?
```
pcucka: Possibly because, prior to the introduction of HBoost, we built against Houdini's version of Boost, and Houdini didn't ship with Boost.Python.

CMakeLists.txt
```diff
@@ -26,58 +26,99 @@
...
   SET ( OPENVDB_BUILD_PYTHON_MODULE OFF )
 ENDIF ()
 
+# Some platform checks
+
 IF ( APPLE )
   IF ( OPENVDB_BUILD_HOUDINI_SOPS AND OPENVDB_BUILD_MAYA_PLUGIN )
-	MESSAGE ( FATAL_ERROR "On OSX, Houdini linked with libc++ whereas Maya is linked with libstdc++, the different runtimes are not mutually compatible, build them Houdini SOPs and Maya plugins separately with the requisite DCC's compiler" )
+  MESSAGE ( FATAL_ERROR "On OSX, Houdini linked with libc++ whereas Maya is linked with libstdc++, "
```
pcucka: Might as well clean this up a little:
"On macOS, Houdini is linked with libc++ whereas Maya is linked with libstdc++.  The different runtimes are not mutually compatible, so build the Houdini and Maya plugins separately with the requisite DCC's compiler."

CMakeLists.txt
```diff
@@ -26,58 +26,99 @@
...
+# Determine the OpenVDB ABI Version. If OPENVDB_ABI_VERSION_NUMBER is defined, use that value,
+# otherwise check the legacy OPENVDB_ENABLE_3_ABI_COMPATIBLE option, falling back to the major
+# library version number
+MESSAGE ( STATUS "Using OpenVDB ABI Version ${OPENVDB_ABI_VERSION_NUMBER}" )
+
+IF ( OPENVDB_ABI_VERSION_NUMBER STREQUAL "" )
+  IF ( OPENVDB_ENABLE_3_ABI_COMPATIBLE )
+    SET ( OPENVDB_ABI_VERSION_NUMBER 3 )
+    ADD_DEFINITIONS ( -DOPENVDB_3_ABI_COMPATIBLE )
+    MESSAGE ( DEPRECATION "OPENVDB_ENABLE_3_ABI_COMPATIBLE is deprecated. Instead use
```
pcucka: Although `OPENVDB_ENABLE_3_ABI_COMPATIBLE` is deprecated, until it is removed it would probably be best to add `-DOPENVDB_3_ABI_COMPATIBLE` if `OPENVDB_ABI_VERSION_NUMBER` is 3.

CMakeLists.txt
Idclip: Ah makes sense, thanks - added as a comment

CMakeLists.txt
Idclip: Done

CMakeLists.txt
Idclip: Done

danrbailey:
Thanks for this Nick, it's great to get an idea of what changes you're proposing, but hard to see quite how to progress with this PR when there are so many changes. I'd rather this doesn't end up sitting open for a long time. 

I suggest that once @zxiiro has made your suggested edit to (#354), we should merge that PR and then consider doing a reformatting pass before discussing further changes to the CMake codebase.

I strongly recommend we align ourselves with typical conventions (http://docs.ros.org/jade/api/catkin/html/user_guide/standards.html) and with the conventions that OpenColorIO follows (https://github.com/imageworks/OpenColorIO/blob/master/CMakeLists.txt). 

That means this style:

cmake_minimum_required(VERSION 3.1)

instead of this:

CMAKE_MINIMUM_REQUIRED ( VERSION 3.1 )

Idclip:
Formatting aside, this contains a number of fixes which should start to make it a viable candidate for CI. There are much larger structural changes and general CMake practises which are still outstanding here, but imo once the python and Houdini build system questions are resolved, there's no reason why we can't start putting together a travis build for this. This should be the priority, allowing us to make additional CMake changes (standards/structural conversations outstanding) with the CI guarantee.

danrbailey:
I totally agree. I suggest a migration plan that looks like this. We transition Travis by introducing two new CMake jobs - one that does all the unit tests, python binaries, etc for ABI=6 and one that does H17 with ABI=5. Then we use that as the basis for making formatting and other larger structural changes, then finally invert the situation - ie leave just two Make-based Travis jobs and turn the rest into CMake.

Perhaps this PR should just be the minimum set of changes needed to get a Travis CMake job up and running for now?

Idclip:
I think I'll have a go at updating travis to use CMake in this PR to demonstrate the how it's currently being used. Should make it much easier to evaluate these changes

Idclip:
Also note that a bunch of further improvements exist here #240, including boost python version options, cmake options for including binaries and switching shard/static library building on/off

Idclip:
Declining this in favour of a new PR which introduces cmake to travis along with all required cmake updates

PR #363, opened 2019-02-20
Limit the scope of the acquired const grid pointers.  

jmlait:
Limit the scope of the acquired const grid pointers.  This will 
cause the grid to not be unique, and trigger asserts as the meta
data is unexpectedly shared.


Idclip:
Nice, worthy of a change log update?

jmlait:
I think it is a bit fine-grained for a change log update?  There is no functional change.

Idclip:
Yeah sure, it's fairly minor though could have noticeable performance improvements for larger PointDataGrids. I don't know where we want to draw the line - I don't think there's any harm with the changelog being all encompassing for any performant/behavioural changes (i.e. improvements which in any way change the data the user will be receiving). I personally kind of prefer it, but maybe that's just because we don't have any other mechanism for storing descriptions of more granular changes, aside from the git history which I don't really think of as an appropriate solution. Anyway, happy for this to go in without an update

PR #364, opened 2019-02-20
Fix a rotation order bug in tools::local_util::decompose().

pcucka:


danrbailey:
Closing. Fixed conflicts and re-submitted as #579 to resolve CLA issues.

PR #365, opened 2019-02-20
Remove trailing whitespace from cmake files

zxiiro:


zxiiro:
Sorry for creating a new PR from #354 I'm not sure why but when I rebased and updated my change it didn't automatically update the old PR.

PR #367, opened 2019-02-28
OVDB-42: TypeList-based alternative to UTvdbProcessTypedGrid

pcucka:
This PR introduces a generic type list class to the core library, along with a `GridBase::apply<TypeList, Op>` method that can be used in place of the `UTvdbProcessTypedGrid` family of functions.  The latter are not obsolete, since they are implemented using a `UT_VDBType` switch, whereas `GridBase::apply` requires string comparisons, but the performance difference is probably insignificant for the typical use case of iterating over input primitives.  A possible improvement, though, would be to add a `UTvdbProcessTypedGrid` variant that accepts a `std::initializer_list<UT_VDBType>` or `std::vector<UT_VDBType>` argument with the allowed types.

I updated the Fill SOP to use `GridBase::apply` for the sake of illustration.  Other SOPs can be updated as needed.


openvdb_houdini/houdini/Utils.h
```diff
@@ -305,10 +305,45 @@ bool isLogForwarding(OP_OpTypeId);
...
+using Vec3GridTypes = openvdb::TypeList<
+    openvdb::Vec3SGrid,
+    openvdb::Vec3DGrid,
+    openvdb::Vec3IGrid>;
+
+using PointGridTypes = openvdb::TypeList<
+    openvdb::tools::PointIndexGrid,
+    openvdb::points::PointDataGrid>;
+
+using NonPointGridTypes = ScalarGridTypes::Append<Vec3GridTypes>;
```
danrbailey: I think "VolumeGridTypes" would be better than "NonPointGridTypes".

openvdb_houdini/houdini/Utils.h
```diff
@@ -305,10 +305,45 @@ bool isLogForwarding(OP_OpTypeId);
 /// @}
 
 
+////////////////////////////////////////
+
+
+// Grid type lists, for use with openvdb::GridBase::apply()
+
+using ScalarGridTypes = openvdb::TypeList<
+    openvdb::BoolGrid,
```
danrbailey: No MaskGrid?

openvdb_houdini/houdini/Utils.h
```diff
@@ -305,10 +305,45 @@ bool isLogForwarding(OP_OpTypeId);
...
+    openvdb::FloatGrid,
+    openvdb::DoubleGrid>;
+
+using Vec3GridTypes = openvdb::TypeList<
+    openvdb::Vec3SGrid,
+    openvdb::Vec3DGrid,
+    openvdb::Vec3IGrid>;
+
+using PointGridTypes = openvdb::TypeList<
+    openvdb::tools::PointIndexGrid,
```
danrbailey: I'm not sure there are many situations where an operation would apply to a PointDataGrid *and* a PointIndexGrid. I think I'd prefer "PointGridTypes" to be used just for a PointDataGrid. Also, not sure if it's relevant but PointIndexGrid isn't exposed in Houdini.

openvdb_houdini/houdini/Utils.h
```diff
@@ -305,10 +305,45 @@ bool isLogForwarding(OP_OpTypeId);
...
+    openvdb::points::PointDataGrid>;
+
+using NonPointGridTypes = ScalarGridTypes::Append<Vec3GridTypes>;
+
+using AllGridTypes = NonPointGridTypes::Append<PointGridTypes>;
+
```
danrbailey:
```
MaskGridTypes = openvdb::TypeList<
    openvdb::MaskGrid,
    openvdb::BoolGrid>;
```

?

openvdb_houdini/houdini/Utils.h
pcucka: Yes, that's better.

openvdb_houdini/houdini/Utils.h
pcucka: It's not one of the supported Houdini grid types.

openvdb_houdini/houdini/Utils.h
pcucka: `UT_VDB_POINTINDEX` is actually one of the supported grid types, though it's true that no SOPs output PointIndexGrids, and they're unlikely to be passed around by base class pointer.

danrbailey: Yes, that's true.

openvdb_houdini/houdini/Utils.h
danrbailey: Ah, I missed that this is in houdini/Utils.h. I've actually had a request to use a method like this outside of Houdini, so I think these import statements should live in the core library and be non-Houdini specific. 

Perhaps they could be defined in an `openvdb::` namespace based on the grid types being registered and then redefined in an `openvdb_houdini::` namespace but with any grids Houdini doesn't support pulled out?

```
namespace openvdb_houdini {
using ScalarGridTypes = openvdb::ScalarGridTypes::Remove<MaskGrid>;
using AllGridTypes = openvdb::AllGridTypes::Remove<MaskGrid, StringGrid>;
}
```

pcucka: I intentionally left the type lists out of the core library so as not to bless certain grid types (any more than we already have) and give the impression that core library code, at least, need not support other types or other tree configurations.

openvdb/Grid.h
```diff
@@ -1619,11 +1707,53 @@ createLevelSet(Real voxelSize, Real halfWidth)
...
+/// @private
+template<typename OpT, typename GridBaseT, typename T, typename ...Ts>
+struct GridApplyImpl { bool operator()(GridBaseT&, OpT&) const { return false; } };
+
+// Partial specialization for (nonempty) TypeSets
+/// @private
+template<typename OpT, typename GridBaseT, typename GridT, typename ...GridTs>
+struct GridApplyImpl<OpT, GridBaseT, TypeList<GridT, GridTs...>>
+{
+    bool operator()(GridBaseT& grid, OpT& op) const
```
Idclip: could make this a non operator so it doesn't require instances of GridApplyImpl. v minor though

openvdb/Types.h
```diff
@@ -337,6 +337,167 @@ struct CanConvertType<ValueMask, T> { enum {value = CanConvertType<bool, T>::val
...
+struct TSRemoveImpl<ListT, TypeList<Ts...>> {
+    using type = typename TSRemoveImpl<ListT, Ts...>::type;
+};
+
+
+template<typename OpT> inline void TSForEachImpl(OpT) {}
+template<typename OpT, typename T, typename... Ts>
+inline void TSForEachImpl(OpT op) { op(T()); TSForEachImpl<OpT, Ts...>(op); }
+
+} // namespace internal
```
Idclip: These could go all into TypeList instead of an internal namespace (and could then become private). Depends how you feel about nested structs

pcucka: ICC 15 (at least) seems to have an issue with parameter pack expansion in nested, templated classes.  I don't see a good way to work around the problem, so I'll leave this as is for now.

PR #370, opened 2019-03-02
OVDB-62: Rename VDB SOPs

danrbailey:


openvdb_houdini/houdini/ParmFactory.h
```diff
@@ -532,6 +532,11 @@ class OPENVDB_HOUDINI_API OpPolicy
     /// factory.@link OpFactory::english() english()@endlink.
     virtual std::string getName(const OpFactory& factory, const std::string& english);
 
+    /// @brief Return a label name for the operator defined by the given factory.
+    /// @details In this base class implementation, this method simply returns
+    /// factory.@link OpFactory::english() english()@endlink.
+    virtual std::string getLabelName(const OpFactory&);
```
pcucka: I don't anticipate any issues with this, but it's safest to add new virtual functions to the end of the class.

openvdb_houdini/houdini/SOP_NodeVDB.cc
```diff
@@ -588,23 +587,42 @@ class SESIOpenVDBOpPolicy: public houdini_utils::OpPolicy
 };
 
 
-/// @brief OpPolicy for OpenVDB operator types at DWA
-class DWAOpenVDBOpPolicy: public houdini_utils::DWAOpPolicy
+/// @brief Default OpPolicy for OpenVDB operator types
```
pcucka: Shouldn't this say "SideFX OpPolicy for OpenVDB operator types"?

openvdb_houdini/houdini/SOP_OpenVDB_Advect.cc
```diff
@@ -295,11 +295,11 @@ newSopOperator(OP_OperatorTable* table)
         .setDefault(temporalIntegrationSchemeToString(TemporalIntegrationScheme(0))));
 
     // Register this operator.
-    hvdb::OpenVDBOpFactory("OpenVDB Advect",
+    hvdb::OpenVDBOpFactory("VDB Advect",
         SOP_OpenVDB_Advect::factory, parms, *table)
         .setObsoleteParms(obsoleteParms)
-        .addAlias("OpenVDB Advect Level Set")
```
pcucka: Just a thought: we could probably safely remove all of these aliases now.

openvdb_houdini/houdini/SOP_OpenVDB_Convert.cc
```diff
@@ -427,8 +427,11 @@ Polygon Soup:\n\
     obsoleteParms.add(hutil::ParmFactory(PRM_INT_J, "activepart", ""));
 
     // Register this operator.
-    hvdb::OpenVDBOpFactory("OpenVDB Convert",
+    hvdb::OpenVDBOpFactory("Convert VDB",
```
pcucka: I don't suppose we could leave this as "VDB Convert", so that it sorts with all the other SOPs?

openvdb_houdini/houdini/SOP_NodeVDB.cc
```diff
@@ -588,23 +587,42 @@ class SESIOpenVDBOpPolicy: public houdini_utils::OpPolicy
...
+class ASWFOpenVDBOpPolicy: public DefaultOpenVDBOpPolicy
+{
+public:
+    std::string getName(const houdini_utils::OpFactory& factory, const std::string& english) override
+    {
+        std::string name = DefaultOpenVDBOpPolicy::getName(factory, english);
+        return "DW_Open" + name;
+    }
+
+    std::string getLabelName(const houdini_utils::OpFactory& factory) override
```
pcucka: For me, at least, GCC 4.8 complains about this:
```
openvdb_houdini/houdini/SOP_NodeVDB.cc:615:17: error: ‘std::string openvdb_houdini::{anonymous}::ASWFOpenVDBOpPolicy::getLabelName(const OpFactory&)’ marked override, but does not override
     std::string getLabelName(const OpFactory& factory) override
                 ^
```
That's nonsense, but I don't know what fix to suggest.

openvdb_houdini/houdini/SOP_OpenVDB_Convert.cc
danrbailey: I think I'd prefer that we align ourselves with Houdini's names as much as possible. Most Houdini users are more familiar with "Convert VDB", so we're breaking with their expectation by naming it "VDB Convert" instead. 

As I see it working, when we add a new SOP to the library, @jmlait decides if it's a good fit for Houdini and then chooses whether to change the label name or not. If we disagree with a rename, we can lobby Jeff to try and change his mind, but ultimately if he decides to rename, we should adopt the new name as the open-source name as well. Ideally this whole process should happen before the PR with the new SOP gets merged so that no-one even gets to see the original proposed name.

There is of course some legacy to resolve, but unless there's a very good reason to do so, I think the renames that have already been decided on by SideFX are the names we should migrate to.

pcucka: What about VDB Create, then?  Should that be renamed to "VDB"?

danrbailey: No, unfortunately there's not a 1:1 in functionality between the VDB SOP and the VDB Create SOP.

The VDB SOP allows you to build a frustum grid without needing a camera, manually specifying taper, depth, etc. That's pretty useful. The VDB Create SOP doesn't offer this.

The VDB Create SOP has a "match topology" option which activates topology matching a reference VDB on creation. This workflow is possible using native VDB SOPs, but it's more fiddly. You have to follow the VDB SOP with a VDB Activate SOP, switch to the reference tab, enable the "activate using VDB" option and then deselect the value toggle. 

Nonetheless, I believe most users are more comfortable with the native SOP workflow, so if SideFX are willing to donate them, I think we should adopt the SideFX SOPs here (VDB SOP and VDB Activate SOP) and then deprecate (ie ophide) the VDB Create SOP rather than simply renaming it.

pcucka: >   No, unfortunately there's not a 1:1 in functionality between the VDB SOP and the VDB Create SOP.

Likewise for Convert VDB and Convert VDB (ASWF), though.

danrbailey: > Likewise for Convert VDB and Convert VDB (ASWF), though.

What are the differences in functionality between these two? I've just been through the options on these two SOPs and the parameters, node inputs and default values look identical to me? (testing on 17.0.456).

pcucka: Through 16.5, the value type conversion options are missing.  If they've been added in 17, then the docs are out of date.

danrbailey: Yeah, it seems like those parameters have been added to the Convert VDB SOP in H17 ("vdbtype", "vdbprecision"), but the Houdini help card hasn't been updated (@jmlait).

openvdb_houdini/houdini/SOP_OpenVDB_Filter_Level_Set.cc
```diff
@@ -635,7 +635,7 @@ and usage examples.\n");
 
         } else if (OP_TYPE_RESHAPE == op) {
 
-            hvdb::OpenVDBOpFactory("OpenVDB Offset Level Set",
+            hvdb::OpenVDBOpFactory("VDB Offset Level Set",
```
pcucka: How about this one?  VDB Offset Level Set corresponds to VDB Reshape SDF, except that the latter lacks the new Trim option.  Should this be renamed now, or only if and when the native SOP gets the new option?

danrbailey: I think we rename now. It'll always be the case that we add new features to the open-source SOPs first, then they propagate to the native SOPs in time. As long as the open-source SOPs contain at minimum all of the functionality that the native SOPs do and that the UI, inputs and parameters are not different enough between the two that it confuses people then I think this is perfectly acceptable. 

I realize there's actually a fair few more of these names that can change, I just put a list up on the JIRA - OVDB-62.

openvdb_houdini/houdini/SOP_OpenVDB_Convert.cc
dokipen3d: Just my thoughts on this. I was going to suggest changing the Convert VDB to VDB Convert mainly because it would be nice to keep them all the same and have them sorted alphabetically in the node list in houdini. The way it is now, they appear at the bottom of the list when you type VDB as the filter.

Idclip:
I still think that the suffix renaming is more trouble than it's worth - I don't think users really need to know from the label whether the SOP comes from the ASWF repository or not. 

With OpenVDB->VDB, I'm fairly indifferent on the interlacing of the SOPs, though if both SE and ASWF SOPs are visible I personally prefer them to be entirely separate - it makes it much easier to see and select the desired tab item when typing something generic like "vdb". Also, can I just double check that someone has tested this does interlace the SOPs? My Houdini config is setup to show SOPs in different tab menu categories completely separately, so I'm wondering if this renaming would actually interlace them. Otherwise we'd also need to add them to the VDB menu.

We discussed implementing policies for hiding SE/ASWF nodes, so the interlacing point becomes less relevant anyway? This seems like something which could be implemented before this and may make this change redundant.

This being said... Instead of a blanket renaming, I'm wondering if there's a way we can leverage Houdini's node versioning system to allow users to switch between the native and ASWF SOPs. Something like prefix all aswf nodes with a ::aswf version. Ideally only one copy of each node should ever exist in the tab menu (which one gets prioritised should be decided at load) but then users could change the version of the node being loaded using the asset bar instead of the policy/ophide stuff.

dnegmw:
Chiming in as I've just caught wind of this and I agree with @Idclip however I'm less indifferent about the interlaced entries as I think that would be confusing/frustrating and easy to misclick and I think should be avoided even as an interim solution.

danrbailey:
@idclip @dnegmw - if you haven't seen it already, there's an open discussion from about two weeks ago between @jmlait and myself about different approaches we could take here for a bit more context (namespacing is briefly considered there):

https://lists.aswf.io/g/openvdb-dev/topic/proposal_unifying_vdb_and/29915127

Adding the (ASWF) suffix I don't particularly have strong opinions about, but it is perhaps still somewhat useful to be able to easily distinguish where this node came from just in terms of being able to report issues.

I have been testing the interlacing of nodes for the last few days and it's great, so much clearer and easier to use. I'll give you a very specific example - you have a mixture of native and open-source nodes deployed and you're looking for a node that does some kind of culling or pruning, but you can't exactly remember the name. You either tab and type some example phrases "cull", "prune", etc and hope you've got it close enough to find it or you just type "VDB" and scan through the list to find it visually. In the second scenario, rather than going through the list one-by-one until you find it, you probably still look for the names that you have in your head first and only resort to an exhaustive search if you don't find it. There's no doubt in my mind that having one consolidated, alphabetical list is better than two disjoint, alphabetical lists. For most users, the least likely piece of information they'll retain is whether the node is native or open-source (that's probably not the case for you Nick!). The interlacing and the scenario I've outlined is still very relevant when talking about hiding nodes through policy.

I also want to highlight that we're only deciding on a default naming scheme for ASWF nodes. For a company that is installing the open-source nodes for the very first time, what should happen? If you have already been installing these nodes for many years and your users have become used to how things are named and organised, this is more a discussion about whether to go through the potential disruption of changing the names or not. In this case, there's always the option to provide your own company policy, it's as simple as adding this to SOP_NodeVDB.cc:

```
class MyCompanyOpenVDBOpPolicy: public ASWFOpenVDBOpPolicy
{
public:
    std::string getLabelName(const houdini_utils::OpFactory& factory) override
    {
        return "Open" + factory.english();
    }
};

using OpenVDBOpPolicy = MyCompanyOpenVDBOpPolicy;
```

openvdb_houdini/houdini/SOP_OpenVDB_Convert.cc
danrbailey: @dokipen3d - I don't disagree with you. However, SideFX has obviously renamed it so that the "Convert VDB" SOP fits in more consistently with the "Convert" SOP. Unless there's a very good reason and I consider your reasoning here as being more of a nice-to-have, I think we should align ourselves with Houdini naming. 

Two immediate benefits of this - an artist that moves from company X where the open-source nodes are not installed to company Y where they are won't have to think twice about where to find this SOP. Also a company where the open-source nodes are suddenly switched on where they weren't before will have less of a jarring experience in making that transition. :)

openvdb_houdini/houdini/ParmFactory.h
danrbailey: Fixed.

openvdb_houdini/houdini/SOP_NodeVDB.cc
danrbailey: Yes, well spotted, fixed.

openvdb_houdini/houdini/SOP_OpenVDB_Advect.cc
danrbailey: Done!

openvdb_houdini/houdini/SOP_OpenVDB_Convert.cc
dokipen3d: @danrbailey yeah good points. One option could be for sidefx to wrap up the convert functionality into the Convert SOP and deprecate Convert VDB. Old scenes could still open but newly created nodes could be the possible VDB Convert. I don't think this should be done now but something to think about for the future. I've just been over the shoulders of many artists who try to drop down a convert and there's always a pause and a mental jump to the fact that its not VDB _X._

jmlait: Convert SOP is a giant kitchen sink, so adding the various VDB specific choices to it is a bit of a nightmare, which is why we were happy with a separate Convert VDB.   The word-order doesn't matter for tab-typing as you can type either way, but it is true it shows up at the bottom.   We picked that one as we already have a class of "Convert Foo" nodes.

Good point about the missing docs on vdbtype and vdbprecision!  That was an oversight on my part.  Unfortunately I have to manually move the parameter-help into a separate .txt help card (as we have to have a single .txt help card for node help as we don't want to force technical writers to edit C++ files!)  Bug: 95181 for the missing help.

PR #373, opened 2019-03-06
Minor fixes for misc. compiler warnings

pcucka:

### DIFF
--- a/openvdb/cmd/openvdb_lod.cc
+++ b/openvdb/cmd/openvdb_lod.cc
@@ -176,7 +176,7 @@ mip(const GridType& inGrid, const Options& opts)
     }
 
     OPENVDB_LOG_INFO("processed grid \"" << inGrid.getName() << "\" in "
-        << std::setprecision(3) << (timer.delta() / 1000.0) << " sec");
+        << std::setprecision(3) << timer.seconds() << " sec");
 
     return outGrids;
 }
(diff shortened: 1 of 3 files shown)


openvdb/cmd/openvdb_lod.cc
```diff
@@ -176,7 +176,7 @@ mip(const GridType& inGrid, const Options& opts)
     }
 
     OPENVDB_LOG_INFO("processed grid \"" << inGrid.getName() << "\" in "
-        << std::setprecision(3) << (timer.delta() / 1000.0) << " sec");
```
pcucka: `CpuTimer::delta` is deprecated.

openvdb/python/pyOpenVDBModule.cc
```diff
@@ -229,7 +229,7 @@ struct MatConverter
             for (int j = 0; j < MatT::size; ++j) { rowObj.append(m(i, j)); }
             obj.append(rowObj);
         }
-        return obj;
+        return std::move(obj);
```
pcucka:
```
python/pyOpenVDBModule.cc:232:16: warning: local variable 'obj' will be copied despite being returned by name [-Wreturn-std-move]
        return obj;
python/pyOpenVDBModule.cc:232:16: note: call 'std::move' explicitly to avoid copying
```

PR #374, opened 2019-03-07
OVDB-67: Document OpenVDB Deprecation Strategy

danrbailey:


tsc/process/deprecation.md
```diff
@@ -0,0 +1,13 @@
+**Deprecation Strategy for OpenVDB**
+
+OpenVDB is committed to supporting the current and last two years of the VFX Reference Platform.
+
+For example, as of writing it is March 2019 which means VFX Reference Platform 2019, 2018 and 2017
+
+This infers the following support:
+
+* OpenVDB ABI=4, ABI=5 and ABI=6
+* Houdini 16.5, 17.0 and 17.5
```
pcucka: Houdini and Maya aren't actually mentioned in the VFX platform.  Do we need to be explicit about how we choose which app versions to support (i.e., by year of release)?

tsc/process/deprecation.md
```diff
@@ -0,0 +1,13 @@
...
+* C++11 and C++14
+
+In January 2020, OpenVDB will drop support for VFX Reference Platform 2017. This will mean deprecating support for Houdini 16.5, ABI=4 and C++11.
```
pcucka: Will we be deprecating that support, or eliminating it?

tsc/process/deprecation.md
```diff
@@ -0,0 +1,13 @@
+**Deprecation Strategy for OpenVDB**
+
+OpenVDB is committed to supporting the current and last two years of the VFX Reference Platform.
```
pcucka: Maybe "current and previous"?

tsc/process/deprecation.md
jmlait: I believe the intent is by whether they support a supported reference platform, rather than their release date.   So if 17.5 supports VFX reference 2018, then it will age out when VDB support for VFX 2018 ages out.

danrbailey: Exactly, I'll clarify this point.

tsc/process/deprecation.md
danrbailey: Eliminating is better, fixed!

tsc/process/deprecation.md
danrbailey: Fixed!

tsc/process/deprecation.md
```diff
@@ -0,0 +1,21 @@
...
+For example, as of writing it is March 2019 which means VFX Reference Platform
+2019, 2018 and 2017
+
+This infers the following support:
+
+* OpenVDB ABI=4, ABI=5 and ABI=6
+* C++11 and C++14
+* Houdini 16.5, 17.0 and 17.5
+
+Although Houdini versions are not included in the VFX Reference Platform, each
```
pcucka: Technically, we support Maya as well (though I haven't tried building for Maya 2019 yet; has anyone?), so perhaps we could collapse this into the above as follows:
```
We are committed to supporting the current year and previous two years of the
[VFX Reference Platform](http://www.vfxplatform.com/)
and all releases of Houdini and Maya based on those versions of the platform.

danrbailey: Your version is better, I've updated the wording here.

danrbailey:
> Do we want this to be in its own file, or is this the kind of thing that belongs in the README?
> 
> And how about the website? Does any of this information belong there?

In the context of "what do we support", probably the website or the README is a better place to put this information. In the context of "what is the policy of the TSC in dropping support for old library dependencies", I think this is the right place for this.

To be honest, I think the README could probably do with an overhaul. It wouldn't be a bad idea to have a section documenting how to get started with building the software on the landing page itself. That feels like a better place than the website for this kind of information and then what we support could naturally live there. Right now, I think it would feel a bit out-of-place putting it in the README or on the website so I suggest we go ahead with this PR and then perhaps get rid of it if and when we make changes there.

We should also include some error checking in the build system itself - checking the version of our dependencies are sufficiently late enough and reporting to the user otherwise. Right now, it'll probably just give you an unexpected error message and leave you to find the INSTALL file and figure out that Boost 1.50 is too old for example.

pcucka:
> We should also include some error checking in the build system itself - checking the version of our dependencies are sufficiently late enough and reporting to the user otherwise.

We should also set `OPENVDB_ABI_VERSION_NUMBER` based on `$HOUDINI_VERSION`.  And now that we're dropping support for older ABIs, we need to check if `OPENVDB_ABI_VERSION_NUMBER` is set to an unsupported version.

PR #375, opened 2019-03-07
Reference Implementation for Proposed Migration to CircleCI

danrbailey:
Note that this is a **draft** pull request, so it cannot be merged. 

This is a reference implementation for migrating from Travis CI to Circle CI as discussed in the CI working group for the TAC today, so I thought I'd share with everyone for transparency. (@lgritz @zxiiro @dheckenberg)

@Idclip, you'll notice I've pulled a few of your refactoring changes across to this branch (thanks for that), otherwise there's not too much that has changed from the Travis configuration. For now, I've put the build scripts into an openvdb/ci sub-directory as I expect there'll be some overlap in using Travis and Circle together during migration. Ultimately the aim should be to deprecate the Travis scripts.

Aside from the ASWF/LF decision to migrate to Circle CI, there are still some outstanding issues to resolve before considering a full PR:
* GCC 6.3.1 issue - the use of -isystem and #include_next is causing some compilation issues, probably a search path issue, but for now I've simply swapped -isystem for -I (I think CMake doesn't encounter this problem)
* Clang 3.8 issue - haven't dug into this in too much detail, but hcustom seems to have an issue with thinking Clang 3.8 is GCC 3.8 with some odd consequences, I've simply swapped the compile flags for the correct ones for now
* The yaml is a bit verbose right now, it would be good to play with some of the code re-use techniques in that language

All builds complete successfully on CircleCI using this configuration.

Note that adopting the new SideFX download API is treated as orthogonal to this migration [OVDB-34](https://jira.aswf.io/browse/OVDB-34)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,217 @@
+version: 2
+jobs:
+  # ABI=6 (VFX Reference Platform 2019) built with Clang 3.8
+  # core library, Python module, binaries, unit tests and header checks
+  testabi6:
+    docker:
+      - image: debian:stretch
+    steps:
+      - checkout
+      - run:
+          name: Install
+          command: bash ci/install.sh none clang++
+      - run:
+          name: Core
+          command: bash ci/run.sh core 6 yes release none clang++
+      - run:
+          name: Extras
+          command: bash ci/run.sh extras 6 yes release none clang++
+      - run:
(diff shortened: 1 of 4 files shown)


danrbailey:
Closing this PR - #382 has now been merged which includes a migration to Circle CI.

PR #376, opened 2019-03-07
As per meeting notes of Jan 17, reword the 25% requirement.

jmlait:
As per meeting notes of Jan 17, reword the 25% requirement.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ The committer role enables the participant to commit code directly to the reposi
 * Monitor Slack (delayed response is perfectly acceptable).
 * Triage GitHub issues and perform pull request reviews for other committers and the community.
 * Make sure that ongoing PRs are moving forward at the right pace or close them.
-* In general continue to be an active contributor to the project in general and the code base in particular.
+* Remain an active contributor to the project in general and the code base in particular.
 
 ### When does a committer lose committer status?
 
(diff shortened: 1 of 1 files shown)


kmuseth:
May I suggest a minor rewording:

"an active contributor to the project in general and the code base in particular"

jmertic:
I think this is reasonable

CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ The committer role enables the participant to commit code directly to the reposi
 * Monitor Slack (delayed response is perfectly acceptable).
 * Triage GitHub issues and perform pull request reviews for other committers and the community.
 * Make sure that ongoing PRs are moving forward at the right pace or close them.
-* In general continue to be willing to spend at least 25% of one's time working on the project (~1.25 business days per week).
+* In general continue to be an active contributor to the project in general and the code base in particular.
```
danrbailey: In general ... in general ?

kmuseth: good catch - let's change it to:

"* Remain an active contributor to the project in general and the code base in particular."

jmlait: But.... Then we lose the wonderful passive voice?  :>

I've made that change.

kmuseth:
just confirming that  I'm good with Jeff's latest edits

PR #378, opened 2019-03-08
OVDB-15: Enable warnings as errors

danrbailey:
Add -Werror flag to strict mode to enable warnings as errors.


Idclip:
As mentioned on the ticket, wrong version of gcc being used for h17, travis isusing 5.4. Should try using 6.3.1, I'm getting warnings with 6.4.

PR #379, opened 2019-03-08
Minor fixes for misc. gcc 6.4 compiler warnings

Idclip:
This should be the complete set of fixes for gcc 6.4 warnings (the ones enabled in the makefile). Note that a lot of these changes may impact the precision of various calculations. I've fixed these in the context of the expression return types - please comment on any where you think this might not be valid.

### DIFF
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1355,12 +1355,7 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    constexpr bool Constructible =
-        openvdb::CanConvertType<float, ValueType>::value;
-    using ToleranceType =
-        typename std::conditional<Constructible, ValueType, float>::type;
-
-    this->tree().prune(ValueType(zeroVal<ValueType>() + ToleranceType(tolerance)));
+    this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
 }
 
 #if OPENVDB_ABI_VERSION_NUMBER >= 3
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1355,10 +1355,11 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    constexpr bool IsString = std::is_same<ValueType, std::string>::value;
-    using ZeroValType =
-        typename std::conditional<IsString, std::string, float>::type;
-    this->tree().prune(ValueType(zeroVal<ZeroValType>() + tolerance));
+    constexpr bool Convertible = std::is_convertible<float, ValueType>::value;
+    using ToleranceType =
+        typename std::conditional<Convertible, ValueType, float>::type;
+
+    this->tree().prune(ValueType(zeroVal<ValueType>() + ToleranceType(tolerance)));
 }
 
 #if OPENVDB_ABI_VERSION_NUMBER >= 3
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1355,9 +1355,10 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    constexpr bool Convertible = std::is_convertible<float, ValueType>::value;
+    constexpr bool Constructible =
+        openvdb::CanConvertType<float, ValueType>::value;
     using ToleranceType =
-        typename std::conditional<Convertible, ValueType, float>::type;
+        typename std::conditional<Constructible, ValueType, float>::type;
 
     this->tree().prune(ValueType(zeroVal<ValueType>() + ToleranceType(tolerance)));
 }
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1355,7 +1355,10 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
+    constexpr bool IsString = std::is_same<ValueType, std::string>::value;
+    using ZeroValType =
+        typename std::conditional<IsString, std::string, float>::type;
+    this->tree().prune(ValueType(zeroVal<ZeroValType>() + tolerance));
 }
 
 #if OPENVDB_ABI_VERSION_NUMBER >= 3
--- a/openvdb/math/Coord.h
+++ b/openvdb/math/Coord.h
@@ -378,12 +378,12 @@ public:
     /// @brief Return @c true if this bounding box is empty (i.e., encloses no coordinates).
     bool empty() const
     {
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__INTEL_COMPILER)
   #pragma GCC diagnostic push
   #pragma GCC diagnostic ignored "-Wstrict-overflow"
 #endif
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
-#if defined(__GNUC__)
+#if defined(__GNUC__) && !defined(__INTEL_COMPILER)
   #pragma GCC diagnostic pop
 #endif
     }
--- a/openvdb/math/Coord.h
+++ b/openvdb/math/Coord.h
@@ -376,7 +376,17 @@ public:
     bool operator!=(const CoordBBox& rhs) const { return !(*this == rhs); }
 
     /// @brief Return @c true if this bounding box is empty (i.e., encloses no coordinates).
-    bool empty() const { return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]); }
+    bool empty() const
+    {
+#if defined(__GNUC__)
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wstrict-overflow"
+#endif
+        return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
+#if defined(__GNUC__)
+  #pragma GCC diagnostic pop
+#endif
+    }
     /// @brief Return @c true if this bounding box is nonempty (i.e., encloses at least one coordinate).
     operator bool() const { return !this->empty(); }
     /// @brief Return @c true if this bounding box is nonempty (i.e., encloses at least one coordinate).
--- a/openvdb/math/Operators.h
+++ b/openvdb/math/Operators.h
@@ -765,9 +765,9 @@ struct Gradient<ScaleMap, CD_2ND>
         using Vec3Type = typename internal::ReturnValue<Accessor>::Vec3Type;
 
         Vec3Type iGradient( ISGradient<CD_2NDT>::result(grid, ijk) );
-        return  Vec3Type(iGradient[0] * ValueType(map.getInvTwiceScale()[0]),
-                         iGradient[1] * ValueType(map.getInvTwiceScale()[1]),
-                         iGradient[2] * ValueType(map.getInvTwiceScale()[2]) );
+        return  Vec3Type(ValueType(iGradient[0] * map.getInvTwiceScale()[0]),
+                         ValueType(iGradient[1] * map.getInvTwiceScale()[1]),
+                         ValueType(iGradient[2] * map.getInvTwiceScale()[2]) );
     }
 
     // stencil access version
@@ -779,9 +779,9 @@ struct Gradient<ScaleMap, CD_2ND>
         using Vec3Type = typename internal::ReturnValue<StencilT>::Vec3Type;
 
         Vec3Type iGradient( ISGradient<CD_2NDT>::result(stencil) );
-        return  Vec3Type(iGradient[0] * ValueType(map.getInvTwiceScale()[0]),
-                         iGradient[1] * ValueType(map.getInvTwiceScale()[1]),
-                         iGradient[2] * ValueType(map.getInvTwiceScale()[2]) );
+        return  Vec3Type(ValueType(iGradient[0] * map.getInvTwiceScale()[0]),
+                         ValueType(iGradient[1] * map.getInvTwiceScale()[1]),
+                         ValueType(iGradient[2] * map.getInvTwiceScale()[2]) );
     }
 };
 
@@ -799,9 +799,9 @@ struct Gradient<ScaleTranslateMap, CD_2ND>
         using Vec3Type = typename internal::ReturnValue<Accessor>::Vec3Type;
 
         Vec3Type iGradient( ISGradient<CD_2NDT>::result(grid, ijk) );
-        return  Vec3Type(iGradient[0] * ValueType(map.getInvTwiceScale()[0]),
-                         iGradient[1] * ValueType(map.getInvTwiceScale()[1]),
-                         iGradient[2] * ValueType(map.getInvTwiceScale()[2]) );
+        return  Vec3Type(ValueType(iGradient[0] * map.getInvTwiceScale()[0]),
+                         ValueType(iGradient[1] * map.getInvTwiceScale()[1]),
+                         ValueType(iGradient[2] * map.getInvTwiceScale()[2]) );
     }
 
     // Stencil access version
@@ -813,9 +813,9 @@ struct Gradient<ScaleTranslateMap, CD_2ND>
         using Vec3Type = typename internal::ReturnValue<StencilT>::Vec3Type;
 
         Vec3Type iGradient( ISGradient<CD_2NDT>::result(stencil) );
-        return  Vec3Type(iGradient[0] * ValueType(map.getInvTwiceScale()[0]),
-                         iGradient[1] * ValueType(map.getInvTwiceScale()[1]),
-                         iGradient[2] * ValueType(map.getInvTwiceScale()[2]) );
+        return  Vec3Type(ValueType(iGradient[0] * map.getInvTwiceScale()[0]),
+                         ValueType(iGradient[1] * map.getInvTwiceScale()[1]),
+                         ValueType(iGradient[2] * map.getInvTwiceScale()[2]) );
     }
 };
 //@}
@@ -1597,9 +1597,7 @@ struct Laplacian<ScaleMap, DiffScheme>
         ValueType iddz = D2<DiffScheme>::inZ(grid, ijk);
         const Vec3d& invScaleSqr = map.getInvScaleSqr();
         // scale them by the appropriate 1/dx^2, 1/dy^2, 1/dz^2 and sum
-        return iddx * ValueType(invScaleSqr[0]) +
-               iddy * ValueType(invScaleSqr[1]) +
-               iddz * ValueType(invScaleSqr[2]);
+        return ValueType(iddx * invScaleSqr[0] + iddy * invScaleSqr[1] + iddz * invScaleSqr[2]);
     }
 
     // stencil access version
@@ -1614,9 +1612,7 @@ struct Laplacian<ScaleMap, DiffScheme>
         ValueType iddz = D2<DiffScheme>::inZ(stencil);
         const Vec3d& invScaleSqr = map.getInvScaleSqr();
         // scale them by the appropriate 1/dx^2, 1/dy^2, 1/dz^2 and sum
-        return iddx * ValueType(invScaleSqr[0]) +
-               iddy * ValueType(invScaleSqr[1]) +
-               iddz * ValueType(invScaleSqr[2]);
+        return ValueType(iddx * invScaleSqr[0] + iddy * invScaleSqr[1] + iddz * invScaleSqr[2]);
     }
 };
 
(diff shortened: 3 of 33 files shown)


openvdb/math/Coord.h
```diff
@@ -376,7 +376,17 @@ class CoordBBox
...
+    {
+#if defined(__GNUC__)
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wstrict-overflow"
+#endif
+        return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
+#if defined(__GNUC__)
+  #pragma GCC diagnostic pop
+#endif
+    }
```
Idclip: Had trouble trying to resolve this :|

openvdb_houdini/houdini/SOP_OpenVDB_Combine.cc
```diff
@@ -1031,8 +1031,8 @@ struct SOP_OpenVDB_Combine::CombineOp
             // For level set grids, use the level set rebuild tool to both resample the
             // source grid to match the reference grid and to rebuild the resulting level set.
             const ValueT halfWidth = ((ref.getGridClass() == openvdb::GRID_LEVEL_SET)
-                ? ValueT(ZERO + this->getScalarBackgroundValue(ref) * (1.0 / ref.voxelSize()[0]))
-                : ValueT(src.background() * (1.0 / src.voxelSize()[0])));
+                ? ValueT(this->getScalarBackgroundValue(ref) * (1.0 / ref.voxelSize()[0]))
+                : src.background() * ValueT(1.0 / src.voxelSize()[0]));
```
Idclip: I've removed some of the zeroval + ... code from a few houdini based methods to resolve conversion warnings. As far as I can tell this was to support string grids? I can change this to a conditional if this needs to stay (the same way I've resolved pruneGrid in Grid.h)

pcucka: The `zeroVal<T>() + x` idiom allows values of arbitrary types to be supported, even types that can't be constructed with `T(x)`.  (For those types, one defines a `zeroVal` specialization and possibly an `operator+` overload.)  It's probably safe to remove it from Houdini code, which uses only a limited set of value types, but definitely not from, say, `Grid.h`.

openvdb/Grid.h
```diff
@@ -1355,7 +1355,10 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
+    constexpr bool IsString = std::is_same<ValueType, std::string>::value;
+    using ZeroValType =
+        typename std::conditional<IsString, std::string, float>::type;
+    this->tree().prune(ValueType(zeroVal<ZeroValType>() + tolerance));
```
danrbailey: @pcucka - do you have any thoughts on this? I had a look at fixing this one a while back but didn't get too far. It feels like the tolerance should only really apply to floating-point value types?

openvdb/math/Coord.h
danrbailey: Yeah, I had a go at this too, I couldn't even figure out where this was coming from, the stack trace is pretty obscure! This seems like a reasonable compromise.

openvdb/Grid.h
pcucka: No, the tolerance is applicable to all numeric types, both scalar and vector, and probably to some user-defined types as well.

openvdb/math/Coord.h
pcucka: ICC, for one, defines `__GNUC__`, but it uses different diagnostic flags, so at a minimum you'll need to `#ifndef __INTEL_COMPILER` (or something like that).

What _was_ the stack trace, though?

openvdb/Grid.h
Idclip: What about this:

```
    constexpr bool Convertible = std::is_convertible<float, ValueType>::value;
    using ToleranceType =
        typename std::conditional<Convertible, ValueType, float>::type;

    this->tree().prune(ValueType(zeroVal<ValueType>() + ToleranceType(tolerance)));
```

pcucka: What warning(s) were you getting with the original code?

Idclip: Conversion warnings from -Wconversion - the expression is wrapped in a cast but the addition is left implicit. This was most of the warnings in this PR, hence the precision concerns: 

```
Grid.h:1361:52: warning: conversion to ‘float’ from ‘long int’ may alter its value [-Wconversion]
     this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
```

pcucka: In that case, I think you're on the right track, but I'm not sure your solution would address the case where `ValueType` is, say, `Vec3i`.  Also, if `ValueType` _is_ convertible, then there's no need to add `zeroVal`.

Idclip: Yeah I don't think is_convertible is returning true for something like Vec3i, but there's never any warnings with the expression for types like Vec3i which have a templated + operator (which forwards to += where the casting is static). The warnings I got from this method were from scalar + float where scalar wasn't a float, resulting in the implicit cast. 

I left add zeroVal in for when ValueType is convertible as introducing a runtime branch would bring back the warnings again.

openvdb/math/Coord.h
Idclip: See below for warnings from the core lib

<details><summary>Stack trace</summary>
<p>

```
In file included from ../openvdb/Types.h:45:0,
                 from ../openvdb/openvdb.h:35,
                 from python/pyGrid.h:62,
                 from python/pyIntGrid.cc:34:
../openvdb/math/Coord.h: In member function ‘void openvdb::v6_0abi5::tree::InternalNode<_ChildNodeType, Log2Dim>::clip(const openvdb::v6_0abi5::math::CoordBBox&, const ValueType&) [with _ChildNodeType = openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::LeafNode<bool, 3u>, 4u>; unsigned int Log2Dim = 5u]’:
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h: In member function ‘void openvdb::v6_0abi5::tree::Tree<_RootNodeType>::readBuffers(std::istream&, const openvdb::v6_0abi5::math::CoordBBox&, bool) [with _RootNodeType = openvdb::v6_0abi5::tree::RootNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::LeafNode<bool, 3u>, 4u>, 5u> >]’:
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h: In member function ‘void openvdb::v6_0abi5::Grid< <template-parameter-1-1> >::readBuffers(std::istream&, const openvdb::v6_0abi5::math::CoordBBox&) [with _TreeType = openvdb::v6_0abi5::tree::Tree<openvdb::v6_0abi5::tree::RootNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::LeafNode<bool, 3u>, 4u>, 5u> > >]’:
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~

In file included from ../openvdb/Types.h:45:0,
                 from ../openvdb/openvdb.h:35,
                 from viewer/RenderModules.h:34,
                 from viewer/RenderModules.cc:31:
../openvdb/math/Coord.h: In member function ‘void openvdb::v6_0abi5::tree::Tree<_RootNodeType>::readBuffers(std::istream&, const openvdb::v6_0abi5::math::CoordBBox&, bool) [with _RootNodeType = openvdb::v6_0abi5::tree::RootNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::InternalNode<openvdb::v6_0abi5::tree::LeafNode<bool, 3u>, 4u>, 5u> >]’:
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:35: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                 ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
../openvdb/math/Coord.h:381:56: warning: assuming signed overflow does not occur when assuming that (X + c) < X is always false [-Wstrict-overflow]
         return (mMin[0] > mMax[0] || mMin[1] > mMax[1] || mMin[2] > mMax[2]);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~

```

</p>
</details>

Idclip: I've added a check for `__INTEL_COMPILER`

openvdb/math/Vec3.h
```diff
@@ -265,79 +265,79 @@ class Vec3: public Tuple<3, T>
     template <typename S>
     const Vec3<T> &operator*=(S scalar)
     {
-        this->mm[0] = static_cast<T>(this->mm[0] * scalar);
-        this->mm[1] = static_cast<T>(this->mm[1] * scalar);
-        this->mm[2] = static_cast<T>(this->mm[2] * scalar);
+        this->mm[0] *= static_cast<T>(scalar);
+        this->mm[1] *= static_cast<T>(scalar);
+        this->mm[2] *= static_cast<T>(scalar);
         return *this;
```
Idclip: I've realised that changing casts like this will break cases where floats are applied to ints. I'm surprised that nothing in the unit tests has caught this. Trivial example. `Vec3i a(2); a *= 1.8f;`. a before would be equal to 3. Now, 1.8f will be truncated prior to multiplication. Silly of me to blanket change these. Thought here welcome

pcucka: Yes, arithmetic should be done wherever possible using the highest-precision operand's type (the `math::promote<>` metafunction is intended to help with that), with only the result being converted to the target type.

Also, the goal is not to suppress all type conversion warnings.  We want to see type conversion warnings, but we also want there to be at least one way to invoke any given operation without warnings if the user determines that the type conversions involved are safe for that operation.

Idclip: Yeah I agree, in which case I suspect the best solution would be to revert these vec3 changes and instead ensure that core vdb logic which invokes vec3 operators operates at the same/suitable precision. I can't remember how many warnings there were though, I'll take another look

openvdb/math/Operators.h
```diff
@@ -1645,7 +1651,9 @@ struct Laplacian<ScaleTranslateMap, DiffScheme>
         ValueType iddz = D2<DiffScheme>::inZ(stencil);
         const Vec3d& invScaleSqr = map.getInvScaleSqr();
         // scale them by the appropriate 1/dx^2, 1/dy^2, 1/dz^2 and sum
-        return ValueType(iddx * invScaleSqr[0] + iddy * invScaleSqr[1] + iddz * invScaleSqr[2]);
+        return iddx * ValueType(invScaleSqr[0]) +
+               iddy * ValueType(invScaleSqr[1]) +
+               iddz * ValueType(invScaleSqr[2]);
```
Idclip: I'm guessing we want methods like these to always operate at double precision? I was a bit wary of casting/converting values away *from* grid `ValueType`s due to the mindset behind the `zeroVal` idiom. There also seems to be a bit of inconsistency between the different specialised structs. Using the Gradient methods as an example, the base, non specialised version stores the return of `ISGradient` at double precision, explicitly also doing the calculation at double precision. All specialised versions store the result as `ValueType`, but are implicitly doing some computation at double precision due to the return type of `getInvTwiceScale()`. An example change could be:

```
    result(const UniformScaleMap& map, const Accessor& grid, const Coord& ijk)
    {
        using ValueType = typename internal::ReturnValue<Accessor>::ValueType;
        using Vec3Type = typename internal::ReturnValue<Accessor>::Vec3Type;

        Vec3Type iGradient( ISGradient<CD_2NDT>::result(grid, ijk) );
        ValueType inv2dx = ValueType(map.getInvTwiceScale()[0]);
        return  iGradient * inv2dx;
    }
```

to

```
    result(const UniformScaleMap& map, const Accessor& grid, const Coord& ijk)
    {
        using Vec3Type = typename internal::ReturnValue<Accessor>::Vec3Type;

        Vec3d iGradient( ISGradient<CD_2NDT>::result(grid, ijk) );
        double inv2dx = map.getInvTwiceScale()[0];
        return  Vec3Type(iGradient * inv2dx);
    }
```

But this would then set the precedent to change **all** operations to double precision. Methods in interpolation also suffer from this (computation at `ValueType` but with a double precision position coef).

openvdb/Grid.h
Idclip: I've ended up with the following:
```
    constexpr bool Constructible =
        openvdb::CanConvertType<float, ValueType>::value;
    using ToleranceType =
        typename std::conditional<Constructible, ValueType, float>::type;
```

pcucka: Unfortunately, that seems to have reintroduced the warning:
https://travis-ci.org/AcademySoftwareFoundation/openvdb/jobs/508949989#L1007

openvdb/math/Operators.h
pcucka: I think if we update `Vec3<T>::operator*()` to use `math::promote<>` internally, then the above could reduce to
```
result(const UniformScaleMap& map, const Accessor& grid, const Coord& ijk)
{
    using Vec3Type = typename internal::ReturnValue<Accessor>::Vec3Type;
    const double inv2dx = map.getInvTwiceScale()[0];
    return Vec3Type(ISGradient<CD_2NDT>::result(grid, ijk) * inv2dx);
}
```

openvdb/Grid.h
Idclip: Well that's annoying, thought I could get away with using `is_constructible` but didn't see clang fail - i'll change it back to `is_convertible`

openvdb/math/Operators.h
Idclip: I think I'd prefer the precision of the calculations to be explicit rather than inferred in the operators i.e. relying on `promote` from the Vec3 op? Also, do we want to add automatic type conversion to math operators? Based on comments above, I was thinking that `Vec3i(1) * 1.0f` should give you the same conversion warning as `1 * 1.0f`

pcucka: Yes, I suppose that's right.  But then I guess for consistency none of the `Vec` operators should do any type conversion, meaning that a few, including `Vec3::operator*=(scalar)`, would need to be changed.

Idclip:
As discussed, I've reverted all non trivial casting changes and made a ticket to represent the other work: https://jira.aswf.io/browse/OVDB-81

openvdb/math/Vec3.h
```diff
@@ -71,9 +71,9 @@ class Vec3: public Tuple<3, T>
     template <typename Source>
     Vec3(Source *a)
     {
-        this->mm[0] = a[0];
-        this->mm[1] = a[1];
-        this->mm[2] = a[2];
+        this->mm[0] = static_cast<T>(a[0]);
```
pcucka: This change should be made in `Vec2.h` and `Vec4.h` as well.

Idclip: Done, and `Mat3/Mat4.h` too

PR #380, opened 2019-03-12
OVDB-77: Fix point conversion rounding error

danrbailey:


openvdb/points/PointConversion.h
```diff
@@ -773,19 +669,70 @@ createPointDataGrid(const PointIndexGridT& pointIndexGrid, const PositionArrayT&
...
+    leafManager.foreach(
+        [&](LeafT& leaf, size_t /*idx*/) {
+
+            // obtain the PointIndexLeafNode (using the origin of the current leaf)
+            const auto* pointIndexLeaf = pointIndexTree.probeConstLeaf(leaf.origin());
+            assert(pointIndexLeaf);
+
+            auto attributeWriteHandle = AttributeWriteHandle<Vec3f, CompressionT>::create(
+                leaf.attributeArray(positionIndex));
+
```
Idclip: I wonder if we should discuss some standards for auto usage as there's none at the moment. Not going to recommend changing anything here, just seen some variable types -> auto during this change.

openvdb/points/PointConversion.h
```diff
@@ -773,19 +669,70 @@ createPointDataGrid(const PointIndexGridT& pointIndexGrid, const PositionArrayT&
...
-    PopulatePositionAttributeOp<PointDataTreeT,
-                                PointIndexTreeT,
-                                PositionArrayT> populate(pointIndexTree,
-                                                        xform,
-                                                        positions);
+    const size_t positionIndex = descriptor->find("P");
+    assert(positionIndex != AttributeSet::INVALID_POS);
+
+    leafManager.foreach(
+        [&](LeafT& leaf, size_t /*idx*/) {
```
Idclip: Is it even necessary to have these as different operators anymore? Can this be merged into the above?

danrbailey: Ooh, good suggestion, I hadn't considered that. Just pushed up a change to address this, thanks!

openvdb/points/PointConversion.h
danrbailey: I tend to try and limit it's use to specific cases - where I'm repeating the name of a type or deducing a type from a nested name, especially when it would require a typename keyword. I don't mind if we include something in our style guide.

openvdb/points/PointConversion.h
Idclip: Looks good!

PR #381, opened 2019-03-12
OVDB-62: Update changes for SOP label names

danrbailey:


danrbailey:
> Please also add something like
> "Added a getLabelName() method to allow an OpPolicy to return a computed name for use in menus."

Done and fixed the changes conflict.

PR #386, opened 2019-03-27
Fixes for tbb::atomic errors reported by GCC 6.3

pcucka:
With the combination of GCC 6.3 and TBB 4.4.6 (not one of the VFX platform configurations), I'm getting errors like this:
```
openvdb/tree/LeafBuffer.h:254:19: error: no matching function for call to ‘tbb::atomic<unsigned int>::atomic(int)’
     , mOutOfCore(0)
                   ^
tbb/4.4.6.0/include/tbb/atomic.h:468:1: note: candidate: tbb::atomic<unsigned int>::atomic()
 __TBB_DECL_ATOMIC(unsigned)
 ^
tbb/4.4.6.0/include/tbb/atomic.h:468:1: note:   candidate expects 0 arguments, 1 provided
tbb/4.4.6.0/include/tbb/atomic.h:468:1: note: candidate: constexpr tbb::atomic<unsigned int>::atomic(const tbb::atomic<unsigned int>&)
tbb/4.4.6.0/include/tbb/atomic.h:468:1: note:   no known conversion for argument 1 from ‘int’ to ‘const tbb::atomic<unsigned int>&’


In file included from openvdb/points/AttributeGroup.h:40:0,
                 from openvdb/points/AttributeGroup.cc:33:
openvdb/points/AttributeArray.h:425:39: error: could not convert ‘0’ from ‘int’ to ‘tbb::atomic<unsigned int>’
     tbb::atomic<Index32> mOutOfCore = 0; // interpreted as bool
```

This PR addresses those and a couple of other such errors by substituting assignment for initialization of TBB atomics.

Anyone else run into this issue?

### DIFF
--- a/openvdb/CHANGES
+++ b/openvdb/CHANGES
@@ -10,13 +10,13 @@ Version 6.0.1 - In development
     - Added a generic TypeList class.
     - Added GridBase::apply(), which invokes a functor on a grid
       if the resolved grid type is a member of a given type list.
-    - Added util::printTime that outputs nicely-formatted time information.
-    - Added a template specialization of std::hash with Coord.
-    - Added CoordBBox::moveMin and CoordBBox::moveMax to move a CoordBBox.
+    - Added util::printTime(), which outputs nicely formatted time information.
+    - Added a std::hash<Coord> template specialization.
+    - Added CoordBBox::moveMin() and CoordBBox::moveMax() to move a CoordBBox.
 
     Improvements:
-    - The API of util::CpuTimer has been improved and is now making use of
-      a new util::printTime function for nicer output of time information.
+    - util::CpuTimer now makes use of util::printTime() for nicer output,
+      and its API has been improved.
     - Significantly improved the performance of point data grid string
       attribute generation.
     - AttributeArray::copy() and the copy assignment operator for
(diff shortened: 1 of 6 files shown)


Idclip:
Not run into this, but could this also be solved by doing `tbb::atomic<Index32> mOutOfCore = Index32(0);`

pcucka:
No, the problem is not the sign conversion, it's the absence (by design) of a `tbb::atomic<T>::atomic(T)` constructor.

danrbailey:
@pcucka - I haven't run into this issue, but I note that the James Reinders TBB book says:

"You can rely on zero initialization to initialize an atomic to 0. To create an atomic with a specific value, default-construct it first, and afterword assign a value to it."

So in fact, for numeric atomics, class member variables and C++11, I think it's not necessary to assign the zero value at all, we can just switch this to use the default constructor?

pcucka:
For non-static data members, zero-assignment can be omitted only if the member is listed explicitly in the class constructor's initializer list.  So, for example, these are equivalent:
```
AttributeArray(): mPageHandle() { mOutOfCore = 0; }  // A

AttributeArray(): mPageHandle(), mOutOfCore() {}  // B
```
but
```
AttributeArray(): mPageHandle() {}  // C
```
would leave `mOutOfCore` uninitialized.

I have a slight preference for assignment in this case, since it makes it clear that the initial value is zero, but I'm OK with either A or B.

danrbailey:
> For non-static data members, zero-assignment can be omitted only if the member is listed explicitly in the class constructor's initializer list. So, for example, these are equivalent:
> 
> ```
> AttributeArray(): mPageHandle() { mOutOfCore = 0; }  // A
> 
> AttributeArray(): mPageHandle(), mOutOfCore() {}  // B
> ```
> but
> 
> ```
> AttributeArray(): mPageHandle() {}  // C
> ```
> would leave `mOutOfCore` uninitialized.
> 
> I have a slight preference for assignment in this case, since it makes it clear that the initial value is zero, but I'm OK with either A or B.

That's fair, I'm ok with the explicit way of doing it.

PR #391, opened 2019-04-12
OVDB-58: Fix bug with filtered point re-bucketing

danrbailey:
Resolve issues with filtered point re-bucketing by removing the optimization that attempts to use a padded vector instead of a hash map within a certain threshold. There are other potential avenues of optimization that are worth exploring here.

### DIFF
--- a/openvdb/unittest/TestPointMove.cc
+++ b/openvdb/unittest/TestPointMove.cc
@@ -265,7 +265,6 @@ TestPointMove::testCachedDeformer()
     CachedDeformer<double> cachedDeformer(cache);
 
     // check initialization is as expected
-    CPPUNIT_ASSERT(cachedDeformer.mLocalLeafVec.empty());
     CPPUNIT_ASSERT(cachedDeformer.mLeafVec == nullptr);
     CPPUNIT_ASSERT(cachedDeformer.mLeafMap == nullptr);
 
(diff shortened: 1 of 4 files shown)


Idclip:
Can you clarify how your changes resolved the case where you're using groups? What part of the padded vector optimisation was at fault?

Idclip:
And follow up questions - there's no new unit test in this PR to test current issue. Can we add one?

danrbailey:
> Can you clarify how your changes resolved the case where you're using groups? What part of the padded vector optimisation was at fault?

The issue is caused by using a combination of advection filtering and a cached deformer. For efficiency reasons, the cached deformer only computes the new positions for points that are actually being moved, populating a hash map as it goes. When using the optimization where this hash map is converted to a vector to accelerate lookups, it wasn't correctly accounting for the spacing caused by the filtering. Rather than fixing this and then verifying all my assumptions still stand with regards to performance, I've simply removed this attempted optimization so that this resolves the seg-fault. My recommendation is that you ensure that you're correctly implementing the filter::state() methods in any custom filters you're using which will automatically drop back to using a vector in the case where no filtering is taking place, otherwise it will use the (slower) hash map.

> And follow up questions - there's no new unit test in this PR to test current issue. Can we add one?

I was a little reluctant to add one because it appeared to require quite a large data set to expose the issue, but with a little bit of work, I managed to produce a simpler test case that does reproduce the issue which I've just added to the unit tests.

openvdb/unittest/TestPointMove.cc
```diff
@@ -1129,6 +1075,48 @@ TestPointMove::testPointData()
         CPPUNIT_ASSERT_EQUAL(short(0), nonZeroGroups2[2]);
         CPPUNIT_ASSERT_EQUAL(short(1), nonZeroGroups2[3]);
     }
+
+    { // larger data set with a cached deformer and group filtering
```
Idclip: As there's no cpp asserts here, perhaps a slightly more verbose comment regarding what this is testing in regards to this PR

Idclip:
Great, thanks @drb! Tested in Houdini with the original scene, it's fixed the issue

openvdb/unittest/TestPointMove.cc
danrbailey: I made a small adjustment to verify the values were as expected.

PR #392, opened 2019-04-12
OVDB-69: Remove code < 16.5 for custom prim and guard on SESI_OPENVDB_PRIM

danrbailey:
Currently, the VDB custom primitive is disabled unless SESI_OPENVDB is set or the Houdini version is lower than H12.5.

This change removes all the logic that applies for Houdini versions less than 16.5 and changes the second guard condition from "lower than H12.5" to "SESI_OPENVDB_PRIM being defined".

### DIFF
--- a/openvdb_houdini/houdini/GEO_PrimVDB.h
+++ b/openvdb_houdini/houdini/GEO_PrimVDB.h
@@ -47,7 +47,13 @@
  */
 
 #include <UT/UT_Version.h>
-#if !defined(SESI_OPENVDB) && (UT_VERSION_INT >= 0x0c050157) // 12.5.343 or later
+
+// Using the native OpenVDB Primitive shipped with Houdini is strongly recommended,
+// as there is no guarantee that this code will be kept in sync with Houdini.
+// However, for debugging it can be useful, so supply -DSESI_OPENVDB_PRIM to
+// the compiler to build this custom primitive.
+
+#if !defined(SESI_OPENVDB) && !defined(SESI_OPENVDB_PRIM)
 
 #include <GEO/GEO_PrimVDB.h>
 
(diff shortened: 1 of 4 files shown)


openvdb_houdini/houdini/GEO_PrimVDB.h
```diff
@@ -47,7 +47,13 @@
  */
 
 #include <UT/UT_Version.h>
-#if !defined(SESI_OPENVDB) && (UT_VERSION_INT >= 0x0c050157) // 12.5.343 or later
+
+// Using the native OpenVDB Primitive shipped with Houdini is strongly recommended,
+// as there is no guarantee that this code will be kept in sync with Houdini.
+// However, for debugging it can be useful, so supply -DSESI_OPENVDB_PRIM to
```
pcucka: Should that be added as an option to the CMake script?

danrbailey: I'm thinking that we don't as we're not considering maintaining it through CI. The extra requirement of having to actually provide it to the compiler yourself (or otherwise modify the build system or source) seems like enough of a barrier that no-one will switch it on by accident. What do you think?

danrbailey:
> If we're now starting to drop support for 16.0 and earlier (even if only in the primitive), we should mention that in the `INSTALL` file and maybe in the release notes.

Yes, agreed.

I just put a comment on the JIRA about this - do you think dropping support for H16.0 and earlier can be done in the next release, ie without the soft migration path of adding a deprecation warning first?

openvdb_houdini/houdini/GEO_PrimVDB.h
pcucka: That's fine.  It's true that making it easy to enable could lead to confusion.

pcucka:
AFAICT, all of the remaining version-specific code in the `openvdb_houdini` public API is for 15.5 or earlier, so I'm OK with removing it without first deprecating it.

danrbailey:
Closing, now included in #395

PR #394, opened 2019-04-12
OVDB-68: Add an error directive if using a deprecated version of the OpenVDB ABI

danrbailey:
I looked at adding deprecations around all functions and classes, but it didn't seem all that useful. Most of the ABI macro branching is to add new functions which is obviously not going to trigger deprecated warnings. In addition, many of the ABI changes are subtly changing the existing function signature (adding a const for example). I think this might not be a particularly obvious way of alerting the user.

Instead I've gone with the following, simply making a compile error with a clear error message. A compiler flag can then optionally suppress the error. I considered using #warning or pragma warnings to achieve the same thing, but this is not very portable, so this approach seemed less problematic.

Looking for some input on the mechanism that the user could use to provide the OPENVDB_USE_DEPRECATED_ABI flag?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,13 +69,9 @@ jobs:
     - script: bash travis/travis.run core $ABI $BLOSC $MODE $HOUDINI_MAJOR $COMPILER
       env: ABI=4 BLOSC=yes MODE=release HOUDINI_MAJOR=none COMPILER=clang
     - script: bash travis/travis.run core $ABI $BLOSC $MODE $HOUDINI_MAJOR $COMPILER
-      env: ABI=3 BLOSC=yes MODE=release HOUDINI_MAJOR=none COMPILER=clang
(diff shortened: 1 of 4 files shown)


pcucka:
> many of the ABI changes are subtly changing the existing function signature (adding a const for example). I think [deprecation] might not be a particularly obvious way of alerting the user.

I see only a handful of functions, all in `Grid.h`, that would need to be deprecated (in addition to the `CopyPolicy` enum in `Types.h`), and making a const argument non-const, as in the shallow copy constructors, is indeed the kind of change we want to alert users to, since it will likely necessitate changes to their code.

danrbailey:
@pcucka - I just updated the PR with those changes (ignore the circle and appveyor failures for the time being)

pcucka:
Would you mind to add a Doxygen comment to each of those deprecated functions?  Something simple like this would do:
```
/// @deprecated ABI versions older than 4 are deprecated.
```

PR #382, opened 2019-03-14
CMake Refactor

Idclip:
<b>Aims of this PR:</b>

 - Validate CMake build system for Linux/MacOs
 - Homogenize FindXxx.cmake modules and all CMakeLists
 - Improve and document dependency requirements for all OpenVDB components
 - Add missing build options for the core, python, binaries and houdini components
 - Doxygen documentation for dependencies and cmake installation.

<b>Incorporates the following PRs:</b>

 - https://github.com/AcademySoftwareFoundation/openvdb/pull/150 - OpenEXR fixes
 - https://github.com/AcademySoftwareFoundation/openvdb/pull/268 - ILMBase fixes

<b>Does not address:</b>

 - https://github.com/AcademySoftwareFoundation/openvdb/pull/160 - I've added NumPy support, but it requires CMake 3.14. This PR adds NumPy support for any version using another custom Find module. We should discuss minimum CMake version.
 - https://github.com/AcademySoftwareFoundation/openvdb/pull/171 - GNU vars. Not incorporated, but should be done
 - https://github.com/AcademySoftwareFoundation/openvdb/pull/240 - Some additional Windows and Maya updates which need investigating

<b>Some important things that have *not* been done in this PR:</b>

 - glfw, blosc, log4cplus. numpy auto options (use if available)

<b>Known issues:</b>

- I am getting cmake warnings on some paltforms when building against houdini and providing my own boost version or my own ilmbase/openexr version (for Houdini 17.5). To be honest I think these warnings are valid as libHalf, for example, is called the same thing in dsolib/ as it is in any old install of ilmbase. But I'd like to try and resolve these as loads are output.
- NumPy support does not work with Boost 1.65 as it needs Boost::numpy. Better NumPy support will be added in a separate PR.

I realise there are a lot of changes here. I can refactor the branch history if it's desired.

----------------------------------------------------------------------------------------------------------------

Test environments include CentOS, travis and MacOS. Note this PR does not include the travis update. This can be found here:

https://github.com/Idclip/openvdb/tree/cmake_travis

- CentOS 7 - g++, runtime tested with and without a Houdini 17 installation. In both cases, uses the same library versions as Houdini.
- MacOS 10.14.3 - clang++, runtime tested with and without a Houdini 17 installation. Of note, tested with Blosc 1.16 where there are issues, Boost 1.68, TBB 2019, ZLIB 1.2.11.
 - Travis - Latest build - https://travis-ci.com/Idclip/openvdb/builds/104392461

[...]

PR #397, opened 2019-04-24
OVDB-15: Fix/Suppress Warnings

danrbailey:
These changes either fix or suppress all warnings on GCC 6.3.0 for the core library, binaries, python module and Houdini SOPs. I've tested on Linux through Circle CI using CMake, which is the first target for enabling this. The changes to enable this in Circle/CMake will be forthcoming once #382 is merged.

This PR represents a kind of amnesty - all existing conversion issues (and there are quite a lot of them!) are simply suppressed using pragmas so that we can switch on warnings as errors. The aim being to ensure that any tools or changes introduced going forward are required to meet the new higher standard. In time, we can discuss the best way to go about resolving the existing issues and removing the pragmas. Where possible, I've refactored slightly to introduce pragmas around the fewest lines of code possible.

The most common conversion issue I encountered is when using an integer grid type and implicitly converting the ValueType into an explicit float/double type, but there are others.

### DIFF
--- a/openvdb/Grid.h
+++ b/openvdb/Grid.h
@@ -1447,7 +1447,16 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
+    // Suppress conversion warning that GCC 6.3+ emits when ValueType is int or long int
+#if defined(__GNUC__)
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wconversion"
+#endif
+    const auto value = zeroVal<ValueType>() + tolerance;
+#if defined(__GNUC__)
+  #pragma GCC diagnostic pop
+#endif
+    this->tree().prune(static_cast<ValueType>(value));
 }
 
 #if OPENVDB_ABI_VERSION_NUMBER >= 3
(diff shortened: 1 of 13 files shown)


openvdb/Grid.h
```diff
@@ -1447,7 +1447,16 @@ template<typename TreeT>
 inline void
 Grid<TreeT>::pruneGrid(float tolerance)
 {
-    this->tree().prune(ValueType(zeroVal<ValueType>() + tolerance));
+    // Suppress conversion warning that GCC 6.3+ emits when ValueType is int or long int
+#if defined(__GNUC__)
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wconversion"
+#endif
```
pcucka: I'd prefer an `OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN`/`OPENVDB_NO_TYPE_CONVERSION_WARNING_END` macro pair (see, e.g., `OPENVDB_NO_UNREACHABLE_CODE_WARNING_BEGIN` in `Platform.h`).  Although these changes are temporary, it might turn out that in some cases suppressing the warnings is the only practical solution, though that should always be a last resort.

openvdb/tools/Interpolation.h
```diff
@@ -738,23 +738,38 @@ template<class ValueT, size_t N>
 inline ValueT
 BoxSampler::trilinearInterpolation(ValueT (&data)[N][N][N], const Vec3R& uvw)
 {
+    struct Local
+    {
+        static ValueT interpolate(const ValueT& a, const ValueT& b, double weight)
```
pcucka: These should probably be lambdas rather than struct members.

danrbailey:
@pcucka - I've addressed your feedback now.

PR #401, opened 2019-04-30
Houdini Deprecations and Cleanup

danrbailey:
* mark methods in Houdini plugin as deprecated that are no longer used as a result of verbification (OP_NodeChain and duplicateSourceStealable)
* tidy up a few Boost includes
* use OP_AutoLockInputs instead of ScopedInputLock (verbification means this class is now only used in a few non-verbified SOPs)


danrbailey:
> Tools in the `houdini_utils` namespace are intended to be general-purpose. We should not assume that they are no longer useful to anyone because we're no longer using them. Not everyone has verbified their SOPs (or intends to).

Deprecating these tools doesn't force verbification, it simply requires switching from duplicateSourceStealable() to duplicateSource(). Verbification is the supported route to achieving the efficiency that used to be available using duplicateSourceStealable and the unload flag. The same applies to the OP_NodeChain tools.

I don't think we should carry around functionality that we ourselves aren't using, particularly as there's a perfectly viable alternative provided by SideFX. Without knowing exactly how useful these general-purpose tools are, it's likely we may be building and maintaining code that is no longer used by anyone.

pcucka:
If I'm not mistaken, verbification applies only to SOPs.  Is there an alternative to `OP_NodeChain` for other types of operators?

`ScopedInputLock` is used extensively at DWA, at least.  It's simpler to use than `OP_AutoLockInputs`, and it's fewer than 20 lines of code.

danrbailey:
> If I'm not mistaken, verbification applies only to SOPs. Is there an alternative to `OP_NodeChain` for other types of operators?

That might be a question for @jmlait. I wasn't aware that anyone had attempted to use OP_NodeChain for any other node types than SOPs? If it's useful outside of SOPs, there's definitely an argument for lobbying SideFX to adopt it - but if they're unwilling, I don't think VDB is the right place to put this.

> `ScopedInputLock` is used extensively at DWA, at least. It's simpler to use than `OP_AutoLockInputs`, and it's fewer than 20 lines of code.

I'm more willing to accept that we could keep this code around than the other two implementations as it is so little code. Is there no internal DWA Houdini library that this could be migrated into instead though?

jmlait:
I don't mind ScopedInputLock.  The OP_NodeChain I've never particularly liked because it involves a node trying to look at the network around it.  Ideally nodes should always cook in a vacuum and make no reference to the surrounding network.  It was a necessary evil, but now with Compile SOPs I think it isn't needed for SOPs.   I'm not sure what other network contexts it was used in, however.

danrbailey:
As discussed - we don't deprecate ScopedInputLock and add a comment to OP_NodeChain explaining the rationale for deprecating this and requesting that anyone using it get in touch with us.

openvdb_houdini/houdini/ParmFactory.h
```diff
@@ -546,6 +546,7 @@ class OPENVDB_HOUDINI_API OpPolicy
 
 
 /// @brief Helper class to manage input locking.
+/// @deprecated Verbification renders this redundant (recommend using OP_AutoLockInputs directly)
```
pcucka: I thought we agreed that `ScopedInputLock` would not be deprecated.  Verbification might render it redundant, but not all of our SOPs are verbified.  I see no reason to stop using it in those SOPs.

danrbailey: Yes, my mistake. Tried again with the update - removed the deprecated comments and reverted back to using ScopedInputLock.

PR #402, opened 2019-04-30
Add ASWF note to SOP help card

danrbailey:
* Add an ASWF toggle to the OpFactory::Impl
* Move some help text up into OpFactory to eliminate duplication
* Use https for OpenVDB website instead of http
* Extend with a mention of the ASWF
* Align Potential Flow help text

@pcucka - I don't know how well this will play with DWA infrastructure? The main motivation behind this change was mentioning the ASWF on the SOP help card so it's clearer to users who maintains the SOP as we try and unify the tab menu labels.

### DIFF
--- a/openvdb_houdini/houdini/ParmFactory.cc
+++ b/openvdb_houdini/houdini/ParmFactory.cc
@@ -895,7 +895,8 @@ public:
         unsigned flags,
         const char** inputlabels,
         const std::string& helpUrl,
-        const std::string& doc)
+        const std::string& doc,
+        const bool isASWF = true)
         : OP_Operator(name, english, construct, multiparms,
             operatorTableName,
             minSources, maxSources, variables, flags, inputlabels)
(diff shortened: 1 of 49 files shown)


openvdb_houdini/houdini/ParmFactory.cc
```diff
@@ -912,8 +913,19 @@ class OP_OperatorDW: public OP_Operator
             os << "= " << english << " =\n\n"
                 << "#type: node\n"
                 << "#context: " << flavorStr << "\n"
-                << "#internal: " << name << "\n\n"
-                << doc << "\n\n";
+                << "#internal: " << name << "\n";
+                if (isASWF) {
```
pcucka: This sort of logic belongs in an `OpPolicy` (in this case, `ASWFOpenVDBOpPolicy`), though how best to expose it is not clear.  One possibility would be to add `OpPolicy::docHeader` and `OpPolicy::docFooter` methods; another would be to add a `docTemplate` method, with some suitable syntax to interpolate the documentation body into the template.

danrbailey: I think that's a good suggestion, I'll have a look at how to go about doing this. The header and footer implementation is probably simplest, so I think I'll start there.

danrbailey: Done - just updated the PR.

danrbailey:
The changes files in this PR keep going out of date. @pcucka - I've addressed your feedback, any further thoughts or concerns here or can we get this merged?

openvdb_houdini/houdini/ParmFactory.cc
pcucka: It's unintuitive at best that the "footer" is inserted in the middle of the page.  Either move it to the bottom or go with a document template instead.  I prefer the flexibility of the latter.

danrbailey: I think you perhaps have a better idea of what you're after with this document template approach, Peter. Do you want to take this task over from me? I just want to try and avoid a bit of back-and-forth if possible.

pcucka: I can take a stab at it, yes.

PR #403, opened 2019-04-30
OVDB-63: Add First Name Policy to Houdini SOPs

danrbailey:
In Houdini, the "first name" is the default name of an instance of the SOP when you create it. Default first name rules (when it's not set) mean that this is based off the operator name. Now that we've changed the label to better align with SideFX naming, this is a bit confusing, so I've added an optional first name policy and then derived the first name of the ASWF policy from the label name instead.


openvdb_houdini/houdini/SOP_NodeVDB.cc
```diff
@@ -603,7 +603,7 @@ namespace {
 class DefaultOpenVDBOpPolicy: public houdini_utils::OpPolicy
 {
 public:
-    std::string getName(const houdini_utils::OpFactory&, const std::string& english) override
+    static std::string getName(const std::string& english)
```
pcucka: What's the motivation for splitting this out into a static member function?

danrbailey: Primarily because ASWFOpenVDBOpPolicy calls SESIOpenVDBOpPolicy::getName() and SESIOpenVDBOpPolicy is not a base class of ASWFOpenVDBOpPolicy.

I thought it was clearer to adopt the same structure for all policy classes with regard to implementing getName() functionality rather than having some be static methods and some be instance methods, or alternatively having ASWFOpenVDBOpPolicy create an instance of SESIOpenVDBOpPolicy in order to call this method.

pcucka: I suggest moving the static `SESIOpenVDBOpPolicy::getName` into `DefaultOpenVDBOpPolicy` as a non-static member named `getLowercaseName`.

danrbailey: Sure, that sounds reasonable to me.

danrbailey: I've addressed your feedback now @pcucka - if you could have a quick look over this and check the changes make sense to you?

PR #404, opened 2019-05-01
OVDB-64: Introduce OpHide Policy to SOPs

danrbailey:
I've implemented this new OpHide Policy, I'll add a little more discussion regarding this approach on the email thread:

https://lists.aswf.io/g/openvdb-dev/topic/proposal_unifying_vdb_and/29915127


danrbailey:
> I would like to at least see what this logic would look like implemented purely as a Python script incorporating a node table of some kind. That approach has the advantage that it would provide a working example that TDs could easily adapt to their needs; a simple all or none choice is unlikely to satisfy everyone.

One of the key features I wished to include by doing it like this is to have the logic about hiding variables and parent names be co-located with the SOP itself. Thus if you only built and installed one SOP it would replace it's parent SOP and no other ophiding would take place. That may still be feasible with your node table suggestion, but if you could outline an example implementation, that would be helpful?

I would suggest that we're not attempting to satisfy absolutely everyone here, the default remains to not hide anything for that reason. If the hiding policies that we're offering doesn't meet someone's needs, the SideFX recommended solution is to introduce an ophide script on startup. I think that remains preferable over introducing our own Python script that duplicates this functionality.

pcucka:
Are ophide scripts still required to be written in HScript?

jmlait:
From the docs, hou.NodeType.setHidden() should do the same as ophide.  So you could run this as a python script.

My objection is that requires installation into the right location.  I think it would be as helpful as just providing an OPcustomize snippet and asking people to append it.  Ie, it won't be able to help with the easy deployment problem.

By building it into the node definitions we can update this policy in the logical place - the node itself - as we add and improve nodes.  We also ensure the policy can be deployed if the nodes are; eliminating another failure point.

openvdb_houdini/CMakeLists.txt
```diff
@@ -334,6 +364,28 @@ if(USE_BLOSC OR OpenVDB_USES_BLOSC)
...
+  # compile-time if OPcustomize is in use
+  if (OPENVDB_HOUDINI_DEFAULT_POLICY)
+    set(OPCUSTOMIZE_ENV "OPENVDB_OPHIDE_POLICY=${OPENVDB_HOUDINI_DEFAULT_POLICY}")
+  else()
+    set(OPCUSTOMIZE_ENV "")
+  endif()
+  add_custom_target(opcustomize ALL
+    COMMAND ${CMAKE_COMMAND} -E env ${OPCUSTOMIZE_ENV}
+      hython ${CMAKE_CURRENT_SOURCE_DIR}/vdbopcustomize.py ${SOP_TARGET_FILES}
+      > ${CMAKE_CURRENT_BINARY_DIR}/OPcustomize
```
pcucka: HScript is deprecated, so it would be preferable to generate a Python script instead.  Installing it as `${OPENVDB_HOUDINI_INSTALL_PREFIX}/python2.7libs/pythonrc.py` should cause it to be executed at startup.

openvdb_houdini/vdbopcustomize.py
```diff
@@ -0,0 +1,52 @@
+import hou
+import sys
+
+sopCategory = hou.sopNodeTypeCategory()
+
+dsoPaths = sys.argv[1:]
```
pcucka: Add a usage message to this script (and preferably name it something like "`buildopcustomize.py`") for anyone who wants to run it outside of CMake:
```
if not dsoPaths or any(help in dsoPaths for help in '-h -help --help'.split()):
    progname = os.path.basename(__file__))
    sys.stderr.write("""\
Usage: %(progname)s PATH [PATH...]
Which: outputs a Houdini customization script for OpenVDB nodes
Where: PATH is the pathname of a compiled OpenVDB Houdini plugin
...
""" % locals())
```
and add a comment header to the output script indicating how it was generated and maybe listing the OpenVDB version and any relevant build options.

openvdb_houdini/vdbopcustomize.py
```diff
@@ -0,0 +1,52 @@
...
+
+aswfKeys.sort()
+nativeKeys.sort()
+
+# print ophide statements for all SOPs, disable with comments
+# those that need not be hidden
+
+opHide = '{comment}ophide Sop {name}'
+
+print '// Native VDB SOPs'
```
pcucka: For Python 3 compatibility, use `print('...')` rather than `print '...'`.

pcucka:
With the new op customization script, you can now remove all of the `OPENVDB_HOUDINI_DEFAULT_POLICY` logic from the CMake script and revert all of your changes to the SOPs and to `ParmFactory` (except maybe the calls to `setOpTabSubMenuPath`, but you can make those conditional on the node name ending with `"(ASWF)"`).  If you must have an `OPENVDB_OPHIDE_POLICY` environment variable, handle it entirely in the `pythonrc.py` customization script.  Even the label suffix logic can go in the Python script, though you might have to do something like
```
if os.getenv('OPENVDB_OPHIDE_POLICY') == '0':
    hou.hscript('oprename Sop {name} "{label} (ASWF)"'.format(...))
```
because there doesn't appear to be a HOM equivalent for `oprename`.

It would of course be better to use meaningful names rather than integers for the `OPENVDB_OPHIDE_POLICY` states.

danrbailey:
Closing this PR now. After much discussion, this has been split out into three separate PRs - #456, #461 and #464.

PR #405, opened 2019-05-02
CMake Style Improvements

danrbailey:
This enables the trailing space check for all CMake files and addresses a few style issues:

* lower-case method names
* no spacing around parentheses
* end parenthesis in-line with method name

Although CMake is very permissive with regards to syntax, the above seems to be the most common style so I suggest we go with convention:

https://community.kde.org/Policies/CMake_Coding_Style
https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/
https://style-guides.readthedocs.io/en/latest/cmake.html
https://github.com/imageworks/OpenColorIO/blob/master/CMakeLists.txt

As we've seen, we tend to get a lot of pull requests around CMake, so I think it makes sense to resolve this before the release.


Idclip:
This PR hasn't updated the find modules? Should we not do those as well?

danrbailey:
> This PR hasn't updated the find modules? Should we not do those as well?

Yeah, I think so. I just updated them (and deleted the FindGLEW module for the time being as we don't yet support GLEW).

Idclip:
Great, thanks a lot for doing this @danrbailey

PR #406, opened 2019-05-02
Futher attribute set copy constructor fixes for non constant strides

Idclip:


openvdb/unittest/TestAttributeSet.cc
```diff
@@ -598,23 +602,35 @@ TestAttributeSet::testAttributeSet()
 
         CPPUNIT_ASSERT_EQUAL(openvdb::Index(50), attrSet.getConst("float1")->size());
 
-        AttributeSet attrSet2(attrSet, /*arrayLength=*/200);
+        // error as the new length is greater than the data size of the
+        // 'boolNonUniform' attribute
```
danrbailey: I'm sure we've talked about this, but a git blame shows that this ValueError has been there since the AttributeArray was first contributed to OpenVDB. What's the reason for this particular check again?

Also, remind me what you're using at DNeg - have you exposed support for dynamic attribute arrays from Houdini? I know in the public repo, it's only available from the API.

openvdb/unittest/TestAttributeSet.cc
```diff
@@ -581,13 +581,17 @@ TestAttributeSet::testAttributeSet()
...
+        attrSet.appendAttribute("boolNonUniform", AttributeB::attributeType(), /*size=*/100, false);
+        attrSet.appendAttribute("intNonUniform", AttributeI::attributeType(), /*size=*/300, false);
 
         CPPUNIT_ASSERT_EQUAL(std::string("float"), attrSet.getConst("float1")->type().first);
         CPPUNIT_ASSERT_EQUAL(std::string("int32"), attrSet.getConst("int1")->type().first);
         CPPUNIT_ASSERT_EQUAL(std::string("float"), attrSet.getConst("float3")->type().first);
         CPPUNIT_ASSERT_EQUAL(std::string("vec3s"), attrSet.getConst("vector")->type().first);
         CPPUNIT_ASSERT_EQUAL(std::string("vec3s"), attrSet.getConst("vector3")->type().first);
         CPPUNIT_ASSERT_EQUAL(std::string("bool"), attrSet.getConst("bool100")->type().first);
+        CPPUNIT_ASSERT_EQUAL(std::string("bool"), attrSet.getConst("boolNonUniform")->type().first);
```
danrbailey: I think I'd refer to this as dynamic rather than non uniform to avoid confusion, as we usually refer to arrays as uniform in the context of uniform value compression.

openvdb/unittest/TestAttributeSet.cc
Idclip: > What's the reason for this particular check again?

I guess that it just doesn't generally make sense to hold a dynamic array with less values than points (size)? And that you can very easily index into invalid data. Perhaps there were more implementation specific reasons but I can't remember them now.

> Also, remind me what you're using at DNeg

afaik we're not actively using non constant (dynamic) strides

Idclip: The reason I caught this is because of our point merge code which does support this (and it threw up a issue on syncing form the public repo)

danrbailey: > I guess that it just doesn't generally make sense to hold a dynamic array with less values than points (size)? And that you can very easily index into invalid data. Perhaps there were more implementation specific reasons but I can't remember them now.

I suppose that's dependent on whether or not you have points with zero values in the dynamic array. I'm not sure it makes sense to disallow that, but this PR probably isn't the right place to change this.

> The reason I caught this is because of our point merge code which does support this (and it threw up a issue on syncing from the public repo)

A point merge unit test? We should really get dynamic attribute array support into Houdini, it's come up a couple of times recently that this is missing.

Idclip: Agreed on both counts

openvdb/unittest/TestAttributeSet.cc
Idclip: done

Idclip:
Don't know whats going on with the ci on this pr?

danrbailey:
> Don't know whats going on with the ci on this pr?

I'm not sure either. I tried merging master into it to kick the builds, but that doesn't seem to have helped.

I suggest closing and opening a new PR. If that doesn't work, it might be related to the fact that as far as I can tell you don't have CircleCI enabled on your fork?

Idclip:
Closing this PR to try and resolve CI issues

PR #407, opened 2019-05-02
Removed -Wfloat-conversion from the conversion warnings as it's impli…

Idclip:
…ed by -Wconversion

Note that `-Wfloat-conversion` was only introduced in GCC 4.9 so 4.8 (for Houdini 16.5) complains that it's unknown. It's also implied by `-Wconversion` so I removed it instead of branching

### DIFF
--- a/openvdb/Platform.h
+++ b/openvdb/Platform.h
@@ -202,7 +202,10 @@
 /// int valueAsInt = value;
 /// OPENVDB_NO_TYPE_CONVERSION_WARNING_END
 /// @endcode
-#if defined __GNUC__
+#if defined __INTEL_COMPILER
+    #define OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN
+    #define OPENVDB_NO_TYPE_CONVERSION_WARNING_END
+#elif defined __GNUC__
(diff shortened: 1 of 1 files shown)


openvdb/Platform.h
```diff
@@ -205,8 +205,7 @@
 #if defined __GNUC__
```
pcucka: Could you please also change this line (205) to the following?
```
#if defined __INTEL_COMPILER
    #define OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN
    #define OPENVDB_NO_TYPE_CONVERSION_WARNING_END
#elif defined __GNUC__
```
ICC actually defines `__GNUC__`, but it doesn't recognize GCC pragmas.

Idclip: Sure, done

danrbailey:
This actually fails in our CI:

https://circleci.com/gh/AcademySoftwareFoundation/openvdb/47

Perhaps this means a pragma warning for conversion _doesn't_ imply float-conversion. Either way, I think we need to branch for GCC >= 4.9 as you suggested. I'm going to re-enable this...

PR #408, opened 2019-05-02
Reverted -Wfloat-conversion change and instead gated by GCC version

danrbailey:


openvdb/Platform.h
```diff
@@ -206,9 +206,17 @@
     #define OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN
     #define OPENVDB_NO_TYPE_CONVERSION_WARNING_END
 #elif defined __GNUC__
-    #define OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN \
-        _Pragma("GCC diagnostic push") \
-        _Pragma("GCC diagnostic ignored \"-Wconversion\"")
+    // -Wfloat-conversion was only introduced in GCC 4.9
```
Idclip: Use the `OPENVDB_CHECK_GCC` macro

danrbailey: Done, I didn't know about this macro.

PR #411, opened 2019-05-03
Conditionally disable POP code for Houdini 17.5

pcucka:
It appears that POPs are no longer supported as of H17.5.


danrbailey:
> It appears that POPs are no longer supported as of H17.5.

I'm curious why builds haven't been failing when building against H17.5 without this change?

pcucka:
That code is enabled only when `OPENVDB_USE_LOG4CPLUS` is defined.  We're not building with log4cplus, are we?

danrbailey:
> That code is enabled only when `OPENVDB_USE_LOG4CPLUS` is defined. We're not building with log4cplus, are we?

Got it. Yeah, I've updated this ticket ([OVDB-89](https://jira.aswf.io/browse/OVDB-89)) with some CI-related tasks and included testing Log4CPlus in that.

PR #415, opened 2019-05-07
Further attribute set copy constructor fixes for non constant strides

Idclip:
Copy of  #406


Idclip:
Creating a new PR hasn't fixed it. I found this https://circleci.com/docs/2.0/workflows-waiting-status/ but as others aren't having this issue I assume it's unrelated? I'm going to try setting up Circle on my fork, but it's pretty ridiculous that I have to give circle read and write access to all organisations and private repositories I'm a member of (travis is the same) and, from what it looks like, set up circle jobs specifically on forks of repositories I want it to run on. Seems to only like running on local branches https://circleci.com/gh/AcademySoftwareFoundation.

Doesn't look Appveyor needs any of this? Are we expecting other users to have to go through this process too?

danrbailey:
> Creating a new PR hasn't fixed it. I found this https://circleci.com/docs/2.0/workflows-waiting-status/ but as others aren't having this issue I assume it's unrelated?

I don't think this is related.

> I'm going to try setting up Circle on my fork, but it's pretty ridiculous that I have to give circle read and write access to all organisations and private repositories I'm a member of (travis is the same) and, from what it looks like, set up circle jobs specifically on forks of repositories I want it to run on. Seems to only like running on local branches https://circleci.com/gh/AcademySoftwareFoundation.
> 
> Doesn't look Appveyor needs any of this? Are we expecting other users to have to go through this process too?

No, I expect that any PR will use the ASWF Circle account (and allowance) for builds and not require you to need to sign up to Circle. That does appear to be the case from what I can tell. It seems like this issue is specific to you for some reason, I only suggested signing up to Circle in case it helped us understand what was going on, I don't think it's the solution.

Appveyor and Circle work quite differently. Appveyor has it's own limitations in that it doesn't understand GitHub organizations so you have to have it tied to a specific GitHub user account.

Idclip:
> It seems like this issue is specific to you for some reason, I only suggested signing up to Circle in case it helped us understand what was going on, I don't think it's the solution.

Signing up for circle seems to have fixed it but I'm still not sure why. Lets get this in for 6.1 then take another look at this.

danrbailey:
> > It seems like this issue is specific to you for some reason, I only suggested signing up to Circle in case it helped us understand what was going on, I don't think it's the solution.
> 
> Signing up for circle seems to have fixed it but I'm still not sure why. Lets get this in for 6.1 then take another look at this.

I encountered this issue myself (https://github.com/AcademySoftwareFoundation/openvdb/pull/418) despite having circle installed on my own fork, so it's probably just coincidence that re-submitting the PR after signing up for circle resolved the issue for you.

On investigation, I think it might be related to us using the old GitHub Commits API for the ASWF rather than the new GitHub Checks API, so I've mailed Andrew/ASWF HelpDesk to ask about enabling it.

PR #416, opened 2019-05-07
Bumping TBB version from 4.4.0 to 4.4.6 for ABI=4 CI build

danrbailey:
There are a number of memory allocation bug fixes in between TBB 4.4.0 and 4.4.6, so bumping to a later version of TBB 4.4 in the hope of improving stability in the ABI=4 build on Circle. I've yet to see a 4.4.6 build fail.

(This is a CI change so not required for the release).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,8 +52,8 @@ _tbb:
     name: Install TBB 2017
     command: bash ci/install_tbb.sh 2017_U6
   tbb4_4: &install_tbb4_4
-    name: Install TBB 4.4
-    command: bash ci/install_tbb.sh 4.4
+    name: Install TBB 4.4.6
+    command: bash ci/install_tbb.sh 4.4.6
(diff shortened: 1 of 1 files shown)


Idclip:
How confident are we that this is the case of the intermittent faults? Can we bump the minimum version requirements?

danrbailey:
> How confident are we that this is the cause of the intermittent faults? Can we bump the minimum version requirements?

Not very. I wouldn't touch the version requirements until we've had builds run against this successfully for a period of time.

Idclip:
Perhaps add a comment under the dependency table along with blosc for tbb 4.4

danrbailey:
> Perhaps add a comment under the dependency table along with blosc for tbb 4.4

This change is specific to the CI. I think we should address any changes we wish to make to our documentation in a separate PR, ideally once we have a bit more data to back it up.

PR #418, opened 2019-05-08
Downgrade Circle CI to use 2.0

danrbailey:
None of the 2.1 features are currently being used, so revert back to the older version to attempt to improve stability.


Idclip:
Thanks for looking into this @danrbailey. Do we know this for certain to be the fix or is this just a test? A comment in the yml would be good too

danrbailey:
> Thanks for looking into this @danrbailey. Do we know this for certain to be the fix or is this just a test? A comment in the yml would be good too

I just don't know - going to re-submit this PR now.

PR #424, opened 2019-05-10
Fix crash with VDB Combine in Copy B

jmlait:
VDB Combine SOP crashes in Copy B if destination is not a VDB.  Avoid this.


openvdb_houdini/houdini/SOP_OpenVDB_Combine.cc
```diff
@@ -807,7 +807,8 @@ SOP_OpenVDB_Combine::Cache::cookVDBSop(OP_Context& context)
 
             // Remove primitives that were copied from input 0.
             for (GU_PrimVDB* vdb: vdbsToRemove) {
-                gdp->destroyPrimitive(*vdb, /*andPoints=*/true);
+                if (vdb)
+                    gdp->destroyPrimitive(*vdb, /*andPoints=*/true);
```
pcucka: Better:
```
vdbsToRemove.erase(nullptr);
for (GU_PrimVDB* vdb: vdbsToRemove) {
    gdp->destroyPrimitive(*vdb, /*andPoints=*/true);
}
```

jmlait: I'm not sure I agree.
First, it is two passes over the array rather than one, so is redundant.
Second, it separates the enforcement of the no-null constraint from where it is needed.  Binding it tightly inside the for loop makes it clear to readers both that there may be nulls in the array *and* that the destroyPrimitive can't handle nulls.  Having it outside the loop hides this.  It creates a tight coupling between the erase() call and the for loop.   Future code edits could cause the erase() call to wander elsewhere, or new code injected between it and where the for loop runs.

Best solution would be to rewrite the logic upstream and instead guarantee that vdbsToRemove never contains nulls.   I'm not even entirely sure what it means to Copy B into a primitive that isn't a VDB.  But there is a lot of logic in needsA and needsB I didn't want to dive into.

pcucka: `vdbsToRemove` is an STL set, so lookups are logarithmic: log N null pointer comparisons vs. N comparisons.

Yeah, I don't want to revisit that upstream logic either; too many special cases to handle: unary vs. binary operations, pairwise vs. one-to-many, more As than Bs, more Bs than As, incompatible A and B grid types, etc.

jmlait: Ah, I usually use vectors for that purpose.  Glad I didn't try to mess with the upper logic :>

I hold with the second argument, however, that is best to make this dependency strongly coupled.  If it is a set, there is even less of a performance issue as that iterator will drown any cost of a test.  I'm more concerned about minimal code and keeping coupled code bound together, both of which the if test achieves.

How strong are your objections?  I'm reluctant to push this change back the other way through the pipeline for a minor stylistic issue.  I kind of hoped also this change could go through quickly  before the changes.txt will trigger a conflict....

danrbailey: Quick reminder - Peter is off this week. I don't mind either approach, so I've given you an approval in case you decide you want to dismiss Peter's review and follow up with him when he's back...

jmlait: Thank you Peter!

PR #435, opened 2019-05-15
OPENVDB_USE_DEPRECATED_ABI usage improvements

Idclip:

Looking for feedback on the `version.h`->`openvdb.cc` move of the deprecated error directive. We have case where we use OpenVDB with a DCC which requires ABI 3 for compatibility. Skipping this in the VDB build is fine, but then **every** piece of software which goes on to require vdb now also needs to specify this flag. imo this seems wrong and this decision should be made once at the time of building the openvdb library. 

Moving this option entirely to the CMake would also solve this problem, however after discussion with @danrbailey, it seems this is an important enough error to keep within the code.


Idclip:
I strongly disagree. The fact that it may be a different developer is even more of a reason to not have to specify this flag for every bit of downstream software. Lets put it in the context of a package manager like pkg-config, or even the current FindOpenVDB.cmake. Querying the expected installation flags from something like pkg-config --cflags should not result in a build setup which errors by default, no matter what's deprecated and what's not. There is a good chance users may now accidentally leave this bypass flags in their configurations due to this error.

> Kind of important to be made aware that this change is coming in my opinion and to be able to schedule some time to have that DCC walled off to use an older version of VDB if necessary or otherwise transition to a later version of the DCC.

Users are already subject to literally tonnes of deprecation errors for ABI 3 (as they should be because it will be dropped in the **next** release). The current behaviour of causing all downstream software to fail is far too intrusive when support hasn't been removed. The progression should be warning->error.

pcucka:
Maybe include the check in both `openvdb.cc` and `version.h`, but downgrade it to a pragma message in the latter?  Just a thought.

Idclip:
> Maybe include the check in both openvdb.cc and version.h, but downgrade it to a pragma message in the latter? Just a thought.

Great idea, I'll do that

Idclip:
Sorry for the delay on this, I've added back the check in `version.h` as a warning

pcucka:
This change probably merits an entry in the "ABI changes" section of the release notes, something like

> ABI versions 3 and older are now deprecated, and support for them will be removed in a future release.  Until then, define the macro `OPENVDB_USE_DEPRECATED_ABI` (or set the CMake `OPENVDB_USE_DEPRECATED_ABI` option to `ON`) to suppress deprecation messages when compiling OpenVDB or dependent code.

Idclip:
> This change probably merits an entry in the "ABI changes" section of the release notes, something like

Sure I can add that - though I think the change that introduced this compile time error exists in 6.1.0

pcucka:
Yes, you're right.  But better late than never, I guess.   :sweat_smile:

Idclip:
Will merge this eod if there are no other comments

PR #436, opened 2019-05-15
OVDB-91: Improve Performance of VDB Delayed Loading

danrbailey:
This is a significant change, so I'll start out with the stats and then explain the details.

I benchmarked using the high-resolution Disney cloud dataset from Moana (https://www.technology.disneyanimation.com/clouds) which comprises a single float grid of 1.49 billion voxels and occupying 2.73GB on disk with Blosc compression. I tested I/O using both a local SSD and an NFS mounted drive on RHEL7 based on timing the following snippet. I re-wrote the cloud dataset to add delayed load metadata and then timed against this new VDB file.

```
openvdb::io::File file("wdas_cloud.vdb");
file.open(/*delayLoad=*/true);
auto grids = file.getGrids();
file.close();
```

Without this change
* SSD - 12.5s
* NFS - 28.3s

With this change
* SSD - 2.35s (5.3x)
* NFS - 3.55s (8.0x)

This change extracts the bytes associated with active mask, Blosc and Zip compression that are scattered throughout the file and puts them into a new Blosc-compressed DelayedLoadMetadata at the head of the file. By doing this, it is possible to seek over the entire leaf node portion of the file while setting up delayed-loading memory map and pull these bytes from this new metadata without having to move the stream head, which is very expensive.

For this 2.73GB cache, the new metadata occupied just 4.5MB (ie 0.16% of the overall file size) and increased the write time by 1.4s using 24 threads (of a total of 28.8s or 4.9%).

Important to mention that this change is both backwards **and** forwards compatible. If you write out a VDB file with this DelayedLoadMetadata and attempt to read it with an older version of VDB that doesn't have this change, it will simply ignore it and read as normal. If you attempt to read a file without this DelayedLoadMetadata in a version of VDB that has this change, it will read as normal. One small subtlety to highlight - if you read and write the VDB with a version < 5.0, the metadata will just be dropped, if you use a version >= 5.0 (but less than 6.2 or whatever version this ends up in), the metadata will be blindly re-written into the new file, but on read with version >= 6.2 ignored as the version the file was written in was < 6.2. We have tested this approach thoroughly at ILM over the last year or so using many different applications and versions and haven't run into any compatibility issues.

Note that many vendor applications have disabled delayed-loading by default (Houdini, Renderman, etc), so I see this as a first step towards making delayed-loading more useful, because until now it has only really offered a benefit to memory usage.


danrbailey:
> I like this idea. We do something similar with .bgeo where we store a header object that allows fast stat.
> 
> Note that GEO_VDBTranslator has a fileStat() function. Could this be used to speed up that function? (Not suggesting it be added to this PR, just asking if there exists enough metadata saved...)

Yes, by enabling the delay-load flag, it should be possible to speed up fileStat() with this change. Related to this, I also have some GEO_VDBTranslator delay-loading changes to propose - right now using an openvdb::io::Stream in fileLoad() wipes out all benefits of delay-loading because it copies the entire file into a temporary location (we've been successfully switching back to using openvdb::io::File under certain conditions at ILM for a while now). I thought I'd tackle these improvements in two stages though, because that'll also open the question of the copy max bytes feature.

> I noticed metadata is capped at 2^32. I'd rather it capped at 2^31. This way in the future you can use -1 as a flag to store that a int64 size will follow, which lets you be unbounded without affecting anything already using this.

That's a good suggestion, I'll make the change to the local limit here. I think the full solution will likely need an ABI=7 change - tweaking the return type of the Metadata::size() method to return Index64 instead of Index32 as well as an update to UnknownMetadata. We'll obviously need to wait for the older methods to age out before being able to take advantage of it, but I agree it's worth putting that mechanism in place now.

danrbailey:
Added a few extra changes to this PR - prefixed the metadata typename with "__" and ignored if read as unknown metadata (ie DelayedLoadMetadata is not registered). 

This also always drops the metadata from the grid now, so it's only available from the StreamMetadata grid metadata. This was a bit tricky to test because I could no longer verify that the version mechanism was working so I've had to add __test() and __setTest(uint32_t) methods to StreamMetadata in order to be able to put it into a mode where the grid metadata is not dropped. By setting this flag from the unit test, I can confirm that the metadata is pre-emptively dropped under certain conditions.

danrbailey:
> > I noticed metadata is capped at 2^32. I'd rather it capped at 2^31. This way in the future you can use -1 as a flag to store that a int64 size will follow, which lets you be unbounded without affecting anything already using this.
> 
> That's a good suggestion, I'll make the change to the local limit here. I think the full solution will likely need an ABI=7 change - tweaking the return type of the Metadata::size() method to return Index64 instead of Index32 as well as an update to UnknownMetadata. We'll obviously need to wait for the older methods to age out before being able to take advantage of it, but I agree it's worth putting that mechanism in place now.

I took a look at this issue and broke it out into a new task ([OVDB-104](https://jira.aswf.io/browse/OVDB-104)). I think doing this is a good idea, but it would be best done as an orthogonal change at the MetaMap level (Metadata::size() returns the intended size of the data to be written out so any limit can be enforced globally).

danrbailey:
@pcucka - I looked into your questions regarding this PR:

1) vdb_print uses io::File::readGrid() which loads the metadata along with the grid and drops the delayed load metadata from the grid as expected, so it's not displayed in the metadata output. I think that's what we would expect here.

2) File::readAllGridMetadata() **doesn't** drop the delayed load metadata as it uses a different routine which doesn't contain the dropping logic. Although it might be somewhat useful to provide one mechanism where you can actually get at this metadata for whatever reason, perhaps it's still more correct for us to drop this metadata from this routine and to be consistent with File::readGrid(). What do you think?

danrbailey:
> I would prefer the two to be consistent (that is, for `readAllGridMetadata` to drop the delayed load metadata).

I agree, I shall make that change. Thank you.

danrbailey:
> I haven't looked through this carefully. I guess no one else has either.

Thanks for approving, I'm going to merge this. The original feature has been through quite a lot of testing at ILM dating back almost two years. The PR has been open over two months, has complete unit test coverage and has been discussed fairly extensively in the TSC. I'll also note that Houdini disables delay-loading by default.

Hopefully we'll catch any unforeseen issues in the interim period before we release 6.2.0.

PR #428, opened 2019-05-11
Fixes for Python3 and loading the python module on MacOS

Idclip:
Signed-off-by: Nick Avramoussis <4256455+Idclip@users.noreply.github.com>

Issue #427, opened 2019-05-11
Build fails with python-3.6


yurivict:
```
/usr/ports/misc/openvdb/work/openvdb-6.1.0/openvdb/python/pyOpenVDBModule.cc:323:14: error: use of undeclared identifier 'PyInt_Check'
        if (!PyInt_Check(obj)) return nullptr; // not a Python integer
             ^
/usr/ports/misc/openvdb/work/openvdb-6.1.0/openvdb/python/pyOpenVDBModule.cc:340:39: error: use of undeclared identifier 'PyInt_AsLong'; did you mean 'PyLong_AsLong'?
        *index = static_cast<IntType>(PyInt_AsLong(obj));
                                      ^~~~~~~~~~~~
                                      PyLong_AsLong
/usr/local/include/python3.6m/longobject.h:23:18: note: 'PyLong_AsLong' declared here
PyAPI_FUNC(long) PyLong_AsLong(PyObject *);
                 ^
2 errors generated.
```

FreeBSD 11.2

Idclip:
Thanks a lot for reporting, PR created here: https://github.com/AcademySoftwareFoundation/openvdb/pull/428

Idclip:
Merged, please feel free to re-open or create a new issue if anything else crops up!

Issue #78, opened 2016-10-14
Support for building python module for multiple versions of python in new CMake files


dracwyrm:
I noticed that the new CMake files don't have support for building the python module, unless I missed that.

Anyways, I would like to request that the build system should somehow take a list of all versions to build the python module for, like 2.7, 3.4, 3.5... I use Gentoo where we can have multiple versions installed to support various packages.

This should also mean that auto-finding the python implementations should be overrided since there is a default that /usr/bin/python, /usr/bin/python2, /usr/bin/python3 link to, so I need to set the path to the correct version which our build system does for us.

Thanks,
Jon

nyue:
Python module is now being built and installed pending

https://github.com/dreamworksanimation/openvdb/pull/77

Cheers

dracwyrm:
@nyue Correct me if I'm wrong, but it doesn't look like there is a way to manually select the python binary and library paths???

nyue:
@dracwyrm I am in the process of verifying Python 3.x with OpenVDB, I need to rebuild existing dependency for Python 3.x before I can commence verification. Meanwhile, I do have a situation where I need to specific specific location for Python 2.x and I do so with the following CMAKE variable that the FindPythonLibs module will use. They are PYTHON_EXECUTABLE, PYTHON_LIBRARY and PYTHON_INCLUDE_DIR

NOTE : I assume you will have your Boost Python built with Python 3.x

A concrete example of one of my test environment (CentOS 6.7) build is achieved via the following

export ILMBASE_ROOT=$RHOME/Systems/OpenEXR/v2.2.0
export OPENEXR_ROOT=$RHOME/Systems/OpenEXR/v2.2.0
export BOOST_ROOT=$RHOME/Systems/boost/v1.61.0
export TBB_ROOT=$RHOME/Systems/tbb/tbb44_20150728oss
export GLFW3_ROOT=$RHOME/Systems/glfw/v3.2.1
export BLOSC_ROOT=$RHOME/Systems/c-blosc/v1.7.0
cmake \
	-D PYTHON_LIBRARY=/sw/external/COS6/python/2.6.4/lib/libpython2.6.so \
	-D PYTHON_INCLUDE_DIR=/sw/external/COS6/python/2.6.4/include/python2.6 \
	-D CMAKE_CXX_COMPILER=/sw/external/COS6/gcc/4.8.2/bin/g++ \
[...]

PR #437, opened 2019-05-17
OVDB-71: Update INSTALL file minimum versions and add link to build documentation

danrbailey:
There's a good argument to remove this file entirely, but for the time being, just update the versions and add a link to point to the CMake build documentation.

### DIFF
--- a/openvdb/INSTALL
+++ b/openvdb/INSTALL
@@ -56,7 +56,7 @@ Optional:
 
 - OpenGL 3.2 or later, for the OpenVDB viewer
 
-- Python 2.7, for the Python module
+- Python 2.7 or later, for the Python module
 
 - NumPy (www.numpy.org), version 1.12.1 or later, for the Python module
 
@@ -69,9 +69,9 @@ Other versions might work but have not been tested.
 Installation
 ------------
 
-This documentation is for building with GNU Make which is now deprecated, it
-is strongly recommended to use CMake to build OpenVDB as this will be removed
-in an upcoming release.
+This documentation is for building with GNU Make. It is strongly recommended to
+use CMake to build OpenVDB as the Makefiles will be removed in an upcoming
+release.
 See the build documentation here -
 https://www.openvdb.org/documentation/doxygen/build.html for more information.
 
--- a/openvdb/INSTALL
+++ b/openvdb/INSTALL
@@ -3,6 +3,11 @@ Installing OpenVDB
 
 Requirements
 ------------
+
+See the dependencies documentation here -
+https://www.openvdb.org/documentation/doxygen/dependencies.html for more
+information.
+
 - A C++11-compatible compiler, such as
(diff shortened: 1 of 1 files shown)


openvdb/INSTALL
```diff
@@ -52,9 +51,9 @@ Optional:
 
 - OpenGL 3.2 or later, for the OpenVDB viewer
 
-- Python 2.5, 2.6 or 2.7, for the Python module
+- Python 2.7, for the Python module
```
pcucka: This can be "2.7 or later", since we support at least some versions of Python 3.

openvdb/INSTALL
```diff
@@ -64,6 +63,11 @@ Other versions might work but have not been tested.
 
 Installation
 ------------
+
+This documentation is for building with GNU Make, it is recommended to use
```
pcucka: Should this be worded more strongly, to say that Make support is deprecated and will be dropped in a future release?

openvdb/INSTALL
```diff
@@ -52,9 +51,9 @@ Optional:
```
pcucka: I think we just dropped support for GLFW 2.  The minimum version is now 3.0.

openvdb/INSTALL
```diff
@@ -4,39 +4,38 @@ Installing OpenVDB
 Requirements
```
pcucka: Maybe add a link here to the new Requirements web page?

openvdb/INSTALL
danrbailey: I thought the point in this file was to only include what we explicitly test for:

`Other versions might work but have not been tested.`

Nick fixed an issue in Python 3 recently (#428) that was reported to us - https://github.com/AcademySoftwareFoundation/openvdb/issues/427. I'm not sure we should include anything here that we don't know is actively being tested, ideally through CI?

danrbailey:
Thanks @pcucka, just addressed your feedback.

openvdb/INSTALL
pcucka: Perhaps, but then we should really drop all of the "or later" qualifiers.

We've supported Python 3 for some time now (see, e.g., PR #78 and the Fedora [openvdb-python3](https://apps.fedoraproject.org/packages/openvdb-python3) package).

openvdb/INSTALL
```diff
@@ -64,8 +69,10 @@ Other versions might work but have not been tested.
 Installation
 ------------
 
-This documentation is for building with GNU Make, it is recommended to use
-CMake to build OpenVDB. See the build documentation here -
+This documentation is for building with GNU Make which is now deprecated, it
+is strongly recommended to use CMake to build OpenVDB as this will be removed
```
pcucka: How about "It is strongly recommended to use CMake to build OpenVDB as the Makefiles will be removed in an upcoming release"?

danrbailey:
@pcucka - just addressed your last two comments.

PR #442, opened 2019-05-21
OVDB-55: In-place Modify Value

danrbailey:
Modify voxel values in-place to improve performance, particularly for custom value types. In adding a unit test that verified no value copy was taking place, I noticed that tree destruction redundantly copies tile values. As expected, fixing this only has a minor impact on performance, but still seems worthwhile so I included it in this PR.


openvdb/tree/Tree.h
```diff
@@ -1487,17 +1487,17 @@ template<typename RootNodeType>
 inline void
 Tree<RootNodeType>::clear()
 {
-    std::vector<LeafNodeType*> leafnodes;
```
pcucka: Rather than trying to change the behavior of `stealNodes`, can't you just reimplement `Tree::clear` using `NodeManager`?

@kmuseth Is the current implementation even correct?  Why does it steal only leaf nodes and first-level internal nodes and not second-level internal nodes?  Would it work for trees with more than four levels?

danrbailey: That's a good point., a NodeManager::foreachBottomUp() should work for this.

danrbailey: I had a go at this and would quite like to hear your thoughts Peter?

pcucka: That looks pretty good.  I'm not sure the `const_cast` is necessary, since `getChild` is const, but it might be safer to leave it in place and do something like this:
```
auto& item = children[iter.pos()];
delete item.getChild();
item.setChild(nullptr);
```
I envisioned the leaf level being a no-op; is there a measurable gain from swapping out the buffer?

openvdb/tree/Tree.h
```diff
@@ -1223,6 +1224,47 @@ class Tree: public TreeBase
...
+            }
+        }
+
+        // bool leaf buffers have no value buffer to be deallocated
+        template <>
+        inline void deallocateLeafBuffer<bool>(typename LeafNodeType::Buffer&) const { }
+
+        void operator()(LeafNodeType& leaf) const
+        {
+            this->deallocateLeafBuffer<ValueType>(leaf.buffer());
```
pcucka: To fix the build failures, maybe use tag dispatch instead:
```
this->deallocateLeafBuffer(leaf.buffer(), ValueType());
```
where `deallocateLeafBuffer` is defined as
```
template<typename ValueT>
void deallocateLeafBuffer(typename LeafNodeType::Buffer& buffer, const ValueT&) const
{
    ...
}
```
and then just overloaded for `bool`:
```
void deallocateLeafBuffer(typename LeafNodeType::Buffer& buffer, bool) const {}
```

danrbailey: Ah yeah, funnily enough, my compiler doesn't complain about this.

How would you feel about adding a LeafBuffer::clear() method that contains the following and could be called from the deconstructor:

```
if (this->isOutOfCore()) {
    this->detachFromFile();
} else {
    this->deallocate();
}
```
I find I use this swapping trick quite a lot for things like streaming where I know I only want to read from each leaf node once. It would be nice to officially support destroying a leaf buffer.

pcucka: That sounds reasonable.  Maybe it deserves a separate PR though?

openvdb/tree/Tree.h
danrbailey: > That looks pretty good. I'm not sure the `const_cast` is necessary, since `getChild` is const, but it might be safer to leave it in place and do something like this:
> 
> ```
> auto& item = children[iter.pos()];
> delete item.getChild();
> item.setChild(nullptr);
> ```

Yeah, that works. Not sure it's necessary to set the child to be a nullptr because it looks like everything checks the child mask before accessing it, but agree that it's still perhaps safer.

> I envisioned the leaf level being a no-op; is there a measurable gain from swapping out the buffer?

Well it depends on the tree configuration. With standard configuration, you could in theory have only 4096 leaf nodes in your tree all of which hang off a single internal node that would otherwise be deallocated from a single thread, right? Even in more typical scenarios, you'll likely still end with a long tail as the grain size is so high. Better to maximize the parallelization by doing a parallel leaf-level pass to drop the buffers first I think.

openvdb/tree/Tree.h
danrbailey: Yes, I think so. Only reason I bundled this change in here is because my value type threw in my unit test and deallocation was causing that to fail. The scope has clearly changed though, so I'll adjust my unit test.

PR #445, opened 2019-05-24
OVDB-94: makeGridUnique fix for Fill SOP

pcucka:
I would actually prefer `GEOvdbApply` to be a member of the primitive (i.e., `GEO_PrimVDB::apply`), but then for older Houdini versions we'd still need a free function.

### DIFF
--- a/openvdb/doc/changes.txt
+++ b/openvdb/doc/changes.txt
@@ -8,22 +8,33 @@
 
 @par
 Improvements:
-- FindOpenVDB.cmake now correctly propagates CXX version requirements.
+- <TT>FindOpenVDB.cmake</TT> now correctly propagates @c CXX version
+  requirements.
 - Improved directory and file path lookups of some CMake commands in
-  the root CMakeLists.txt <I>[Reported by Daniel Elliott]</I>
+  the root <TT>CMakeLists.txt</TT>.
+  <I>[Reported&nbsp;by&nbsp;Daniel&nbsp;Elliott]</I>
 
 @par
 Bug fixes:
-- Fixed a bug in @vdblink::tools::computeScalarPotential() computeScalarPotential
-  @endlink that could produce a corrupt result due to invalid memory access.
-  <I>[Reported by Edwin Braun]</I>
+- Fixed a bug in
+  @vdblink::tools::computeScalarPotential() computeScalarPotential@endlink
+  that could produce a corrupt result due to invalid memory access.
+  <I>[Reported&nbsp;by&nbsp;Edwin&nbsp;Braun]</I>
 
 @par
 Python:
-- Updated python module CMake to always produce a .so for the python plugin
+- CMake now always produces a <TT>.so</TT> for the Python module
   on UNIX platforms.
-- Python 3 support improvements to python/pyOpenVDBModule.cc
-  <I>[Reported by yurivict]</I>
+- Fixed a compile-time error when building the Python module for Python&nbsp;3.
+  <I>[Reported&nbsp;by&nbsp;yurivict]</I>
+
+@par
+Houdini:
+- Added <B>openvdb_houdini::GEOvdbApply</B>, which invokes a functor
+  on a VDB primitive if the resolved grid type is a member of
+  a given type list.
+- Fixed a regression in the Fill&nbsp;SOP that caused it to modify VDBs
+  in the input detail.
 
 
 @htmlonly <a name="v6_1_0_changes"></a>@endhtmlonly
--- a/openvdb_houdini/Utils.h
+++ b/openvdb_houdini/Utils.h
@@ -369,11 +369,8 @@ GEOvdbApply(GEO_PrimVDB& vdb, OpT& op, bool makeUnique = true)
             if (treePtr.use_count() > 2) { // grid + treePtr = 2
                 // If the grid resolves to one of the listed types and its tree
                 // is shared with other grids, replace the tree with a deep copy.
-                gridPtr->apply<GridTypeListT>([](Grid& baseGrid) {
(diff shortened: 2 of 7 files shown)


openvdb/doc/changes.txt
```diff
@@ -8,31 +8,40 @@
...
 
 @par
 Python:
-- Updated python module CMake to always produce a .so for the python plugin
+- CMake now always produces a <TT>.so</TT> for the Python module
   on UNIX platforms.
-- Python 3 support improvements to python/pyOpenVDBModule.cc
-  <I>[Reported by yurivict]</I>
+- Fixed a compile-time error when building the Python module for Python&nbsp;3.
+  <I>[Reported&nbsp;by&nbsp;yurivict]</I>
```
Idclip: @pcucka thanks for improving these. bar the re-wording, what should I be doing on my end to catch these issues? Referring to the new lines you've added and the use of &nbsp

openvdb_houdini/Utils.h
```diff
@@ -339,6 +339,68 @@ using VolumeGridTypes = ScalarGridTypes::Append<Vec3GridTypes>;
...
+template<typename GridTypeListT, typename OpT>
+inline bool
+GEOvdbApply(GEO_PrimVDB& vdb, OpT& op, bool makeUnique = true)
+{
+    if (vdb.hasGrid()) {
+        if (!makeUnique) {
+            return vdb.getGrid().apply<GridTypeListT>(op);
+        }
+        internal::MakeUniqueOp<OpT> uniqueOp(op);
+        return vdb.getGrid().apply<GridTypeListT>(uniqueOp);
```
Idclip: Instead of having `MakeUniqueOp` rely on the entire functor, can it not be a separate process whose only job is to copy the grid? i.e.
```
if (vdb.hasGrid()) {
    if (makeUnique) {
        internal::MakeUniqueOp uniqueOp;
        vdb.getGrid().apply<GridTypeListT>(uniqueOp);
    }
    return vdb.getGrid().apply<GridTypeListT>(op);
}
```

danrbailey:
> I would actually prefer `GEOvdbApply` to be a member of the primitive (i.e., `GEO_PrimVDB::apply`), but then for older Houdini versions we'd still need a free function.

I agree. Could we at least add the new function to the primitive as well and simply call the free function from it, then we can transition to the member primitive and refactor this in the future?

openvdb_houdini/Utils.h
pcucka: That's a good question.  I wanted to avoid iterating through the type list twice, since the type checks involve string comparisons, but that overhead is of course negligible compared to the cost of copying the tree when the grid is not unique.  So perhaps the best solution is to check for uniqueness before applying `MakeUniqueOp`: 
```
if (vdb.hasGrid()) {
    auto gridPtr = vdb.getGridPtr();
    if (makeUnique) {
        auto treePtr = gridPtr->baseTreePtr();
        if (treePtr.use_count() > 2) { // grid + treePtr = 2
            internal::MakeUniqueOp uniqueOp;
            gridPtr->apply<GridTypeListT>(uniqueOp);
        }
    }
    return gridPtr->apply<GridTypeListT>(op);
}

```
Then `MakeUniqueOp` can be simplified to
```
struct MakeUniqueOp
{
    template<typename GridT>
    void operator()(GridT& grid)
    {
        if (auto treePtr = grid.constTreePtr()) {
            grid.setTree(treePtr->copy());
        }
    }
};
```
Does that seem reasonable?

openvdb/doc/changes.txt
pcucka: They're just stylistic preferences.  For text that is not primarily intended to be human-readable, I like to start each sentence or record on a new line.  That makes diffs easier to read, since it limits the scope of each change.

The non-breaking spaces are just to avoid ugliness like "Python" and "3" being split across lines.

pcucka:
> Could we at least add the new function to the primitive as well and simply call the free function from it, then we can transition to the member primitive and refactor this in the future?

To do that, I think we would have to introduce a header dependency to `GEO_PrimVDB.h` on a file like `openvdb_houdini/Utils.h`, which is a bit awkward since the `openvdb_houdini` headers don't ship with Houdini.

openvdb_houdini/Utils.h
```diff
@@ -339,6 +339,48 @@ using VolumeGridTypes = ScalarGridTypes::Append<Vec3GridTypes>;
...
+{
+    if (vdb.hasGrid()) {
+        auto gridPtr = vdb.getGridPtr();
+        if (makeUnique) {
+            auto treePtr = gridPtr->baseTreePtr();
+            if (treePtr.use_count() > 2) { // grid + treePtr = 2
+                // If the grid resolves to one of the listed types and its tree
+                // is shared with other grids, replace the tree with a deep copy.
+                gridPtr->apply<GridTypeListT>([](Grid& baseGrid) {
+                    if (auto treePtr = baseGrid.constBaseTreePtr()) {
```
Idclip: don't think this assignment check is needed? `baseGrid.setTree(baseGrid.constBaseTreePtr()->copy());`

openvdb_houdini/Utils.h
Idclip: Yeah this looks good to me!

openvdb_houdini/Utils.h
pcucka: It's not needed; I've removed it now.

pcucka:
The `GEO_PrimVDB::apply` implementation is untested (I couldn't get the primitive to build), but note that it is potentially more efficient than `GEOvdbApply`, since it avoids incrementing the primitive's unique IDs unnecessarily.

PR #447, opened 2019-05-28
OVDB-96: CMake Memory Allocators

danrbailey:
Add support for Jemalloc and TBB malloc and enable Jemalloc by default.

This seems to build for me on Linux and Windows fine. @jmlait - a quote from your chapter in "Multithreading for Visual Effects":

> On platforms where linking to jemalloc is not practical, we resort to tbb::malloc_proxy.

Can you expand on this? Are there any platforms that you'd suggest we make TBB malloc the default? Also curious to know if there is any configuration in which Houdini is shipped _without_ using Jemalloc?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,13 +116,17 @@ option(DISABLE_DEPENDENCY_VERSION_CHECKS [=[
 Disable minimum version checks for OpenVDB dependencies. It is strongly recommended that this remains disabled.
 Consider updating your dependencies where possible if encountering minimum requirement CMake errors.]=] OFF)
 
+set(_CONCURRENT_MALLOC_OPTIONS None Auto Jemalloc Tbbmalloc)
 if(NOT CONCURRENT_MALLOC)
   set(CONCURRENT_MALLOC Auto CACHE STRING
     "Choose a concurrent malloc library, options are: None Auto Jemalloc Tbbmalloc.
     Although not required, it is strongly recommended to use a concurrent memory allocator.
-    Auto is the default and implies Tbbmalloc for Maya and Jemalloc for non-MAYA,
-    with a fallback of Tbbmalloc if unavailable." FORCE
+    Auto is the default and implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable,
+    in which case Tbbmalloc is used." FORCE
   )
+elseif(NOT ${CONCURRENT_MALLOC} IN_LIST _CONCURRENT_MALLOC_OPTIONS)
+  message(WARNING "Unrecognized value for CONCURRENT_MALLOC, using Auto instead.")
+  set(CONCURRENT_MALLOC Auto CACHE STRING FORCE)
 endif()
 
 ###### Deprecated options
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,16 +100,6 @@ option(USE_BLOSC [=[
 Use blosc while building openvdb components. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the located
 openvdb build configuration to decide on blosc support. You may set this to on to force blosc to be used if you
 know it to be required.]=] ON)
-option(USE_JEMALLOC [=[
-Use Jemalloc while building openvdb components. If USE_JEMALLOC is OFF, CMake attempts to query the
-located openvdb build configuration to decide on Jemalloc support. You may set this to off to disable Jemalloc.
-Although not strictly required, it is strongly recommended to use a concurrent memory allocator.
-Note that you must choose between Jemalloc or TBB malloc, or otherwise disable them both.]=] ON)
-option(USE_TBBMALLOC [=[
-Use TBB malloc while building openvdb components. If USE_TBBMALLOC is OFF, CMake attempts to query the
-located openvdb build configuration to decide on TBB malloc support. You may set this to on to force TBB malloc
-to be used. Although not required, it is strongly recommended to use a concurrent memory allocator.
-Note that you must choose between Jemalloc or TBB malloc, or otherwise disable them both.]=] OFF)
 option(USE_LOG4CPLUS [=[
 Use log4cplus while building openvdb components. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the
 located openvdb build configuration to decide on log4cplus support. You may set this to on to force log4cplus
@@ -125,6 +115,15 @@ option(USE_CCACHE "Build using Ccache if found on the path" ON)
 option(DISABLE_DEPENDENCY_VERSION_CHECKS [=[
 Disable minimum version checks for OpenVDB dependencies. It is strongly recommended that this remains disabled.
 Consider updating your dependencies where possible if encountering minimum requirement CMake errors.]=] OFF)
+
+if(NOT CONCURRENT_MALLOC)
+  set(CONCURRENT_MALLOC Auto CACHE STRING
+    "Choose a concurrent malloc library, options are: None Auto Jemalloc Tbbmalloc.
+    Although not required, it is strongly recommended to use a concurrent memory allocator.
+    Auto is the default and implies Tbbmalloc for Maya and Jemalloc for non-MAYA,
+    with a fallback of Tbbmalloc if unavailable." FORCE
+  )
+endif()
 
 ###### Deprecated options
 
@@ -162,12 +161,12 @@ mark_as_advanced(
   OPENVDB_ENABLE_RPATH
   USE_HOUDINI
   USE_MAYA
-  USE_JEMALLOC
   USE_LOG4CPLUS
   USE_SYSTEM_LIBRARY_PATHS
   USE_CCACHE
   OPENVDB_BUILD_HOUDINI_SOPS
   DISABLE_DEPENDENCY_VERSION_CHECKS
+  CONCURRENT_MALLOC
 )
 
 # Configure minimum version requirements
@@ -427,6 +426,16 @@ endif()
 
 if(USE_MAYA)
   include(OpenVDBMayaSetup)
+endif()
+
+# Configure malloc library. Use Tbbmalloc for Maya and Jemalloc for non-Maya.
+
+if(CONCURRENT_MALLOC STREQUAL "Auto")
+  if(USE_MAYA)
+    set(CONCURRENT_MALLOC "Tbbmalloc")
+  else()
+    set(CONCURRENT_MALLOC "Jemalloc")
+  endif()
 endif()
 
 #########################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,16 @@ option(USE_BLOSC [=[
 Use blosc while building openvdb components. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the located
 openvdb build configuration to decide on blosc support. You may set this to on to force blosc to be used if you
 know it to be required.]=] ON)
+option(USE_JEMALLOC [=[
+Use Jemalloc while building openvdb components. If USE_JEMALLOC is OFF, CMake attempts to query the
+located openvdb build configuration to decide on Jemalloc support. You may set this to off to disable Jemalloc.
+Although not strictly required, it is strongly recommended to use a concurrent memory allocator.
+Note that you must choose between Jemalloc or TBB malloc, or otherwise disable them both.]=] ON)
+option(USE_TBBMALLOC [=[
+Use TBB malloc while building openvdb components. If USE_TBBMALLOC is OFF, CMake attempts to query the
+located openvdb build configuration to decide on TBB malloc support. You may set this to on to force TBB malloc
+to be used. Although not required, it is strongly recommended to use a concurrent memory allocator.
+Note that you must choose between Jemalloc or TBB malloc, or otherwise disable them both.]=] OFF)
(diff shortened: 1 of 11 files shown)


pcucka:
When building for Houdini or Maya will this prefer the version of jemalloc or tbbmalloc that ships with the app?  If not, should it?

danrbailey:
> When building for Houdini or Maya will this prefer the version of jemalloc or tbbmalloc that ships with the app? If not, should it?

Yes, it should. I just added a change which sets the Jemalloc library directory path in OpenVDBHoudiniSetup.cmake so it will now prefer to pick it up from the Houdini install. TBB malloc was already being picked up from the Houdini install if USE_TBBMALLOC was set to on.

I also updated Circle CI to only install jemalloc for non-Houdini builds as well, just to be sure it doesn't fall back to the pre-installed version. I don't know about Maya.

CMakeLists.txt
```diff
@@ -152,6 +162,7 @@ mark_as_advanced(
   OPENVDB_ENABLE_RPATH
   USE_HOUDINI
   USE_MAYA
+  USE_JEMALLOC
```
Idclip: USE_TBBMALLOC here as well?

Idclip: And actually, if we decide to set USE_JEMALLOC to ON by default, USE_JEMALLOC probably shouldn't be an advanced option.

CMakeLists.txt
```diff
@@ -100,6 +100,16 @@ option(USE_BLOSC [=[
 Use blosc while building openvdb components. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the located
 openvdb build configuration to decide on blosc support. You may set this to on to force blosc to be used if you
 know it to be required.]=] ON)
+option(USE_JEMALLOC [=[
+Use Jemalloc while building openvdb components. If USE_JEMALLOC is OFF, CMake attempts to query the
+located openvdb build configuration to decide on Jemalloc support. You may set this to off to disable Jemalloc.
+Although not strictly required, it is strongly recommended to use a concurrent memory allocator.
+Note that you must choose between Jemalloc or TBB malloc, or otherwise disable them both.]=] ON)
```
Idclip: This is on by default, is that correct? is jemalloc usually available on mac/centos/windows outside of houdini/maya builds?

Idclip:
This looks great, thanks @danrbailey. Do we have any more specific advice for choosing a mem allocator with vdb aside from "use a concurrent one" - i.e. none vs je vs tbb? Not part of this PR, but it would be great to run some tests and get some numbers out so we can support the recommendation.

CMakeLists.txt
pcucka: FWIW, Maya ships with both tbbmalloc and (as of Maya 2017) "awjemalloc", which seems to be jemalloc with renamed symbols.  Since we already require TBB, it might be better to make tbbmalloc the default allocator outside of Houdini builds.

danrbailey: I have found jemalloc to provide a decent performance improvement over tbbmalloc, so perhaps we make jemalloc the default outside of non-Maya builds?

danrbailey:
> This looks great, thanks @danrbailey. Do we have any more specific advice for choosing a mem allocator with vdb aside from "use a concurrent one" - i.e. none vs je vs tbb? Not part of this PR, but it would be great to run some tests and get some numbers out so we can support the recommendation.

I happen to have some benchmarks I ran recently. I did a very simple test case where I was comparing a vector of around 100 million or so "512 float" buffers to mimic the data storage in OpenVDB and attempt to isolate allocator performance (run across 24 logical cores):

Allocator | Allocate (seconds) | Deallocate (seconds)
--------- | ------------------ | ---------------------
glibc malloc (serial) | 22.6 | 24.6
glibc malloc (parallel) | 90.0 | 12.9
tbb malloc (serial) | 8.33 | 31.5
tbb malloc (parallel) | 2.73 | 15.4
je malloc (serial) | 2.21 | 1.64
je malloc (parallel) | 1.15 | 0.37

I haven't looked at any tuning, in particular jemalloc offers options to balance memory usage vs performance. 

I think what surprised me most was how much the deallocation time varied. I ran an additional test with jemalloc allocating in serial and then deallocating in parallel (related to OVDB-93) and the deallocation time remained the same as the serial one. I'm still trying to understand what causes this. Right now, we always deserialize .vdb files in serial, so this is an interesting aspect to consider regarding how we do IO.

CMakeLists.txt
danrbailey: As I understand it, mark_as_advanced is just about controlling what shows up when using CMake from the UI? So we should include both jemalloc and tbbmalloc in that?

CMakeLists.txt
danrbailey: @Idclip - do you have a good idea how we would do this in the CMake? Make USE_TBBMALLOC default to ON and USE_JEMALLOC default to OFF when USE_MAYA is ON otherwise OFF and ON respectively?

pcucka: Does CMake support enum options?  jemalloc and tbbmalloc should be mutually exclusive, so what we probably really want is a CONCURRENT_MALLOC option for which the choices are NONE, JEMALLOC, TBBMALLOC, and APPLICATION, where the latter specifies jemalloc for Houdini builds and tbbmalloc for Maya builds.

danrbailey: Hmm. On the one hand, I think I prefer OFF/ON options because then the common use case of enabling one (USE_JEMALLOC=ON) fits in better with how the other variables work.

The contrary argument is that if we were to add support for another memory allocator, this gets pretty fiddly to support, so I suppose the common use case should be *not* changing the default at all.

I think your suggestion is best, except perhaps using "AUTO" instead of "APPLICATION" which can imply jemalloc when building outside of Houdini and Maya.

CMakeLists.txt
Idclip: Yeah that's correct, I'd classify choosing a concurrent allocator as an advanced option, one which we should try to automate or at least choose a good default for (in this case jemalloc). But if jemalloc isn't always available (and can thus cause CMake errors) it probably shouldn't be hidden from the user.

CMakeLists.txt
Idclip: > Does CMake support enum options? 
afaik no - you have to use strings i.e.

```
if(NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE Release CACHE STRING
    "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE
  )
endif()
```
You can have dependent CMake options - we use them a bit:

```
CMAKE_DEPENDENT_option(USE_JEMALLOC "use jemalloc" ON "USE_MAYA" OFF)
CMAKE_DEPENDENT_option(USE_TBBMALLOC "use tbbmalloc" OFF "USE_MAYA" ON)
```

https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

Idclip:
> I happen to have some benchmarks I ran recently. I did a very simple test case where I was comparing a vector of around 100 million or so "512 float" buffers to mimic the data storage in OpenVDB and attempt to isolate allocator performance (run across 24 logical cores):

This is really interesting, thanks! Maybe even worth adding to the documentation, though perhaps we should do some more vdb specific testing first. Either way this definitely seems like enough of a reason to continue to default to jemalloc

CMakeLists.txt
pcucka: For AUTO I would say the logic should be jemalloc for Houdini builds; tbbmalloc for Maya builds; and for all other builds jemalloc if it is available, otherwise tbbmalloc.

danrbailey: I agree that's ideal, but I'm not sure how easy it is to implement the "if it is available" part.

Idclip: You can remove the REQUIRED keyword from the find_package call so that it doesn't error when searching for jemalloc. Then you can use if(TARGET Jemalloc::jemalloc) ... To check to see if the imported target was created/found

danrbailey: Thanks for the suggestion, I'll give that a go.

CMakeLists.txt
```diff
@@ -429,6 +428,16 @@ if(USE_MAYA)
...
+# Configure malloc library. Use Tbbmalloc for Maya and Jemalloc for non-Maya.
+
+if(CONCURRENT_MALLOC STREQUAL "Auto")
+  if(USE_MAYA)
+    set(CONCURRENT_MALLOC "Tbbmalloc")
+  else()
+    set(CONCURRENT_MALLOC "Jemalloc")
+  endif()
+endif()
+
```
pcucka: Can you add a warning here if CONCURRENT_MALLOC is something other than "Auto", "Tbbmalloc", "Jemalloc", or "None", and then maybe set it to either "Auto" or "None"?

CMakeLists.txt
```diff
@@ -126,6 +116,15 @@ option(DISABLE_DEPENDENCY_VERSION_CHECKS [=[
 Disable minimum version checks for OpenVDB dependencies. It is strongly recommended that this remains disabled.
 Consider updating your dependencies where possible if encountering minimum requirement CMake errors.]=] OFF)
 
+if(NOT CONCURRENT_MALLOC)
+  set(CONCURRENT_MALLOC Auto CACHE STRING
+    "Choose a concurrent malloc library, options are: None Auto Jemalloc Tbbmalloc.
+    Although not required, it is strongly recommended to use a concurrent memory allocator.
+    Auto is the default and implies Tbbmalloc for Maya and Jemalloc for non-MAYA,
```
pcucka: How about "Auto is the default and implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable, in which case Tbbmalloc is used"?

jmlait:
> Can you expand on this? Are there any platforms that you'd suggest we make TBB malloc the default? Also curious to know if there is any configuration in which Houdini is shipped _without_ using Jemalloc?

JE Malloc is only used on Linux.   Windows uses tbb_malloc.  It is very hard to inject into the windows runtime to replace the system allocator.  TBB actually has to do some very suspicious binary hacking at the machine code level to replace the base allocator.  (And it is a patch that windows updates sometimes breaks...)  We've not managed to get such a patch for Windows.

We're also still running jemalloc 3.6.0.  I've tried several times to move up to 5.0, 5.1, and 5.2.  The problem is that the 5.X series uses lots of address space.  I've got nothing against this personally, it is a wise way to get faster performance with no significant cost (other than TLB allocation).  But users will freak out and submit bugs if the VIRT column of top goes to 128GB, even if it is only using 20GB of actual physical RAM.

CMakeLists.txt
danrbailey: Done, thanks.

CMakeLists.txt
danrbailey: I made it fallback on Auto.

danrbailey:
I think I've addressed all the feedback now, let me know if there's anything else outstanding in this implementation.

Idclip:
I've just tried to run this on MacOS against Houdini 17.0.506 and immediately get the following error:
`vdb_test(65605,0x1177d65c0) malloc: *** malloc_zone_unregister() failed for 0x7fff95729000`
The last address doesn't seem to change.

danrbailey:
> I've just tried to run `vdb_test` on MacOS against Houdini 17.0.506 and immediately get the following error:
> `vdb_test(65605,0x1177d65c0) malloc: *** malloc_zone_unregister() failed for 0x7fff95729000`
> The last address doesn't seem to change.
> 
> Seems like a related issue here: [facebook/rocksdb#1616](https://github.com/facebook/rocksdb/issues/1616) [jemalloc/jemalloc#420](https://github.com/jemalloc/jemalloc/issues/420)
> and the solution they went with: [rust-lang/jemalloc#15](https://github.com/rust-lang/jemalloc/pull/15)

Thanks for highlighting this - I think the simplest solution is to default to using tbbmalloc on Mac, which would leave just Linux on jemalloc by default. I'll give that a go and verify that fixes this issue.

danrbailey:
I've pushed a change to use TBB malloc on Mac and that seems to be working correctly now.

CMakeLists.txt
```diff
@@ -417,6 +432,16 @@ if(USE_MAYA)
   include(OpenVDBMayaSetup)
 endif()
 
+# Configure malloc library. Use Tbbmalloc for Windows or Maya and Jemalloc otherwise.
```
Idclip: Update this comment to include the macos decision with a reference to the problem (as looks like we can actually use jemalloc on mac after a specific version). Also perhaps expand on why windows uses tbbmalloc too for future reference.

openvdb/doc/changes.txt
```diff
@@ -17,6 +17,8 @@ Improvements:
   @vdblink::tree::LeafNode::modifyValueAndActiveState()
   LeafNode::modifyValueAndActiveState@endlink now modify voxel values
   in place for improved performance.
+- Add CMake support for linking against Jemalloc and TBB malloc and enable
+  Jemalloc by default.
```
Idclip: Should probably add windows/mac notes here

Idclip:
Re-tested on mac, looks good

CMakeLists.txt
danrbailey: Done.

openvdb/doc/changes.txt
danrbailey: Done.

PR #448, opened 2019-05-29
TSC minutes

Idclip:

### DIFF
--- a/tsc/meetings/2019-05-23.md
+++ b/tsc/meetings/2019-05-23.md
@@ -1,0 +1,126 @@
+Minutes from 17th OpenVDB TSC meeting, May 23, 2019
+
+Attendees: *Nick* A., *Dan* B., *Peter* C., *Jeff* L.
+
+Additional Attendees: Bruce Chernia (Intel), John Mertic (LF)
+
+Regrets: *Ken* M.
+
+Agenda:
+
+1) Quorum Confirmation
+2) Secretary Selection
+3) CI Update (Dan)
+4) CLA Update (John)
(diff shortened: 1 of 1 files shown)


pcucka:
Does that also disable the merge check?  (If so, that seems like a Mack truck-sized hole in the merge check mechanism.)

Do Circle and Azure have some equivalent of a .gitignore file to which we could add the tsc/ directory and the .md files?

danrbailey:
> Does that also disable the merge check? (If so, that seems like a Mack truck-sized hole in the merge check mechanism.)

I'm not sure, I'd have to experiment with it.

> Do Circle and Azure have some equivalent of a .gitignore file to which we could add the tsc/ directory and the .md files?

Azure does have the concept of path filters, which I think we can make use of:

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#paths

Let's attempt to complete the transition to Azure then we can talk about stuff like this. I agree this is a nice feature to have.

pcucka:
Sure, it can wait until we make the transition.

Idclip:
Stopping unnecessary CI build would be good to solve for sure. I agree with @danrbailey that the notes will most likely be more visible in the repo but it would be nice to have them on the website too. My main concern with them here is that I don't think people will generally want them checked out for development. I'm not really a fan of git sub modules for development, but perhaps the tsc folder is a good use for them.

jmlait:
I prefer them in the repo.  Thought I do agree the CI check could best be skipped.

PR #451, opened 2019-05-31
OVDB-95: VDB to Spheres leaves bounds of input VDB

pcucka:


jmlait:
Thank you, it now builds the spheres inside the geometry in my test cases!

Thank you also for splitting into two commits as that made cherry-picking easy as I didn't have to deal with ghost conflicts on the release notes.

PR #452, opened 2019-05-31
OVDB-53: Add position compression option to Scatter SOP and default to use 16-bit default

danrbailey:


danrbailey:
> Won't this change the default behaviour for existing nodes? We should add support so that the initial value is set to `none` for these cases. Think you can use the obsolete parameters for this but I may be wrong (as they're mainly for changing existing params). I'm sure its possible though

I know it won't take a _long_ time to do what you're suggesting, but after our discussion in the last TSC meeting about our limited resources and prioritizing the right development work - do you think it's really worth going to this effort to preserve the exact same behaviour? This SOP isn't shipped with Houdini so we're only talking about open-source installations, the codec isn't query-able from Houdini and the difference between 32-bit->16-bit is negligible for I'm guessing >99% of real world use cases.

Idclip:
> I know it won't take a long time to do what you're suggesting, but after our discussion in the last TSC meeting about our limited resources and prioritizing the right development work - do you think it's really worth going to this effort to preserve the exact same behaviour? This SOP isn't shipped with Houdini so we're only talking about open-source installations, the codec isn't query-able from Houdini and the difference between 32-bit->16-bit is negligible for I'm guessing >99% of real world use cases.

Yes I believe so, especially if it's a trivial change. We cannot change default behaviour. It's going to take me some time to trawl through all our Houdini tools and make sure nothing is relying on position compression - and that is still no guarantee without fully testing. Consider our tools like point merging and P2LS - it's very difficult to say with confidence this won't impact other peoples. The worst case would be where it breaks something due to the value changes and that has to be tracked down. P2LS is an example where small differences can have a large impact depending on the relative grid scales. Our surfacer works explicitly off VDB Points.

I of course don't want to create more work than is absolutely necessary. But in almost all cases I think changes will always have to be backwards compatible. This should have been specced out on the ticket, probably one of the reasons why discussing work during TSC meetings will help in the future. But we can't use resourcing as an excuse to skip compatibility (regardless of how much we assume them to be in use) because the work has already been almost completed. I would rather wait another month for this change that for it to have breaking consequences.

Happy to discuss this more, perhaps other members can way in too.

danrbailey:
> > I know it won't take a long time to do what you're suggesting, but after our discussion in the last TSC meeting about our limited resources and prioritizing the right development work - do you think it's really worth going to this effort to preserve the exact same behaviour? This SOP isn't shipped with Houdini so we're only talking about open-source installations, the codec isn't query-able from Houdini and the difference between 32-bit->16-bit is negligible for I'm guessing >99% of real world use cases.
> 
> Yes I believe so, especially if it's a trivial change. We cannot change default behaviour. It's going to take me some time to trawl through all our Houdini tools and make sure nothing is relying on position compression - and that is still no guarantee without fully testing. Consider our tools like point merging and P2LS - it's very difficult to say with confidence this won't impact other peoples. The worst case would be where it breaks something due to the value changes and that has to be tracked down. P2LS is an example where small differences can have a large impact depending on the relative grid scales. Our surfacer works explicitly off VDB Points.
> 
> I of course don't want to create more work than is absolutely necessary. But in almost all cases I think changes will have to be backwards compatible. This should have been specced out on the ticket, probably one of the reasons why discussing work during TSC meetings will help in the future. But we can't use resourcing as an excuse to skip compatibility (regardless of how much we assume them to be in use) because the work has already been almost completed. I would rather wait another month for this change that for it to have breaking consequences.
> 
> Happy to discuss this more, perhaps other members can way in too.

That's ok, if you think it's that important, I shall make the change you're suggesting. I still believe compatibility should be handled on a case-by-case basis as maintaining this does have a non-trivial cost. And you're right we should have caught this when extending this SOP to support VDB Points. Now that we have JIRA in place, hopefully this won't slip through the cracks as much in future.

With regards to planning and speccing out work, perhaps now is a good time to highlight this ticket - [OVDB-52](https://jira.aswf.io/browse/OVDB-52). I put up full details of the issue with some screenshots and a suggestion of how to resolve it a few months back. I didn't receive any feedback, so as I've been in this codebase anyway, I've been progressing with fixing this one too and have updated the ticket as I go. I'm now almost ready to submit a PR and have just added a comment to explicitly describe the upcoming change in behavior.

danrbailey:
> Won't this change the default behaviour for existing nodes? We should add support so that the initial value is set to `none` for these cases. Think you can use the obsolete parameters for this but I may be wrong (as they're mainly for changing existing params). I'm sure its possible though

I just had a look at this and I don't see a way to use obsolete parameters because we're not removing any parameters. The easiest solution is to just use the version/namespace solution that SideFX recommends(http://www.sidefx.com/docs/houdini/assets/namespaces.html), ie add:

`.setInternalName("DW_OpenVDBScatter::2.0")`

If we're going to go this route, probably it's better to add support for SOP versioning through OpenVDBOpFactory as we haven't done this with any other nodes yet.

openvdb_houdini/SOP_OpenVDB_Scatter.cc
```diff
@@ -186,6 +186,19 @@ newSopOperator(OP_OperatorTable* table)
             "For scalar-valued VDBs other than signed distance fields,"
             " use voxel values as local multipliers for point density."));
 
+    parms.add(hutil::ParmFactory(PRM_ORD, "poscompression", "Position Compression")
```
pcucka: Please use string-valued parameters for menus, so that the menu items can be reordered if necessary.

Idclip:
>  That's ok, if you think it's that important, I shall make the change you're suggesting. I still believe compatibility should be handled on a case-by-case basis as maintaining this does have a non-trivial cost. And you're right we should have caught this when extending this SOP to support VDB Points. Now that we have JIRA in place, hopefully this won't slip through the cracks as much in future.

Thanks, it will definitely make me a lot more confident releasing it here as this is the main node we use for scattering vdb points in our tools. 

> I just had a look at this and I don't see a way to use obsolete parameters because we're not removing any parameters.

@jmlait can you suggest anything here?

openvdb_houdini/SOP_OpenVDB_Scatter.cc
danrbailey: Shouldn't it match SOP_OpenVDB_Points_Convert::poscompression? We wouldn't change one menu without also changing the other.

pcucka: Points Convert should also have used string-valued parameters.

danrbailey: I agree. That being said, I think I'd rather it match, then we can go through the effort of changing both if we ever need to.

pcucka: OK, but then would it not be a good idea to add a function (or functions) to `PointUtils.h` to provide the mapping from integers to menu item strings, so that the two SOPs can't get out of sync?

danrbailey: We haven't changed the options in this drop-down in at least five years and I don't have any plans to change it in the future, so that's not a big concern to me.

jmlait:
You can use the syncNodeVersion to handle new defaults that are not backwards compatible.

For example, this is the sync code for UV Layout.  As of 15.5.438 we added two new parameters that had default values that would change old nodes.   The syncNodeVersion is called whenever a node is loaded from an old .hda or .hip file into a newer session.   It then has permission to change the values of parameters (including inside locked assets) to be compatible with the new version of the node.

void
SOP_UVLayout::syncNodeVersion(const char *old_version,
    const char *cur_version, bool *node_deleted)
{
    if (UT_String::compareVersionString(old_version, "15.5.438") < 0)
    {
        setInt("axisalignislands", 0, 0.0, 0);
        setInt("optimizewastedspace", 0, 0.0, 0);
    }

    SOP_Node::syncNodeVersion(old_version, cur_version, node_deleted);
}

The two things that can still get you:
1) If you bring a new node back into older versions, obviously you lose the new parameter value.
2) Scripts which put down nodes will behave differently as they'll use the new defaults.  (But OTOH, this is often what is desired)

I would not recommend a ::2.0 node for this.  This would require having two installed operators, the original and the 2.0, so bifurcate the code (at least semantically if not in terms of actual C++)

Idclip:
Thanks @jmlait. Can we still leverage `syncNodeVersion` when it's not the Houdini version that it depends on? For example we build and use this Node in 16.5, 17 and 17.5, so we'd want the `old_version/cur_version` values to relate to the node version rather than the Houdini version (similair to how you can do this with otls/hdas)

danrbailey:
Yeah - I had a look at what you suggested @jmlait and I agree with @Idclip, it's not clear to me how we could use SOP_Node::syncNodeVersion as that tends to assume that we can gate by Houdini version, not the version of the node itself?

We could perhaps support something similar on our side using userdata, but would be nice if there's something in Houdini we can hook into here.

jmlait:
syncNodeVersion uses the OP_Operator::getVersion.

If it is an HDA, this comes from the Version field stored there.
If it is not, it just uses the current Houdini Version.

But since VDB already overrides OP_Operator, it can just return a getVersion() string there that matches its own rules.   Going forward, we can treat anything that looks like a Houdini string as being "old" and anything matching the new VDB strings as being from the relevant version of VDB.

danrbailey:
> syncNodeVersion uses the OP_Operator::getVersion.
> 
> If it is an HDA, this comes from the Version field stored there.
> If it is not, it just uses the current Houdini Version.
> 
> But since VDB already overrides OP_Operator, it can just return a getVersion() string there that matches its own rules. Going forward, we can treat anything that looks like a Houdini string as being "old" and anything matching the new VDB strings as being from the relevant version of VDB.

Ah, I like that approach! Let's discuss in the TSC meeting to make sure everyone is happy with this. Perhaps it's worth adding a SOP_Node::getVersion() override now even if we still have to support old being Houdini version whenever we add a new syncNodeVersion.

pcucka:
To summarize from our June 13 meeting: `getVersion` is an `OP_Operator` method, not an `OP_Node` method, so the override would have to be added to `OP_OperatorDW`.  But by design neither that class nor any other in `houdini_utils` has any OpenVDB dependencies, so it has no access to the library version number.  The operator spare data mechanism proposed in [OVDB-106](https://jira.aswf.io/browse/OVDB-106) solves the problem by allowing `OpenVDBOpFactory` to attach version metadata that can be returned by an `OP_OperatorDW::getVersion` method, without introducing any new dependencies.

danrbailey:
I've just updated this PR with an implementation of `SOP_OpenVDB_Scatter::syncNodeVersion()` with the hope of getting this one finished up for 7.0.0.

There's probably a good argument for moving some of this logic to a convenience method somewhere in `SOP_NodeVDB`, but I don't want to spend time on that with the release so soon. All the logic is contained in the `syncNodeVersion()` method, so we can always make this change when we next need this functionality.

openvdb_houdini/SOP_OpenVDB_Scatter.cc
```diff
@@ -264,6 +279,38 @@ and usage examples.\n");
...
+    bool disableCompression = false;
+    size_t spacePos = oldVersionStr.find_first_of(' ');
+    if (spacePos == std::string::npos) {
+        // no space in VDB versions prior to 6.2.0
+        disableCompression = true;
+    } else if (oldVersionStr.size() > 3 && oldVersionStr.substr(0,3) == "vdb") {
+        std::string vdbVersion = oldVersionStr.substr(3,spacePos-3);
+        // disable compression in VDB version 6.2.1 or earlier
+        if (UT_String::compareVersionString(vdbVersion.c_str(), "6.2.1") < 0) {
+            disableCompression = true;
```
Idclip: I think this needs to be `<= 0` as `compareVersionString` returns 0 on version match?

danrbailey: This is to handle the specific case where you've built and released an "in development" version of 6.2.1 that _doesn't_ have this change right? I think you're correct, I'll make that change.

Idclip: This is to handle all existing versions of 6.2.1 which have this node saved in it. In these cases, oldVersionStr will be 6.2.1 i.e. we don't want to change the parm interface

danrbailey: Yes, that's what I understood. I've made that change now.

danrbailey:
> I haven't built and tested this since it was originally requested, so I haven't tested the sync logic. Assuming that's all legit, this looks great

I've tested the sync mechanism and everything works as expected so I'm going to merge this. One small caveat regarding the previous comment - if the VDB version **and** the Houdini version are the same, I don't believe syncNodeVersion() is called, so you may need to be careful with regards to choosing the point to deploy this change to production to ensure that at least one of those two versions **is** changing.

PR #453, opened 2019-05-31
OVDB-99: Fix an integer Tuple/Mat compile bug on Windows 

danrbailey:


openvdb/CHANGES
```diff
@@ -15,6 +15,10 @@ Version 6.1.1 - In development
     - Fixed a bug in tools::computeScalarPotential() that could produce
       a corrupt result due to invalid memory access.
       [Reported by Edwin Braun]
+    - Fixed a bug in isInfinite() and isNan() methods of math::Tuple and
```
pcucka: This is a workaround for a bug in VS, not a bug in OpenVDB.  C++11 requires `std::isfinite` and `std::isnan` to be overloaded for integral types.

danrbailey: To clarify - you're happy with the change itself but you'd like it to be recorded as a VS workaround in the release notes as opposed to a VDB bug?

pcucka: I would prefer that we deprecate functions like `math::isFinite`, `math::Sqrt`, etc. for which there are standard library equivalents, but it sounds like that's not going to be possible any time soon.

danrbailey: It's still an issue in Visual Studio 2019, so no, I don't think this is going away any time soon.

I've improved the change notes.

danrbailey:
I just merged master in and fixed the conflicts in the release notes. @pcucka - can we merge this?

openvdb/CHANGES
```diff
@@ -12,6 +12,8 @@ Version 6.1.1 - In development
       now modify voxel values in place for improved performance.
     - Add CMake support for linking against Jemalloc and TBB malloc and enable
       Jemalloc by default for Linux and non-Maya, otherwise use TBB malloc.
+    - Made a few improvements related to the use of std::inf and std::nan with
```
pcucka: How about "Added math::isInfinite() and math::isNaN() to resolve Visual Studio compatibility issues with integer types"?

openvdb/math/Math.h
```diff
@@ -351,6 +351,26 @@ inline bool
...
+
+/// Return @c true if @a x is an infinity value (either positive infinity or negative infinity).
+template<typename Type, typename std::enable_if<std::is_arithmetic<Type>::value, int>::type = 0>
+inline bool
+isInfinite(const Type& x) { return std::isinf(static_cast<double>(x)); }
+
+
+/// Return @c true if @a x is a NaN (Not-A-Number) value.
+inline bool
+isNan(const float x) { return std::isnan(x); }
```
pcucka: Let's call this "isNaN".

openvdb/CHANGES
danrbailey: Sure

openvdb/math/Math.h
danrbailey: I'm not keen on the inconsistency between math::isNaN() and math::Tuple::isNan(), but that's a minor thing. I'll make that change now.

pcucka: You're right, they should be consistent.  But I think I would prefer that we add `Tuple::isNaN` and `Mat::isNaN` and deprecate the `isNan` methods.

danrbailey: Do you not think we might annoy users forcing them to change their code when they update simply to capitalize one letter? The current capitalization is at least consistent with Half:

https://github.com/openexr/openexr/blob/master/IlmBase/Half/half.h#L171

danrbailey: I changed it back to isNan so this is consistent and users don't have to change their code.

PR #455, opened 2019-06-04
Sharpening filter

pcucka:
This is a preliminary implementation of a sharpening filter—currently only for level sets, but the intention is to extend it to fog volumes.  This PR is not ready to be merged; at this point I'm mainly looking for feedback on whether the basic approach is reasonable before I invest more time in it.

The foundation is the new `ConvolutionStencil` class, which is constructed from a 3D array of kernel coefficients.  The class is similar to the existing `DenseStencil`, but with the addition of a `convolve` method and an optimization to avoid repopulating the entire stencil for the common case (for convolution) of translation along the  _z_ axis.  The performance difference for the latter is significant.

The `LevelSetFilter` class gets a new `sharpen` method, implemented using a Laplacian of Gaussian-inspired kernel, but what wasn't clear to me is whether the kernel should be exposed as an operator in `math/Operators.h`.  @kmuseth, can you comment on that?

Finally, on the Houdini side I've added a new Sharpen SDF SOP.  A new node seemed best to me, since the only existing one to which the new functionality could reasonably be added is Smooth SDF, and it would be odd for a node with that name to include a Sharpen mode.


pcucka:
Sample output:

sharpening example

openvdb/math/Stencils.h
```diff
@@ -1701,6 +1702,119 @@ class DenseStencil: public BaseStencil<DenseStencil<GridT, IsSafe>, GridT, IsSaf
...
+        const Coord offset = shape >> 1;
+
+        if (mInitialized && (ijk == this->getCenterCoord().offsetBy(0, 0, 1))) {
+            // If the stencil buffer has been initialized and its new position differs
+            // from the old by +1 along the z axis (the common case for convolution),
+            // preserve the contents of the buffer and sample only the new z slice.
+            // (Note that BaseStencil::moveTo() sets mCenter before calling init(),
+            // inhibiting this optimization.)
+            /// @todo This optimization could be applied for any offset that leaves some
+            /// overlap between the old and new stencils, but the bookkeeping gets messy.
```
Idclip: We've also run into these kind of problems in the past too, especially for larger kernel sizes - typically 4x4 for cubic b splines which cache a 5x5 region for a single voxel which may sample positions either side. It would be awesome if the stencil framework supported optimizations for trivial stepping through neighbours!

kmuseth: @Idclip are you referring to incrementally modifying the stencil buffer as it is translated (as oppose to repopulating the entire buffer from scratch)? If so, this is something we've though about as well but never found an efficient implementation (the logic for checking which grid points can be reused easily outweighs the benefit of avoiding a reload - at least that's my experience).

kmuseth: oh I see - this optimization is limited to the special case where the stencil is translated by one voxel along the z-axis. Assuming that's a sufficiently common case that's a great idea (and obviously much more tractable then the general case I was referring to)

Idclip: > @Idclip are you referring to incrementally modifying the stencil buffer as it is translated (as oppose to repopulating the entire buffer from scratch)? 
Correct, I'd be interested to know how much time this optimisation of handling a single +z step shaves off, as although arbitrary size steps with arbitrary width stencils could be pretty complex, +/- stepping a single voxel in any direction (or perhaps just faces) could be worth investigating for all stencils.

pcucka: Some Houdini timings for the sharpening filter both with and without the +z optimization and, for comparison, for the existing Gaussian smoothing filter:
<table>
<tr><td></td><td align=center><b>crawler</b><br><sub>181 MVox</sub></td><td align=center><b>dragon</b><br><sub>23 MVox</sub></td><td align=center><b>venus</b><br><sub>6 MVox</sub></td></tr>
<tr><td><b>5&times;5&times;5</b><br>&nbsp;&nbsp;Sharpening<br>&nbsp;&nbsp;Optimized Sharpening<br>&nbsp;&nbsp;Gaussian Smoothing</td><td><br>61.3s<br>34.3s<br>17.8s</td><td><br>8.0s<br>4.6s<br>2.8s</td><td><br>2.0s<br>1.2s<br>1.0s</td></tr>
<tr><td><b>7&times;7&times;7</b><br>&nbsp;&nbsp;Sharpening<br>&nbsp;&nbsp;Optimized Sharpening<br>&nbsp;&nbsp;Gaussian Smoothing</td><td><br>144.6s<br>76.6s<br>20.1s</td><td><br>18.6s<br>10.3s<br>3.5s</td><td><br>4.7s<br>2.6s<br>1.1s</td></tr>
<tr><td><b>9&times;9&times;9</b><br>&nbsp;&nbsp;Sharpening<br>&nbsp;&nbsp;Optimized Sharpening<br>&nbsp;&nbsp;Gaussian Smoothing</td><td><br>292.3s<br>146.3s<br>22.3s</td><td><br>37.8s<br>20.0s<br>3.5s</td><td><br>9.4s<br>5.0s<br>1.1s</td></tr>
</table>

Idclip: Nice results, thanks @pcucka! Are you able to calculate/estimate the probability of stepping in the more common case (+z) for the convolution stencil (as I'm not quite clear algorithmically why stepping in +z is more common)?  For example, 50% of all stencil moves are in +z. Or, to put it another way, I'm wondering if we can estimate the potential performance gains from introducing this type of optimisation more generally, as I think you could support -z simply by changing the inner loop to `for (int k = 0; k < shape[2] - 1; ++k)` and other face offsets with a similar tweak.

pcucka: > I'm not quite clear algorithmically why stepping in +z is more common

All of the existing filters, as well as the sharpening filter, iterate over leaf-level active values using a `ValueOnIter`, which is arguably the "natural" way to implement a filter.  `ValueOnIter` loops over leaf buffer indices (via `NodeMask::OnIterator`), and that corresponds to _xyz_ iteration with _z_ as the fastest-changing coordinate.  So that's why I chose to optimize the +_z_ case first.

The other common case, I suppose, would be _xyz_ iteration over a bounding box, but there too the _z_ loop is typically the inner loop.

pcucka: Just to clarify, the sharpening kernel is not a separable kernel, so I would expect an unsharp mask filter to significantly outperform this sharpening filter.  (That is what I was trying to suggest by including timings for the Gaussian filter.)  That said, in none of my experiments with unsharp masking was I able to produce surfaces without holes.

openvdb/math/Stencils.h
```diff
@@ -1701,6 +1705,177 @@ class DenseStencil: public BaseStencil<DenseStencil<GridT, IsSafe>, GridT, IsSaf
...
+}; // class ConvolutionStencil
+
+
+////////////////////////////////////////
+
+
+/// Return a 3D sharpening kernel of the given radius, for use with ConvolutionStencil.
+template<typename ValueType>
+ConvolutionKernel<ValueType>
+sharpeningKernel(int r)
```
pcucka: I'm not thrilled about moving `sharpeningKernel` here, but it's needed by both `tools::Filter` and `tools::LevelSetFilter`, and I don't see a better place to put it, other than in a header of its own.

danrbailey:
What a mess. I'm going to have to re-submit this. :(

I attempted to add a new branch to our ASWF repo to roughly where our master branch was when Peter originally created this PR and changed the target branch to this new branch, but it failed due to a CI issue trying to build Houdini 17.5. It was a compile issue in our codebase that was fixed a few commits later, so I attempted to move the branch forwards until after the issue was resolved. I couldn't move it because it requires a DCO signoff check to have completed successfully in order to force push to a protected branch. So I instead dropped the branch and re-created it in the new position. However, dropping the branch caused the PR to auto-close because it was pointing at a branch that was deleted. When I re-opened the PR, it triggered the CLA to re-run and now as Peter has left DWA since submitting this PR, the CLA check now fails.

Moral of the story - don't try and live in the past, as desirable as it may be to look back with nostalgia on the golden days of 2019...

PR #456, opened 2019-06-07
OVDB-64: Add SOP Pairing Logic

danrbailey:
This adds the ability to hide an open-source SOP from the tab menu using OpFactory::setInvisible(). It also adds the ability to "pair" an open-source SOP with a native SOP shipped by SideFX. This is achieved through a getParentName() rule in the OpPolicy classes and by using OpenVDBOpFactory::setParentName("...") for cases where it deviates from this logic (currently Advect and Filter SOP).

I have included setParentName("") with an empty string for cases where the open-source SOP **isn't** shipped with Houdini (OpenVDB Densify SOP for example). If the sole use-case is for ophiding, this isn't strictly necessary because it shouldn't matter if the native SOP doesn't exist, but I included it anyway so the pairing is accurate. 

No SOPs are being marked invisible with this PR, so this is just a stepping stone to figuring out the best way to proceed with #404.

### DIFF
--- a/openvdb_houdini/SOP_NodeVDB.cc
+++ b/openvdb_houdini/SOP_NodeVDB.cc
@@ -635,9 +635,9 @@ public:
         return factory.flavorString() + "_OpenVDB";
     }
 
-    /// @brief Return the name of the parent operator.
-    /// @note This is typically the name of the native operator if shipped with Houdini.
-    virtual std::string getParentName(const houdini_utils::OpFactory& factory)
+    /// @brief Return the name of the equivalent native operator as shipped with Houdini.
+    /// @details An empty string indicates that there is no equivalent native operator.
+    virtual std::string getNativeName(const houdini_utils::OpFactory& factory)
     {
         return "";
     }
@@ -674,7 +674,7 @@ public:
         return this->getLowercaseName(this->getValidName(this->getLabelName(factory)));
     }
 
-    std::string getParentName(const houdini_utils::OpFactory& factory) override
+    std::string getNativeName(const houdini_utils::OpFactory& factory) override
     {
         return this->getLowercaseName(this->getValidName(factory.english()));
     }
@@ -698,35 +698,38 @@ OpenVDBOpFactory::OpenVDBOpFactory(
     houdini_utils::OpFactory::OpFlavor flavor):
     houdini_utils::OpFactory(OpenVDBOpPolicy(), english, ctor, parms, table, flavor)
 {
-    mParentName = OpenVDBOpPolicy().getParentName(*this);
+    mNativeName = OpenVDBOpPolicy().getNativeName(*this);
 }
 
 
 OpenVDBOpFactory::~OpenVDBOpFactory()
 {
-    // hide the parent node if marked invisble
-    if (!mParentName.empty()) {
-        bool invisible = mParentInvisible;
+    // hide the native node if marked invisble
+    if (!mNativeName.empty()) {
+        bool invisible = mNativeInvisible;
 
         if (invisible) {
-            this->table().addOpHidden(mParentName.c_str());
+            this->table().addOpHidden(mNativeName.c_str());
         }
     }
 }
 
 
 OpenVDBOpFactory&
-OpenVDBOpFactory::setParentName(const std::string& name)
+OpenVDBOpFactory::setNativeName(const std::string& name)
 {
-    mParentName = name;
+    // SideFX nodes have no native equivalent.
+#ifndef SESI_OPENVDB
+    mNativeName = name;
+#endif
     return *this;
 }
 
 
 OpenVDBOpFactory&
-OpenVDBOpFactory::setParentInvisible()
+OpenVDBOpFactory::setNativeInvisible()
 {
-    mParentInvisible = true;
+    mNativeInvisible = true;
     return *this;
 }
 
--- a/openvdb_houdini/SOP_NodeVDB.cc
+++ b/openvdb_houdini/SOP_NodeVDB.cc
@@ -634,6 +634,13 @@ public:
     {
         return factory.flavorString() + "_OpenVDB";
     }
+
+    /// @brief Return the name of the parent operator.
+    /// @note This is typically the name of the native operator if shipped with Houdini.
+    virtual std::string getParentName(const houdini_utils::OpFactory& factory)
+    {
+        return "";
+    }
 };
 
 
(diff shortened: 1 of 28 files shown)


openvdb_houdini/SOP_OpenVDB_Filter.cc
```diff
@@ -326,6 +326,9 @@ Offset:\n\
 
     // Register this operator.
     hvdb::OpenVDBOpFactory("VDB Filter", SOP_OpenVDB_Filter::factory, parms, *table)
+#ifndef SESI_OPENVDB
+        .setParentName("vdbsmooth")
+#endif
```
Idclip: Is "parent" the right terminology for this? I wouldn't, for example, describe sesis vdbsmooth node as the parent of the aswf's vdb filter node. Is this intended to be used in other ways? If not, maybe sometime more specific would be appropriate e.g. setSESIPairName() / setLinkedNodeName() / setNodeTwinName(). Also, do we actually need the ifdef guards?

danrbailey: I'm not particularly happy with the name parent either, though I'm not sure your suggestions are any more intuitive. Perhaps native or upstream?

I think we should use the ifdef guards, because otherwise if SideFX were to build this code, they'd end up making a node be it's own a parent. While we could ensure that there were appropriate guards around ophiding to ensure that logic doesn't get executed for SideFX, I'm not sure it's a good idea because this could be used in other ways in the future.

Regarding the VDB Filter/Smooth, my other PR renamed OpenVDB Filter to VDB Smooth (along with others like OpenVDB Convert -> Convert VDB). I didn't include this change here because I wanted this PR to not alter any existing behavior.

openvdb_houdini/SOP_NodeVDB.h
```diff
@@ -61,6 +61,22 @@ class OPENVDB_HOUDINI_API OpenVDBOpFactory: public houdini_utils::OpFactory
     /// Construct an OpFactory that on destruction registers a new OpenVDB operator type.
     OpenVDBOpFactory(const std::string& english, OP_Constructor, houdini_utils::ParmList&,
         OP_OperatorTable&, houdini_utils::OpFactory::OpFlavor = SOP);
+
+    /// Register the node
+    ~OpenVDBOpFactory();
+
+    /// @brief Set the name of the parent operator.
+    /// @details This is typically the name of the operator shipped with Houdini and is
+    /// only needed where the parent name policy doesn't provide the correct name.
```
pcucka: Add something like "Pass an empty string to indicate that there is no equivalent native operator."

openvdb_houdini/SOP_NodeVDB.h
```diff
@@ -61,6 +61,22 @@ class OPENVDB_HOUDINI_API OpenVDBOpFactory: public houdini_utils::OpFactory
     /// Construct an OpFactory that on destruction registers a new OpenVDB operator type.
     OpenVDBOpFactory(const std::string& english, OP_Constructor, houdini_utils::ParmList&,
         OP_OperatorTable&, houdini_utils::OpFactory::OpFlavor = SOP);
+
+    /// Register the node
+    ~OpenVDBOpFactory();
```
pcucka: Unfortunately, `~OpFactory` is non-virtual; but since we don't delete `OpenVDBOpFactory` objects through base class pointers, this is probably OK for now.  Make a note, though, to make `~OpFactory` virtual in OpenVDB 7.

openvdb_houdini/SOP_OpenVDB_Filter.cc
pcucka: Let's go with "Native".  "Parent" is definitely not the right term; there's no hierarchical relationship.

pcucka: The `ifdef` can be moved inside the method body, I think.

openvdb_houdini/SOP_NodeVDB.h
danrbailey: That's true. I made a ticket to track this one - https://jira.aswf.io/browse/OVDB-102

danrbailey:
@pcucka - I just pushed up a change to address all your feedback, thanks. Also added native names for the convert SOPs which were missing.

openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -686,6 +698,39 @@ OpenVDBOpFactory::OpenVDBOpFactory(
...
+    mNativeName = OpenVDBOpPolicy().getNativeName(*this);
+}
+
+
+OpenVDBOpFactory::~OpenVDBOpFactory()
+{
+    // hide the native node if marked invisble
+    if (!mNativeName.empty()) {
+        bool invisible = mNativeInvisible;
+
```
pcucka: Maybe collapse this to
```
if (mNativeInvisible && !mNativeName.empty()) {
    this->table().addOpHidden(mNativeName.c_str());
}
```
(Testing the boolean first is faster.)

openvdb_houdini/SOP_NodeVDB.h
```diff
@@ -61,6 +61,26 @@ class OPENVDB_HOUDINI_API OpenVDBOpFactory: public houdini_utils::OpFactory
...
+
+    /// @brief Return the name of the equivalent native operator as shipped with Houdini.
```
pcucka: This doesn't belong here, does it?

openvdb_houdini/SOP_NodeVDB.cc
danrbailey: This isn't in the hot code path, so I don't think it's necessary to optimize this. The main reason I did it like this is so that changing the logic to adjust the invisible variable can be done without modifying any existing lines. Adding, reverting and combining changes as we figure out the right logic here should be less likely to go wrong as a result.

openvdb_houdini/SOP_NodeVDB.h
danrbailey: Good catch, deleted.

danrbailey:
Any further comments? I think I've addressed all feedback or responded to all comments now.

pcucka:
This was merged prematurely.  The "nativeInvisible" logic is needed only if we opt to implement support for the environment variable at the C++ level, and that has not yet been decided.

danrbailey:
> This was merged prematurely. The "nativeInvisible" logic is needed only if we opt to implement support for the environment variable at the C++ level, and that has not yet been decided.

Ah I see, well we can opt to pull that bit out depending on what we decide - nothing is calling that method yet. I'll send round some agenda suggestions and include this as one of the items.

PR #459, opened 2019-06-11
Create MAINTAINERS.md

jmertic:
Solves one of the graduation requirements :-)

Signed-off-by: John Mertic <jmertic@linuxfoundation.org>

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,5 +1,5 @@
 <!-- SPDX-License-Identifier: CC-BY-4.0 -->
-<!-- Copyright Contributors to the ODPi Egeria project. -->
+<!-- Copyright Contributors to the OpenVDB project. -->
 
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,15 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
```
danrbailey: Should this not be 

As the project uses this license?

(Also think the copyright contributors might need to change too)

jmertic: Good catch on the copyright statement - fixed now.

For the license, since this is technically documentation it would fall under the documentation license ( CC-BY-40 ).

danrbailey: Ah right, I see. Yes we haven't been adding a license to our documentation, but perhaps we should.

PR #460, opened 2019-06-13
Use hash map with Coord key for faster performance

danrbailey:
This is a simple performance improvement to switch a couple of std::maps to std::unordered_maps as an openvdb::Coord hash is now available.

### DIFF
--- a/openvdb/points/IndexFilter.h
+++ b/openvdb/points/IndexFilter.h
@@ -79,6 +79,7 @@
 
 #include <random> // std::mt19937
 #include <numeric> // std::iota
+#include <unordered_map>
 
 
 class TestIndexFilter;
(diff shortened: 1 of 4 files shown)


openvdb/points/IndexFilter.h
```diff
@@ -255,7 +256,7 @@ class RandomLeafFilter
 {
 public:
     using SeedCountPair = std::pair<Index, Index>;
-    using LeafMap       = std::map<openvdb::Coord, SeedCountPair>;
+    using LeafMap       = std::unordered_map<openvdb::Coord, SeedCountPair>;
```
pcucka: This is technically an API change, since it's a public declaration in a public class.

danrbailey: Yes, it is and they really should have been private declarations in the beginning. Having said that these aliases aren't being used in used in the public class interface or accessible through any public member variable so I'm not too concerned. If we want to be strict with this though, I can add a new private alias with a new name and mark this one as deprecated?

pcucka: I agree that it's not critical.  But another option would be to bump the library version to 6.2.0 (which we were pretty much expecting to have to do anyway) and mention the change in the release notes.

danrbailey: Yeah, ok.

PR #461, opened 2019-06-13
OpenVDB SOPs Tab Sub-Menu

danrbailey:
Add a tab sub-menu option to OpPolicy/OpFactory and set OpenVDB SOPs to use "ASWF" instead of "Custom".

### DIFF
--- a/openvdb_houdini/SOP_NodeVDB.cc
+++ b/openvdb_houdini/SOP_NodeVDB.cc
@@ -686,7 +686,7 @@ public:
 
     std::string getTabSubMenuPath(const houdini_utils::OpFactory& factory) override
     {
-        return "ASWF";
+        return "VDB/ASWF";
     }
 };
 
--- a/openvdb_houdini/SOP_NodeVDB.cc
+++ b/openvdb_houdini/SOP_NodeVDB.cc
@@ -652,6 +652,11 @@ public:
     {
         return this->getLowercaseName(this->getValidName(english));
     }
+
+    std::string getTabSubMenuPath(const houdini_utils::OpFactory& factory) override
+    {
+        return "VDB";
+    }
 };
 
 
@@ -677,6 +682,11 @@ public:
     std::string getNativeName(const houdini_utils::OpFactory& factory) override
     {
         return this->getLowercaseName(this->getValidName(factory.english()));
+    }
+
+    std::string getTabSubMenuPath(const houdini_utils::OpFactory& factory) override
+    {
+        return "ASWF";
     }
 };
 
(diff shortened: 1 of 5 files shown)


openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -678,6 +683,11 @@ class ASWFOpenVDBOpPolicy: public DefaultOpenVDBOpPolicy
     {
         return this->getLowercaseName(this->getValidName(factory.english()));
     }
+
+    std::string getTabSubMenuPath(const houdini_utils::OpFactory& factory) override
+    {
+        return "ASWF";
```
pcucka: Wouldn't something more self-explanatory like "VDB - ASWF" be better?  It would have the additional advantage of sorting next to the native submenu.

danrbailey: I don't have a big problem with this, but just to mention that the main downside is that it makes the tab sub-menu suffix quite large when both are being shown and we are relying on the implicit SideFX label renaming:

* VDB LOD (VDB)
* VDB LOD (VDB - ASWF)

How do you feel about that?

pcucka: In what contexts would the suffix be visible?

danrbailey: You have to drop the "Open" prefix from the label names to see this, but if you tab and click on the sub-menu option ("VDB" or "ASWF"), you get a list without any suffixes ("VDB LOD", "VDB...", etc), however if you tab and type "VDB" it puts both SOPs in one big list and attempts to disambiguate where there are two nodes with identical label names by suffixing the sub-menu option to the end of the label name in parentheses. 

The point in taking advantage of this feature is not having to do any of our own renaming, they will only be renamed by Houdini when they show up alongside another SOP with the same label name.

(Although I'll mention that there's currently a bug in this feature, as of 17.5.280).

pcucka: Another option to consider: set the submenu path to `"VDB/ASWF"`.  That moves the ASWF SOPs into a submenu of the VDB submenu (which is not ideal), but causes search results to be displayed as desired, e.g., as "VDB LOD (VDB)" and "VDB LOD (ASWF)".

danrbailey: Interesting. It might make sub-menu navigation a bit strange if there are no VDB SOPs visible (once the remaining SideFX SOPs have been contributed to the repo). Perhaps we discuss this in the meeting this week though, I'm not sure I have a strong preference here.

@jmlait - I submitted a bug report for what I believe is a problem in the suffix representation (Houdini Bug #97465).

danrbailey: As discussed in the TSC meeting, I've just updated this PR to go with your suggestion here (VDB/ASWF), thanks!

PR #462, opened 2019-06-14
Security Policy.

jmlait:
New security policy

### DIFF
--- a/tsc/process/security.md
+++ b/tsc/process/security.md
@@ -1,0 +1,100 @@
+**Security and OpenVDB**
+
+The OpenVDB project takes security very seriously.
+
+OpenVDB was not originally built as an outward-facing library.
+Users should exercise caution when working with untrusted data.
+Code injection bugs have been found in much simpler data structures,
+so it would be foolish to assume OpenVDB is immune.
+
+OpenVDB is also focused on high performance.   It will rely on
+incoming parameters being valid.  For example, array bounds checking is
+intentionally avoided.  Likewise, integer overflow concerns are
+intentionally not addressed.
+
+***Reporting***
+
+If you discover a security vulnerability that you feel cannot
+be disclosed publicly, please submit it to security@openvdb.org.
+This will go to a private mailing list of the TSC where we will
+endeavour to triage and respond to your issue in a timely manner.
+
+***Outstanding Security Issues***
+
+None
+
+***Addressed Security Issues***
+
+None
+
+***File Format Expectations***
+
+Attempting to read a .vdb file will:
+* Return success and produce a valid VDB data structure in memory
+* Fail with an error
+* Execute forever
+* Run out of memory
+
+The last two options may be surprising.  VDBs, however, are designed
+as open-ended containers of production data that may be terabytes in size.
+
+It is a bug if some file causes the library to crash.  It is a serious
+security issue if some file causes arbitrary code execution.
+
+***Runtime Library Expectations***
+
+We consider the library to run with the same privilage as the linked
+code.  As such, we do not guarantee any safety against malformed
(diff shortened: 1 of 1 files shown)


tsc/process/security.md
```diff
@@ -0,0 +1,100 @@
...
+
+The last two options may be surprising.  VDBs, however, are designed
+as open-ended containers of production data that may be terabytes in size.
+
+It is a bug if some file causes the library to crash.  It is a serious
+security issue if some file causes arbitrary code execution.
+
+***Runtime Library Expectations***
+
+We consider the library to run with the same privilage as the linked
```
danrbailey: typo - privilage

tsc/process/security.md
```diff
@@ -0,0 +1,100 @@
...
+code.  As such, we do not guarantee any safety against malformed
+arguments.   Provided functions are called with well-formed
+parameters, we expect the same set of behaviours as with file
```
danrbailey: I struggle a lot with this rule - but aren't we supposed to be using American English? Ie behavior vs behaviour.

tsc/process/security.md
jmlait: I had it spelled even worse in the original :>  When I fixed I obviously didn't fix far enough :>

tsc/process/security.md
jmlait: Canadian English allows the use of either at the author's discretion.  Like you, my default is to keep the 'u's.  Which leads to embarrassments like the neighbour() function in VEX.  Americanized it in any case.

tsc/process/security.md
```diff
@@ -0,0 +1,100 @@
...
+
+The best practice for building a clean VDB is populate an
+empty grid voxel-by-voxel with the desired data and only
+copy known and trusted metadata fields.
+
+****Inactive Voxels****
+
+When voxels are marked inactive in the grid, they are not cleared
+to the background value.  If you rely on the data being deleted, you
+should set it as well as setting it inactive.  In addition, calling
```
pcucka: How about "you should overwrite voxels' values as well as deactivating them"?

tsc/process/security.md
```diff
@@ -0,0 +1,100 @@
+**Security and OpenVDB**
+
+The OpenVDB project takes security very seriously.
```
pcucka: "The OpenVDB Technical Steering Committee (TSC) takes security very seriously" might be a better way to phrase this.

danrbailey:
@jmlait - I think you missed a dco.

danrbailey:
@kmuseth - think we're just waiting on your approval here, would you have a few minutes to review? There's a couple of CII best practice items we can tick off once this gets merged. :)

tsc/process/security.md
```diff
@@ -0,0 +1,101 @@
...
+
+****Steganographic****
+
+Most image-based steganographic techniques can be applied to VDBs.
+Narrow band SDF have an additional concern, however.   Since only
+the zero crossing affects the perceived geometry, there is considerable
+room for information hiding in the off-band voxels.  Conversion to
+polygon and reconstruction from the polygonal mesh should eliminate
+those channels.
+
```
kmuseth: I would emphasize that by design most of the algorithms provided in the OpenVDB library ignore inactive values. Narrow band level sets is just one example of this type of data. As such virtually any VDB grid could be constructed to contain hidden information encoded as inactive values.

kmuseth: The easiest way to mediate this is to first run tools::changeBackground, which sets all inactive values to the same value, followed by tools::pruneInactive, which eliminates nodes with inactive values only (note that this method will ignore any nodes with active values, which is why the first method must be called as well)

PR #463, opened 2019-06-14
OVDB-106: Add spare data support to OpenVDB operators

pcucka:
This PR adds a spare data mechanism to `OP_Operator`s created with `OpFactory`.  I deliberately kept the API simple (we can extend it if the need arises, but multi-gigabyte spare data is already supported), and I have not included any ophide logic (I will file my `pythonrc.py` implementation as a separate PR).

To complement the C++ API, I've added HScript and Python functions to retrieve (but not set) spare data:

**HScript**
```
> opsparedata
opsparedata

    List spare data associated with an operator type.

    USAGE
      opsparedata <networktype> <opname> [<token>]

    When the token is omitted, all (token, value) pairs
    associated with the operator type are displayed.

    Currently, only operator types defined with OpenVDB's OpFactory
    can have spare data.  See www.openvdb.org for more information.

    EXAMPLES
      > opsparedata Sop DW_OpenVDBConvert
        lists all spare data associated with the Convert VDB SOP
      > opsparedata Sop DW_OpenVDBClip nativename
        displays the VDB Clip SOP's native name

> opsparedata Sop DW_OpenVDBClip
nativename vdbclip

> opsparedata Sop DW_OpenVDBClip nativename
vdbclip

> opsparedata Sop vdbclip

```

**HOM**
```
>>> help(hou.NodeType.spareData)
Help on method _spareData in module __main__:

_spareData(self, name) unbound hou.NodeType method
    spareData(name) -> str or None
    
    Return the spare data with the given name, or None
    if no data with that name is defined for this node type.
    
    Currently, only node types defined with OpenVDB's OpFactory
    can have spare data.  See www.openvdb.org for more information.

>>> help(hou.NodeType.spareDataDict)
Help on method _spareDataDict in module __main__:

_spareDataDict(self) unbound hou.NodeType method
    spareDataDict() -> dict of str to str
    
    Return a dictionary of the spare data for this node type.
    
    Currently, only node types defined with OpenVDB's OpFactory
    can have spare data.  See www.openvdb.org for more information.

>>> hou.nodeType('Sop/DW_OpenVDBClip').spareDataDict()
{'nativename': 'vdbclip'}

>>> hou.nodeType('Sop/DW_OpenVDBClip').spareData('nativename')
'vdbclip'

>>> hou.nodeType('Sop/vdbclip').spareDataDict()
{}
```

### DIFF
--- a/openvdb_houdini/ParmFactory.cc
+++ b/openvdb_houdini/ParmFactory.cc
@@ -127,6 +127,17 @@ getSparePointer(const PRM_SpareData* spare, const char* token, const void* deflt
     return deflt;
 }
 
+
+// Copy elements from one spare data map to another,
+// overwriting any existing elements with the same keys.
+inline void
+mergeSpareData(SpareDataMap& dst, const SpareDataMap& src)
+{
+    for (const auto& it: src) {
+        dst[it.first] = it.second;
+    }
+}
+
 } // anonymous namespace
 
 
@@ -1058,7 +1069,7 @@ struct OpFactory::Impl
 
         if (mVerb) SOP_NodeVerb::registerVerb(mVerb);
 
-        op->spareData().insert(mSpareData.begin(), mSpareData.end());
+        mergeSpareData(op->spareData(), mSpareData);
 
         return op;
     }
@@ -1123,6 +1134,7 @@ del nt, _spareData, _spareDataDict\n");
         const char* const token = args[3];
 
         if (!networkType || !opName) {
+            /// @todo Install this as a command.help file?
             args.out() << kSpareDataCmdName << "\n\
 \n\
     List spare data associated with an operator type.\n\
@@ -1482,7 +1494,7 @@ OpFactory::setInvisible()
 OpFactory&
 OpFactory::addSpareData(const SpareDataMap& spare)
 {
-    mImpl->mSpareData.insert(spare.begin(), spare.end());
+    mergeSpareData(mImpl->mSpareData, spare);
     return *this;
 }
 
@@ -1504,7 +1516,7 @@ void
 addOperatorSpareData(OP_Operator& op, SpareDataMap& spare)
 {
     if (auto* opdw = dynamic_cast<OP_OperatorDW*>(&op)) {
-        opdw->spareData().insert(spare.begin(), spare.end());
+        mergeSpareData(opdw->spareData(), spare);
     } else {
         throw std::runtime_error("spare data cannot be added to the \""
             + op.getName().toStdString() + "\" operator");
--- a/openvdb_houdini/ParmFactory.cc
+++ b/openvdb_houdini/ParmFactory.cc
@@ -34,6 +34,8 @@
 #include "ParmFactory.h"
 
 #include <CH/CH_Manager.h>
+#include <CMD/CMD_Args.h>
+#include <CMD/CMD_Manager.h>
 #include <GOP/GOP_GroupParse.h>
 #include <GU/GU_Detail.h>
 #include <GU/GU_PrimPoly.h>
@@ -644,12 +646,11 @@ ParmFactory&
 ParmFactory::setRange(const PRM_Range* r) { mImpl->range = r; return *this; }
 
 ParmFactory&
-ParmFactory::setSpareData(const std::map<std::string, std::string>& items)
+ParmFactory::setSpareData(const SpareDataMap& items)
 {
-    using StringMap = std::map<std::string, std::string>;
     if (!items.empty()) {
         if (!mImpl->spareData) { mImpl->spareData = new PRM_SpareData; }
-        for (StringMap::const_iterator i = items.begin(), e = items.end(); i != e; ++i) {
+        for (SpareDataMap::const_iterator i = items.begin(), e = items.end(); i != e; ++i) {
             mImpl->spareData->addTokenValue(i->first.c_str(), i->second.c_str());
         }
     }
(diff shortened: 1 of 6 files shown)


openvdb_houdini/ParmFactory.cc
```diff
@@ -1047,9 +1069,200 @@ struct OpFactory::Impl
...
+        // The operator's name
+        const char* const opName = args[2];
+        // An optional spare data token
+        const char* const token = args[3];
+
+        if (!networkType || !opName) {
+            /// @todo Install this as a command.help file?
+            args.out() << kSpareDataCmdName << "\n\
+\n\
+    List spare data associated with an operator type.\n\
```
pcucka: I know this isn't the right way to document an HScript command, but I really don't want to move these twenty lines to a separate `command.help` file.

danrbailey:
I think in general, this idea is good, but I do have a few concerns.

This introduces quite a bit of complexity in adding an hscript _and_ a python accessor for this spare data. This also adds a Python dependency to the houdini_utils module that didn't previously have it. Would it be acceptable to consider adding just the hscript accessor for now and relying on using this from Python with `hou.hscript()` until we have a better idea of how useful/important a mechanism this is going to be? 

As I understand it, spare data is both stored on each instance of the node and is serialized and deserialized with it, neither of which is desirable for native name. It can (and should) be re-generated on starting a Houdini session and it only needs to be stored once with the operator definition rather than on every node instance. I think the VDB version is a much better example of where this mechanism is needed as discussed here (#452), so perhaps we could start there?

openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -720,19 +707,11 @@ OpenVDBOpFactory::setNativeName(const std::string& name)
 {
     // SideFX nodes have no native equivalent.
 #ifndef SESI_OPENVDB
-    mNativeName = name;
+    addSpareData({{"nativename", name}});
```
danrbailey: I'd prefer to keep the existing solution where we update a member variable on the factory and then populate the spare data just once during factory deconstruction to avoid multiple read and write operations having to go through this dict.

pcucka:
> This introduces quite a bit of complexity in adding an hscript and a python accessor for this spare data.

This PR adds fewer than 300 lines of code, including comments and documentation.  That's fewer lines than the solution proposed in PR #404, and this API is general-purpose, not specific to ophiding.  What complexity there is is in private code, not in the public API.

> This also adds a Python dependency to the houdini_utils module that didn't previously have it.

This doesn't add any new library dependencies (the Python API used here is implemented in libHoudiniUT and libHoudiniPRM, AFAICT), and the header dependencies are on HDK and HBoost headers only, not on any libpython headers.

> As I understand it, spare data is both stored on each instance of the node and is serialized and deserialized with it, neither of which is desirable for native name. It can (and should) be re-generated on starting a Houdini session and it only needs to be stored once with the operator definition rather than on every node instance.

The spare data is stored with `OP_OperatorDW` instances, not with node instances.  Nodes can access the data with `getOperatorSpareData(*this->getOperator())`.

openvdb_houdini/SOP_NodeVDB.cc
pcucka: The native name will be set at most twice per operator, so the overhead is negligible.

danrbailey:
> > This also adds a Python dependency to the houdini_utils module that didn't previously have it.
> 
> This doesn't add any new library dependencies (the Python API used here is implemented in libHoudiniUT and libHoudiniPRM, AFAICT), and the header dependencies are on HDK and HBoost headers only, not on any libpython headers.

I believe you're right regarding the libraries we would link against. It was really more a question of whether providing an implementation for both languages (HScript and Python) was essential from the perspective of the extra code maintenance and the new headers and classes we would pull in and have to adapt our code to if they change. 

The right answer could be only offering the Python implementation, I don't know. It just seemed like a worthwhile question to ask considering how easy it is to access hscript functionality from python. In many cases SideFX doesn't even offer a Python implementation such as the `oprename` function you pointed out the other day.

danrbailey:
> > This introduces quite a bit of complexity in adding an hscript and a python accessor for this spare data.
> 
> This PR adds fewer than 300 lines of code, including comments and documentation. That's fewer lines than the solution proposed in PR #404, and this API is general-purpose, not specific to ophiding. What complexity there is is in private code, not in the public API.

I wasn't comparing this to PR #404, that's very out-of-date now. I've closed that one and put in a new PR #464 to contain the much more lightweight implementation I'm now proposing.

> > As I understand it, spare data is both stored on each instance of the node and is serialized and deserialized with it, neither of which is desirable for native name. It can (and should) be re-generated on starting a Houdini session and it only needs to be stored once with the operator definition rather than on every node instance.
> 
> The spare data is stored with `OP_OperatorDW` instances, not with node instances. Nodes can access the data with `getOperatorSpareData(*this->getOperator())`.

I think where I'm confused is that my expectation was that to solve the VDB version issue we would need to store a version with each node instance and put that into the hip file. On re-opening we could override `OP_Node::getVersion()` to pull from spare data for each node instance and use `syncNodeVersion` to maintain previous behavior. I guess I'm still not clear how exactly this solves this problem if this data doesn't live with the node instance or indeed get serialized with the file? I'm going to experiment with this to better understand how we might use this to solve #452.

pcucka:
The spare data API doesn't solve the versioning problem on its own.  It solves the problem of communication between a node and its operator without introducing an OpenVDB dependency into `houdini_utils`.  Depending on when `syncNodeVersion` gets called during deserialization, that might not be sufficient.

danrbailey:
@pcucka - are you waiting on any more feedback to merge this? I'm waiting for this to go in to be able to make updates to #452 and #464.

PR #464, opened 2019-06-17
OVDB-64: Unify SOP Labels and Hiding

danrbailey:
After much discussion, this is a much simpler PR to replace my original proposal #404.

This renames the labels from "OpenVDB" to "VDB" and relies on the native SideFX behavior for label suffix based on sub-menu. It also adds a compile flag and an environment variable to control the hiding behavior. My current proposal is to include this _in addition_ to an OPcustomize/pythonrc solution which could hold any more complicated logic that we need.

(#461 will need to be merged first to ensure that the tab sub-menu is correct).


openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -708,6 +708,16 @@ OpenVDBOpFactory::~OpenVDBOpFactory()
     if (!mNativeName.empty()) {
         bool invisible = mNativeInvisible;
 
+#ifdef OPENVDB_HOUDINI_HIDE_NATIVE_SOPS
+        // hide the native nodes unless overridden
```
pcucka: As discussed in the TSC meeting, it would be good to log a warning if the environment variable is set but support for it was disabled at compile time.

openvdb_houdini/ParmFactory.cc
```diff
@@ -1316,6 +1317,12 @@ OpFactory::~OpFactory()
     if (mImpl->mInvisible) {
         mImpl->mTable->addOpHidden(mImpl->mName.c_str());
     }
+
+    // hide other nodes in invisible name list
+
+    for (const auto& name : mImpl->mInvisibleNames) {
+        mImpl->mTable->addOpHidden(name.c_str());
+    }
```
danrbailey: @pcucka - I could see two main options as to how to provide node names to hide to the operator table. This option (the one I went with) simply stores a vector of names to be hidden on the factory implementation class. The alternative I considered was to add a protected non-const OpFactory::table() method so as to be able to modify it directly from the derived class.

I don't really have a preference here though, so if you'd prefer the other way of doing this or have any other ideas, I'm happy to do this differently.

openvdb_houdini/SOP_NodeVDB.cc
danrbailey: My understanding was that we'd log an error if the opcustomize and the factory compile flag was enabled togther (because they'd both essentially be doing the same thing and that could be confusing).

In my new implementation, I'm using the same environment variable as in your pythonrc file, so there's actually never a situation where support for the environment variable is disabled at compile time.

openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -711,13 +711,48 @@ OpenVDBOpFactory::OpenVDBOpFactory(
...
+    }
+}
 
 OpenVDBOpFactory&
 OpenVDBOpFactory::setNativeName(const std::string& name)
 {
     // SideFX nodes have no native equivalent.
 #ifndef SESI_OPENVDB
-    addSpareData({{"nativename", name}});
+    mNativeName = name;
```
danrbailey: @pcucka - It's possible I missed something, but I didn't see a way to retrieve the "nativename" value from the base class, so I've modified this implementation slightly to re-introduce the native name member variable and then add the spare data in the deconstructor of this class.

danrbailey:
@pcucka - I'd like to get this merged in as soon as possible. Could you possibly spare a few minutes to look over the changes I've made? It's not that many lines of code, so hopefully won't take too long.

openvdb_houdini/SOP_NodeVDB.cc
pcucka: All of the logic that's currently in the destructor can be moved into `setNativeName`, can't it?

danrbailey: No because setNativeName() is called multiple times for many operators - once from the OpenVDBOpFactory constructor and then overridden from the operator registration, which would add multiple hide names to the OP_Table. The only option to handle this cleanly would be to remove the old name from the OP_Table when it is modified which seems like a lot of extra redundant logic.

pcucka: I don't see any problem with checking whether the native SOP exists before changing visibilities or with calling `OP_OperatorTable::delOpHidden` to undo a change.

danrbailey: Ok, then I shall do it like that.

openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -711,10 +711,24 @@ OpenVDBOpFactory::OpenVDBOpFactory(
...
+        this->table().isOpHidden(mNativeName.c_str())) {
+        this->table().delOpHidden(mNativeName.c_str());
+    }
+
+    mNativeName = name;
 
+    if (!name.empty()) {
         // environment variable takes precedence over compiler flag
 
         const char* opHidePolicy = std::getenv("OPENVDB_OPHIDE_POLICY");
```
pcucka: You should `#include <cctype>` for `std::tolower`.

Also, how about moving all of the string manipulation logic (lines 734–749) into a function?
```
namespace {

const std::string&
getOpHidePolicy()
{
    static bool sInitialized = false;
    static std::string sOpHidePolicy;
    if (!sInitialized) {
        const char* opHidePolicy = std::getenv("OPENVDB_OPHIDE_POLICY");
        ...
        sOpHidePolicy = opHidePolicy;
        std::transform(sOpHidePolicy.begin(), sOpHidePolicy.end(), sOpHidePolicy.begin(),
            [](unsigned char c) { return std::tolower(c); });
        sInitialized = true;
    }
    return sOpHidePolicy;
}

} // anonymous namespace
```

danrbailey: I've just implemented this (with a minor change to use std::call_once instead of a boolean).

pcucka: Doesn't the Mako comment belong with the `node_info_text` namespace?

danrbailey: Not really, the mako comment belongs with this define that immediately follows it:

//#define OPENVDB_CUSTOM_MAKO

which refers to the block of code captured by #ifdef OPENVDB_CUSTOM_MAKO much later in the file (https://github.com/danrbailey/openvdb/blob/unifysops/openvdb_houdini/SOP_NodeVDB.cc#L283).

pcucka: Might be good at some point to rephrase that comment to something like "Define OPENVDB_CUSTOM_MAKO to enable custom UT_InfoTree data...".

danrbailey: Yes, we could reword that a little so it's clearer.

PR #466, opened 2019-06-18
OVDB-64: pythonrc.py ophide implementation

pcucka:
This is the promised `pythonrc.py` implementation of an environment variable-based ophiding mechanism.  The advantage of this approach, as I see it, is that it is a single implementation that makes the logic explicit and easily accessible to TDs.

This PR makes use of the spare data API proposed in PR #463.


openvdb_houdini/CMakeLists.txt
```diff
@@ -372,3 +372,8 @@ install(FILES
   DW_OpenVDBRasterizePoints.cmd
   DESTINATION ${OPENVDB_HOUDINI_INSTALL_PREFIX}/scripts/sop
 )
+
+install(FILES
```
danrbailey: This should probably be an opt-in mechanism so that it doesn't wipe out an existing pythonrc.py file that already exists in this location. By default I think the base installation directory is $HOUDINI_USER_PREF_DIR which is obviously not specific to VDB.

pcucka: Agreed.

PR #471, opened 2019-06-20
Update dependencies table to use TBB 4.4.6

danrbailey:
As suggested by @Idclip in this PR (https://github.com/AcademySoftwareFoundation/openvdb/pull/416), bumping the minimum version dependencies table for TBB from 4.4 to 4.4.6.

We have been successfully building against 4.4.6 for over a month in our CI and I haven't seen a single memory allocation bug that we sporadically ran into using 4.4.0, so it looks like this was a good guess.


Idclip:
Can you also bump the `MINIMUM_TBB_VERSION` cmake variable in the root cmake?

danrbailey:
> Can you also bump the `MINIMUM_TBB_VERSION` cmake variable in the root cmake?

Good catch, done.

danrbailey:
> > Can you also bump the `MINIMUM_TBB_VERSION` cmake variable in the root cmake?
> 
> Good catch, done.

I had to revert this change unfortunately - include/tbb/tbb_stddef.h doesn't define the patch version so CMake only knows major.minor for it's version.

Idclip:
> I had to revert this change unfortunately - include/tbb/tbb_stddef.h doesn't define the patch version so CMake only knows major.minor for it's version.

Ah annoying, thanks for trying!

PR #474, opened 2019-06-21
CMake GLFW Improvements

danrbailey:
* Bump minimum version to 3.1
* Accept cmake/glfw as well as cmake/glfw3 paths for CMake config
* Add SYSTEM_LIBRARY_PATHS to GLFW search paths
* Include glfwTargets.cmake or glfw3Targets.cmake to setup targets
* Support X11 for GLFW/vdbview on Linux and add an option which defaults to on
* Added X11 dependencies on Circle

Tested on:

* Ubuntu16.04 with apt-get glfw
* CentOS7 with 3.1
* CentOS7 with 3.4 (ie latest)
* Mac OSX 14.04 (works, but lots of deprecation warnings due to OpenGL being deprecated)


cmake/OpenVDBGLFW3Setup.cmake
```diff
@@ -89,18 +89,33 @@ if(PC_glfw3_FOUND)
   endforeach()
 endif()
 
+list(APPEND _GLFW3_ROOT_SEARCH_DIR ${SYSTEM_LIBRARY_PATHS})
```
Idclip: oops I must have forgotten this, good catch

openvdb/cmd/CMakeLists.txt
```diff
@@ -199,6 +205,20 @@ if(OPENVDB_BUILD_VDB_VIEW)
...
+      list(APPEND OPENVDB_X11_LIBS
+        ${X11_Xrandr_LIB}
+        ${X11_Xxf86vm_LIB}
+        ${X11_Xcursor_LIB}
+        ${X11_Xinerama_LIB}
+        ${X11_Xi_LIB}
+        ${X11_LIBRARIES}
+        ${CMAKE_DL_LIBS}
+      )
+    endif()
```
Idclip: doesn't glfw's cmake pull these in for you?

Idclip: Running find_package(glfw) without CONFIG mode sets up these interface libraries on Linux
```
set_target_properties(glfw PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "/usr/lib64/librt.so;/usr/lib64/libm.so;dl;/usr/lib64/libX11.so;-lpthread;/usr/lib64/libXrandr.so;/usr/lib64/libXinerama.so;/usr/lib64/libXxf86vm.so;/usr/lib64/libXcursor.so"
)
```

What there a particular reason we want this to be optional? Is this applicable for mac too? Usually it's better if we let the 3rd party cmake sort out it's own deps.

If we do choose to do this, a comment about updating to imported targets in cmake 3.14 for x11 would also be handy

danrbailey: Unfortunately, INTERFACE_LINK_LIBRARIES is only defined from GLFW 3.2 onwards, see this commit where the export line was introduced:

https://github.com/glfw/glfw/commit/7f7ad39e150320c53b3970e906c12f8bd31e1531#diff-af3b638bc2a3e6c650974192a53c7291R421

In order to make this work with GLFW 3.1, we have to define these libraries ourselves. I've added a change that sets this property ourselves if not defined, so at least this logic is isolated to the GLFW CMake wrapper.

Regarding making X11 optional - GLFW allows for compiling against X11, Wayland or Mesa (https://www.glfw.org/docs/latest/compile.html), so X11 isn't technically a mandatory dependency. That being said, because this only applies on 3.1 due to the lack of the property and X11 is by far the most common dependency, I've just removed the option and added a warning stating that we only support X11 on Unix when that property isn't set. I tested a vanilla installation of GLFW 3.1.2 on Mac OSX 14.04 and it seems the view application builds and runs without adding any extra libraries so I'm not going to planning to do any more here. (Anyway Mac support is very tentative right now due to the phasing out of OpenGL in favour of Metal).

Idclip: Nice, thanks for the explanation

danrbailey:
> Seems like the `openvdb/INSTALL` file has crept back in? Otherwise looks good, thanks!

Well spotted, I must have made a mistake with the merge. I've just fixed that and I'll merge it once the CI completes.

PR #475, opened 2019-06-24
Bug fix paged array

kmuseth:
Fixed a bug in util::PagedArray caused by a race condition (due to an incorrect template argument)

### DIFF
--- a/openvdb/CHANGES
+++ b/openvdb/CHANGES
@@ -17,8 +17,10 @@ Version 6.2.0 - In development
     - Minor performance improvements to moving and filtering VDB points.
 
     Bug fixes:
-    - Fixed a bug in util::PagedArray that could produce a crash due to a race 
-      condution.
+    - Fixed a (rare) crash in util::PagedArray caused by the fact that std::vector
+      (unlike std::deque) does not allow mutiple threads to call operator[] while one thread
+      calls push_back. The last template argument of PageArray is now elimited
+      and PageTableT is set to std::deque.
     - Fixed a bug in tools::computeScalarPotential() that could produce
       a corrupt result due to invalid memory access.
       [Reported by Edwin Braun]
--- a/openvdb/CHANGES
+++ b/openvdb/CHANGES
@@ -17,6 +17,8 @@ Version 6.2.0 - In development
     - Minor performance improvements to moving and filtering VDB points.
 
     Bug fixes:
+    - Fixed a bug in util::PagedArray that could produce a crash due to a race 
+      condution.
     - Fixed a bug in tools::computeScalarPotential() that could produce
       a corrupt result due to invalid memory access.
       [Reported by Edwin Braun]
--- a/openvdb/util/PagedArray.h
+++ b/openvdb/util/PagedArray.h
@@ -43,7 +43,8 @@
 #define OPENVDB_UTIL_PAGED_ARRAY_HAS_BEEN_INCLUDED
 
 #include <openvdb/version.h>
-#include <vector>
+#include <openvdb/Types.h>// SharedPtr
+#include <deque>
 #include <cassert>
 #include <iostream>
 #include <algorithm>// std::swap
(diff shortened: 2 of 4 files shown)


openvdb/util/PagedArray.h
```diff
@@ -185,7 +186,7 @@ namespace util {
 
 template<typename ValueT,
          size_t Log2PageSize = 10UL,
-         template<typename ...> class TableT = std::vector>
+         template<typename ...> class TableT = std::deque>// must allow mutiple threads to call operator[] as long as only one thread calls push_back
```
pcucka: If it's not safe to use `std::vector`, would it not be wise to add a static assertion (ideally, one that invokes an `IsThreadsafeRWContainer` type trait)?

Or maybe this class doesn't need to be templated on `TableT` at all?

openvdb/util/PagedArray.h
```diff
@@ -200,9 +201,10 @@ class PagedArray
 
 public:
     using ValueType = ValueT;
+    using Ptr       = SharedPtr<PagedArray>;
 
     /// @brief Default constructor
-    PagedArray() { mSize = 0; }
+    PagedArray() = default;
```
pcucka: Revert this change; `tbb::atomic` needs to be initialized by assignment.  (See PR #386.)

openvdb/util/PagedArray.h
```diff
@@ -234,7 +239,7 @@ class PagedArray
     /// @param value value to be added to this PagedArray
     ///
     /// @details Constant time complexity. May allocate a new page.
-    size_t push_back(const ValueType& value)
+    inline size_t push_back(const ValueType& value)
```
pcucka: Inline qualifiers are redundant for member functions defined inside the class declaration.

openvdb/util/PagedArray.h
```diff
@@ -510,8 +515,8 @@ class PagedArray
         }
     }
     PageTableT mPageTable;//holds points to allocated pages
-    tbb::atomic<size_t> mSize;// current number of elements in array
-    size_t mCapacity = 0;//capacity of array given the current page count
+    tbb::atomic<size_t> mSize{0};// current number of elements in array
```
pcucka: Revert (see PR #386).

openvdb/CHANGES
```diff
@@ -17,6 +17,8 @@ Version 6.2.0 - In development
     - Minor performance improvements to moving and filtering VDB points.
 
     Bug fixes:
+    - Fixed a bug in util::PagedArray that could produce a crash due to a race 
+      condution.
```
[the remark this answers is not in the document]
kmuseth: done

openvdb/util/PagedArray.h
kmuseth: removed last template argument

openvdb/util/PagedArray.h
kmuseth: done

openvdb/util/PagedArray.h
kmuseth: I'm finding conflicting information on this subject matter. For instance, one of the very first google search results on "C++ inline keyword" includes and a example of an application identical to mine (see code in http://www.cplusplus.com/articles/2LywvCM9/). Other forum discussions seem to suggest that the C++ standard does indeed guarantee that member functions that are defined in the declaration will be treated as inline functions. I'm happy to make the change but I'd feel much better is there was definitive proof that it is indeed redundant. Can you point me to something that settles this?

openvdb/util/PagedArray.h
kmuseth: done

kmuseth:
addressed review comments

openvdb/util/PagedArray.h
pcucka: cplusplus.com is notoriously unreliable; use cppreference.com instead.

Here's the cppreference entry on inline functions: https://en.cppreference.com/w/cpp/language/inline.

And here's the C++ standard: http://open-std.org/JTC1/SC22/WG21/docs/papers/2019/n4800.pdf (see line 4 of section 9.1.6, "The inline specifier").

kmuseth: cool - second sentence in https://en.cppreference.com/w/cpp/language/inline settled it!

I've already made the code changes

PR #476, opened 2019-06-24
OVDB-44: Move documentation to a top-level directory

pcucka:
This PR moves documentation files from the `openvdb/` subdirectory to a top-level `doc/` directory and adds documentation on the Houdini API, with some examples.  (We can add Maya API docs similarly, but I would prefer that to be a separate PR.)  The makefile has been updated to reflect the new directory structure, but it does not build the Houdini docs; CMake does, but only if `OPENVDB_BUILD_HOUDINI_PLUGIN` is enabled.

Also included are new `@vdblink`, `@hvdblink`, and `@hulink` Doxygen commands with a simplified syntax: "`@vdblink{tree::LeafNode::origin(),LeafNode::origin}`" instead of "`@vdblink::tree::LeafNode::origin() LeafNode::origin@endlink`".  The old syntax is deprecated but still supported.


CMakeLists.txt
```diff
@@ -276,19 +306,43 @@ if(OPENVDB_BUILD_DOCS)
...
   file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/openvdb/cmake-doxygen-config ${DOXYGEN_CONFIG_CONTENT})
+  if(OPENVDB_BUILD_HOUDINI_PLUGIN)
+    # Append Houdini-specific settings to the Doxygen config file.
+    file(READ ${CMAKE_CURRENT_SOURCE_DIR}/doc/doxygen-config-houdini DOXYGEN_CONFIG_CONTENT)
+    file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/openvdb/cmake-doxygen-config ${DOXYGEN_CONFIG_CONTENT})
+    if(Houdini_VERSION)
+      # Extract the components of the Houdini version number into a list.
+      string(REPLACE "." ";" HOUDINI_VERSION_COMPONENTS ${Houdini_VERSION})
+      list(GET HOUDINI_VERSION_COMPONENTS 0 HVERS_MAJOR)
+      list(GET HOUDINI_VERSION_COMPONENTS 1 HVERS_MINOR)
```
pcucka: This change required me to move the DCC configuration section up to ensure that `Houdini_VERSION` is defined.  I'm not sure whether it's safe to do that—@Idclip, can you comment?

appveyor.yml
```diff
@@ -27,7 +27,7 @@ build_script:
   - cmake \
     -G"Visual Studio 15 2017 Win64" \
     -DCMAKE_CONFIGURATION_TYPES=%CONFIGURATION% \
-    -DCMAKE_INSTALL_PREFIX=%APPVEYOR_BUILD_FOLDER%\install \
+    -DCMAKE_INSTALL_PREFIX=%APPVEYOR_BUILD_FOLDER%\cmake_install \
```
pcucka: `install` now conflicts with the top-level `INSTALL` file (formerly `openvdb/INSTALL`).

danrbailey: The contents of the INSTALL file are a bit odd IMO. There's the dependency information which is useful, but less verbose than the information stored in doc/dependencies.txt and then there's the Makefile build instructions, which I feel should at least live alongside openvdb/Makefile rather than being moved up into the parent directory where there is no Makefile.

Perhaps we consider moving the Makefile build instructions into doc/build.txt and then either removing all the dependency information in favour of a direct website link from README.md or otherwise append it to that readme in a new dependency section?

pcucka: Agreed.  The INSTALL file has outlived its usefulness.

pcucka:
That's a good idea, but I think I'll have to defer to someone more knowledgeable about CMake on that.

CMakeLists.txt
Idclip: Moving the DCC configuration to here is fine, the docs previously didn't depend on the houdini or maya install so if you just wanted to build a cmake configuration with the docs in then you could get unrelated errors if it couldn't find those installations. 

They still don't rely on the maya installation, so a configuration with all build options off which uses `OPENVDB_BUILD_DOCS=ON USE_HOUDINI=ON USE_MAYA=ON` will fail due to the `if(USE_MAYA AND USE_HOUDINI)` build catch. Perhaps move that branch to after the early build exit? (line 354) This will support any future behaviour where any maya docs which rely on the maya version.

Idclip:
> This is looking great! Perhaps it make sense to modularise all of the documentation CMake logic into a new doc/CMakeLists.txt and then call out into that from the root CMakeLists.txt:

Yep this would definitely be worthwhile as that doc branch is getting pretty big

PR #477, opened 2019-06-26
Optimization of level set sphere

kmuseth:
Multi-threading of LevelSetSphere


danrbailey:
Those old commits are showing up on this PR, try this to clean them up:

git fetch
git reset --hard upstream/master
git cherry-pick 799075efbf077e5fa51aacfb2772046ac62a3784
git cherry-pick b63870351d4aada1209e777332145f65b8a14c4c
git push origin optimization_of_LevelSetSphere -f

pcucka:
Some comments in `LevelSetSphere::rasterSphere` describing the threading logic would be helpful (for example, to explain why you use TLS when generating the voxel values but not when merging).

danrbailey:
It depends a little on the problem, but in my experience, doing parallel reductions tends to be slower than serial reductions when handling large data structures like trees as you end up spending more time generating intermediate temporaries than the time you save from the early reduction phases being in parallel.

I wonder if you've looked at tbb::combinable? It's still thread-local storage, but the API is well suited to this type of operation IMO.

kmuseth:
> Some comments in `LevelSetSphere::rasterSphere` describing the threading logic would be helpful (for example, to explain why you use TLS when generating the voxel values but not when merging).

I'm not sure what you mean? Both the parallel_for and parallel_reduce are both based on the TLS container of trees (denoted "pool"). The former constructs them and the latter reduces them, exactly because this allows for a minimum on concurrent tree is be constructed (one per thread), minimizing the cost of subsequently reducing them into a single tree. This is nothing new so why is called out as a new technique that needs justification?

kmuseth:
> It depends a little on the problem, but in my experience, doing parallel reductions tends to be slower than serial reductions when handling large data structures like trees as you end up spending more time generating intermediate temporaries than the time you save from the early reduction phases being in parallel.
> 
> I wonder if you've looked at tbb::combinable? It's still thread-local storage, but the API is well suited to this type of operation IMO.

I have benchmark this algorithm against several other approaches, including serial reduction and a custom concurrent reduction, and the submitted implementation was the most performant.

pcucka:
> Both the parallel_for and parallel_reduce are based on the TLS container of trees (denoted "pool"). The former constructs them and the latter reduces them, exactly because this allows for a minimum on concurrent tree is be constructed (one per thread), minimizing the cost of subsequently reducing them into a single tree.

The merge step iterates over the pool of trees, but it constructs a new tree for each Op instance.  Why did you do it that way instead of merging trees in the pool directly?

kmuseth:
> > Both the parallel_for and parallel_reduce are based on the TLS container of trees (denoted "pool"). The former constructs them and the latter reduces them, exactly because this allows for a minimum on concurrent tree is be constructed (one per thread), minimizing the cost of subsequently reducing them into a single tree.
> 
> The merge step iterates over the pool of trees, but it constructs a new tree for each Op instance. Why did you do it that way instead of merging trees in the pool directly?

I tried exactly that and this was faster

pcucka:
I'm not saying that your code is suboptimal, just that it's not clear why it was written the way it is.  A comment providing some intuition as to why it performs better than the more "obvious" approach of not constructing additional trees would help.

kmuseth:
> I'm not saying that your code is suboptimal, just that it's not clear why it was written the way it is. A comment providing some intuition as to why it performs better than the more "obvious" approach of not constructing additional trees would help.

Cool, I can certainly add that. I didn't me to come across as defensive, I was genially confused as to what you meant (as in maybe I have overlooked an obvious change for further optimization).  Adding comment ...

PR #478, opened 2019-06-27
Add hboost thread as dependency to resolve linking error

danrbailey:
Fix to a new linking error on Mac OSX that's just started happening:

```
Undefined symbols for architecture x86_64:
  "hboost::this_thread::interruption_point()", referenced from:
      hboost::condition_variable::wait(hboost::unique_lock<hboost::mutex>&) in ParmFactory.cc.o
      hboost::condition_variable::do_wait_until(hboost::unique_lock<hboost::mutex>&, timespec const&) in ParmFactory.cc.o
  "hboost::detail::get_current_thread_data()", referenced from:
      hboost::detail::interruption_checker::interruption_checker(_opaque_pthread_mutex_t*, _opaque_pthread_cond_t*) in ParmFactory.cc.o
ld: symbol(s) not found for architecture x86_64
```

On investigation, it seems that it's coming from introducing HOM_AutoLock in #463 which itself uses UT_TaskLockT which uses boost::thread.

### DIFF
--- a/cmake/OpenVDBHoudiniSetup.cmake
+++ b/cmake/OpenVDBHoudiniSetup.cmake
@@ -197,18 +197,21 @@ if(APPLE)
   list(APPEND _HOUDINI_EXTRA_LIBRARIES
     ${_HOUDINI_LIB_DIR}/libHoudiniRAY.dylib
     ${_HOUDINI_LIB_DIR}/libhboost_regex.dylib
+    ${_HOUDINI_LIB_DIR}/libhboost_thread.dylib
   )
 else()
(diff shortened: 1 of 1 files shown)


danrbailey:
> Good catch. This is something I left out, but might be nice to put comments next to each of these deps as to why they're necessary. i.e. RAY for mantra points procedural, regex for vector merge and thread for HOM/UT_TaskLockT.

Yeah, would be nice to have. I'm just going to merge this as-is because the master branch is broken for Mac right now, but we can come back and document this bit better...

(Note that I didn't include a comment in the release notes because this is merely a fix to the spare data PR which was very recent).

PR #481, opened 2019-07-02
Add option to use colored output through CMake

danrbailey:
Not sure how much collective time has been spent over the years wrangling colored output from compilers. Many google hits still seem to reference a perl script to color the output.

One simple solution is to define GCC_COLORS in the environment which will be recognized by GCC >= 4.9. This is a convenient alternative that simply sets -fdiagnostics-color=always in GCC or -fcolor-diagnostics in Clang. Perhaps when we deprecate support for GCC 4.8 we could even enable this by default.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,8 @@ Version 6.2.0 - In development
       multithreading.
     - Increase minimum supported version of GLFW to 3.1 and improve CMake
       support for GLFW versions >= 3.1.
+    - Add a USE_COLORED_OUTPUT option to CMake to add the compiler flag to
+      color the compiler output.
 
     Bug fixes:
     - Replaced std::vector with std::deque as the underlying container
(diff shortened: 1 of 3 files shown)


Idclip:
afaik clang does this automatically?

danrbailey:
> afaik clang does this automatically?

It tries to do it automatically. The logic is based around checking that the terminal supports ANSI colors and that stderr is going to a file descriptor. I'm looking at this using Clang on a Mac xterm and it's not working for some reason, but I didn't look further than setting this compiler flag to fix it. I'll note that it also doesn't work with Clang when using other CMake generators such as ninja (because they use a pipe for their stderr).

Idclip:
> It tries to do it automatically. The logic is based around checking that the terminal supports ANSI colors and that stderr is going to a file descriptor. I'm looking at this using Clang on a Mac xterm and it's not working for some reason, but I didn't look further than setting this compiler flag to fix it. I'll note that it also doesn't work with Clang when using other CMake generators such as ninja (because they use a pipe for their stderr).

Cool thanks, haven't tested but look good

PR #482, opened 2019-07-03
Add OPENVDB_LOG statements to empty catch blocks

danrbailey:
This is to address CWE-391 as picked up by SonarCloud. The main concern is that the exception can be caught silently which could potentially lead to a security vulnerability. Adding a log statement is sufficient to address the SonarCloud vulnerability.

These five empty catch statements seem to fall into two categories:

* an exception that we don't believe should ever be thrown, but we catch for safety as method is marked "noexcept" - I've added OPENVDB_LOG_ERROR to these
* a specific exception that is an allowed and expected method failure and the logic is designed to handle this being thrown - I've added OPENVDB_LOG_INFO to these

### DIFF
--- a/openvdb/Exceptions.h
+++ b/openvdb/Exceptions.h
@@ -32,6 +32,7 @@
 #define OPENVDB_EXCEPTIONS_HAS_BEEN_INCLUDED
 
 #include <openvdb/version.h>
+#include <openvdb/util/logging.h> // OPENVDB_LOG_ERROR
 #include <exception>
 #include <sstream>
 #include <string>
@@ -52,7 +53,11 @@ public:
 
     const char* what() const noexcept override
     {
-        try { return mMessage.c_str(); } catch (...) {}
+        try {
+            return mMessage.c_str();
+        } catch (...) {
+            OPENVDB_LOG_ERROR("Unexpected exception thrown in io::Exception::what().");
+        }
         return nullptr;
     }
 
@@ -63,7 +68,9 @@ protected:
         try {
             if (eType) mMessage = eType;
             if (msg) mMessage += ": " + (*msg);
-        } catch (...) {}
+        } catch (...) {
+            OPENVDB_LOG_ERROR("Unexpected exception thrown in io::Exception constructor.");
+        }
     }
 
 private:
--- a/openvdb/math/Maps.cc
+++ b/openvdb/math/Maps.cc
@@ -232,7 +232,8 @@ approxInverse(const Mat4d& mat4d)
         try {
             return mat4d.inverse();
         } catch (ArithmeticError& ) {
-            // Mat4 code couldn't invert.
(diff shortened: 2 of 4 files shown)


openvdb/math/Maps.cc
```diff
@@ -232,7 +232,8 @@ approxInverse(const Mat4d& mat4d)
         try {
             return mat4d.inverse();
         } catch (ArithmeticError& ) {
-            // Mat4 code couldn't invert.
+            OPENVDB_LOG_INFO(
```
pcucka: I'm concerned that this is going to generate misleading log messages.  A singular matrix is the expected input to the `approxInverse` function.

openvdb/Exceptions.h
```diff
@@ -52,7 +53,11 @@ class OPENVDB_API Exception: public std::exception
 
     const char* what() const noexcept override
     {
-        try { return mMessage.c_str(); } catch (...) {}
+        try {
+            return mMessage.c_str();
+        } catch (...) {
+            OPENVDB_LOG_ERROR("Unexpected exception thrown in io::Exception::what().");
```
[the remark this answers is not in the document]
danrbailey: Good point, I think we just return this string as the result.

openvdb/math/Maps.cc
danrbailey: I agree. I think the only viable alternative is to adjust the API here (with the appropriate deprecation warnings):

`bool approxInverse(Mat4d& result, const Mat4d& mat4d);`

And communicate failure through a return value rather than an exception? I see very minimal value in changing this code that has been in use for many years just to satisfy the requirements of static analysis, but it's really the only warning that's not trivial to fix so we should probably just do it.

pcucka: `approxInverse` doesn't communicate failure through an exception, at least not in the case of a singular input matrix.  It's swallowing the exception, not throwing one.

If you're proposing instead to change the `Mat::inverse` API just to silence this static analyzer, I'm opposed to that.

danrbailey: Yes, oops, that's what I meant. A change to the API such as:

`bool inverse(Mat4& result, T tolerance = 0) const;`
`Mat4 inverse(T tolerance = 0, bool& result) const;`

What's your opposition? Sounds like you think we should just not worry about fixing this particular warning?

pcucka: Changing the API would mean that `inverse` could no longer be used within expressions ("`b = a.inverse() * x * a`") and/or that failure would be too easily ignored.

Is not fixing any of these warnings an option?  I mean, why is "`catch (...) { status = false; }`" any more secure than "`catch (...) {}; status = false;`"?

danrbailey: I totally agree. I'm going to close this PR. This is only a warning after all, if it was easy to fix we

PR #483, opened 2019-07-03
Adding a Code Coverage Option to CMake

danrbailey:
This also overrides the CMAKE_BUILD_TYPE to set it to DEBUG which is required for code coverage.


CMakeLists.txt
```diff
@@ -323,15 +324,36 @@ if(USE_CCACHE)
...
+# Code coverage configuration
+
+if(OPENVDB_CODE_COVERAGE)
+  # set gcov compile and link flags
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} --coverage")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} --coverage")
+
+  # use .gcno extension instead of .cc.gcno
+  set(CMAKE_CXX_OUTPUT_EXTENSION_REPLACE 1)
```
Idclip: Why is this necessary?

CMakeLists.txt
```diff
@@ -323,15 +324,36 @@ if(USE_CCACHE)
...
   set(CMAKE_BUILD_TYPE Release CACHE STRING
     "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE
   )
 endif()
 message(STATUS "CMake Build Type: ${CMAKE_BUILD_TYPE}")
 
+# Code coverage configuration
+
+if(OPENVDB_CODE_COVERAGE)
+  # set gcov compile and link flags
```
Idclip: Where does gcov actually get invoked?

CMakeLists.txt
danrbailey: It's common practice using CMake and code coverage to handle out-of-source builds:

https://texus.me/2015/09/06/cmake-and-gcov/

CMakeLists.txt
danrbailey: Nothing that's natively supported in our CMake (we use it in our CI only). This is mainly because there are a suite of tools that can process the .gcno files that are generated to turn the data into an html report (for example). It's left up to the user to process the data that's generated.

Idclip: Ah I meant where are the CI changes which use these options, are they still to be added?

CMakeLists.txt
Idclip: Hmm I dunno how common this is, can't find any docs for that variable! But seems like the only solution

CMakeLists.txt
danrbailey: Ah, I understand what you're asking. It's in the master branch of the repo already, included as part of the Azure changes. It currently gets called but doesn't find any code coverage files because this change hasn't been included, so reports code coverage of 0%.

PR #485, opened 2019-07-08
Added FindActiveValues

kmuseth:
Finds the active values in a tree which intersects a bounding box.
Two methods are provided, one that counts the number of active values
and one that simply tests if any active values intersect the bbox.

### DIFF
--- a/openvdb/unittest/TestFindActiveValues.cc
+++ b/openvdb/unittest/TestFindActiveValues.cc
@@ -191,33 +191,33 @@ void
 TestFindActiveValues::testSparseBox()
 {
     {//test active tiles in a sparsely filled box
-      const int half_dim = 256;
-      const openvdb::CoordBBox bbox(openvdb::Coord(-half_dim), openvdb::Coord(half_dim));
-      openvdb::FloatTree tree;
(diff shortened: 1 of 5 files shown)


openvdb/unittest/TestFindActiveValues.cc
```diff
@@ -0,0 +1,309 @@
...
+        CPPUNIT_ASSERT( bbox.max() == openvdb::Coord( n));
+        CPPUNIT_ASSERT( bbox.min() == openvdb::Coord(-n));
+    }
+}
+
+void
+TestFindActiveValues::testSparseBox()
+{
+    {//test active tiles in a sparsely filled box
+      const int half_dim = 256;
```
danrbailey: Minor style thing, could you fix this 2-space indented code?

kmuseth:
> Minor formatting thing, otherwise looks good to me.
> 
> I tested FindActiveValues::count() using the Disney cloud data-set on my 24-core machine:
> 
> Single-threaded ... completed in 712.971 milliseconds
> Multi-threaded ... completed in 101.882 milliseconds
> 
> Well worth doing the threading! :)

thanks - I removed the serial implementation

kmuseth:
merged

PR #490, opened 2019-07-11
Add TSC Meeting Nodes for July 11th

danrbailey:


tsc/meetings/2019-07-11.md
```diff
@@ -0,0 +1,118 @@
...
+
+PR 455 - Some issues raised on the Sharpening PR that need consideration.
+Optimization to repopulate stencil in Z-axis could be generalized to more axes
+and extended to support jumps of more than one voxel. Also, Peter has been asked
+by artists at DWA for a more generic convolution kernel and this PR could be
+seen as a proof of concept of doing that.
+
+An alternative to this sharpening approach could be to subtract the results of
+computing the gaussian filter and would be desirable due to the improved
+performance. However, Peter did not get usable results when attempting this, Ken
```
kmuseth: This paragraph is misrepresenting what was said and certainly what I meant. The alternative technique I proposed (and have already implemented) is a well established (actually dates back to the 1930s) referred to as "unsharp masking". The fact that Peter wasn't able to get satisfactory results does not mean that this technique doesn't work - it certainly does for me. However, it suggests that he might have found a better solution which of course remains to be proven. Both methods are based on  separable filter kernels so I don't see why there should be a significant difference in the computational efficiency, but I obviously cannot speak to the difference in quality since I haven't had a change to do a direct comparison.

danrbailey: Thanks for the clarification, I've updated the notes with your paragraph here more or less verbatim but for being in the third person.

tsc/meetings/2019-07-11.md
```diff
@@ -0,0 +1,122 @@
...
+process to loading the rest of the file. Dan to investigate and reply to this
+then Peter to do a final pass before we merge.
+
+PR 402 - Peter to look into improving VDB SOP documentation, low priority item.
+
+Main upcoming feature includes an extension to point moving to provide merging
+grid functionality. Two options for exposing in Houdini, extending the VDB
+Combine SOP and adding a new VDB Point Merge SOP. Dan presented Nick's argument
+for the latter and all agreed that this is the better option considering compile
+times and the existing node parameterization.
```
Idclip: Disclaimer: I wasn't at this meeting

I'm a bit confused by the comments here which are pro the Combine SOP due to compile times and parameterization. With compile times, I'm fairly confident that splitting it into a new SOP will be more efficient than what you're proposing. The build can be parallelized (given the existing one translation unit=one sop structure) and thus essentially overlapped with the build of the Combine SOP which is **by far** the slowest building node. Shoving more functionality into it is guaranteed to increase the maximum possible build time. Note that as this new functionality will only work on OpenVDB Points grids, only a single grid type will need to be considered, which, given the time complexity of building other nodes, becomes negligible in comparison with more than 1 thread.

In regards to node parameterization, the argument for a new node was to more closely match Houdini workflow for "points" rather than trying to augment a SOP where almost none of it's existing behaviour (maybe none at all?) is actually relevant to OpenVDB Points grids. This will definitely not be immediately intuitive to users wanting to combine (the Houdini terminology for this operation is "merge") OpenVDB Point grids. Note that if we, say, supported activity operations in Houdini on OpenVDB Points grids, this would make sense to live on the Combine node. The act of "merging" however perhaps doesn't. An additional minor point is the nice multi input advantage the merge has vs using an extra Houdini merge.

I almost missed this point, please tag me in future!

jmlait:
I also wasn't at the meeting, but I read:

Dan presented Nick's argument
+for the latter and all agreed that this is the better option
considering compile
+times and the existing node parameterization.

to meant that all agreed to use a separate node?  I agree that it should be
a separate node for the reasons you specify, so was happy when I thought
that was the conclusion :>

tsc/meetings/2019-07-11.md
danrbailey: Yes, Jeff is right, decision was for a separate SOP. With extra clarification:

Two options for exposing in Houdini, extending the VDB Combine SOP and adding a new VDB Point Merge SOP. Dan presented Nick's argument for the latter **(ie a new Point Merge SOP)** and all agreed that this is the better option...

Idclip: Thanks @jmlait @danrbailey, sorry for the confusion!! I'm not sure how I managed to miss-read this :|

tsc/meetings/2019-07-11.md
```diff
@@ -0,0 +1,122 @@
...
+
+PR 483 - Adds code coverage support to CMake, Nick currently reviewing. Minor
+issue where code coverage takes a very long time to run (2 hours+) so can only
+be run sporadically. A number of options to resolve this. Ken suggested running
+the unit tests concurrently, though there may be issues if the tests themselves
+run in parallel. Dan suggested using a smaller data set for some of the heavy
+tests, perhaps with a flag that enables the more expensive tests for a less
+frequent test run. Peter suggested using a mode that he added to the unit tests
+where you can supply a file that defines which unit tests to execute would allow
+for filtering out some of the expensive ones.
```
pcucka: My suggestion was actually to use the file mechanism to partition the tests into groups to be run concurrently (i.e., across multiple `vdb_test` processes).  I'm not in favor of excluding tests as a standard procedure.

danrbailey: Just addressed this, thanks and sorry for the delay.

PR #492, opened 2019-07-17
Replace the PROPRIETARY notices with copyright notices to make

jmlait:
Replace the PROPRIETARY notices with copyright notices to make
the intention clear.  This should make it clear that the wrapped
Dreamworks license applies to the content.

### DIFF
--- a/openvdb_houdini/GEO_PrimVDB.cc
+++ b/openvdb_houdini/GEO_PrimVDB.cc
@@ -29,12 +29,9 @@
 ///////////////////////////////////////////////////////////////////////////
 
 /*
- * PROPRIETARY INFORMATION.  This software is proprietary to
- * Side Effects Software Inc., and is not to be reproduced,
- * transmitted, or disclosed in any way without written permission.
+ * Copyright (c) Side Effects Software Inc.
  *
  * Produced by:
- *      Jeff Lait
  *      Side Effects Software Inc
  *      477 Richmond Street West
  *      Toronto, Ontario
--- a/openvdb_houdini/GEO_PrimVDB.h
+++ b/openvdb_houdini/GEO_PrimVDB.h
@@ -29,12 +29,9 @@
 ///////////////////////////////////////////////////////////////////////////
 
 /*
- * PROPRIETARY INFORMATION.  This software is proprietary to
- * Side Effects Software Inc., and is not to be reproduced,
- * transmitted, or disclosed in any way without written permission.
+ * Copyright (c) Side Effects Software Inc.
  *
  * Produced by:
- *      Jeff Lait
  *      Side Effects Software Inc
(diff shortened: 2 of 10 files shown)


danrbailey:
Thanks Jeff! There are two more Houdini files in the repo which have slightly different copyright to the others which currently requires an explicit BSD exclusion:

* GEO_VDBTranslator.cc
* SOP_VDBVerbUtils.h

Any chance you could bring these two in line as well? It looks like HDK sample text which is being included:

```
/*
 * Copyright (c) 2012
 *      Side Effects Software Inc.  All rights reserved.
 *
 * Redistribution and use of Houdini Development Kit samples in source and
 * binary forms, with or without modification, are permitted provided that the
 * following conditions are met:
 * 1. Redistributions of source code must retain the above copyright notice,
 *    this list of conditions and the following disclaimer.
 * 2. The name of Side Effects Software may not be used to endorse or
 *    promote products derived from this software without specific prior
 *    written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY SIDE EFFECTS SOFTWARE `AS IS' AND ANY EXPRESS
 * OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
 * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.  IN
 * NO EVENT SHALL SIDE EFFECTS SOFTWARE BE LIABLE FOR ANY DIRECT, INDIRECT,
 * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
 * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
 * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
 * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 *----------------------------------------------------------------------------
 */
```

jmlait:
Yes, that looks like the HDK sample header.  I think we intentionally used that rather than our internal proprietary notice as it should be compatible.

What is the BSD exclusion that is required?  Is it the explicit reference to the MPL 2.0 that is required?

Reading through it, the standard dreamworks disclaimer/non-endorsement includes contributors, so the additional text asserting those rights for Side Effects are redundant.  So I should be able to reduce this to just the copyright.  (Minus the year, of course)

PR #493, opened 2019-07-26
Using x86 instructions instead of software emulation

Const-me:
I’ve noticed you use lookup tables and bit tricks in performance-critical parts of the code, for functionality that’s implemented in hardware.

Intel introduces `BSR` and `BSF` instructions in 80386, i.e. they support them since 1985.

`POPCNT` is slightly newer, but it’s reasonable to assume it’s present, introduced in SSE4a by AMD in 2007, supported by Intel since 2008. Most people don't use 12 years old computers.

This pull request implements them for vc++, gcc and clang compilers. For unknown ones it falls back to your original software implementation.

Please let me know if you’re willing to accept the PR, is yes I should probably implement a couple of tests, so far I've only built with vc++ but haven't tested yet. Also you wrote you want me to sigh something.

If you’re not interested, I’ll just keep these changes in my private branch.

### DIFF
--- a/openvdb/util/NodeMasks.h
+++ b/openvdb/util/NodeMasks.h
@@ -43,23 +43,32 @@
 #include <openvdb/Types.h>
 //#include <boost/mpl/if.hpp>
 //#include <strings.h> // for ffs
-
+#ifdef _MSC_VER
+#include <intrin.h>
+#endif
 
 namespace openvdb {
 OPENVDB_USE_VERSION_NAMESPACE
 namespace OPENVDB_VERSION_NAME {
 namespace util {
 
+#ifdef _MSC_VER
+    // Compiling with vc++
+#include "NodeMasksMSVC.h"
+#elif defined(__GNUC__) || defined(__clang__)
+    // Compiling with clang or gcc
+#include "NodeMasksGccClang.h"
+#else
+    // Compiling with something else (Intel?), using old software-based routines.
+    // Software routines take many instructions to execute, some of them even access RAM for lookup tables.
+    // Hardware instructions are faster.
+
 /// Return the number of on bits in the given 8-bit value.
 inline Index32
 CountOn(Byte v)
 {
-    // Simple LUT:
-#ifndef _MSC_VER // Visual C++ doesn't guarantee thread-safe initialization of local statics
-    static
-#endif
     /// @todo Move this table and others into, say, Util.cc
-    const Byte numBits[256] = {
+    static const Byte numBits[256] = {
 #define COUNTONB2(n)  n,            n+1,            n+1,            n+2
 #define COUNTONB4(n)  COUNTONB2(n), COUNTONB2(n+1), COUNTONB2(n+1), COUNTONB2(n+2)
 #define COUNTONB6(n)  COUNTONB4(n), COUNTONB4(n+1), COUNTONB4(n+1), COUNTONB4(n+2)
@@ -79,9 +88,6 @@ CountOn(Byte v)
     //return (v * UINT64_C(0x200040008001) & UINT64_C(0x111111111111111)) % 0xF;
 }
 
-/// Return the number of off bits in the given 8-bit value.
-inline Index32 CountOff(Byte v) { return CountOn(static_cast<Byte>(~v)); }
-
 /// Return the number of on bits in the given 32-bit value.
 inline Index32
 CountOn(Index32 v)
@@ -90,9 +96,6 @@ CountOn(Index32 v)
     v = (v & 0x33333333U) + ((v >> 2) & 0x33333333U);
     return (((v + (v >> 4)) & 0xF0F0F0FU) * 0x1010101U) >> 24;
 }
-
-/// Return the number of off bits in the given 32-bit value.
-inline Index32 CountOff(Index32 v) { return CountOn(~v); }
 
 /// Return the number of on bits in the given 64-bit value.
 inline Index32
@@ -104,18 +107,12 @@ CountOn(Index64 v)
         (((v + (v >> 4)) & UINT64_C(0xF0F0F0F0F0F0F0F)) * UINT64_C(0x101010101010101)) >> 56);
 }
 
-/// Return the number of off bits in the given 64-bit value.
-inline Index32 CountOff(Index64 v) { return CountOn(~v); }
-
 /// Return the least significant on bit of the given 8-bit value.
 inline Index32
 FindLowestOn(Byte v)
 {
     assert(v);
-#ifndef _MSC_VER // Visual C++ doesn't guarantee thread-safe initialization of local statics
-    static
-#endif
-    const Byte DeBruijn[8] = {0, 1, 6, 2, 7, 5, 4, 3};
+    static const Byte DeBruijn[8] = {0, 1, 6, 2, 7, 5, 4, 3};
     return DeBruijn[Byte((v & -v) * 0x1DU) >> 5];
 }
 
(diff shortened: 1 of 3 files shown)


danrbailey:
Thanks for the PR. This is a change that will likely need discussion among the TSC so we should get this onto an upcoming meeting agenda. Your PR has come at a time of the year when we're otherwise engaged with the Siggraph conference and Summer vacations that has disrupted our regular TSC meeting schedule. Bear with us and we'll get back to you when we can.

e4lam:
We might consider first doing a separate commit to change all the static const tables (everywhere) into static constexpr, removing the #ifdef _MSC_VER since we have since moved on to require C++14, even on Windows.

danrbailey:
> We might consider first doing a separate commit to change all the static const tables (everywhere) into static constexpr, removing the #ifdef _MSC_VER since we have since moved on to require C++14, even on Windows.

Technically we're still intending to support C++11 until the end of the year as we're currently maintaining compatibility for the last two years worth of the VFX Reference Platform (ie as far back as 2017). That being said, we did briefly discuss at Siggraph about adjusting this policy to deprecate based on when the next major release of VDB is released. I understood that to mean that once we've pushed out ABI 7 + VFX 2020, we'd soon thereafter drop support for ABI 5 + VFX 2017. However, even if we change that, it'll likely still be Q4 before we can start using C++14 features in the core library unfortunately.

e4lam:
Sure, but using static constexpr is probably C++11 and doesn't require C++14.

jmlait:
Thank you again for this PR.  We had a chance to discuss it at the last TSC meeting.

POPCOUNT is a troublesome one as a surprising number of such old CPUs are still around.  2008 is also when CPUs stopped getting faster,  so stopped turning into bricks every two years. (In particular, if I recall properly, because of a strange interaction of sse 4.1, 4.0a, and 4.2, you end up having to set 4.2 as your minimal platform to gain popcount, which causes the compiler to start generating all the 4.2 instructions, thereby disabling the early AMD Phenoms.)  All that said, there were no objections to changing VDB to default to SSE 4.2+ only.

We would very much appreciate, however, if you could run a quick performance test on your favorite CPU/OS to verify that these intrinsics are faster than the equivalent C++.  While a glance at Agner's tables suggests BSR and POPCOUNT aren't some of them, the x86 architecture is lousy with intrinsics that are slower than the equivalent assembly, so we would like to see some wall-clock numbers to justify the additional maintenance of supporting the intrinsics.

As for signing something, there are two things that need to be signed.
1) Each commit has to be amended with a Signed-off-by: line to meet the requirements of the DCO.  The --signoff command can be used for this.  Unfortunately, it must be in all commits, so I sometimes find it easier to squash and resubmit as a single commit with the right --signoff than try to amend, but you may be more skilled at Git than I.
2) You will need to sign a Contributor License Agreement.  I think this is still in the air as we haven't turned on the system yet.

Const-me:
@jmlait In VC++, when you use intrinsics like `__popcnt()`, the compiler just emits the instruction you want. No need to change any project-wide configs. But the code will crash in runtime when not supported, with “invalid instruction” runtime error (happened to me once, years ago, for an SSSE3 instruction not being supported by some old AMD my customer was running).

Gcc and clang are different in that regard, the `__builtin_popcount` can compile into `popcnt` if allowed by compiler options, if not allowed the compiler will fall back to bit tricks, probably very similar ones to what you have in the master branch. E.g. `-march=native` will emit `popcnt` if building on a CPU newer than 12 years. There’s also a separate option in gcc to enable just the popcnt, `-mpopcnt`

OK, I’ll try to make a small performance benchmark later this week.

BTW, I have no idea who uses vdb, but I would be very surprised if more than 0.1% of users are using CPUs without hardware `popcnt` support. To find one, you need something like Core 2 Quad Q6600, or Athlon 64 X2. Even Intel Atoms support it for many years, since Silvermont microarch.

jmlait:
It likely is the GCC case I was thinking of then.  I do recall that to enable using the intrinsics at all you *had* to enable the compiler flag. And as soon as you did that, the compiler would start using the intrinsics itself in random parts of your code.  Which is frustrating as we are then unable to have runtime switches to enable those intrinsics on certain critical code paths.   And POPCOUNT in particular needs 4.2 because I don't think you can get the correct intersection of 4.0a and 4.1 otherwise.  Coincidentally, just heard from someone unable to move from 16.5 on their laptop because of the 4.2 requirements :>

danrbailey:
I happen to have been in a similar part of the code recently and found an opportunity to run a quick test as I was curious. On my particular hardware, I found that the lookup table was actually around 10-20% faster in the typical case where you're doing a lot of repeated lookups (ie once the table is in the cache). That appears to be supported by other benchmarks such as this one - http://dalkescientific.com/writings/diary/archive/2011/11/02/faster_popcount_update.html. 

It does look like there _may_ be a faster implementation around if you use SSE intrinsics carefully, but it's not trivial and I wasn't able to find such an implementation in the time I put towards this. There are also some copyright considerations with some of these other methods too. 

In summary - we still need some convincing here otherwise I think we'll stick with the current implementation and close out this PR. We do appreciate you submitting this though as this is a highly critical routine that we would be very happy to find a faster alternative for, so well worth investigating.

Const-me:
@danrbailey It’s your project, not mine. Do whatever you see fit. Meanwhile I’m still using my private branch.

> On my particular hardware, I found that the lookup table was actually around 10-20% faster in the typical case

I’m pretty sure you have used software fallbacks implemented by gcc/clang. Hardware `popcnt` can’t possibly be slower than lookup table. Intel says the instruction has 3 cycles latency, and for repeated computations it’s even faster because the throughput is just a single CPU cycle. A lookup table is slower than that by an order of magnitude, the typical latency of L1 data cache is 4 cycles / read, and you need multiple reads, e.g. 4 reads for 32-bit popcnt with 256 values in the table.

The article you’ve linked doesn’t include the hardware implementation I’m using in the PR. Here’s what it says about it:

> if you have a chip with the POPCNT instruction built-in then use it.

Also note the PR it’s not just about `popcnt`. The first/last set bit algorithm has similar performance characteristics of hardware versus software.

danrbailey:
Yes, you're right, I was accidentally using the software fallbacks when testing. Thanks for the info and the articles. It _is_ faster, around 40% on my hardware which is very exciting.

Here are some suggestions about how we can tweak your implementation to try and get this into OpenVDB.

Firstly, as @jmlait has already outlined, there may be people out there who are still using VDB with older hardware so at the very least I think this should be exposed through a CMake compile option so it can be toggled on/off, something like (using GCC as an example here):

```
if (OPENVDB_USE_SSE42)
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2")
  add_definitions(-DOPENVDB_USE_SSE42)
endif()
```

An additional reason for including this compiler toggle - for most studios, this code will be compiled on a build machine and then used by others on potentially quite different hardware. This rules out checking bit manipulation flags in CPUID feature bits or relying on software fallbacks.

For the sake of getting this PR in, I'd recommend that we default it to off at first, then we discuss whether to switch it on in our TSC meetings by default as a subsequent change. Perhaps we'd even take a very cautious approach by leaving it off for a while, enabling it in a CI build then changing the default at a later date.

I think we should put any logic that includes headers related to intrinsics into openvdb/Platform.h, that feels like the right place for this to go.

Because this is a fairly significant change, I suggest we focus this on just four methods at first:

* CountOn(Byte)
* CountOn(Index64)
* FindLowestOn(Byte)
* FindLowestOn(Index64)

These are the ones that will have the biggest impact. The Index32 variants and FindHighestOn() are only used for an old legacy file format which I don't believe we are even testing in our CI.

As a result, I think we should include ifdef branching for these four methods in NodeMasks.h rather than introducing new header files as that will keep all the code in one place. It would be nice, but I don't think it's necessary to include new unit tests as these methods are called fairly directly from NodeMask and are tested pretty extensively there.

@pcucka - please comment on my suggestions above if you think we should go about this differently.

Now on to the question of DCO and CLA. To be able to accept this PR, all of your commits have to be signed using a DCO which Jeff mentioned above:

> Each commit has to be amended with a Signed-off-by: line to meet the requirements of the DCO. The --signoff command can be used for this. Unfortunately, it must be in all commits, so I sometimes find it easier to squash and resubmit as a single commit with the right --signoff than try to amend, but you may be more skilled at Git than I.

You can either go back and edit every commit in the chain and sign them, collapse the commits and sign that and then force push to your branch or just re-submit the PR with signed commits. 

The next step is the CLA. We now have a system called EasyCLA turned on, any more commits to this PR and you'll get a new GitHub CLA check that will run with a link to allow you to sign a corporate or individual CLA.

Thanks again for the contribution and your patience with explaining the details of this change! :)

danrbailey:
Hi @Const-me, we have discussed this PR in TSC meeting and are happy to move forward with introducing some of the specific hardware instructions you're proposing. However, we would like to see this go into the next release of OpenVDB which is coming up in the next few weeks. Would you either have time to prepare this PR for merging and we'll guide you through the process or otherwise would you be happy to hand this off for us to complete (and we'll credit you for the idea in the change notes)?

Const-me:
@danrbailey Don’t have much time lately but I hope it won’t take too long.
As you see I’ve squashed the commits and added “Signed-off-by”.
I’ve signed the CLA, got the e-mail with a PDF which says “Your signed document is attached to this email” but this issue wasn’t updated. Can forward you the PDF if it helps. When I try again, it doesn’t offer me to sign, redirects back to this PR.

danrbailey:
> @danrbailey Don’t have much time lately but I hope it won’t take too long.
> As you see I’ve squashed the commits and added “Signed-off-by”.
> I’ve signed the CLA, got the e-mail with a PDF which says “Your signed document is attached to this email” but this issue wasn’t updated. Can forward you the PDF if it helps. When I try again, it doesn’t offer me to sign, redirects back to this PR.

The EasyCLA check doesn't appear to have been updated, perhaps an empty commit will trigger it? (This system is still pretty new).

Const-me:
@danrbailey Nothing changed. The CLA web site has remembered my github account, they know I have already signed and they just redirect me back. New commits don’t change anything because the PDF says “all Contributions to the Project made by You”, it doesn’t list commits nor defines a time interval.

danrbailey:
> @danrbailey Nothing changed. The CLA web site has remembered my github account, they know I have already signed and they just redirect me back. New commits don’t change anything because the PDF says “all Contributions to the Project made by You”, it doesn’t list commits nor defines a time interval.

I think you're confusing the system because you appear to have two email addresses you're using here:

* const@const.me (the DCO email)
* soonts@live.com (the Git author email)

Perhaps only one of these is linked to your GitHub account?

It looks like b2a61d7399cbeb4a657b811721893f6a200b6e48 did pass the EasyCLA check (so you're close!), but when you subsequently force-pushed, it then failed again.

I'd recommend using one email address for your DCO, Git author and GitHub registered emails and trying again? Thanks for your patience.

danrbailey:
Looks like everything is working now, great! I've just pushed up a PR that switches on a conditional flag to enable SIMD from our CMake:

https://github.com/AcademySoftwareFoundation/openvdb/pull/557

The intention is to use this to gate your changes by #ifdef OPENVDB_USE_SSE42 to add the popcount intrinsic.

jmlait:
I feel your pain.  I was stuck in that limbo for a couple of weeks before I
got all addresses verified.  I had to link the email used for the DCO with
the github account; that was a bit tricky as its hostname wasn't actually
an email address :>  Fortunately our IT could invent it for the duration.
Thank you for persevering through this!

danrbailey:
I've just merged a PR (#557) that adds OPENVDB_SIMD CMake support. 

To complete this PR, we'd need you to merge master back, fix the conflicts, add gated conditionals using `ifdef OPENVDB_USE_SSE42` and to remove all but the following intrinsic changes:

* CountOn(Byte)
* CountOn(Index64)
* FindLowestOn(Byte)
* FindLowestOn(Index64)

We'd like to get this into our upcoming release which is a few weeks out, but as mentioned before I'm happy to help finish this for you if you don't have the time to do so.

danrbailey:
Hi @Const-me,

I've just pushed up a PR from my own fork with the changes I was proposing on this PR (#574). Due to the big benefits in performance this represents, we're trying to get this into the 7.0 major release that we'll be pushing out in the next week or so, which means we've had to rather rush to get this ready. I appreciate you may not have had time to look at addressing my feedback yet so apologies for having to take this over to get it ready in time. If you do have a few moments to review my PR (#574) and verify I haven't made any mistakes in pulling your logic across, it would be much appreciated.

I'm going close down this PR now. Thanks again for the contribution, it's exciting to get these changes into OpenVDB! :)

PR #498, opened 2019-08-09
cmake: use GNUInstallDirs and other enhancements

davvid:
Use the standard GNUInstallDirs variables to avoid hard-coding platform-specific paths.
This provides standard cmake user knobs and general improvements to the cmake scripts.


danrbailey:
This all seems reasonable to me. I've assigned to @Idclip to look over before we consider merging, but I believe we briefly discussed using GNUInstallDirs and agreed in principal that this was a good idea.

(Note that we're currently transitioning to a new CLA system which will likely take a little more time before we can provide an automated way of signing CLAs that's required in order to accept these kind of PRs).

davvid:
In case it helps, these commits have DCO sign-offs, and can be taken under the older Disney CLA otherwise if needed.  Thanks Dan.

danrbailey:
> In case it helps, these commits have DCO sign-offs, and can be taken under the older Disney CLA otherwise if needed. Thanks Dan.

I don't think we'll be able to use the older CLA as we're now part of the ASWF and I assume you would have filled out the DWA CLA? Incidentally, the documentation for the new system is here:

https://github.com/swinslow/cla-tool-docs/blob/master/walkthroughs/4-CLA-Manager.md

openvdb/CMakeLists.txt
```diff
@@ -36,12 +36,16 @@ cmake_minimum_required(VERSION 3.3)
...
+include(GNUInstallDirs)
 
 ###### OpenVDB Core Options
 
 option(OPENVDB_CORE_SHARED "Build dynamically linked version of the core library." ON)
 option(OPENVDB_CORE_STATIC "Build statically linked version of the core library." ON)
 
+set(OPENVDB_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}/openvdb" CACHE STRING
+    "The prefix-relative path for installed openvdb headers."
+)
```
Idclip: Any reason you chose to make this a cached item? Do you require the ability to modify this? If we are to use the unified standard (GNUInstallDirs) then ideally these shouldn't be advertised as modifiable

openvdb_houdini/CMakeLists.txt
```diff
@@ -119,12 +120,15 @@ endif()
 set(OPENVDB_HOUDINI_INSTALL_PREFIX ${HOUDINI_INSTALL_LOCATION}
   CACHE PATH "Base install path for OpenVDB Houdini shared libraries, dsos and headers."
 )
+set(OPENVDB_HOUDINI_CONFIGDIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/config")
+set(OPENVDB_HOUDINI_DSODIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/dso")
+set(OPENVDB_HOUDINI_INCLUDEDIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}")
+set(OPENVDB_HOUDINI_LIBDIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+set(OPENVDB_HOUDINI_PYTHONRC_DIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/python2.7libs")
+set(OPENVDB_HOUDINI_SCRIPTSDIR "${OPENVDB_HOUDINI_INSTALL_PREFIX}/scripts")
```
Idclip: If we're to set these (which I think we 100% should) we should consider exposing the first folder item as cache variables. Lost of studios have different ways they are deploying Houdini/its plugins, I'm already symlinking stuff from the current vdb builder folders to get things working here

CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ endif()
 ###### OpenVDB Build/Component Options
 
 include(CMakeDependentOption)
+include(GNUInstallDirs)
```
Idclip: Does this need to be included in every lists? Can we include it once in the root (here)?

openvdb/CMakeLists.txt
davvid: Good question -- I can remove the cache var if you prefer.  The reason I did this was because of the slightly-hard-to-maintain feature where we allow the `openvdb_houdini` component to install into a different prefix.

That alone is enough of a departure from the "natural cmake semantics" that it trickles down into needing to expose a lot more to deal with that duality.  IMO that feature can be dropped because it significantly complicates the build scripts, but I did not want to disrupt existing use cases.  Let me know if it's worth keeping it in light of that use case.

openvdb_houdini/CMakeLists.txt
davvid: Ditto -- would the "first folder item" be the existing `OPENVDB_HOUDINI_INSTALL_PREFIX`?

Would it be okay to make all of these cache variables?  It seems like we should.  Pardon my cmake ignorance -- can any variable be overriden via `cmake -D FOO=BAR` or can only cache variables be overridden?  If only cache variables, then these should definitely be made into cache variables.  I'll go ahead and make that tweak shortly and push it up while I wait to hear back about the other questions.

CMakeLists.txt
davvid: The multiple inclusion can be considered a remnant from back when there were separate `openvdb_houdini` and `openvdb` repos, and now they are combined.

`CMakeLists.txt` specifies a different `project(...)` in each subdir, which to me means, "these are in a single repo, but really they're different projects and they should be able to be built independently".

That was my assumption -- that they should be able to build independently.  That assumption led me to add the include to each component subdir.   Let me know if that's not important and I'll go ahead and include it from the top-level only instead.

openvdb_houdini/CMakeLists.txt
Idclip: > can any variable be overriden via cmake -D FOO=BAR or can only cache variables be overridden?

It's possible to have two of the same variable at any given time. For example, with a CMake variable `MY_VAR`, you can have a "normal" (for lack of better CMake terminology) version of the variable and a `CACHE` version of the variable. The "normal" version will take precedence unless you request the `CACHE` version. I believe that whenever you specify a variable from the command line, it's inserted as a `CACHE` variable:
```shell
cmake -DMY_VAR="bar" ../
```

```cmake
...
# MY_VAR is fetched from the CACHE as no local variable has been set
message(STATUS ${MY_VAR}) # prints bar
# This creates a new local MY_VAR and sets it to "foo". The cache is untouched
set(MY_VAR "foo")
# MY_VAR exists as a local which is prioritised
message(STATUS ${MY_VAR}) # prints foo
# set() sees the CACHE keyword so tries to set the CACHE version of 
# MY_VAR to "cached". However, as MY_VAR already contains a value it is 
# left untouched
set(MY_VAR "cached" CACHE STRING "some help text" )
# The above set() command forces CMake to drop all local values of MY_VAR
# so "bar" is printed... :|
message(STATUS ${MY_VAR}) # prints bar
...
```

I can't find up to date documentation on this, but see the * Variable types in CMake * section on the below link which summarises it pretty well in one section. 

https://cmake.org/cmake/help/v3.0/command/set.html

openvdb/CMakeLists.txt
Idclip: > IMO that feature can be dropped because it significantly complicates the build scripts

Are you referring to the inclusion of the CACHE variable here or the installation behaviour of `openvdb_houdini`? Either way, I think making the install path of the core library the ground truth (`set(OPENVDB_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}/openvdb"`) and allowing more modifications on the install paths of  `openvdb_houdini` is perhaps the way to help resolve this?

CMakeLists.txt
Idclip: > CMakeLists.txt specifies a different project(...) in each subdir, which to me means, "these are in a single repo, but really they're different projects and they should be able to be built independently".

Right, that's a fair enough assumption - unfortunately most of the CMakeLists have been put together under the assumption you're running cmake from the root repo and using the available configuration controls to isolate individual projects. I can foresee a number of issues with the current structure were you to try and use them as you describe. That being said it's probably not a bad idea to try and support going forward. I'm happy to have them included as you've done so

davvid:
Having a single top-level place to manage includes and common setup is easier to maintain, and I don't want to encourage the opposite, so all of the lower-level includes of GNUInstallDirs have been removed; only the top-level one remains.

The `CACHE` specifier for `OPENVDB_INCLUDE_DIR` has been removed since we don't need that be configurable.

Thanks for the review.

davvid:
I was notified today that the corporate CLA has been signed, so hopefully this is good to merge.  cheers.

danrbailey:
Hi @davvid, we've just started using a new automated CLA system called EasyCLA so we're no longer keeping track of paper CLAs. If you push an empty commit to this PR, it'll trigger a new set of GitHub checks, which will include the EasyCLA check and a link to initiate the process there if you haven't already.  This system is being rolled out to all the ASWF projects.

davvid:
Alrighty, we got the easycla stuff passing now.  cheers

Idclip:
Thanks all. @danrbailey I'll take another look asap

openvdb/python/CMakeLists.txt
```diff
@@ -226,7 +227,7 @@ install(TARGETS
   ${PYOPENVDB_INSTALL_DIRECTORY}
 )
 
-install(FILES ${PYTHON_PUBLIC_INCLUDE_NAMES} DESTINATION include/openvdb/python)
+install(FILES ${PYTHON_PUBLIC_INCLUDE_NAMES} DESTINATION ${OPENVDB_INCLUDE_DIR}/python)
```
Idclip: There's a bug here. `OPENVDB_INCLUDE_DIR` is defined in openvdb/CMakeLists.txt so is unset in this conetxt and results in CMake attempting to perform a absolutely path install to /python. I suggest for now that we just move `set(OPENVDB_INCLUDE_DIR "${CMAKE_INSTALL_INCLUDEDIR}/openvdb")` to the root CMake.

Idclip: Actually, I think this variable needs to be renamed. The root CMake allows you to build components of vdb without building the core library. One of the variables in the FindOpenVDB module which lets you search for a VDB installation is `OPENVDB_INCLUDE_DIR` which will conflict with its intended use here, especially if we move this to the root. Can we also rename this variable to `OPENVDB_INSTALL_INCLUDEDIR`?

davvid: Ah, good catch.  I pushed a new commit that resolves that issue by moving `OPENVDB_INSTALL_INCLUDEDIR` to the top-level.  cheers.

davvid:
This series has been rebased on top of `master`.  The only new change since last time is the top-most commit.

danrbailey:
I've tested this on Windows 7 with Visual Studio 7 and all is working.

PR #499, opened 2019-08-09
openvdb/unittest: blosc 1.15 compatibility

davvid:
blosc >= 1.15 requires `const char**`, older versions require `char**`.

### DIFF
--- a/openvdb/unittest/TestFile.cc
+++ b/openvdb/unittest/TestFile.cc
@@ -2568,7 +2568,12 @@ TestFile::testBlosc()
 
     for (int compcode = 0; compcode <= BLOSC_ZLIB; ++compcode) {
         char* compname = nullptr;
-        if (0 > blosc_compcode_to_compname(compcode, &compname)) continue;
+#if BLOSC_VERSION_MAJOR > 1 || (BLOSC_VERSION_MAJOR == 1 && BLOSC_VERSION_MINOR >= 15)
+        if (0 > blosc_compcode_to_compname(compcode, const_cast<const char**>(&compname)))
+#else
+        if (0 > blosc_compcode_to_compname(compcode, &compname))
+#endif
+            continue;
         /// @todo This changes the compressor setting globally.
         if (blosc_set_compressor(compname) < 0) continue;
 
--- a/openvdb/unittest/TestStreamCompression.cc
+++ b/openvdb/unittest/TestStreamCompression.cc
@@ -75,6 +75,11 @@ namespace boost { namespace interprocess { namespace detail {} namespace ipcdeta
 #if BLOSC_VERSION_MAJOR > 1 || (BLOSC_VERSION_MAJOR == 1 && BLOSC_VERSION_MINOR > 10)
 #define BLOSC_HCR_BLOCKSIZE_OPTIMIZATION
 #endif
+// Blosc 1.14+ writes backwards-compatible data by default.
+// http://blosc.org/posts/new-forward-compat-policy/
+#if BLOSC_VERSION_MAJOR > 1 || (BLOSC_VERSION_MAJOR == 1 &&  BLOSC_VERSION_MINOR >= 14)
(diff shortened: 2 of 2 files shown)


davvid:
Nevermind, this is actually a blosc API change from `char**` to `const char**`.  I'll see if there's a way to ifdef it.

danrbailey:
Thanks for the PR. In fact, I've made this very change myself when testing later versions of Blosc. However, this is not quite as simple as it seems. As one example, certain versions of Blosc are not supported due to the way the library broke backward compatibility at one stage (http://blosc.org/posts/new-forward-compat-policy/).

A while back, I ran an automated test with every version of Blosc that has been released since 1.5.0 and got sporadic and unpredictable failures for specific versions that I couldn't explain. I didn't have enough time to dig into this further so suffice to say, our official line on this is that **we only support 1.5.0 currently** which is the version shipped natively with Houdini.

To be able to accept this PR, I'd like to first update our documentation to say with some confidence which versions of Blosc we do and do not support and to back this up with some continuous integration testing.

davvid:
Sounds good.  Later when we have CI testing against 1.15 (or some other version) we can document that it's acceptable and supported (indeed, the changes in this PR didn't touch the cmake warnings or docs).

In case it helps, the test suite does pass here on debian/testing with some small tweaks around the hard-coded file offsets in `TestStreamCompression.cc` :: `testPagedStreams()` so this latest version at least does seem to behave correctly.  The offset of 42424 becomes 51480, and offset 4422 becomes 5400, which is unexpected when compatibility across versions might be desired.  Is it an issue that the offsets are larger (implying larger file sizes)?

Yeah, this is a bit complicated -> http://blosc.org/posts/new-forward-compat-policy/

VDBs often end up on disk, so the forwards-compatibility guarantee in newer versions of blosc seem like a good thing going forward.  That blog post does seem to suggest that we should discourage blosc 1.11 through (but not including) 1.14 -- 1.14 is the first version that introduced the compat guarantees.

Testing against 1.15 in CI will be a little of a chore unless you build from debian/testing instead of ubuntu/bionic so it's understandable to park this until the future (debian/testing reports are often issues from the future).  Nonetheless, ubuntu/bionic does have 1.14, which is past the 1.11-1.13 bad range, so getting CI up and running there should point out the missing bits in the test suite that need changing once newer bloscs are verified.  For me those were lines 450, 479, and 568.  We'd have to `#ifdef` that too.

We can always come back to this in the future so I'll leave the branch around until we have consensus on a plan.  If we do want to take the code sooner rather than later (but not change the policy about 1.5 being the only officially supported version) then this will need ifdefs in the test suite for now, so I'll push those up shortly in case it helps anyone in the future.

Thanks for reviewing.

danrbailey:
Thanks, that's really helpful. I'll aim to repeat my Blosc tests soon and run that against your changes. As you've suggested, I was expecting to be able to just explicitly exclude 1.11-1.13 versions, but it didn't seem to be as simple as that in practice.

Regarding the versions we're using in CI, we're actually building Blosc from source using CentOS 7 rather than using yum, so we can pick up any versions we might want to test against. 

> Is it an issue that the offsets are larger (implying larger file sizes)?

I don't believe so because I think the offset sizes are also stored in the file for forwards-compatibility. The unit tests are really only written like that because we've been wary of changing the version of Blosc we support, so explicit seemed better.

davvid:
I pushed a new commit to trigger the new CLA process for this PR as well.

Idclip:
lgtm - if this has also fixed the unit tests for the latest blosc version you can probably close the following out too https://jira.aswf.io/browse/OVDB-78

danrbailey:
> lgtm - if this has also fixed the unit tests for the latest blosc version you can probably close the following out too https://jira.aswf.io/browse/OVDB-78

Thanks Nick. Yep, I believe we can close out that ticket when this is merged.

danrbailey:
Merged, thanks again for the contribution @davvid! :)

PR #502, opened 2019-08-14
OVDB-53: Provide OP_Operator::getVersion override to return VDB version

danrbailey:
By introducing this override we need to be able to distinguish between the Houdini default (such as "17.5.204") and the VDB version (such as "6.2.0") in a syncNodeVersion() override.

I considered adding a prefix (such as "vdb:6.2.0") then settled on using the packed integer representation (such as "100794368"). Simply checking the string is non-empty and contains no "." characters is thus sufficient to determine this is a VDB version and not a Houdini version. I also couldn't think of any reason to need to preserve which version of Houdini this node was created in, in addition to which version of VDB was used. Native Houdini VDB nodes will still use the Houdini version.

Of course if there's any logic expecting the Houdini default form (X.Y.Z.W) then this will break, but I'm not aware of anything here.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -89,6 +89,9 @@ Version 6.2.0 - In development
     - Added a houdini_utils::OpPolicy::getTabSubMenuPath() method to allow
       OpPolicy subclasses to provide their own tab sub-menu path.
     - OpenVDB SOPs are now displayed in an ASWF sub-menu of the VDB tab menu.
+    - Added an houdini_utils::OP_OperatorDW::getVersion() override which
+      returns the operator version that is set in the OpenVDBOpFactory
+      constructor.
 
 
 Version 6.1.0 - May 8, 2019
--- a/doc/changes.txt
+++ b/doc/changes.txt
@@ -113,6 +113,9 @@ Houdini:
 - Added an @hulink{OpPolicy::getTabSubMenuPath(),OpPolicy::getTabSubMenuPath}
   method to allow @b OpPolicy subclasses to provide their own tab sub-menu path.
 - OpenVDB SOPs are now displayed in an ASWF sub-menu of the VDB tab menu.
+- Added an @hulink{OP_OperatorDW::getVersion(),OP_OperatorDW::getVersion}
+  override which returns the operator version that is set in the
+  @hulink{OpenVDBOpFactory,OpenVDBOpFactory} constructor.
 
 
 @htmlonly <a name="v6_1_0_changes"></a>@endhtmlonly
--- a/openvdb_houdini/ParmFactory.cc
+++ b/openvdb_houdini/ParmFactory.cc
@@ -957,6 +957,19 @@ public:
         return !mDoc.empty();
     }
 
+#ifndef SESI_OPENVDB
+    bool getVersion(UT_String &version) override
+    {
+        auto it = spareData().find("operatorversion");
+        if (it != spareData().end()) {
+            version = it->second;
+            return true;
+        }
+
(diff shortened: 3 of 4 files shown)


danrbailey:
@pcucka - any thoughts on this implementation?

pcucka:
I'm waiting to hear @jmlait's thoughts, specifically on the intended use(s) of `getVersion` outside of `syncNodeVersion`.

jmlait:
I'd think the vdb prefix would be cleaner, but have no objections to the packed version either.  All uses of these nodes in the future should get the vdb library version, not the houdini version, so match the behaviour of HDAs.  The packed version will certainly simplify people adding tests to the code.

danrbailey:
> I'd think the vdb prefix would be cleaner, but have no objections to the packed version either. All uses of these nodes in the future should get the vdb library version, not the houdini version, so match the behaviour of HDAs. The packed version will certainly simplify people adding tests to the code.

I don't really mind either to be honest. If your preference is strong enough for vdb prefix, I'm happy to make the change? Otherwise if you want to approve this, I'll simply merge using the packed version.

jmlait:
Whoever parses it has to parse a string anyways, so I think it is reasonable to have a vdb: prefix to parse out.  If it were an integer, a packed version is easier, but since it is a string there might as well be a helper function to do the comparison when we need it.

UT_String::compareVersionString can be used after the vdb: is skipped.

So put me down for vdb:6.1.0

Idclip:
FWIW I agree with @jmlait. I've pulled these changes and am using them in AX with "vdb_ax:X.Y.Z" for a similar issue to the vdb points scatter stuff which started all of this. Works really well! +1

pcucka:
Would it make sense to append "`vdbX.Y.Z`" (or whatever) to the string returned by the base class (i.e., `OP_Operator::getVersion`), so that we retain the Houdini version?

danrbailey:
> Would it make sense to append "`vdbX.Y.Z`" (or whatever) to the string returned by the base class (i.e., `OP_Operator::getVersion`), so that we retain the Houdini version?

Can you perhaps give an example of what the Houdini version might be used for? I don't believe we've ever built a different version of a SOP for one version of Houdini as another.

pcucka:
Well, we've never needed to call `syncNodeVersion` before either.  But an example (from the past) might be changing the default To Polygons output mode based on the existence of poly soups.

danrbailey:
> Well, we've never needed to call `syncNodeVersion` before either. But an example (from the past) might be changing the default To Polygons output mode based on the existence of poly soups.

Right, I guess it doesn't cost much and it's a lot less effort to add it now than if we ever need it in the future.

As for the format, how about:

17.5.314:vdb6.2.0 ?

pcucka:
It would be nice if the format were such that a base class `syncNodeVersion` (for example) could parse the Houdini version without needing to be aware of the VDB component.  It doesn't look like `UT_String::compareVersionString` allows that, though, at least not when the separator is either a space or a colon.

danrbailey:
> It would be nice if the format were such that a base class `syncNodeVersion` (for example) could parse the Houdini version without needing to be aware of the VDB component. It doesn't look like `UT_String::compareVersionString` allows that, though, at least not when the separator is either a space or a colon.

@jmlait - I don't think the definition of compareVersionString is visible to us. Do you know if there's a way to achieve what @pcucka is suggesting here?

danrbailey:
Bump - I need agreement on the version format to use to proceed here.

@jmlait @pcucka

PR #503, opened 2019-08-22
CMake NumPy/Python3 Improvements

Idclip:
Main changes in this PR fix issues using python3, different versions of boost python (which can end up being called boost_python, boost_pythonX or boost_pythonXx) and adds support for boost numpy post boost 1.65.


doc/build.txt
```diff
@@ -758,4 +758,34 @@ Older versions in particular are not supported by OpenVDB.
...
+for all its dependencies. For imported Boost compatibility, the following
+versions of CMake are required:
+
+  - Boost 1.63 requires CMake 3.7 or newer.
+  - Boost 1.64 requires CMake 3.8 or newer.
+  - Boost 1.65 and 1.65.1 require CMake 3.9.3 or newer.
+  - Boost 1.66 requires CMake 3.11 or newer.
+  - Boost 1.67 requires CMake 3.12 or newer.
+  - Boost 1.68, 1.69 require CMake 3.13 or newer.
+  - Boost 1.70 requires CMake 3.14 or newer.
```
Idclip: The Boost/CMake relationship is a mess. It sounds like they are attempting to review this post 1.70 and 1.70 does come with a different CMake implementation (I think). I've chosen to add this comment here rather than implement this logic in our CMake as I'm not sure how strict we should be.

An interesting point is that if we do want to continue to use imported targets, our minimum CMake version will jump to 3.11 for VFX18. Worth discussion now.

Idclip: Yes, it was more whether or not 3.11 is the correct version to bump to i.e why not higher?

Idclip:
> I'm tentatively approving because this is already an improvement on what's there, but we should really be testing against Python3 + Boost 1.70 in an Azure build, so that we can ensure it builds going forwards.

I've tested this on CentOS and MacOS

doc/build.txt
danrbailey: I think the question is why not lower?

Right now it feels like CMake 3.11 is a good balance between picking up the more recent features of CMake and not being too bleeding edge so as to minimize the chances that users won't need to install a newer version that's already available to them. It also nicely ties in with Boost 1.66 which is the lowest supported version for VFX 2019. If there's a good reason to go higher than 3.11 to get access to a particular feature, that's worth talking about though.

danrbailey:
> > I'm tentatively approving because this is already an improvement on what's there, but we should really be testing against Python3 + Boost 1.70 in an Azure build, so that we can ensure it builds going forwards.
> 
> I've tested this on CentOS and MacOS

It should still be automated through CI.

Idclip:
> It should still be automated through CI.

Obviously - I'm happy to wait for the docker image

danrbailey:
> > It should still be automated through CI.
> 
> Obviously - I'm happy to wait for the docker image

I was suggesting you could put in a PR to add this Docker image yourself as it's a such a trivial change. :)

Here, I've done it for you:

https://github.com/AcademySoftwareFoundation/aswf-docker/pull/11

Idclip:
> I was suggesting you could put in a PR to add this Docker image yourself as it's a such a trivial change. :)

This is not all that's required though - the VFX 2020 job needs to be configured in VDB, is that correct? I was not planning to do this as part of this PR.

doc/build.txt
Idclip: > I think the question is why not lower?

Because of the boost version (1.66) requiring 3.11 for imported targets.

3.12 would allow us to remove some of the python branching logic AND (I think) brings native support for `XXX_ROOT` variables which would be incredibly useful. 3.14 would remove all conditional branching logic. So as 3.14 seems a bit too much, the choice is between 3.11 and 3.12.

aloysbaillet:
Hi, just a quick note to say that the `aswf/ci-openvdb:2020` is now available!

Idclip:
I've probably done something from, but the CI build I've tried to add immediately fails:

https://github.com/AcademySoftwareFoundation/openvdb/pull/503/checks?check_run_id=228261877

aloysbaillet:
No you haven't done anything wrong, some tags are missing at the moment, I've got a PR pending with a fix, for now you need to add a ".0" at the end of the image name: `aswf/ci-openvdb:2020.0` should be working.

Idclip:
Hey thanks @aloysbaillet - I don't mind waiting if  `aswf/ci-openvdb:2020` is the intended image name? Just let me know when it's available and I'll try again

danrbailey:
> Hey thanks @aloysbaillet - I don't mind waiting if `aswf/ci-openvdb:2020` is the intended image name? Just let me know when it's available and I'll try again

I think we should just use 2020:0 for now, I'd rather we get this PR built, validated and merged in. We still need to do a tidy-up pass over all these image names and switch from aswftesting to aswf anyway, so adjusting this tag can be included as part of this change.

danrbailey:
@Idclip - FYI, you can check which images are available by searching on dockerhub, here's a link to the available tags for aswf/ci-openvdb:

https://hub.docker.com/r/aswf/ci-openvdb/tags

Idclip:
I've run the install_boost.sh script for 1.70.0 and at least for me it's installing the shared libs correctly. Will have to investigate further as to why CMake is choosing the static libraries here...

aloysbaillet:
I've published a set of images in `aswftesting` where boost is compiled with "link=shared" explicitly set, which should only build shared libs... you can try by using this image: `aswftesting/ci-openvdb:2020` instead of `aswf/ci-openvdb:2020.0` and see if cmake is less confused?

Idclip:
Thanks @aloysbaillet I'll give this a shot

aloysbaillet:
Actually, the branch was merged and `aswf/ci-openvdb:2020` does not have static boost libraries anymore!

Idclip:
Thanks @aloysbaillet. This originally failed with:

```
CMake Error at /usr/local/lib/cmake/Boost-1.70.0/BoostConfig.cmake:95 (find_package):
  Found package configuration file:

    /usr/local/lib/cmake/boost_iostreams-1.70.0/boost_iostreams-config.cmake

  but it set boost_iostreams_FOUND to FALSE so package "boost_iostreams" is
  considered to be NOT FOUND.  Reason given by package:

  No suitable build variant has been found.
```

I think the issue is due to the joy which is the boost+cmake relationship, specifically a bug in the boost 1.70 cmake config:

http://boost.2283326.n4.nabble.com/CMake-config-scripts-broken-in-1-70-td4708957.html
https://github.com/boostorg/boost_install/commit/160c7cb2b2c720e74463865ef0454d4c4cd9ae7c

I believe I can work around this for 1.70 by explicitly setting both:
```
set(BUILD_SHARED_LIBS ON)
set(Boost_USE_STATIC_LIBS OFF)
```

cmake/backports/FindNumPy.cmake
```diff
@@ -0,0 +1,115 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# IN NO EVENT SHALL THE COPYRIGHT HOLDERS' AND CONTRIBUTORS' AGGREGATE
+# LIABILITY FOR ALL CLAIMS REGARDLESS OF THEIR BASIS EXCEED US$250.00.
+#
+#[=======================================================================[.rst:
```
danrbailey: I don't think this license is correct. If this module has been backported from the CMake project, then it should have the same license as cmake/backports/FindOpenGL.cmake ?

Idclip: This module doesn't exist in CMake, it's various bits and pieces of the numpy find logic which I've copied and modified. See https://github.com/Kitware/CMake/blob/master/Modules/FindPython/Support.cmake

PR #504, opened 2019-08-23
CMake default path improvements

Idclip:
Removing the manual system path list in favour of delegating to CMakes automatic search.

As discussed I'm still considering if this is the best way to do this. Waiting on results from appveyor and may make additional changes.


CMakeLists.txt
```diff
@@ -113,7 +113,10 @@ openvdb build configuration to decide on OpenEXR support. You may set this to on
 know it to be required.]=] OFF)
 CMAKE_DEPENDENT_option(OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING
   "Disable the implicit linking of Boost libraries on Windows" ON "WIN32" OFF)
-option(USE_SYSTEM_LIBRARY_PATHS "Build with system library paths" ON)
+option(USE_SYSTEM_LIBRARY_PATHS [=[
+Prioritise CMakes default system library/installation paths when attempting to locate dependencies. When disabled,
+only explicitly provided paths will be searched. Note the list variable SYSTEM_LIBRARY_PATHS can be populated with
+paths to use as fallbacks when calling any find_xxx() commands.]=] ON)
```
Idclip: This change uses CMake's search process as described in the documentation for `find_xxx` commands. Note that previous system paths were 'fallbacks' where as now they will be prioritised over explicitly provided variables. I'm not sure if this is ideal as this essentially means that you either have to configure every location yourself with `USE_SYSTEM_LIBRARY_PATHS=OFF` or ensure the configuration is completely correct with `USE_SYSTEM_LIBRARY_PATHS=ON`. For example, if you have openexr installed via vcpkg but want to use a custom build installed in a different location, you will have to set `USE_SYSTEM_LIBRARY_PATHS=OFF` and define every package location.

Idclip: Interestingly the `XXX_ROOT` variables become part of the default search path AND are prioritised from CMake 3.12. So perhaps this isn't a big deal!

https://cmake.org/cmake/help/v3.12/command/find_package.html

> Search paths specified in the <PackageName>_ROOT CMake variable and the <PackageName>_ROOT environment variable, where <PackageName> is the package to be found. The package root variables are maintained as a stack so if called from within a find module, root paths from the parent’s find module will also be searched after paths for the current package. This can be skipped if NO_PACKAGE_ROOT_PATH is passed. See policy CMP0074.

I'll re run some tests with different versions of CMake with this change.

Idclip:
I've updated this PR with additional improvements and changes. The core change remains the same (removing the `NO_DEFAULT_PATH`) option from `find_` calls, but I've reverted the variable logic from `USE_SYSTEM_LIBRARY_PATH` TO `DISABLE_CMAKE_SEARCH_PATHS`. I've also added support for root variables which match the find module letter case as this is what CMake 3.12 does. The new behaviour is the same as before with the addition of 4), which is essentially what was stopping the windows build from working with a toolchain file.

1) Search _INCLUDEDIR/_LIBRARYDIR variables
2) Search Package_ROOT variables (PACKAGE_ROOT still supported)
3) Search SYSTEM_LIBRARY_PATH entries
4) With DISABLE_CMAKE_SEARCH_PATHS=OFF, do cmake search

Backwards compatible with all older ROOT variables, the value of USE_SYSTEM_LIBRARY_PATH and any additions to SYSTEM_LIBRARY_PATH.

Idclip:
@danrbailey do you want to give this another once over?

PR #506, opened 2019-08-23
Configure OpenVDB for Houdini with _GLIBCXX_USE_CXX11_ABI=0

Idclip:
Automatically detect the current CXX11 ABI of GCC if in use and set _GLIBCXX_USE_CXX11_ABI. Report the appropriate message/warning accordingly.


cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -329,3 +329,50 @@ else()
...
+# ------------------------------------------------------------------------
+#  Configure GCC CXX11 ABI
+# ------------------------------------------------------------------------
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if((CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5.1) OR
+     (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1))
+    message(STATUS "GCC >= 5.1 detected. Configuring GCC CXX11 ABI for Houdini compatibility...")
+
+    execute_process(COMMAND echo "#include" " " "<string>"
```
davvid: Curious, does this work with simply `echo "#include <string>"` instead of separate arguments?  I tested with cmake 3.13.4 and it worked there.

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -329,3 +329,50 @@ else()
...
+        "Disabling for all OpenVDB components for Houdini compatibility. Make sure all "
+        "dependencies have been build with older GCC ABI. "
+        "See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html and "
+        "https://vfxplatform.com/#footnote-gcc6 for more information.")
+    else()
+      message(WARNING "Unable to determine the current ABI configuration of GCC CXX11."
+        " Disabling for all OpenVDB components for Houdini compatibility. Make sure all "
+        "dependencies have been build with older GCC ABI. "
+        "See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html and "
+        "https://vfxplatform.com/#footnote-gcc6 for more information.")
```
davvid: Minor grammar fix, perhaps, `Make sure all dependencies have been built with the older GCC ABI` in both of these warning messages.

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -329,3 +329,50 @@ else()
...
+#  Configure GCC CXX11 ABI
+# ------------------------------------------------------------------------
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if((CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5.1) OR
+     (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1))
+    message(STATUS "GCC >= 5.1 detected. Configuring GCC CXX11 ABI for Houdini compatibility...")
+
+    execute_process(COMMAND echo "#include" " " "<string>"
+      COMMAND g++ "-x" "c++" "-E" "-dM" "-"
```
davvid: It might be worth using `${CMAKE_CXX_COMPILER}` to avoid hard-coding `g++` here.

cmake/OpenVDBHoudiniSetup.cmake
Idclip: I think you're right, I'll change this - thanks

cmake/OpenVDBHoudiniSetup.cmake
Idclip: agreed, done

cmake/OpenVDBHoudiniSetup.cmake
Idclip: done!

Idclip:
> However, if there is a simple way to check which ABI Houdini is built against and to use that, it might prove to be more robust for the future?

`D_GLIBCXX_USE_CXX11_ABI` is defined as a definition form the Houdini imported target - I perhaps could grep for that value. If the value we're providing and the value they're providing differ the compile will complain a lot, so there's that. I'll take a look and see how easy this would be.

Idclip:
Updated to also check the definition being set in the Houdini targets `INTERFACE_COMPILE_OPTIONS`, defaulting to the newer ABI if it cant be found. All this logic is perhaps a bit unnecessary - if we don't care about promoting any warnings about mismatching values it can most likely be simplified to something like:

```
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  if((CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5.1) OR
     (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1))

    if(NOT DEFINED HOUDINI_CXX11_ABI)
      get_target_property(houdini_interface_compile_options
        Houdini INTERFACE_COMPILE_OPTIONS)
      set(HOUDINI_CXX11_ABI "UNKNOWN")
      if("-D_GLIBCXX_USE_CXX11_ABI=0" IN_LIST houdini_interface_compile_options)
        set(HOUDINI_CXX11_ABI 0)
      elseif("-D_GLIBCXX_USE_CXX11_ABI=1" IN_LIST houdini_interface_compile_options)
        set(HOUDINI_CXX11_ABI 1)
      else()
        set(HOUDINI_CXX11_ABI 1)
    endif()

    add_definitions(-D_GLIBCXX_USE_CXX11_ABI=${HOUDINI_CXX11_ABI})
  endif()
endif()

```

With a few messages

Idclip:
Leaving as is, merging on successfully build

PR #509, opened 2019-08-30
Do not try to make metadata out of non-tuples as we don't support them

jmlait:
Do not try to make metadata out of non-tuples as we don't support them


jmlait:
Thank you for the reminder!   Added to CHANGES and doc/changes.txt

jmlait:
... My brain just caught up to me though.... These changes actually aren't in the core library as we don't replace GU_PrimVDB by default.   So you need to get the new Houdini version, not these changes, to stop the crash.   I just added these to get the reference implementation in sync....

PR #510, opened 2019-08-30
Marking base virtual functions deprecated is painful in Windows, so removing it.

jmlait:
I do not like this change overmuch, but it was required for our builds to avoid deprecation failures.  When the compress() method is instantiated in the derived templated class, MSVC will generate deprecation failures since the pure virtual base method is deprecated.  This is clearly wrong as the overridden version has also been marked deprecated.

We have not found a good solution internally to deal with this with our deprecations.  Usually we are forced to just disable deprecations on the entire derived class.

I've also added a missing deprecation on the derived set() method.

### DIFF
--- a/openvdb/points/AttributeArray.h
+++ b/openvdb/points/AttributeArray.h
@@ -277,7 +277,9 @@ public:
     /// @deprecated From ABI 6 on, use copyValues() with source-target index pairs.
 #if OPENVDB_ABI_VERSION_NUMBER >= 6
     // Windows does not allow base classes to be easily deprecated.
-    // OPENVDB_DEPRECATED
+#ifndef _MSC_VER
+    OPENVDB_DEPRECATED
+#endif
 #endif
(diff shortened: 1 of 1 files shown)


Idclip:
Is it worth wrapping this in `_MSC_VER` ifdefs so the warnings are maintained on other platforms? Perhaps in this instance it can be replaced with a pragma so at least there's some output. Though I suspect these methods will be removed soon anyway.

jmlait:
Hmm... We may want to have an OPENVDB_BASECLASS_DEPRECATED that disables on relevant platforms if this becomes a more common issue.  It is rather annoying as it can't show up on our CI until the CI has a copy of Houdini on the future ABI; which means it likely will always be something I get to hit :>  I can do the _MSC_VER tests for now.  As you note, I think these functions are first out in 7.

jmlait:
It only happens if you compile against Houdini with ABI 6+.  No public versions of Houdini have this, so the CI can't detect.  (And when I ran into it, of course it didn't show up for me on Linux, waiting till I had pushed internally to break all of our Windows CI :>)

PR #511, opened 2019-09-01
TSC Meeting notes August 29th, 2019

jmlait:
Meeting Notes

### DIFF
--- a/tsc/meetings/2019-08-29.md
+++ b/tsc/meetings/2019-08-29.md
@@ -1,4 +1,4 @@
-Minutes from 26th OpenVDB TSC meeting, July 25, 2019
+Minutes from 26th OpenVDB TSC meeting, August 29th, 2019, (GMT)
 
 Attendees: *Nick* A., *Dan* B., *Jeff* L., *Ken* M., *Peter* C.
 
--- a/tsc/meetings/2019-08-29.md
+++ b/tsc/meetings/2019-08-29.md
@@ -1,0 +1,130 @@
+Minutes from 26th OpenVDB TSC meeting, July 25, 2019
+
+Attendees: *Nick* A., *Dan* B., *Jeff* L., *Ken* M., *Peter* C.
+
+Additional Attendees: Sean McDuffee
+
+Agenda:
+
+1) Quorum
+2) Secretary
+3) Round Robin
+4) Release Schedule for 6.2.0
+5) levelSetVolume
+6) VDB Smooth doesn't densify
+7) Bitscan/popcount
+8) Deprecation
+9) Combine SOP Performance
+10) Tree Refactoring
+11) Delayed Loading
+12) Next Meeting
+
+1) Quorum was confirmed.
+
+2) Secretary was Jeff Lait
+
+3) Round Robin
+
+A quick round robin to identify what is being worked on and what may
+make it for 6.2.
+
+Ken: Working on velocity extension.  Read-Only grid may end up being internal
+only.  Has found interesting results by making a write-only grid that allows
+better multithreaded scaling than usual multiple grids + merge.
+
+Nick: AX integration.  Feature branch is hoped to exist in a week or so.
+As well, CMake as required.  Should anistropic rasterization be higher
+priority?  In response to a later question, MPI will not be available with
+current priorities.
+
+Dan: Point partitioning improvements.  C++11 changes.  30-40% better performance
+on some datasets.  Merging of grids is open questions.  Likewise, refactoring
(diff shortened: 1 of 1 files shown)


tsc/meetings/2019-08-29.md
```diff
@@ -0,0 +1,130 @@
+Minutes from 26th OpenVDB TSC meeting, July 25, 2019
```
kmuseth: wait what, July 25?!? Isn't this minutes for last week's meeting on August 29/30? Since I am regularly moving between totally different time-zone I'd like for us to start indicating time-zones when we talked about dates and times (e.g. GMT). Technically our last meeting was on August 30 for me :)

tsc/meetings/2019-08-29.md
```diff
@@ -0,0 +1,130 @@
...
+
+7) Bitscan/popcount
+
+The TSC was unsure of the performance difference, so recommended asking
+the submitter to verify a measurable difference.  POPCOUNT is the more
+debatable as SSE 4.2 is more recent (Only with Houdini 17.0 did that become
+a requirement).
+
+We noted that this method does not belong in Tree, but belongs in Utils.
+No objections to moving to Utils for 7.
```
danrbailey: I think this comment was about moving util/NodeMasks.h => tree/NodeMasks.h. Perhaps by retaining a dummy header for util/NodeMasks.h with a deprecated pragma.

kmuseth: that's exactly right. Incidentally that's also the header file in which the intrinsic popcount was recently proposed (remember the response you recently wrote Jeff :)

tsc/meetings/2019-08-29.md
```diff
@@ -0,0 +1,130 @@
...
+
+No one had any critical items to be done for 6.2.0.
+
+5) levelSetVolume
+
+VDB Clip in 6.X leaves different values in inactive voxels, causing volume
+computation to change.  Consensus is that such half-open SDF don't have
+a proper volume so this is acceptable.
+
+Peter notes that there is an error computing volume of entirely empty
```
pcucka: Not exactly: the issue is that empty bounding boxes are represented by _max_ < _min_ (because `CoordBBox` bounds are closed intervals rather than half-open intervals), but the bounding box volume computation doesn't check for that case.

tsc/meetings/2019-08-29.md
```diff
@@ -0,0 +1,130 @@
...
+
+Is this a good chance to rethink the whole tree?  There seems to be
+some cruft that could be removed as is for obsolete file formats.
+General agreement that we should always be ready to re-think our structures.
+
+11) Delayed Loading
+
+This is still in a state of limbo between the File SOP and VDB.  One
+sticking point has been Windows support.  Another is the difficulty
+to pass options to the IO translator.
```
pcucka: Is enabling/disabling delayed loading per file a requirement, actually?

tsc/meetings/2019-08-29.md
jmlait: My secret of always starting from the last meeting notes is exposed!  I was so busy trying to get the meeting number right (because the 25th meeting hadn't been merged yet) I didn't notice I didn't change the date.  The filename at least I got right.

Not sure what to do about timezone there as there is no time, so adding (GMT) to the end is a bit odd.

tsc/meetings/2019-08-29.md
jmlait: Okay, definitely my fault here.  My notes said that it was utils -> tree.  But I didn't believe that as it didn't make sense since it is a bunch of bit twiddling tools, so should be in utils.  So I assumed my notes were wrong!  Bad secretary!   I have no objections either way.

tsc/meetings/2019-08-29.md
jmlait: Hopefully I have this clarified.

tsc/meetings/2019-08-29.md
jmlait: Maximum size should definitely be a global setting.

I guess enabling can be a global setting too, maybe we just add an env variable to turn on delayed loading so studios can enable as they desire?

pcucka: I created [OVDB-112](https://jira.aswf.io/browse/OVDB-112) so that we can continue this discussion.

tsc/meetings/2019-08-29.md
pcucka: PR #512 includes the fix.

PR #512, opened 2019-09-03
Fixes for empty CoordBBox bug and ICC warning

pcucka:


openvdb/math/Coord.h
```diff
@@ -398,7 +404,7 @@ class CoordBBox
     /// @brief Return the dimensions of the coordinates spanned by this bounding box.
     /// @note Since coordinates are inclusive, a bounding box with min = max
     /// has dimensions of (1, 1, 1).
-    Coord dim() const { return mMax.offsetBy(1) - mMin; }
+    Coord dim() const { return empty() ? Coord() : (mMax.offsetBy(1) - mMin); }
```
danrbailey: Probably worth including a test for the empty case.

pcucka: Done.  (The empty case was actually being tested implicitly via `CoordBBox::volume`, but it makes sense to add explicit `dim` tests.)

kmuseth: I would prefer:
return empty() ? Coord(0) : (mMax.offsetBy(1) - mMin);
to make it obvious what is returned for empty bbox'es

jmlait: I agree with Ken.  Maybe it is because of using classes that don't have proper initialization, but reading that I feel compelled to double check the constructor to make sure it zero-initializes :>

PR #515, opened 2019-09-16
CMake deprecation warnings for VFX 18 versions

Idclip:


CMakeLists.txt
```diff
@@ -212,10 +212,25 @@ if(NOT DISABLE_DEPENDENCY_VERSION_CHECKS)
...
 
+# VFX 18 deprecations to transition to MINIMUM_* variables in OpenVDB 7.0.0
+
+set(DEPRECATED_CMAKE_VERSION 3.12)
+set(DEPRECATED_GCC_VERSION 6.3.1)
+set(DEPRECATED_ICC_VERSION 17)
+set(DEPRECATED_MSVC_VERSION 2017)
+set(DEPRECATED_TBB_VERSION 2017.6)
+set(DEPRECATED_NUMPY_VERSION 1.12.1)
+set(DEPRECATED_HOUDINI_VERSION 17)
```
Idclip: Should Maya also be here? Note that MSVC is "2017 15.0", I'm not sure exactly how the version is set in CMake. Based on @danrbailey's reply r.e. default versions with certain OS distributions, I've left GLFW and log4-cplus alone.

Idclip: After looking at the appveyor build it looks like I'll need to change the MSVC version

pcucka: Could these variables perhaps be named "`FUTURE_MINIMUM_*_VERSION`"?

doc/changes.txt
```diff
@@ -42,6 +42,7 @@ Improvements:
 - Added a @c USE_COLORED_OUTPUT option to CMake to display compiler output
   in color.
 - Added an @c OPENVDB_CODE_COVERAGE option to CMake.
+- Marked VFX 17 platform version dependencies as deprecated.
```
pcucka: Maybe change this to "CMake now issues deprecation warnings for 2017 VFX Reference Platform version dependencies.  In particular, C++11-only compilers are now deprecated; OpenVDB 7.0 will require a C++14-compatible compiler."

CMakeLists.txt
Idclip: done

doc/changes.txt
Idclip: Thanks, done

PR #517, opened 2019-09-18
Starting OpenVDB 7.0.0 development

pcucka:


Idclip:
Hmm, when we released 6.1.0 we bumped to 6.1.1 during dev: 130fe81444b9bfa393bdf8cecd507a4c6569a068
Should we be bumping to 6.2.1? This would help if, say, we made changes which depended on the library version prior to a 7.0.0 release.

pcucka:
Ordinarily, yes, we'd bump the micro version number, but with such a short release window for 7.0.0, I don't see us releasing another 6.2 version, unless it's for a critical bug fix.

But I'm OK with putting this PR on hold, if everyone's in agreement.

danrbailey:
I agree with Peter, I think it's right to jump to 7.0.0 now. There's no plan to do another 6.x release and we need adequate time to stabilise any new changes, particularly related to ABI. Any critical issues that do come up in the next couple of months can always be added to a branch off 6.2.0 if necessary.

PR #518, opened 2019-09-19
bug fix during group conversion with non point cloud inputs

Idclip:
- Fixed a bug during conversion from Houdini geometry to OpenVDB Points with the use of point groups, where group information could end up corrupted or cause a crash with non-contiguous point offsets (mesh inputs). 
 - Threaded the population of point group memberships during conversion from Houdini geometry to OpenVDB Points

I'll update the change logs with the above soon


danrbailey:
Can we merge this now?

Idclip:
@danrbailey I was waiting pending yesterdays discussion on the 6.2.1 patch. As we're not re-releasing 6.2.0 I can merge this now

PR #520, opened 2019-09-25
Fix StreamMetadata segfault

danrbailey:


danrbailey:
I've updated this PR with some change notes and added an extra comment preceding the member variables in the stream-related classes. If this can be approved by the end of this week, I'll cherry-pick this change back into a 6.2 branch and initiate the release on Monday.

PR #526, opened 2019-10-04
Update Docker Containers for CI

danrbailey:
* Migrate from aswftesting -> aswf docker containers
* Use vfx2020 as the main image 
* Drop Houdini 16.5 build


danrbailey:
@Idclip @pcucka - would one of you mind approving please?

PR #530, opened 2019-10-10
Add a new GitHub Actions build workflow

danrbailey:


.github/workflows/build.yml
```diff
@@ -0,0 +1,122 @@
...
+
+on: [push]
+
+jobs:
+  testabi7:
+    runs-on: ubuntu-16.04
+    container:
+      image: aswf/ci-openvdb:2020
+    steps:
+    - uses: actions/checkout@v1
```
Idclip: what is `actions/checkout@v1`?

danrbailey: It checks out the repo - https://github.com/actions/checkout

Essentially, these "actions" are open-source scripts written in javascript that can be re-used across any project, you can see a full list of the official ones here - https://github.com/actions. There's an API and community developed ones available too.

PR #532, opened 2019-10-10
Drop Support for VFX Reference Platform 2017 and Houdini 16.5

danrbailey:
This does a number of things:

* Increases minimum compiler to a C++14 compiler
* Drops support for Houdini 16.5 and removes some conditional guards
* Applies minimum versions from VFX Reference Platform 2018


CMakeLists.txt
```diff
@@ -235,7 +230,7 @@ if(${CMAKE_VERSION} VERSION_LESS ${FUTURE_MINIMUM_CMAKE_VERSION})
     "is deprecated and will be removed.")
 endif()
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
Idclip: This also needs to be set in FindOpenVDB.cmake with the `INTERFACE_COMPILE_FEATURES` property (right at the bottom of the file during target creation)

CMakeLists.txt
```diff
@@ -219,12 +220,6 @@ endif()
 # VFX 18 deprecations to transition to MINIMUM_* variables in OpenVDB 7.0.0
 
 set(FUTURE_MINIMUM_CMAKE_VERSION 3.12)
-set(FUTURE_MINIMUM_GCC_VERSION 6.3.1)
-set(FUTURE_MINIMUM_ICC_VERSION 17)
-set(FUTURE_MINIMUM_MSVC_VERSION 19.10)
-set(FUTURE_MINIMUM_TBB_VERSION 2017.6)
-set(FUTURE_MINIMUM_NUMPY_VERSION 1.12.1)
-set(FUTURE_MINIMUM_HOUDINI_VERSION 17)
```
Idclip: These will cause problems if they don't exist when they're being checked - and it seems like this PR only comments out the Houdini ones? Maybe it's better to leave these in but set them to their above counterparts

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -163,28 +163,10 @@ find_package_handle_standard_args(Houdini
...
-endif()
-
-if(Houdini_VERSION VERSION_LESS ${FUTURE_MINIMUM_HOUDINI_VERSION})
-  message(DEPRECATION "Support for Houdini versions < ${FUTURE_MINIMUM_HOUDINI_VERSION} "
-    "is deprecated and will be removed.")
-endif()
+# if(Houdini_VERSION VERSION_LESS ${FUTURE_MINIMUM_HOUDINI_VERSION})
+#   message(DEPRECATION "Support for Houdini versions < ${FUTURE_MINIMUM_HOUDINI_VERSION} "
+#     "is deprecated and will be removed.")
+# endif()
```
Idclip: If we decide to leave this in, this should be wrapped in a `if(FUTURE_MINIMUM_HOUDINI_VERSION)` otherwise it causes problems when other packages attempt to include this file

CMakeLists.txt
```diff
@@ -219,12 +220,6 @@ endif()
 # VFX 18 deprecations to transition to MINIMUM_* variables in OpenVDB 7.0.0
 
 set(FUTURE_MINIMUM_CMAKE_VERSION 3.12)
```
Idclip: Why not also bump this one too?

CMakeLists.txt
danrbailey: Good catch, thanks

CMakeLists.txt
danrbailey: Perhaps it's best to have a global flag that disables them which we only switch on one or two minor releases before we finally remove support? I understand VERSION<MINIMUM to mean "no longer supported" and VERSION<FUTURE_MINIMUM to mean "deprecated but still supported".

If that's the case, this would mean that Houdini 17.0 and Houdini 17.5 are "deprecated" which is a hard argument to make considering Houdini 18.0 has yet to be released.

CMakeLists.txt
danrbailey: I tried, but it immediately broke our CI because the default version of CMake being pulled in there is 3.9.

We can bump this, but we'll need to get the Docker containers upgraded so I figured this can be a separate effort. I'd like to consider whether we can get away with a lower CMake version as well, if are asking our users to install CMake from source, that's going to be a hurdle I'd rather not insist on unless it's absolutely necessary.

Question to @aloysbaillet and @michdolan - I notice that OpenColorIO uses CMake 3.11 as a minimum - how is this working through the ASWF Docker containers?

CMakeLists.txt
danrbailey: Ok, so I've added an OPENVDB_FUTURE_DEPRECATION flag which currently defaults to OFF. If you switch it on, you can see which third-party libraries/applications you're using that are due to be deprecated in the future. I figure that we can switch the default value of this flag as we get closer to the next major release. I've also guarded by these variables existing as they won't always be in use.

cmake/OpenVDBHoudiniSetup.cmake
danrbailey: Agreed, done.

CMakeLists.txt
Idclip: Nice work :+1:

CMakeLists.txt
Idclip: Okay, fine by me to leave this out of this PR for now

michdolan: OpenColorIO is using the 2019 tag for all builds currently, which is CMake 3.12.

danrbailey: @michdolan - ah that's good to know, thanks!

PR #535, opened 2019-10-11
OVDB-54: Modify Transform and/or Metadata on Copy of Const Grid

danrbailey:


openvdb/Grid.h
```diff
@@ -501,6 +510,11 @@ class OPENVDB_API GridBase: public MetaMap
     /// @brief Initialize with an identity linear transform.
     GridBase(): mTransform(math::Transform::createLinearTransform()) {}
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 7
+    /// @brief Initialize with metadata and a transform.
+    GridBase(math::Transform::Ptr xform, const MetaMap& meta): MetaMap(meta), mTransform(xform) { }
```
pcucka: This needs to throw (see `GridBase::setTransform`) if the transform pointer is null.

openvdb/Grid.h
```diff
@@ -128,6 +128,15 @@ class OPENVDB_API GridBase: public MetaMap
...
     virtual GridBase::ConstPtr copyGrid() const = 0;
+#if OPENVDB_ABI_VERSION_NUMBER >= 7
+    /// @brief Return a new grid of the same type as this grid whose tree is shared with
+    /// this grid, whose metadata is a deep copy of this grid's and whose transform is
+    /// provided as an argument.
+    virtual GridBase::ConstPtr copyGrid(math::Transform::Ptr xform) const = 0;
+    /// @brief Return a new grid of the same type as this grid whose tree is shared with
+    /// this grid and whose transform and metadata are provided as arguments.
+    virtual GridBase::ConstPtr copyGrid(math::Transform::Ptr xform, const MetaMap& meta) const = 0;
+#endif
```
pcucka: How about an overload to specify the metadata but not the transform?

One option might be to change `copyGrid(math::Transform::Ptr, const MetaMap&)` to `copyGrid(const MetaMap&, math::Transform::Ptr = math::Transform::createLinearTransform())`.

danrbailey: I think I'd expect `gridBase->copyGrid(newMetaMap)` to re-use the transform from gridBase, not a new transform? 

What about `copyGrid(const MetaMap&, math::Transform::Ptr = this->transformPtr())` ?

danrbailey: I like the idea of making it more explicit and I agree it is confusing if the metadata comes from another grid. Those are **_very_** long method names though. :/

What about `copyGridAndReplace(...)` and then we provide overloads for all three options:

`copyGridAndReplace(const MetaMap&)`
`copyGridAndReplace(math::Transform::Ptr)`
`copyGridAndReplace(const MetaMap&, math::Transform::Ptr)`

I'm trying to find a balance between making it more explicit and keeping the method names down. The only small ambiguity is whether `copyGridAndReplace(otherGrid)` also replaces the transform, but it's at least consistent in that the transform is only ever replaced if provided as an explicit parameter.

openvdb/Grid.h
```diff
@@ -501,6 +510,12 @@ class OPENVDB_API GridBase: public MetaMap
     /// @brief Initialize with an identity linear transform.
     GridBase(): mTransform(math::Transform::createLinearTransform()) {}
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 7
+    /// @brief Initialize with metadata and a transform.
+    /// @throw ValueError if the transform pointer is null
```
pcucka: Add this `@throw` comment to the copy methods as well.

danrbailey: Think I got them all. I didn't do the private Grid constructor as that appears to be the convention.

pcucka: I think you missed `GridBase::copyGrid`.

danrbailey: Done, thanks.

openvdb/Grid.h
danrbailey: @kmuseth @jmlait @Idclip - looking for some input on what to do here?

Idclip: I don't think the confusion with providing a grid as a metamap for metadata replacement is a particularly big problem. Agreed that it's less clear if you're looking at the usage and are unaware of how `copyGrid` is working, but adding +3 methods to alleviate it seems a bit much. Personally I'd just use:

```
    virtual GridBase::ConstPtr copyGrid(math::Transform::Ptr newxform = nullptr, MetaMap::Ptr newmeta = nullptr) const = 0;
```

Where the transform and/or meta are used in place of a deep copy if they're not null.
 - Only requires a single new method and completely replaces the old one
 - Handles nullptr without requiring a runtime error

If however you are set on functions with a more explicit name to avoid this confusion without reading documentation, then the name needs to absolutely reflect the behaviour. So `copyGridAndReplace` isn't going to work and you need something more like `copyGridReplacingMetadataAndTransformWith`.

Alternatively, perhaps `copyGrid` isn't the right method to be replacing here and instead new static `create` methods are needed e.g.

`static Ptr create(TreePtrType, math::Transform::Ptr newxform = nullptr, MetaMap::Ptr newmeta = nullptr);`

This also avoid an ABI change.

danrbailey: Thanks for weighing in @Idclip.

I prefer making the Metadata argument a const reference as it makes it very clear that the map is deep copied, whereas including it as a shared ptr doesn't do that.

I also prefer a non-static method - the copyGrid() methods exist on GridBase not Grid so you wouldn't need to know the grid type to copy it and swap out the transform.

I like the idea of using explicit method names, I'm just not keen on a 40-character method name if we can find a way to avoid that.

danrbailey: I've changed these method names as discussed in our last meeting. This should be ready to be merged now.

danrbailey:
> You'll want to update the release notes, though.

Just updated the release notes.

PR #536, opened 2019-10-14
Added Tree::nodeCount

kmuseth:

This method returns a list (std::vector) with node counts organized by node type. It will improve the performance of Tree::print (which currently uses a slow Tree iterator to count nodes as oppose to bit-mask counting in Tree::nodeCount) and can optimize node allocations in classes like openvdb::tree::NodeManager.

Counting tree nodes in a grid with a level set sphere of size 4096^3:

Old technique ... completed in 118.182 milliseconds
New technique ... completed in 5.841 milliseconds

The "old techniques" (used in Tree::print) is based on a Tree::Iterator whereas the "new technique" employs bit-mask counting (both are serial). For this example the speedup is over 20x!

This example can easily be reproduced in the unit test TestTree::testNodeCount


doc/changes.txt
```diff
@@ -6,12 +6,16 @@
 @par
 <B>Version 7.0.0</B> - <I>In development</I>
 
-@par
```
danrbailey: Think this was deleted by accident.

openvdb/tree/InternalNode.h
```diff
@@ -302,6 +302,15 @@ class InternalNode
     void setOrigin(const Coord& origin) { mOrigin = origin; }
 
     Index32 leafCount() const;
+    void nodeCount(std::vector<Index32> &vec) const
```
danrbailey: Probably best to follow convention here and put the implementation outside of the function declaration.

openvdb/tree/RootNode.h
```diff
@@ -51,7 +51,8 @@
 #include <map>
 #include <set>
 #include <sstream>
-#include <deque>
+#include <vector>
+#include <memory> // for std::make_unique
```
danrbailey: Think this header is not needed anymore? (You're no longer using a std::unique_ptr?)

openvdb/tree/RootNode.h
```diff
@@ -516,6 +517,19 @@ class RootNode
     Index64 onLeafVoxelCount() const;
     Index64 offLeafVoxelCount() const;
     Index64 onTileCount() const;
+    void nodeCount(std::vector<Index32> &vec) const
```
danrbailey: Likewise here, moving the implementation out of the declaration.

openvdb/unittest/TestGrid.cc
```diff
@@ -38,6 +38,7 @@
 #include <openvdb/util/CpuTimer.h>
 #include <cppunit/extensions/HelperMacros.h>
 #include <iostream>
+#include <memory> // for std::make_unique
```
danrbailey: Likewise - I don't think this header is needed any more.

PR #537, opened 2019-10-15
Add the correct pragmas to disable deprecation warnings on MSVC.

jmlait:
Attempt to get proper deprecation pragmas on Windows.

### DIFF
--- a/openvdb/Platform.h
+++ b/openvdb/Platform.h
@@ -189,7 +189,9 @@
 #elif defined _MSC_VER
     #define OPENVDB_NO_DEPRECATION_WARNING_BEGIN \
         __pragma(warning(push)) \
-        __pragma(warning(disable : 4996))
+        __pragma(warning(disable : 4996)) \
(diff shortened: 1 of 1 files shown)


openvdb/Platform.h
```diff
@@ -186,6 +186,12 @@
         _Pragma("message(\"NOTE: ignoring deprecation warning\")")
     #define OPENVDB_NO_DEPRECATION_WARNING_END \
         _Pragma("GCC diagnostic pop")
+#elif defined _MSC_VER
+    #define OPENVDB_NO_DEPRECATION_WARNING_BEGIN \
+        __pragma(warning(push)) \
+        __pragma(warning(disable : 4996))
```
pcucka: Does MSVC have a message pragma?  I think we decided that we want to know, if possible, when a deprecation warning is being suppressed.

jmlait: Yes it does.  I'll look into verifying the syntax.

PR #540, opened 2019-10-23
Modify CMake to treat AppleClang like Clang

danrbailey:
Compiler id for Apple Clang is now AppleClang.

https://cmake.org/cmake/help/v3.0/policy/CMP0025.html


danrbailey:
@Idclip - would you mind reviewing this? It's a trivial change.

PR #547, opened 2019-11-06
Explicitly write out the name attribute into the grids

jmlait:
Explicitly write out the name attribute into the grids before
processing, as they will internally use the name metadata to generate
the output names.  Otherwise our names will go out of sync and
not respect up-stream Name SOPs.

Bug: 96367

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_LOD.cc
+++ b/openvdb_houdini/SOP_OpenVDB_LOD.cc
@@ -294,6 +294,7 @@ SOP_OpenVDB_LOD::Cache::cookVDBSop(OP_Context& context)
         std::vector<std::string> skipped;
 
         hvdb::Interrupter boss("Creating VDB LoD pyramid");
+        GA_RWHandleS name_h(gdp, GA_ATTRIB_PRIMITIVE, "name");
 
         const auto lodMode = evalInt("lod", 0, 0);
         if (lodMode == 0) {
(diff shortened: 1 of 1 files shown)


pcucka:
Should we be trying to address grid metadata/attribute mismatches (due to the use of nodes like Attribute Create and Name) in some more comprehensive way, rather than on a case-by-case basis?  And in this case, shouldn't we be syncing all of the attributes, not just the name?

I added syncing functionality to the OpenVDB Metadata SOP in 5.2.0, but I admit that requiring the insertion of an extra node is not an ideal solution—and anyway the Metadata SOP doesn't ship with Houdini.  Is there perhaps a signalling mechanism by which primitives can be notified when their attributes change?

jmlait:
Yes, I agree this should be some more systematic solution rather than patching it when we encounter it.  There should be no signalling mechanism as attributes should be detached from primitives.  This allows operations on attributes to operate generically and in parallel.  There is currently an exception with P and VDBs as their transform contains the position offset, but I would not want to continue in that path.

We could do something at the SOP_NodeVDB level as a pre-roll on the GDP.  It probably is not a performance consideration to just detect and re-sync if necessary.

Until then, however, we should probably patch the issues we do run into.  I can't do anything more involved for 18.0 in any case.

dokipen3d:
I can confirm that we have had to const_cast the vdb grid that we get from Houdini to be able to call setName() to synchronize the primitive name and grid name. Would be nice if Houdini did this. There was a reason given in the sidefx ticket which I can't remember off the top of my head.  Rather than set the name before processing, could it be set after processing so that the next SOP ( which might be a custom plugin that ingests the vdb grid) can reliably use the gridname.

jmlait:
I think it needs to be as a pre-op in the wrapper.  Most of the createVDB methods already properly synchronize back the name attribute, so usually when leaving a VDB SOP the name and grid name match.  The problem is that even if they match on leaving, the next SOP may be an Attribute Wrangle that knows nothing about VDBs and just writes to name directly.

The const_cast is an issue if you aren't operating in-place, so don't have a writeable copy of the vdb.  I don't think a SOP should ever modify an upstream node's grid with const_cast as that might violate thread safety.

pcucka:
We probably should do both (that is, transfer attributes to metadata pre-op and metadata to attributes post-op), since any call to a core library routine could result in updated metadata.  And keep in mind that we want to keep all metadata in sync (grid class, vector type, etc.), not just the grid name.

PR #548, opened 2019-11-07
Rename the Translate parameter and update the help to reflect truth.

jmlait:
Rename the Translate parameter and update the help to reflect that it is stuck in index space.

Bug: 87783

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_Resample.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Resample.cc
@@ -174,11 +174,11 @@ Using Voxel Scale Only:\n\
             " for the new transform"));
 
     // Translation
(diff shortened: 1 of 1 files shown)


pcucka:
If we're going to make any change to this now, my preference would be to add a "Use World Space Units" toggle, like we have in several other SOPs, and have it apply to both the Translate and Pivot parameters.

jmlait:
I agree that would be the perfect.  But I don't think this change in any way hinders us from doing that at some point.  In the meantime, it can hopefully reduce confusion from people just glancing over the parameter interface.  (I also can't do anything larger at this point as 18 is locked)

jmlait:
Good point.  I've added OVDB-125 to track this request.

Issue #496, opened 2019-08-03
Linker Tools Error vdb_render


jonassorgenfrei:
Hey,

I've got compilation complications when trying to compile vdb_render for Windows 10.

Error LNK2019 Reference to undefined symbol ""private: static union half::uif const * const half::_toFloat" (?_toFloat@half@@0QBTuif@1@B)" in Funktion ""public: static void __cdecl openvdb::v6_2::io::HalfReader<1,float>::read(class std::basic_istream<char,struct std::char_traits<char> > &,float *,unsigned int,unsigned int,class openvdb::v6_2::io::DelayedLoadMetadata *,unsigned __int64)" (?read@?$HalfReader@$00M@io@v6_2@openvdb@@SAXAEAV?$basic_istream@DU?$char_traits@D@std@@@std@@PEAMIIPEAVDelayedLoadMetadata@234@_K@Z)".	vdb_render	D:\Bibliotheken\openvdb\build\openvdb\cmd\openvdb_render.obj	1	

I'm not quite sure if it's a general problem or just a wrong setup in my system.

Best
Jonas

jonassorgenfrei:
The error seems to come from line 343 in io/Compression.h.
When I comment the std::copy out it compiles and seems to work, but since I just started diving into openvdb im not quit sure what may can go wrong internally, when this copy to the data parameter ist missing.

danrbailey:
This has been raised before - https://github.com/AcademySoftwareFoundation/openvdb/issues/247

That thread should offer some explanation as to why it's happening. I was under the impression that this issue has since been resolved by the CMake refactor, but please do let us know if that's not the case (I'm not a Windows user).

jonassorgenfrei:
Well it's not the same error as mentioned in the thread. 
I double checked that the HALF_EXPORTS wasn't set in the the project.

It seems to be a problem in the Half-2_3.lib. 
Linking against the static lib (Half-2_3_static.lib) worked fine, only the dynamic lib produced problems,

Idclip:
@danrbailey our appveyor CI is caching the build dir and seems to be content with keeping openexr at 2.2. I think 2.3 now exists in vcpkg - perhaps we could try that?

danrbailey:
I have reproduced the issue and found a fix - https://github.com/AcademySoftwareFoundation/openvdb/pull/550

Thanks for reporting.

danrbailey:
This fix has now been merged in, closing out this issue.

PR #550, opened 2019-11-12
Fix OpenEXR VDB Render issue on Windows

danrbailey:
Paraphrasing from an OpenEXR change log:

> If you are using any OpenEXR DLLs on Windows, you must define OPENEXR_DLL in your project's preprocessor directives.

My understanding is that we're handling this with the following logic in openvdb/PlatformConfig.h:

```
    // By default, assume that we're dynamically linking OpenEXR, unless
    // OPENVDB_OPENEXR_STATICLIB is defined.
    #if !defined(OPENVDB_OPENEXR_STATICLIB) && !defined(OPENEXR_DLL)
        #define OPENEXR_DLL
    #endif
```

However in the vdb_render binary, the OpenEXR headers are included before this PlatformConfig header so this OPENEXR_DLL is not defined early enough which causes an undefined symbol issue as described in detail here (https://github.com/AcademySoftwareFoundation/openvdb/issues/496).

I believe a pending change in OpenEXR 2.4 will solve this properly using a new cmake module.

(@e4lam)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,7 @@ Version 7.0.0 - In development
     Bug fixes:
     - Fixed a bug in FindJemalloc.cmake where paths were not being handled
       correctly.
+    - Fixed a Windows build issue in openvdb_render.
 
     ABI changes:
     - OpFactory destructor is now virtual as of ABI=7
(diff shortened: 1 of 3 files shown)


e4lam:
I see, sorry I think I misread the code thinking it was OPENVDB_DLL.

meshula:
Chiming in from the OpenEXR project, this looks like a good fix to me.

PR #552, opened 2019-11-12
Rasterize Points SOP non-finite value fix

pcucka:
When transferring attributes, the Rasterize Points SOP computes density-weighted sums over neighboring points, and if there are clusters of points with extremely small (subnormal) densities, normalization by the sum of those densities can produce NaNs or infs.  This PR adds logic to check for and suppress non-finite values.

This also serves as a DWA test of EasyCLA.

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_Rasterize_Points.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Rasterize_Points.cc
@@ -2185,8 +2185,10 @@ struct RasterizePoints
 
                 if (transferAttributes) {
                     for (size_t nn = 0; nn < BoolLeafNodeType::SIZE; ++nn) {
-                        voxelWeightArray[nn] =
-                            voxelWeightArray[nn] > 0.0f ? 1.0f / voxelWeightArray[nn] : 0.0f;
+                        const float weight = (voxelWeightArray[nn] > 0.0f) ?
+                            (1.0f / voxelWeightArray[nn]) : 0.0f;
+                        // Subnormal input values are nonzero but can result in infinite weights.
+                        voxelWeightArray[nn] = openvdb::math::isFinite(weight) ? weight : 0.0f;
                     }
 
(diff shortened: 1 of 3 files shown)


openvdb_houdini/SOP_OpenVDB_Rasterize_Points.cc
```diff
@@ -2185,8 +2185,10 @@ struct RasterizePoints
 
                 if (transferAttributes) {
                     for (size_t nn = 0; nn < BoolLeafNodeType::SIZE; ++nn) {
-                        voxelWeightArray[nn] =
-                            voxelWeightArray[nn] > 0.0f ? 1.0f / voxelWeightArray[nn] : 0.0f;
+                        const float weight = (voxelWeightArray[nn] > 0.0f) ?
+                            (1.0f / voxelWeightArray[nn]) : 0.0f;
+                        // Subnormal input values are nonzero but can result in infinite weights.
+                        voxelWeightArray[nn] = openvdb::math::isFinite(weight) ? weight : 0.0f;
```
jmlait: Can we simplify this to
const float weight = 1.0f  / voxelWeightArray[nn];
voxelWeightArray[nn] = openvdb::math::isFinite(weight) ? weight : 0.0f;

because division by zero also reliably generates INF, so the guard against exact zero isn't needed...  This does assume that negative weights aren't possible, but if they are computed as coverage that should be the case.

pcucka: The weights are derived from particle densities, and those are user-supplied values, so there's no guarantee that they're nonnegative (although one would hope that they always are).  Removing the test would change a behavior that users might have come to rely on, but if everyone else is in agreement, I'm OK with making that change.

jmlait: If they are user supplied values then I agree that we should keep the > 0 guard, so retract my suggestion.

PR #553, opened 2019-11-13
Various Visual Studio 2017 Improvements

danrbailey:
(@e4lam)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,8 @@ Version 7.0.0 - In development
       dependency searches.
     - Improvements to OpenVDB's CMake module setup order in regards to CMake
       policy, minimum version and project calls.
+    - Made various small changes to improve Visual Studio compatiblity and
+      deprecate some logic for compilers older than Visual Studio 2017.
 
     Bug fixes:
     - Fixed a bug in FindJemalloc.cmake where paths were not being handled
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -437,7 +437,7 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "App
   if(OPENVDB_CXX_STRICT)
     message(STATUS "Configuring Clang CXX warnings")
     set(HAS_AVAILABLE_WARNINGS TRUE)
-    add_definitions(
+    add_compile_options(
       -Werror
       -Wall
       -Wextra
@@ -464,7 +464,7 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   if(OPENVDB_CXX_STRICT)
     message(STATUS "Configuring GCC CXX warnings")
     set(HAS_AVAILABLE_WARNINGS TRUE)
-    add_definitions(
+    add_compile_options(
       -Werror
       -Wall
       -Wextra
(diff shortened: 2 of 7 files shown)


e4lam:
In trying to use these changes to compile with cmake on Windows, I found that I couldn't build the python bindings unless I made this change:

```diff --git a/openvdb/python/CMakeLists.txt b/openvdb/python/CMakeLists.txt
index 58acb33c..0b69ee34 100644
--- a/openvdb/python/CMakeLists.txt
+++ b/openvdb/python/CMakeLists.txt
@@ -149,6 +149,11 @@ set(OPENVDB_PYTHON_DEPENDENT_LIBS

 ##########################################################################

+# @todo Should be target definitions
+if(WIN32)
+  add_definitions(-DNOMINMAX)
+endif()
+
 set(OPENVDB_PYTHON_MODULE_SOURCE_FILES
   pyFloatGrid.cc
   pyIntGrid.cc
```

e4lam:
See also https://github.com/sideeffects/openvdb_dev/commit/f20bd835e9a9ccfeb7b285d102f0bf90ddabf9a7

e4lam:
Another note addendum here is that we have various _MSC_VER exceptions where we do something like this:

e4lam:
> 
> 
> In trying to use these changes to compile with cmake on Windows, I found that I couldn't build the python bindings unless I made this change:
> 
> ```diff
> index 58acb33c..0b69ee34 100644
> --- a/openvdb/python/CMakeLists.txt
> +++ b/openvdb/python/CMakeLists.txt
> @@ -149,6 +149,11 @@ set(OPENVDB_PYTHON_DEPENDENT_LIBS
> 
>  ##########################################################################
> 
> +# @todo Should be target definitions
> +if(WIN32)
> +  add_definitions(-DNOMINMAX)
> +endif()
> +
>  set(OPENVDB_PYTHON_MODULE_SOURCE_FILES
>    pyFloatGrid.cc
>    pyIntGrid.cc
> ```

Ah nevermind, I think this was just me trying to build with clang-cl. I didn't realize it was going into the non-Windows compiler code path.

e4lam:
There are some weird things here like:

e4lam:
There are some weird things in `openvdb/CMakeLists.txt` like:  add_definitions("/bigobj")

danrbailey:
> There are some weird things in `openvdb/CMakeLists.txt` like:
> 
> ```
> add_definitions("/bigobj")
> ...
> add_definitions("/wd4244")
> ```
> 
> where I would expect one to use `add_compile_options()` and it's just not in the MSVC cmake code paths either.

That's true, those should definitely be in the MSVC code path. I'll fix that and switch them to add_compile_options too. Thanks!

danrbailey:
> Another note addendum here is that we have various _MSC_VER exceptions where we do something like this:
> 
> ```
> inline void
> callTypedGrid(const GEO_PrimVDB& prim, OpType& op)
> {
> #ifdef _MSC_VER
>     op.operator()<GridType>(*(UTverify_cast<const GridType*>(&prim.getConstGrid())));
> #else
>     op.template operator()<GridType>(*(UTverify_cast<const GridType*>(&prim.getConstGrid())));
> #endif
> }
> ```
> 
> I think VS2017 does not require these anymore either.

Ah, that's good to know, I couldn't find anything explicitly stating whether that was the case or not so I didn't make this change.

I can change this in a separate PR, it would be nice to get rid of all of these.

danrbailey:
I've updated this PR, @Idclip can you approve?

PR #555, opened 2019-11-13
Replace boost::scoped_array with std::unique_ptr

danrbailey:


danrbailey:
> Some of these changes are API-breaking, like the changes to the return values of the `VolumeToMesh` methods. It would be good to call those out explicitly in the release notes.

Done.

PR #561, opened 2019-11-21
OVDB-124: Adding RootNode::addChild and InternalNode::addChild

danrbailey:
I had a go at implementing this, some quick comments:

* I changed the proposal in OVDB-124 slightly to remove the offset/coord argument and use the one in the child node to avoid any issues with them not being in sync.
* InternalNode returns a bool as I think it can/should gracefully fail if you're attempting to insert a child node that doesn't belong in that InternalNode (not an issue for the RootNode)
* InternalNode and RootNode are inconsistent in that the former only asserts that the child pointer is not a nullptr and then dereferences, the latter checks and early-exits. I kept these consistent with other methods in these classes (but merely pointing out that they're not consistent with each other).
* @kmuseth - I'm adding to TestTree here because I didn't want to reorganize this unit test as part of this PR, but still agree that this is worth doing.


danrbailey:
Any other reviews before I merge this? (@Idclip @kmuseth)

kmuseth:
Dan, I'm confused - is this not my PR? If so why are you as a reviewer make changes - that's my job as the submitter.

danrbailey:
> Dan, I'm confused - is this not my PR? If so why are you as a reviewer making changes - that's my job as the submitter.

Nope, this one is my PR - introducing RootNode::addChild and InternalNode::addChild methods. The linked JIRA ticket has a very similar number to your own though (yours is OVDB-123, mine is OVDB-124), perhaps that's the confusion here?

kmuseth:
> ?

okay, that explains my confusion - all go then :)

kmuseth:
> okay, that explains my confusion - all good then :)

Dan, can you approve this PR - it's currently blocked by you:
https://github.com/AcademySoftwareFoundation/openvdb/pull/536

openvdb/tree/RootNode.h
```diff
@@ -2587,6 +2591,20 @@ RootNode<ChildT>::addLeafAndCache(LeafNodeType* leaf, AccessorT& acc)
     child->addLeafAndCache(leaf, acc);
 }
 
+template<typename ChildT>
+inline void
+RootNode<ChildT>::addChild(ChildT* child)
+{
+    if (!child) return;
```
kmuseth: The behavior of this method is inconsistent with InternalNode::addChild which I find to be confusing. If the incoming pointer is nullptr the method is a noop, whereas InternalNode::addChild assumes this to never happen!

danrbailey: Hi Ken,

Thanks for the feedback, I've addressed everything except this comment. I find this situation confusing too, however this inconsistency already exists in other methods across these two classes such as **addLeaf**:

RootNode methods are no-op:
```
RootNode<ChildT>::addLeaf(LeafNodeType* leaf)
{
    if (leaf == nullptr) return;
```

InternalNode methods assume not a nullptr:
```
InternalNode<ChildT, Log2Dim>::addLeaf(LeafNodeType* leaf)
{
    assert(leaf != nullptr);
```

So I had to make a choice here. I don't mind changing one of them if you have a preference, but just highlighting the existing confusion. :/

kmuseth: Hi Dan,

Good point - I wonder if now (with a major update) we should change the signature of RootNode::addLeaf? The new Boolean return value should not break existing client code so I don't see the risk - do you?

danrbailey: Just to clarify, are you saying that you're ok with the current behavior when passing in a nullptr to all of these addLeaf/addChild methods?

Regarding the return types, there's a somewhat unrelated hole in that LeafNode::addLeaf() does not take ownership of the child, whereas InternalNode and RootNode do. I can address this, but perhaps it's better to do so in a new PR?

kmuseth: yes I am okay with the current behavior

danrbailey: ok, thanks @kmuseth, I'm going to merge this then.

PR #563, opened 2019-11-21
Add isGridUnique.

jmlait:
Duplicate the makeGridUnique as isGridUnique that allows testing if grids are shared before necessarily uniquing them.

This is a non-compiling reference change to track the internal GEO_PrimVDB version of Houdini 18.

### DIFF
--- a/openvdb_houdini/GEO_PrimVDB.h
+++ b/openvdb_houdini/GEO_PrimVDB.h
@@ -429,6 +429,11 @@ public:
     void                        makeGridUnique()
                                     { myGridAccessor.makeGridUnique(); }
 
+    /// @brief Returns if the tree is not shared.  If it is not shared,
+    /// one can make destructive edits without makeGridUnique.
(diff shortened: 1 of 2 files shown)


openvdb_houdini/GEO_PrimVDB.h
```diff
@@ -429,6 +429,11 @@ class OPENVDB_HOUDINI_API GEO_PrimVDB : public GEO_Primitive
     void                        makeGridUnique()
                                     { myGridAccessor.makeGridUnique(); }
 
+    /// @brief Returns if the tree is not shared.  If it is not shared,
```
danrbailey: Returns _**true**_ if the tree...

Thanks for adding this. I think I might even add this check to GridBase as it seems pretty useful.

jmlait: Note that at the GridBase level it should be called isTreeUnique.

PR #564, opened 2019-11-22
[OVDB-117] Mesh To Volume bugfix

Idclip:
See the ticket for more information https://jira.aswf.io/browse/OVDB-117

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,8 @@ Version 7.0.0 - In development
     - Fixed a bug in FindJemalloc.cmake where paths were not being handled
       correctly.
     - Fixed a Windows build issue in openvdb_render.
+    - Fixed a non deterministic threading bug in tools::meshToVolume() during
+      polygon voxelization which could result in different distance values.
 
     ABI changes:
     - OpFactory destructor is now virtual as of ABI=7
--- a/doc/changes.txt
+++ b/doc/changes.txt
@@ -33,6 +33,9 @@ Bug fixes:
 - Fixed a bug in <TT>FindJemalloc.cmake</TT> where paths were not being handled
   correctly.
 - Fixed a Windows build issue in openvdb_render.
+- Fixed a non deterministic threading bug in @vdblink::tools::meshToVolume()
+  mesh to volume@endlink during polygon voxelization which could result in
+  different distance values.
 
 @par
 ABI changes:
--- a/openvdb/tools/MeshToVolume.h
+++ b/openvdb/tools/MeshToVolume.h
@@ -1960,6 +1960,7 @@ struct VoxelizationData {
         if (mPrimCount == MaxPrimId || primIdTree.leafCount() > 1000) {
             mPrimCount = 0;
             primIdTree.clear();
+            mPrimCount = 0;
         }
 
         return mPrimCount++;
--- a/openvdb/tools/MeshToVolume.h
+++ b/openvdb/tools/MeshToVolume.h
@@ -1957,10 +1957,25 @@ struct VoxelizationData {
 
     unsigned char getNewPrimId() {
 
+        /// @warning Don't use parallel methods here!
+        /// The primIdTree is used as a "scratch" pad to mark visits for a given polygon
+        /// into voxels which it may contribute to. The tree is kept as lightweight as
+        /// possible and is reset when a maximum count or size is reached. A previous
+        /// bug here occurred due to the calling of tree methods with multi-threaded
+        /// implementations, resulting in nested parallelization and re-use of the TLS
+        /// from the initial task. This consequently resulted in non deterministic values
+        /// of mPrimCount on the return of the initial task, and could potentially end up
+        /// with a mPrimCount equal to that of the MaxPrimId. This is used as the background
+        /// value of the scratch tree.
+        /// @see jira.aswf.io/browse/OVDB-117, PR #564
+        /// @todo Consider profiling this operator with tree.clear() and Investigate the
+        /// chosen value of MaxPrimId
+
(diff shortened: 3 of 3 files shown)


tristanbarback:
I hope it's not presumptive of me to comment, but it seems to me that this will result in the scratch tree effectively being cleared more often than we would want. As you describe with this change the clear() call is interleaved with subsequent calls to the VoxelizePolygons task due to task stealing. 

So we get as a possible (likely?) scenario:
iteration 100 on thread A starts, sets mPrimID to 0, starts clear(), leaf nodes are stolen and iteration 100 blocks which they are deallocated.
iteration 101 on thread A starts, uses mPrimID 0. and uses the scratch tree, creating new leaf nodes as needed.
stolen leaf node deallocation completes on other threads.
iteration 101 on thread A completes.
iteration 100 resumes, steals internal nodes and then blocks while internal nodes are deallocated.
iteration 102 on thread A starts, uses mPrimID 1, and uses the scratch tree, creating new internal and leaf nodes as needed.
stolen internal node deallocation completes on other threads.
iteration 102 on thread A completes.
iteration 100 resumes, serially clears root node and any created nodes, uses mPrimId0, and uses the scratch tree, creating new internal and leaf nodes as needed.
iteration 100 completes.
iteration 103 starts, uses mPrimID 1. Execution then proceeds as expected.

I'd be inclined to just prefer a serial clear(), which I think would look like this (but please don't take my word for it):

primIdTree.root().clear();
primIdTree.clearAllAccessors();

It's much easier to understand why it's safe, and will I think lead to more efficient usage of the scratch tree. The downside of course is there might be some inefficiency in the task packing when a clear() is called near the end of the parallel voxelisation operation, but given the number of primitives is expected to be much much greater than the number of threads the effect should be minimal.

Idclip:
Hey @tristanbarback, thanks a lot for your input! This was a solution that I considered as well. Really, the only reason I opted for this quick naive fix is due to the upcoming release deadline - but I agree with your analysis, I think clear() might have been multithreaded *after* this implementation (need to double check, might be mistaken) so was intended to be a serial implementation here. There are a few more points that I was planning to discuss with the VDB TSC (tomorrow) but I'll put them here now!

1) root.clear(), clearAllAccessors(), leafCount()

Whilst your proposed fix is definitely a better solution, what now happens if any of the above calls are multithreaded in the future? Whilst this isn't something we can predict, it's not something we can easily track either. The subsequent 0 assignment will guarantee no issues arise from any changes to these methods, but is still not ideal as it's not obvious at all. It seems to me that two TLS objects are needed; 1) A single struct which holds the distance and poly index trees and 2) A stack of primitive id trees which threads can push and pop as needed, handling the case where methods are changed in the future.

2) Changing the order of clear()

One question whose answer is still a bit... unclear is the exact performance impact of this double parallel call. If there is a benefit to keeping this multithreaded in this context, then perhaps we can simply re-work clear to ensure the tree is empty prior to performing the threaded deallocation:

```
    std::vector<LeafNodeType*> leafnodes;
    this->stealNodes(leafnodes);

    std::vector<typename RootNodeType::ChildNodeType*> internalNodes;
    this->stealNodes(internalNodes);

    mRoot.clear();
    this->clearAllAccessors();

    tbb::parallel_for(tbb::blocked_range<size_t>(0, leafnodes.size()),
        DeallocateNodes<LeafNodeType>(leafnodes));
    tbb::parallel_for(tbb::blocked_range<size_t>(0, internalNodes.size()),
        DeallocateNodes<typename RootNodeType::ChildNodeType>(internalNodes));
```
In this way, subsequent invocations won't have contended results with their primIdTree - but the 0 assignment is still necessary.

3) Clear being called in destructors

This is slightly tangential but something I want to bring up. Tree destructors call clear() - which means that this potentially has an impact in *any* algorithm which implements TLS structs for temporary trees or types of parallel reductions. In fact one of my first attempts at debugging this issue was to replace the call to clear with a pointer reset (which obviously doesn't help at all). This might stem to a larger conversation about controlling how tree data can be destroyed.


If others are happy with the serial implementation approach I'll update this PR

jmlait:
I think if we want to have a serialClear, it is important to make it part of the mandated function signature to signal it must remain single threaded.

This is very tricky to do, however, for anything but the simplest functions.  Our own history has been rife with code breaking because a deeply nested utility function was suddenly threaded.  This is why holding any TLS or Lock across any non-trivial code becomes risky.

I think the reordering in #2 only works if you do *not* do the mPrimCount reset?   Because you've returned to the scheduler after the mRoot.clear(), there is the chance someone else has used the scractch pad and incremented up to a higher prim idcount.  If you reset the primid, you'll re-use whatever they had put in there.

I don't like the extra assignment, but it would be good to close this non-determinism that has plagued us for a while.  I would, however, insist on an lengthy comment explaining why the apparently nonsensical pair of mPrimId = 0 are there, or they are bound to be removed by some linter in the future.

Idclip:
Another option is to add a bool threaded argument to the clear function instead of a method with a new name - though I know this has been a point of dispute in the past. 

> I think the reordering in #2 only works if you do not do the mPrimCount reset?

Oops yes you're correct - which means we'd have to wrap the maxid/leafcount check in a while loop to ensure that the max prim id hasn't been reached by an overlapping task.

Will definitely add a verbose comment with whatever approach we take. I double checked and clear() was indeed threaded **after** this implementation, in VDB 3.2. I don't believe this particular logic in MeshToVolume has changed since then.

Idclip:
As discussed, opted for the serial implementation of clear() with a nice long comment (suggestions on the wording welcome)

Idclip:
I'll be merging this tomorrow

PR #567, opened 2019-11-25
Ovdb 123 gaussian curvature

kmuseth:
Added: Gaussian curvature, principal curvatures, Euler characteristic, genus, and new intersection tests (needed for sparse fast sweeping)


openvdb/tools/LevelSetMeasure.h
```diff
@@ -57,67 +58,62 @@ namespace tools {
...
-///                      world space units, else in voxel units.
+///                      closed level set surfaces
+/// @warning The genus is only well defined for a single connected surface
 ///
-/// @throw TypeError if @a grid is not scalar or not floating-point or not a level set.
+/// @throw TypeError if @a grid is not scalar or not floating-point or not a level set or empty.
 template<class GridType>
-inline void
-levelSetMeasure(const GridType& grid, Real& area, Real& volume, Real& avgCurvature,
-                bool useWorldSpace = true);
```
Idclip: As discussed, if we can bring these back, mark them as deprecated (perhaps move to the end of the file?) and alias them to the new methods that would be good. Ideally we'd do this in the LevelSetMeasure class too, but I'm unclear on how much work that will be

openvdb/tools/LevelSetMeasure.h
```diff
@@ -162,77 +156,122 @@ class LevelSetMeasure
...
+            if (parent->mInterrupter) parent->mInterrupter->start("Measuring area and volume of level set");
             if (parent->mGrainSize>0) {
                 tbb::parallel_for(parent->mLeafs->leafRange(parent->mGrainSize), *this);
             } else {
                 (*this)(parent->mLeafs->leafRange());
             }
+            std::thread t1([&](){parent->mArea   = parent->reduce(0);});
+            std::thread t2([&](){parent->mVolume = parent->reduce(1)/3.0;});
+            t1.join();
+            t2.join();
```
Idclip: Do we have a preference to using stl here versus tbb task groups?

openvdb/tools/LevelSetMeasure.h
```diff
@@ -162,77 +156,122 @@ class LevelSetMeasure
...
+            if (parent->mInterrupter) parent->mInterrupter->end();
         }
-        Measure3(const Measure3& other) : mParent(other.mParent), mAcc(*mParent->mTree) {}
+        MeasureCurvatures(const MeasureCurvatures& other) : mParent(other.mParent), mStencil(*mParent->mGrid) {}
         void operator()(const LeafRange& range) const;
         LevelSetMeasure* mParent;
-        typename GridT::ConstAccessor mAcc;
-    };
-    inline double reduce(double* first, double scale)
+        mutable math::CurvatureStencil<GridT, false> mStencil;
```
Idclip: minor: you could opt to move the member mStencil into the operator of these mini struct to avoid have to explicitly define the copy constructor

openvdb/tools/LevelSetMeasure.h
kmuseth: no we have never made that a policy (and if we did other code would have to change). I would argue that since we're officially on C++11 and std::threads are one of the major new features in this version of the language it would be odd to disallow it. In fact I would argue that we should phase out TBB where std::threads are easily replace it.

openvdb/tools/LevelSetMeasure.h
kmuseth: I will add this function back but deprecate it

openvdb/tools/LevelSetMeasure.h
kmuseth: That will imply that a new instance of the stencil is constructed each time the operator method is called. If on the other hand it's a member a stencil is only created for each new instance of the struct. Since the stencil is internally caching grid nodes it's not exactly lightweight so I want to re-use it as much as possible (without actually resorting to TLS)

Idclip: > That will imply that a new instance of the stencil is constructed each time the operator method is called. 

Isn't this also the case currently due the copy constructor of these structs re-creating it from the grid? I assume this is to *avoid* the copying of its cache unnecessarily. As tbb will copy this object for each subrange, isn't this equal to putting it in the operator?

openvdb/tools/LevelSetMeasure.h
jmlait: It might be I'm misunderstanding what std::thread does.  But from my viewpoint std::thread should not be used.

How can we get composition of multithreading if everything generates a new set of threads?  If we use std::thread, don't we have to then re-create our own thread pool?   And how do we restrict the maximum number of active threads if we have both tbb:tasks grabbbing from a tbb pool, and std::thread with its own custom pool?

std::thread is nice to have in the language as it lets you build something higher level, like TBB, out of it.   It means that tbb::thread is now deprecated.  The rest of TBB's task scheduler, however, is very much needed until C++ has a built-in work-stealing task scheduler.

openvdb/tools/LevelSetMeasure.h
kmuseth: is it thought? It's my impression that tbb can reuse the same instance of the function on multiple ranges. If that's never the case then yes you're of course right but I'm pretty sure I recall reading somewhere that tbb is clever about when it is required to copy construct an instance and when it can be re-used (for tbb::parallel_for)

openvdb/tools/LevelSetMeasure.h
kmuseth: I really don't see the problem here. In this case I have exactly two tasks (that by design each have to be serial because they address the "problem of catastrophic cancellation") so why would I need (or even benefit) from tbb here? I agree that in general we should be careful of not mixing the two since they are likely to use different thread-schedulers (and hence might cannibalize and impair performance of each other), but in this case I have a fixed (low) number of tasks that cannot internally be parallelized so the patterns I use seems just fine to me. Have I missed something or were you merely responding to my comment as if I said we should generally switch from tbb to std::thread - which is certainly not what I meant. However, there are simple cases like this where it seems strange to be that we're not embracing a major new feature of the language. std::thread also offer several great feature that simply are not available in tbb as far as I know (e.g. std::future and std::promise) - are you suggesting we never use those as well?

jmlait: This is a very good discussion to have.  We should be trying to move to standard features rather than external library features where we can.  So I'm glad you thought to use std::thread.

I think my problem is that std::thread is meant to replace pthreads, not replace tasks.   So if this were a pthread implementation, I'd object to it because creating & destroying threads is expensive. We'd want those threads to come from a pool.

I found this discussion on the issues:
https://stackoverflow.com/questions/12993451/c11-stdthread-pooled

In theory someone could implement std::thread to use a pool.  But as noted there, it is expected each std::thread has its own thread-local-storage, etc.  So in practice it would be very difficult to do that, so it is almost always going to be just a thin wrapper over the local system threading (win threads or pthreads).  This means each measure computation we generate & destroy two OS-level threads.  I can't imagine this being pretty if we try and do it hundreds of times in parallel due to measuring a few hundred VDBs at once.

std::thread makes a lot of sense for me for long-term threads.  Stuff like IO threads, Networking transfer threads, UI threads, etc.  These are where probably where std::future may be useful.  But as worker threads, I think they shouldn't be used.

1) Pooling.   We want to cap total thread creation.  Both for sanity of TLS size and for performance costs.

2) Control Core Count.  Houdini clamps the number of active cores to the user-specified range.  It does this by limiting the TBB task scheduler.  If we spawn multiple threads inside of a task we defeat this.  A -j4 program might end up using 8 cores rather than 4.  A bit of over-subscription we can get away with, but if this pattern continues we'll lose all control.  This is a commonly used feature to split up blades between multiple processes.

3) Single Threaded.  Explicitly using std::thread makes it impossible to devolve to a single-threaded algorithm.  I always want to have the ability to go serial.  It is very useful for debugging.  And it is equally important for performance.  There is a grainsize check on the leafRange() function, but no such grain size guarding the threading to the two tasks.   Admittedly, usually we need two separate calls for TBB as well to achieve this (as shown in the parallelfor example).  In Houdini we wrap tbb::parallel_for specifically so we can guarantee the serial variant is just inlined with no creation of tasks.

4) Locked main thread.  We aren't using the main thread as a worker in this pattern.  Probably need similar to what was used in PointScatter if we were to use std::thread:

```
std::thread t1([&](){parent->mArea   = parent->reduce(0);});
parent->mVolume = parent->reduce(1)/3.0;
t1.join();
```

One advantage of using tbb tasks is that they can be immediately picked up by the main thread without returning to the OS scheduler.

...

All this said, I do like how clean the std::thread implementation reads!  We'd definitely want to make sure a replacement pattern is as easy to write.  Might be useful to develop a VDB parallel util to make it easy to invoke multiple lambadas in this sort of fork/join pattern.  It can then also handle the serial versions, etc, in a single location so we don't have to keep having if (threaded) { parellel } else { serial } in the code everywhere.

danrbailey: I agree with Jeff, we should be using the thread pool that TBB provides. The PointScatter seems to be the only place we're not doing that currently, so we should probably change that too.

I believe Nick mentioned this too, but TBB's task groups seem like a good fit here and are probably only marginally more complicated than this code?

Regarding the larger issue of threading, there's lots to talk about here, we should put it in the agenda for a meeting soon! I'm leaning towards a simple header-only TBB wrapper that offers parallel_for, parallel_reduce, parallel_sort, enumerable_thread_specific and task groups and a serial, non-TBB implementation that would make TBB an optional dependency. This kind of abstraction is already offered to an extent by leafManager.foreach(...), so it's just taking this further. There have been discussions ongoing in the ASWF for a while now about thread pools, even better if we can tie into such a shared effort rather than a VDB parallel util:

https://lists.aswf.io/g/tac/topic/tac_topic_request_thread/42377613?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,42377613

openvdb/tools/LevelSetMeasure.h
```diff
@@ -135,77 +138,122 @@ class LevelSetMeasure
...
+
+    /// @brief Compute the volume of the level set surface.
+    /// @param useWorldUnits Specifies if the result is in world or voxel units.
+    /// @note Performs internal caching so only the initial call incurs actual computation.
+    Real volume(bool useWorldUnits = true);
+
+    /// @brief Compute the total mean curvature of the level set surface.
+    /// @param useWorldUnits Specifies if the result is in world or voxel units.
+    /// @note Performs internal caching so only the initial call incurs actual computation.
+    Real totMeanCurvature(bool useWorldUnits = true);
```
danrbailey: I think I'd prefer the more explicit "totalMeanCurvature" to the abbreviation of "tot". Likewise "average" instead of "avg".

openvdb/math/Stencils.h
```diff
@@ -1511,8 +1534,34 @@ class CurvatureStencil: public BaseStencil<CurvatureStencil<GridT, IsSafe>, Grid
...
+        Real alpha, normGrad;
+        return this->gaussianCurvature(alpha, normGrad) ? ValueType(alpha*mInvDx2/math::Pow4(normGrad)) : 0;
+    }
+
+    /// @brief Return both the mean and the Gaussian curvature at the
+    ///        previously buffered location.
+    ///
+    /// @note This method should not be called until the stencil
+    /// buffer has been populated via a call to moveTo(ijk).
+    inline void curvatures(ValueType &mean, ValueType& gauss)
```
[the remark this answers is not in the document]
kmuseth: this and several of the other methods - fixed

openvdb/math/Stencils.h
```diff
@@ -1511,8 +1534,34 @@ class CurvatureStencil: public BaseStencil<CurvatureStencil<GridT, IsSafe>, Grid
...
+    }
+
+    /// @brief Return the Gaussian curvature at the previously buffered location.
+    ///
+    /// @note This method should not be called until the stencil
+    /// buffer has been populated via a call to moveTo(ijk).
+    inline ValueType gaussianCurvature()
+    {
+        Real alpha, normGrad;
+        return this->gaussianCurvature(alpha, normGrad) ? ValueType(alpha*mInvDx2/math::Pow4(normGrad)) : 0;
```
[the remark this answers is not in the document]
kmuseth: fixed

openvdb/tools/LevelSetMeasure.h
Idclip: This should definitely be discussed at the next TSC. @jmlait @danrbailey, do you have any additional thoughts on the rest of the PR bar the these specific changes? If we want this in for 7.0.0 and we're confident on the rest of the PR, I vote we update this implementation to use tbb::task_groups and continue this discuss after the release. @kmuseth thoughts?

kmuseth: I've resolved this with tbb::parallel_invoke

openvdb/tools/LevelSetMeasure.h
```diff
@@ -30,67 +31,71 @@ namespace tools {
...
+/// @throw TypeError if @a grid is not scalar or not floating-point or not a level set or empty.
+template<class GridType>
+inline int
+levelSetGenus(const GridType& grid);
+
+////////////////////////////////////////////////////////////////////////////////////////
+
+/// @deprecated Use the @a LevelSetMeasure class instead.
 template<class GridType>
+OPENVDB_DEPRECATED
```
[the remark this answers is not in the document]
kmuseth: removed forward deceleration

openvdb/tools/LevelSetMeasure.h
kmuseth: Do "grep avg openvdb/tools/*" and you'll find plenty of hits (all of them code). Now do "grep average openvdb/tools/*" and you get far fewer hits and more importantly none of them are actual code (only documentation). So, in the interest of consistency I will not make the suggested change (just like you pushed back on my recent suggested for a different signature of addChild - using the same argument of inconsistency with existing code :).

openvdb/tools/LevelSetMeasure.h
kmuseth: There seems to be some uncertainty here, so can we agree that the worst case scenario of your suggested change is less efficiency whereas the worst case scenario of mine is a few extra lines of code? If so I think we should keep it as is. I'm fine revisiting this again once we're certain the change doesn't reduce performance, but as of now I'm not convinced.

openvdb/tools/LevelSetMeasure.h
danrbailey: Consistency is always a good argument. :)

I'm going to approve this, but with a small caveat to your reasoning - most of those hits for "avg" are used in variable names not method names, meaning that they are implementation detail and not public API (implementation detail is less critical IMO). There is in fact only one method name which uses an abbreviated "avg()" which is in Stats.h which also includes an equivalent mean() method. Where possible, I prefer longer but more explicit method names.

kmuseth:
I believe all review comments are now addressed, either with actual code changes or with comments rebutting the proposed changes. Please someone review and (hopefully) approve so I can merge this into v7.0.

openvdb/tools/LevelSetMeasure.h
danrbailey: That seems like a good approach to me.

openvdb/tools/LevelSetMeasure.h
kmuseth: thanks and yes I agree that in general descriptive names are to be preferred over cryptic abbreviations. I just think that in this case there is a precedence in the code for using avg so at least I would be familiar with (not to say expecting) this naming scheme.

openvdb/tools/LevelSetMeasure.h
Idclip: Yes that's absolutely fine

openvdb/tools/LevelSetMeasure.h
jmlait: Thank you very much for switching to consistent tbb usage.

PR #574, opened 2019-11-28
Add SIMD Instructions to NodeMask

danrbailey:
This is an implementation derived from #493 to be included in the 7.0 release. Differences between that PR include:

* OPENVDB_SSE42 flag used to opt-in to using hardware intrinsics
* Intrinsic headers included in Platform.h
* Hardware intrinsics only used for the two most common methods (CountOn() and FindLowestOn() with Byte and Index64 variants)
* 32-bit Windows uses software implementation for CountOn(Index64)
* Implementation introduced directly into NodeMasks.h using conditional macros instead of additional header files
* CI extended to evaluate the unit tests on SSE 4.2 builds of OpenVDB

@Const-me - I'd much appreciate you reviewing this change. Also, we typically use the contributors full name in the release notes, but I only have your first name - so let me know if you'd like me to change the way you're credited.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,3 +129,5 @@ jobs:
     - uses: actions/checkout@v1
     - name: build
       run: ./ci/build.sh clang++ Debug 7 ON SSE42
+    - name: test
+      run: cd build && ctest -V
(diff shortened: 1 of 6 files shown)


Const-me:
Hi, @danrbailey. The change looks OK.
BTW, if you want to further improve performance of VDB by utilizing modern hardware, you might find this interesting: http://const.me/articles/simd/simd.pdf

danrbailey:
> Hi, @danrbailey. The change looks OK.
> BTW, if you want to further improve performance of VDB by utilizing modern hardware, you might find this interesting: http://const.me/articles/simd/simd.pdf

That's great, thanks! (@jmlait @Idclip @kmuseth - can someone approve please and I'll merge).

PR #576, opened 2019-11-29
Python Module Option Now Off in CMake

danrbailey:
As discussed here:

https://github.com/AcademySoftwareFoundation/openvdb/blame/master/tsc/meetings/2019-10-10.md#L46

Disable the Python Module in CMake by default so that it's more likely that a first time user attempting to build OpenVDB will succeed.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -50,6 +50,7 @@ Version 7.0.0 - In development
     - Added support for Boost versions 1.65 and later when building
       the Python module with NumPy support through CMake.
     - Improved CMake Python3 support.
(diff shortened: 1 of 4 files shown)


Idclip:
Just noticed, there's a comment at the top of the root CMakeLists.txt which states that the python module is built by default - needs to be updated with this.

danrbailey:
> Just noticed, there's a comment at the top of the root CMakeLists.txt which states that the python module is built by default - needs to be updated with this.
> 
> in build.txt, the OpenVDB Components table also needs updating

Thanks, updated!

PR #578, opened 2019-12-03
Add Houdini 18.0 to CI

danrbailey:
@jmlait - I just tried adding H18 to the CI and aside from having to install the future and lxml python modules, am now getting this error:

> CMake Error at hou/toolkit/cmake/HoudiniConfig.cmake:130 (message):
>   Error running /__w/openvdb/openvdb/hou/bin/hython: No licenses could be
>   found to run this application.
> 
>   	Please check for a valid license server host

Any ideas what might be going on? This works for H17 and H17.5, I'm not aware there's been any license server changes. :/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,17 @@ jobs:
     - name: test
       run: cd build && ctest -V
 
+  testhou180:
+    runs-on: ubuntu-16.04
+    container:
+      image: aswf/ci-base:2019
+    steps:
+    - uses: actions/checkout@v1
+    - name: houdini
+      run: ./ci/install_houdini.sh 18.0 ${{ secrets.HOUPASS }}
+    - name: build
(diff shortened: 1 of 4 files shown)


Idclip:
The CMake function `houdini_get_default_install_dir` (and subsequently `houdini_configure_target`) now call hython to get the default install directory. In 17.5 it was implemented manually and there weren't any calls to houdini binaries (aside from sesitag).

Idclip:
I think we can bypass this by providing a value for `-D OPENVDB_HOUDINI_INSTALL_PREFIX` when executing cmake. But I do wonder whether hython should be used here at all for this task?

danrbailey:
> I think we can bypass this by providing a value for `-D OPENVDB_HOUDINI_INSTALL_PREFIX` when executing cmake. But I do wonder whether hython should be used here at all for this task?

Thanks for this suggestion, this seems to have worked fine now. Good to be testing against H18. :)

PR #579, opened 2019-12-04
Fixed a rotation order bug in tools::local_util::decompose().

danrbailey:
Re-submission of #364 with conflicts resolved (Peter's CLA checks were failing).


openvdb/tools/GridTransformer.h
```diff
@@ -267,14 +267,24 @@ class GridTransformer: public GridResampler
...
 template<typename T>
-inline bool
+inline int
 decompose(const math::Mat4<T>& m, math::Vec3<T>& scale,
     math::Vec3<T>& rotate, math::Vec3<T>& translate)
 {
-    if (!math::isAffine(m)) return false;
+    if (!math::isAffine(m)) return DECOMP_INVALID;
+
+    const auto posAngle = [](T angle) {
```
kmuseth: can you explain or motivate this functor in a brief comment. It's certainly not self explanatory to me. It seems to be replace an abs operation further down in the code - was that a bug?

openvdb/tools/GridTransformer.h
```diff
@@ -310,15 +318,14 @@ decompose(const math::Mat4<T>& m, math::Vec3<T>& scale,
         const math::Vec3<T> tmpAngle = math::eulerAngles(mat, math::XYZ_ROTATION);
 
         const math::Mat3<T> rebuild =
-            math::rotation<math::Mat3<T> >(math::Vec3<T>(1, 0, 0), tmpAngle.x()) *
-            math::rotation<math::Mat3<T> >(math::Vec3<T>(0, 1, 0), tmpAngle.y()) *
             math::rotation<math::Mat3<T> >(math::Vec3<T>(0, 0, 1), tmpAngle.z()) *
+            math::rotation<math::Mat3<T> >(math::Vec3<T>(0, 1, 0), tmpAngle.y()) *
```
kmuseth: this is a typo/bug in the old code?

CHANGES
```diff
@@ -39,6 +39,8 @@ Version 7.0.0 - In development
     - Fixed a Windows build issue in openvdb_render.
     - Fixed a non deterministic threading bug in tools::meshToVolume() during
       polygon voxelization which could result in different distance values.
+    - Fixed a rotation order bug in tools::local_util::decompose()
+      that caused it to fail to decompose some matrices.
```
Idclip: these notes need to be moved to 7.0.1

danrbailey:
@kmuseth - I don't know the answers to your questions here. I tried running the new unit test that Peter added, but without the implementation change to attempt to better understand the details, but it still passes so I'm none the wiser.

If we don't have confidence in accepting this PR without Peter's input here, we'll have to close it. 

(@pcucka)

openvdb/tools/GridTransformer.h
jmlait: From the Math.h:
Note that in our convention, XYZ means Rz * Ry * Rx.

Thus I think this new order is correct.

openvdb/tools/GridTransformer.h
jmlait: My understanding of the code as a whole:  when you have a reflection, you have the issue of how to apply it - it could be a reflection through the origin, or it could be a plane reflection + 180 degree turn.   In Houdini I know we always select reflection through the origin as the most unbiased choice, but it is also rarely the one people "mean".

The code seems to be looking for the euler angles with the minimial change from identity to find the best choice of plane reflections to introduce.

This function seems to be doing a poor-man's modulus, understandable since C++ has a poorly defined modulus for negative numbers.    However, I'm not sure that's what one wants if one wants minimal rotation.  A slight negative rotation is going to return M_PI*2 so flag as being a large rotation, so not acceptable as a solution.

I think the abs() version should be kept instead?

danrbailey:
I've modified this PR to revert back to the abs() behavior as we discussed. In fact, we ran into this issue in production recently and I can confirm that the fixing of the rotation order resolved the problem. 

@kmuseth has already approved so unless anyone has any further feedback, I'll merge this tomorrow.

PR #581, opened 2019-12-05
fixed bugs

kmuseth:


kmuseth:
@Idclip, I have tested this with ABI 6 and 7 (and actually fixed an unknown bug) but you still need to check it against Houdini builds

PR #583, opened 2019-12-09
Minor CMake + Release Process updates

Idclip:


openvdb_houdini/CMakeLists.txt
```diff
@@ -48,11 +48,6 @@ houdini_get_default_install_dir():
   Windows: $ENV{HOMEDRIVE}$ENV{HOMEPATH}\Documents\houdiniX.X
 If OFF, CMAKE_INSTALL_PREFIX is used.]=] ON)
 
-# SESITAG CMake options
-option(OPENVDB_HOUDINI_ENABLE_SESI_TAG [=[
```
danrbailey: Shouldn't we wait until we deprecate support for H17 and H17.5 to actually remove this option?

doc/CMakeLists.txt
```diff
@@ -14,6 +14,14 @@ if(POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW)
 endif()
 
+# If OPENVDB_DOXYGEN_HOUDINI is set, the Houdini documentation is also
+# built. You can provide this module a value for Houdini_VERSION which will
+# take into account any conditional compiled blocks. Whilst this doesn't
+# require you to build the Houdini plugin, building from the root CMake
+# with OPENVDB_BUILD_HOUDINI_PLUGIN or USE_HOUDINI will define this value.
```
danrbailey: I would expect this value to only be set to ON if OPENVDB_BUILD_HOUDINI_PLUGIN=ON **and** OPENVDB_BUILD_DOCS=ON.

tsc/process/release.md
```diff
@@ -7,8 +7,42 @@ The following assumes that the current OpenVDB library version number is 6.0.0 a
...
-- [ ] Build the documentation (both the `doc` and `pydoc` targets) and replace the contents of `openvdb-website/contents/documentation/doxygen/` with the output.  [_This step should be automated, and the thousands of files it generates should preferably not be committed to the repository._]
+- [ ] Update `openvdb-website/index.html` with a news item announcing the release, and delete the oldest news item.  Open that page in a browser and check that the website renders correctly and that there are no broken links.
+- [ ] Build the documentation (both the `doc` and `pydoc` targets) and replace the contents of `openvdb-website/documentation/doxygen/` with the output. There are two sub folders, /search and /python. /search will be generated by the core doxygen commands. The /python folder will be built by epydoc. The python documentation requires the python module to be built and installed. Make sure you build with OPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON, otherwise documentation will not be generated for some grid types. An example CMake command list (assuming within a nested build folder):
+
```
danrbailey: Nice to have extra clarity here, thanks!

doc/CMakeLists.txt
Idclip: The point of this option is so you don't have to have `OPENVDB_BUILD_HOUDINI_PLUGIN=ON` set, as building the Houdini plugin isn't connected to these docs

openvdb_houdini/CMakeLists.txt
Idclip: It's not removed, just deprecated (and no longer visible as an option). In my opinion it makes sense to do this now seeing as this does nothing for Houdini 18/future versions

doc/CMakeLists.txt
danrbailey: Understood. I'm saying that if you _do_ have OPENVDB_BUILD_HOUDINI_PLUGIN=ON set, it shouldn't build the Houdini docs _unless_ you also have OPENVDB_BUILD_DOCS=ON set.

openvdb_houdini/CMakeLists.txt
danrbailey: Ok, makes sense, thanks.

doc/CMakeLists.txt
Idclip: Ah gotchya - but that's the case anyway as unless OPENVDB_BUILD_DOCS is set this CMakeLists file never gets added. So no docs are built unless OPENVDB_BUILD_DOCS=ON, where by all docs are built by default irrespective of the OPENVDB_BUILD_HOUDINI_PLUGIN option

danrbailey: Sure, it's perhaps just worth adding a small caveat to make that explicit in this comment. Although it's convention for the project, the wording:

> building from the root CMake with OPENVDB_BUILD_HOUDINI_PLUGIN or USE_HOUDINI will define this value

introduces some uncertainty that could still require tracing back to the root CMakelists to find out that this file is only in use if OPENVDB_BUILD_DOCS is set. 

It's a minor thing though, I'll mark the PR as approved anyway and you can decide what to do here.

PR #584, opened 2019-12-10
The interface for evalActiveVoxelBoundingBox suggests it returns false.

jmlait:
The interface for evalActiveVoxelBoundingBox suggests it returns false.
if empty, but only does so if the tree is empty.  This will change
it so it will also do so if the computed bounds are empty.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,9 @@ OpenVDB Version History
 
 Version 7.0.1 - In Development
 
+    Bug fixes:
+    - Fixed a bug where empty grids were still returning true from
+      evalActiveVoxelBoundingBox when leaf nodes were still present.
 
 Version 7.0.0 - December 6, 2019
 
--- a/CHANGES
+++ b/CHANGES
@@ -4,8 +4,9 @@ OpenVDB Version History
 Version 7.0.1 - In Development
 
     Bug fixes:
-    - Fixed a bug where empty grids were still returning true from
-      evalActiveVoxelBoundingBox when leaf nodes were still present.
+    - Fixed a bug where grids with no active values might return true when the
+      method evalActiveVoxelBoundingBox is called. The correct behavior is to
+      only return true if the grid contains any active values.
 
 Version 7.0.0 - December 6, 2019
 
--- a/doc/changes.txt
+++ b/doc/changes.txt
@@ -8,8 +8,9 @@
 
 @par
 Bug fixes:
-- Fixed a bug where empty grids were still returning true from
-  <TT>evalActiveVoxelBoundingBox</TT> when leaf nodes were still present.
+- Fixed a bug where grids with no active values might return true when the
+  method <TT>evalActiveVoxelBoundingBox</TT> is called. The correct behavior is to
+  only return true if the grid contains any active values.
 
 @htmlonly <a name="v7_0_0_changes"></a>@endhtmlonly
 @par
--- a/doc/changes.txt
+++ b/doc/changes.txt
@@ -2,6 +2,15 @@
 
 @page changes Release Notes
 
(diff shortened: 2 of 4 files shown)


jmlait:
Change notes added.  Any suggestions which unit test to add to?  There is nothing testing this function per se, only indirectly.   I also need to make sure it builds an empty-but-full tree...

danrbailey:
> Change notes added. Any suggestions which unit test to add to? There is nothing testing this function per se, only indirectly. I also need to make sure it builds an empty-but-full tree...

I believe there are specific tests for these two methods in TestGridBbox, so I think we should just extend these two unit tests.

jmlait:
Thank you, that is what I was hoping for.

CHANGES
```diff
@@ -3,6 +3,9 @@ OpenVDB Version History
 
 Version 7.0.1 - In Development
 
+    Bug fixes:
+    - Fixed a bug where empty grids were still returning true from
+      evalActiveVoxelBoundingBox when leaf nodes were still present.
```
kmuseth: This description was confusing to me since an empty grid (as defined by the Grid::empty() = Tree::empty()) is a Tree without any nodes. However, this method computes the bounding box of active values (irrespective of where they live in the tree). May I therefore suggest this wording instead:

- Fixed a bug where grids with no active values might return true when the method evalActiveVoxelBoundingBox is called. The correct behavior is to only return true if the grid contains any active values.

doc/changes.txt
```diff
@@ -2,6 +2,15 @@
 
 @page changes Release Notes
 
+@htmlonly <a name="v7_0_1_changes"></a>@endhtmlonly
+@par
+<B>Version 7.0.1</B> - <I>In Development</I>
+
+@par
+Bug fixes:
+- Fixed a bug where empty grids were still returning true from
```
kmuseth: This description was confusing to me since an empty grid (as defined by the Grid::empty() = Tree::empty()) is a Tree without any nodes. However, this method computes the bounding box of active values (irrespective of where they live in the tree). May I therefore suggest this wording instead:

- Fixed a bug where grids with no active values might return true when the method evalActiveVoxelBoundingBox is called. The correct behavior is to only return true if the grid contains any active values.

Idclip:
Not to try and extend this PR, but we should probably update the brief for `evalLeafBoundingBox` to say that only leaf nodes with active voxels are included too, as it currently reads a bit weird. i.e. `Return in bbox the axis-aligned bounding box of all leaf nodes and active tiles.` -> `Return in bbox the axis-aligned bounding box of all active tiles and leaf nodes with active values.`

jmlait:
> Not to try and extend this PR, but we should probably update the brief for `evalLeafBoundingBox` to say that only leaf nodes with active voxels are included too, as it currently reads a bit weird. i.e. `Return in bbox the axis-aligned bounding box of all leaf nodes and active tiles.` -> `Return in bbox the axis-aligned bounding box of all active tiles and leaf nodes with active values.`

I just looked at the code for LeafNode.h and I believe your description is more accurate.

CHANGES
jmlait: True, this confusion is precisely how the issue arose :>

In my mind, a tree with no active values *is* empty.  It is a purely internal matter whether there any leaf nodes allocated or not.   This viewpoint is likely why I also don't like "densify" functions as they are meaningless to me :>

I'll try and clarify the wording as you suggest.

kmuseth: thanks

Idclip:
> I just looked at the code for LeafNode.h and I believe your description is more accurate.

Yeah this has caught me out before, I have a use case where I actually want to include all leaf nodes, irrespective of their mask. I'll make a ticket for this.

PR #585, opened 2019-12-11
OVDB-135: Point Partitioner Determinism Test

danrbailey:
Fix a typo and add a unit test to validate that the point partitioner is deterministic (it's easy to accidentally break determinism when refactoring)

To be clear - the point partitioner **is** currently deterministic (to the best of my knowledge).

### DIFF
--- a/openvdb/tools/PointPartitioner.h
+++ b/openvdb/tools/PointPartitioner.h
@@ -709,10 +709,10 @@ struct OrderSegmentsOp
     using IndexArray = boost::scoped_array<PointIndexType>;
     using SegmentPtr = typename Array<PointIndexType>::Ptr;
 
-    OrderSegmentsOp(SegmentPtr* indexSegments, SegmentPtr* offestSegments,
+    OrderSegmentsOp(SegmentPtr* indexSegments, SegmentPtr* offsetSegments,
         IndexArray* pageOffsetArrays, Index binVolume)
         : mIndexSegments(indexSegments)
-        , mOffsetSegments(offestSegments)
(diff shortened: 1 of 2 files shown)


Idclip:
This is a bit of an odd test - imo testing determinism can usually be solved by "regression" tests (i.e. testing against a "hard coded"/expected value) rather than an arbitrary amount of consecutive testing (for example, you're constructing 10 additional partitioners which doesn't really mean anything). You could argue that this then changes the function of the test as it also tests the value result rather than simply testing the results match; but really `TestPointPartitioner::testPartitioner()` should implicitly handle this by checking the results of `indices` :\

danrbailey:
> This is a bit of an odd test - imo testing determinism can usually be solved by "regression" tests (i.e. testing against a "hard coded"/expected value) rather than an arbitrary amount of consecutive testing (for example, you're constructing 10 additional partitioners which doesn't really mean anything). You could argue that this then changes the function of the test as it also tests the value result rather than simply testing the results match; but really `TestPointPartitioner::testPartitioner()` should implicitly handle this by checking the results of `indices` :\

This test is better than not having one at all and it caught a number of times I accidentally broke determinism. 

That being said, I get what you're saying and I don't fundamentally disagree. I have a subsequent change coming that exposes a threading parameter which sorts the points on a single-thread. If you can approve this one as-is, when I push that PR up, I'll modify this to test multi-threaded vs single-threaded instead?

danrbailey:
I'll close this out and try a different approach (I have a number of pretty major changes coming to this tool, trying to figure out the easiest order in which to push them up for reviewing).

PR #589, opened 2019-12-11
OVDB-135: Point Partitioner Leaf Origins

danrbailey:
This is a batch of (relatively) small improvements as a starting point, it's nicely broken out into commits to hopefully make reviewing simpler. 

The most significant change in this PR is computing the leaf node origins / page coordinates from cached page indices. The current implementation re-runs the partition algorithm for the first point in the page to figure out which leaf node it belongs to. This is problematic because it assumes the point positions are immutable and thus can't be used in a way that changes or streams the point position data during the partitioning. This has no measurable impact to performance or memory usage from the various cases I've tested - the number of page coordinates is typically much smaller than the number of voxels or points and similar processing time goes into evaluating the partition algorithm as to computing the leaf node origins from the page indices.

Aside from this, some other minor changes include fixing a typo, adjusting some indexing for readability and migrating from boost::scoped_array to std::unique_ptr<[]>. This last change has resulted in a small change to the API, so bumping the version from 7.0.1 to 7.1.0.

Following this PR, some more major refactoring to come...


openvdb/tools/PointPartitioner.h
```diff
@@ -465,13 +409,13 @@ template<typename PointIndexType>
 struct MergeBinsOp
 {
     using Segment = Array<PointIndexType>;
-    using SegmentPtr = typename Segment::Ptr;
+    using SegmentPtr = std::unique_ptr<Array<PointIndexType>>;
```
Idclip: Is there a need for this change seeing as you've changed the `Ptr` typedef? We use `Ptr` and `UniquePtr`, though the `UniquePtr` typedef is something we've only used in /points/. We should decide on how we want these to be promoted, either all through `Ptr` or introducing specific types for unique, weak etc

openvdb/tools/PointPartitioner.h
```diff
@@ -918,11 +878,49 @@ inline void partition(
...
+        [&](tbb::blocked_range<size_t>& range)
+        {
+            for (size_t n = range.begin(); n < range.end(); n++)
+            {
+                Index segmentOffset = segmentOffsets[n];
+                PointIndexType* indices = pageIndexArrays[n].get();
+
+                const Coord& segmentCoord = segmentCoords[n];
+
+                tbb::blocked_range<size_t> copyRange(0, pageOffsetArrays[n][0] - 1);
```
Idclip: A comment here that the array size is stored in the first element, otherwise it's really not clear

openvdb/tools/PointPartitioner.h
```diff
@@ -918,11 +878,49 @@ inline void partition(
...
+                        {
+                            Index pageIndex = indices[i];
+                            Coord& ijk = pageCoordinates[segmentOffset+i];
+
+                            ijk[0] = pageIndex >> (2 * binLog2Dim);
+                            Index pageIndexModulo = pageIndex - (ijk[0] << (2 * binLog2Dim));
+                            ijk[1] = pageIndexModulo >> binLog2Dim;
+                            ijk[2] = pageIndexModulo - (ijk[1] << binLog2Dim);
+
+                            ijk = (ijk << bucketLog2Dim) + segmentCoord;
```
Idclip: Can you elaborate on how this logic maps to the original behaviour of `LeafNodeOriginOp`

Idclip:
> The current implementation re-runs the partition algorithm for the first point in the page to figure out which leaf node it belongs to. This is problematic because it assumes the point positions are immutable and thus can't be used in a way that changes or streams the point position data during the partitioning.

A call to construct->partition will always assume that the point positions are immutable. You don't want to change them mid way through or even after the binning, as that will cause a whole range off issues. Instead, as far as I can tell, your changes seem to be focused on only accessing the positions once. I obviously can't see how you intend to leverage this, so can you elaborate on why this is useful for your use case?

openvdb/tools/PointPartitioner.h
danrbailey: > Is there a need for this change seeing as you've changed the Ptr typedef?

Not really, I can change this back if you would prefer? This is only exposed in an "internal" namespace, so I'm not sure it matters too much in this specific case, though I agree there's a more general discussion to be had here (which I'd rather not get into here).

openvdb/tools/PointPartitioner.h
danrbailey: I'll put it into a new temporary variable for now to make it clearer, though I believe this is changed in the next round of refactoring.

openvdb/tools/PointPartitioner.h
danrbailey: The original behavior of `LeafNodeOriginOp` re-computed the leaf node by re-applying the binning operation to the first point in the bin. Instead, this algorithm stores the page indices computed in OrderSegmentsOp (which would otherwise be thrown away) and then uses the inverse of the index portion of the algorithm in BinPointIndicesOp to deduce the leaf node origin from this stored index.

PS I'm really not fan of the terminology in this tool - in particular **buckets** means the same as **pages** and **leafs**, **bins** means the same as **segments** and the frequent interchange of these terms makes the algorithm quite hard to follow IMO. Nevertheless, I've attempted to follow existing conventions in this isolated change. There's more refactoring to come to try and make the algorithm as a whole more readable.

openvdb/tools/PointPartitioner.h
danrbailey: Fixed.

openvdb/tools/PointPartitioner.h
danrbailey: Fixed.

danrbailey:
@Idclip - any further thoughts, I'd like to get this one merged.

Idclip:
@danrbailey Sorry it's taken me so long to get back to you on this one.

> I don't believe this is true, perhaps you can elaborate on what you see to be the problem here?

Isn't the idea behind this partitioner that points are partitioned spatially? If you're going to subsequently modify the positions it suggests to me that perhaps this isn't the best use of this class. As you've pointed out though, nothing like that is being performed here.

Everything else looks good, but If possible it would be great if @kmuseth or @jmlait could take a quick look too as it's used heavily in `ParticlesToLevelSet`.

openvdb/tools/PointPartitioner.h
Idclip: Okay great, thanks - I haven't verified the inverse calculation here but your explanation sounds good.

danrbailey:
> @danrbailey Sorry it's taken me so long to get back to you on this one.
> 
> > I don't believe this is true, perhaps you can elaborate on what you see to be the problem here?
> 
> Isn't the idea behind this partitioner that points are partitioned spatially? If you're going to subsequently modify the positions it suggests to me that perhaps this isn't the best use of this class. As you've pointed out though, nothing like that is being performed here.

Ah, perhaps this is the source of the confusion here. 

Imagine an array of 100 points - (1) you could iterate over all the point positions and apply a modification (say +1 unit in Y) and _then_ spatially partition or (2) you could spatially partition based on the original point positions with the modification applied (+1 unit in Y). In (1), you would read the point positions twice, in (2) you would only read them once. For cases where reading is cheap (like world-space linear arrays), this doesn't make much sense why you would need this, but for cases where reading is (more) expensive, it's worthwhile.

Idclip:
Okay thanks @drb, I assumed the modification that you were talking about would be done after partitioning. @kmuseth @jmlait any further thoughts here? Otherwise this looks good to me

openvdb/tools/PointPartitioner.h
kmuseth: @danrbailey Given that you found the terminology confusing I suggest you do one of two - either document it (the easiest) or attempt to fix it (I assume this is a tall order). Anyway, confusing terminology in code is a really bad thing and should be addressed in some way (it makes the code hard to maintain).

danrbailey: Thanks @kmuseth! I will merge this PR as-is and then follow up with a new PR that refactors portions of this algorithm to fix the terminology as well as introduce some C++11 improvements that together will hopefully improve maintainability.

PR #590, opened 2019-12-12
OVDB-136: Attribute Default Value Metadata

danrbailey:
AttributeArray default value metadata has never been properly supported, it's mainly been used as a means to ensure 1-to-1 consistency with data coming from Houdini. This change resolves this so that the default value is applied from the metadata when appending attributes and during construction and copy-construction of the AttributeSet.

Note that this also requires a bump to the version from 7.0.1 to 7.1.0 as it represents an API change.


richhones:
General comment : Is there a need to change use of Metadata::Ptr to Metadata*? Because this means a bunch of deprecations and it doesn't match the API of the functions in PointAttribute which I imagine is suboptimal.

danrbailey:
> General comment : Is there a need to change use of Metadata::Ptr to Metadata*? Because this means a bunch of deprecations and it doesn't match the API of the functions in PointAttribute which I imagine is suboptimal.

I think using a raw pointer here is preferred because it's clearer that the method isn't storing the metadata whereas a shared pointer can be a little more ambiguous (not to mention avoiding the extra cost of copying the shared pointer). A const reference would be the best option, but as the metadata can be optional, a const raw pointer seems like the next best thing.

It's not ideal that we have to do a few deprecations. You're right though, I think I should make the same change in PointAttribute so that everything is consistent. I'll update the PR.

openvdb/points/AttributeArray.h
```diff
@@ -1266,9 +1274,14 @@ TypedAttributeArray<ValueType_, Codec_>::unregisterType()
...
-TypedAttributeArray<ValueType_, Codec_>::create(Index n, Index stride, bool constantStride)
+TypedAttributeArray<ValueType_, Codec_>::create(Index n, Index stride, bool constantStride,
+    const Metadata* metadata)
 {
-    return Ptr(new TypedAttributeArray(n, stride, constantStride));
+    const TypedMetadata<ValueType>* typedMetadata = metadata ?
+        dynamic_cast<const TypedMetadata<ValueType>*>(metadata) : nullptr;
+
+    return Ptr(new TypedAttributeArray(n, stride, constantStride,
+        typedMetadata ? typedMetadata->value() : zeroVal<ValueType>()));
```
Idclip: Perhaps two ::create methods on the derive types should exist so you can avoid this dynamic_cast. For example, `appendAttribute` already knows the type of the attribute, so the metadata shouldn't be dynamically cast for every leaf. Any signature which is templated on the attr type should enforce it on the metadata as well, i.e:

```
template <typename ValueType,
          typename CodecType = NullCodec,
          typename PointDataTreeT = PointDataTree>
inline void appendAttribute(PointDataTreeT& tree,
                            const std::string& name,
                            const ValueType& uniformValue =
                                point_attribute_internal::Default<ValueType>::value(),
                            const Index strideOrTotalSize = 1,
                            const bool constantStride = true,
                            const TypedMetadata<ValueType>* defaultValue = nullptr,
                            const bool hidden = false,
                            const bool transient = false);
```

Idclip: I guess through the current factory methods this isn't possible more the higher level API functions which don't runtime branch. Really what would be great is templated methods at every stage of the hierarchy. i.e. appendAttribute<ValueType> on the AttributeSet which can call the derived AttributeArray constructor

danrbailey: I'll make the change to enforce the derived Metadata type in this value-typed `appendAttribute` (this is only currently used in `PointUtils.cc`). However, the ripple-out effect of enforcing this at all levels just to avoid the dynamic_cast is quite large (and potentially not even possible without some runtime branching), so I don't think it's worth addressing that at this stage.

Idclip: You shouldn't need any runtime branching through the templated methods. You only need templated method on AttributeSet::appendAttribute and potentially on the LeafNode::appendAttribute too, as well as a new constructor on the TypedAttributeArray. The problem currently is that the templated methods generate the value type strings and then call the non-templated methods - which makes them a bit redundant (performance wise) bar their ease of use. They shouldn't ever need to go to their "registered" string type and instead resolve all calls at compile time. Were it the other way around (with the bulk of the logic living in the templated methods and providing the base lib support via runtime branching for the non-templated ones) you'd get the optimal behaviour at the cost of limiting support through the NamePair API.

Idclip: I'm not suggesting this be done now btw

danrbailey: In theory yes, but the types are still stored as strings in the `Descriptor `which you'll almost always want to be modifying too. Aside from the dynamic_cast, the only thing you would really save is the map lookup of the functor that creates the `AttributeArray `(https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/points/AttributeArray.cc#L104). By far the bigger performance issue was the locking of the mutex in accessing this map (which we've now resolved), so I'd say it's a pretty minor issue to address.

I believe what you're suggesting would need to be an ABI=8 anyway as it requires adding a new constructor to the `TypedAttributeArray`.

(As an aside, I think we could probably get a bigger performance win by making the `std::map` of the AttributeFactoryMap into a `std::unordered_map`.)

danrbailey: @Idclip - Addressed your feedback regarding the function signature of appendAttribute() including the derived metadata type.

openvdb/points/PointAttribute.h
```diff
@@ -531,6 +532,45 @@ inline void compactAttributes(PointDataTreeT& tree)
...
+                const ValueType& uniformValue,
+                const Index strideOrTotalSize,
+                const bool constantStride,
+                Metadata::Ptr metaDefaultValue,
+                const bool hidden = false,
+                const bool transient = false)
+{
+    // default metadata value must now be provided as a raw pointer
+    appendAttribute<ValueType, CodecType>(tree, name, uniformValue, strideOrTotalSize, constantStride,
+        metaDefaultValue.get(), hidden, transient);
```
Idclip: Does this work? Do you now not get compiler errors without an additional dynamic cast on the raw pointer here?

danrbailey: Good catch, we obviously don't unit test the deprecated methods. Fixed this now.

PR #592, opened 2019-12-13
Add inlines and fix MSVC clang paths.

jmlait:
This is a reference commit to keep the Houdini & VDB versions in sync.

Primary change is a sprinkling of force inline.  To ensure abi compatibility it is important these functions are always inlined so as to use the caller's version of the VDB grid.

### DIFF
--- a/openvdb_houdini/GEO_PrimVDB.h
+++ b/openvdb_houdini/GEO_PrimVDB.h
@@ -143,9 +143,11 @@ public:
                             bool normalize = false) const;
 
     /// Get the storage type of the grid
+    SYS_FORCE_INLINE
     UT_VDBType          getStorageType() const
                                 { return myGridAccessor.getStorageType(); }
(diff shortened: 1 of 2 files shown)


openvdb_houdini/GEO_PrimVDB.h
```diff
@@ -712,7 +736,7 @@ callTypedGrid(GEO_PrimVDB& prim, OpType& op)
 {
     prim.makeGridUnique();
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
```
danrbailey: @e4lam mentioned that this conditional logic might not be needed any more now that our minimum is Visual Studio 17. I haven't investigated this yet though, do you have any further insights on this?

jmlait: Good point, I'll pass it on to him.

PR #593, opened 2019-12-13
OVDB-136: Various VDB Point Changes

danrbailey:
* Adding `GroupWriteHandle::setUnsafe()` for faster performance when group is in-core and non-uniform
* Deleted some point methods that have been deprecated for more than a year now
* Added UniquePtr aliases to many of the convenience classes
* Tidied up some documentation in AttributeArray to remove mentions of in-memory compression


Idclip:
@danrbailey I've noticed through syncing VDB here that these deprecated methods that were removed do not appear in the change logs. Can you add these? Perhaps it doesn't have to be explicit, but it should definitely be mentioned

danrbailey:
> @danrbailey I've noticed through syncing VDB here that these deprecated methods that were removed do not appear in the change logs. Can you add these? Perhaps it doesn't have to be explicit, but it should definitely be mentioned

Sure, I've added this on to #590

PR #594, opened 2019-12-14
Verbify Platonic SOP

danrbailey:


danrbailey:
> That looked suspiciously easy :>

Indeed! Four left - Create, Rasterize Points, Read, Write. I think only Rasterize Points will be difficult from that list.

PR #595, opened 2019-12-17
OVDB-136: Point Group Inspection

danrbailey:
Move group inspection functionality from point_group_internal::GroupInfo up into the AttributeSet and Descriptor so that it becomes part of the public API and can be used outside of the point group methods.


openvdb/points/AttributeSet.h
```diff
@@ -375,7 +381,8 @@ class OPENVDB_API AttributeSet::Descriptor
     /// Return @c true if group exists
     bool hasGroup(const Name& group) const;
     /// Define a group name to offset mapping
-    void setGroup(const Name& group, const size_t offset);
+    void setGroup(const Name& group, const size_t offset,
+        const bool checkValidOffset = false);
```
Idclip: `@param` for these arguments

openvdb/points/AttributeSet.h
```diff
@@ -402,6 +409,27 @@ class OPENVDB_API AttributeSet::Descriptor
...
+
+    /// Return @c true if there are sufficient empty slots to allow compacting
+    bool canCompactGroups() const;
+
+    /// Return the next empty group slot
+    size_t nextUnusedGroupOffset() const;
+
+    /// Determine if a move is required to efficiently compact the data and store the
+    /// source name, offset and the target offset in the input parameters
+    bool requiresGroupMove(Name& sourceName, size_t& sourceOffset, size_t& targetOffset) const;
```
Idclip: same here

openvdb/points/PointGroup.h
```diff
@@ -641,14 +521,14 @@ inline void compactGroups(PointDataTree& tree)
         LeafManagerT leafManager(tree);
         tbb::parallel_for(leafManager.leafRange(), copy);
 
-        descriptor->setGroup(sourceName, targetOffset);
+        descriptor->setGroup(sourceName, targetOffset, /*checkValidOffset=*/true);
```
Idclip: Why does `checkValidOffset` need to be true here - doesn't this algorithm guarantee this?

openvdb/points/PointGroup.h
```diff
@@ -641,14 +521,14 @@ inline void compactGroups(PointDataTree& tree)
...
     }
 
     // drop unused attribute arrays
 
-    std::vector<size_t> indices = groupInfo.populateGroupIndices();
+    std::vector<size_t> indices = attributeSet.groupAttributeIndices();
```
Idclip: can be const

Idclip:
Minor comments, main one about the use of `checkValidOffset` otherwise lgtm

openvdb/points/PointGroup.h
danrbailey: I believe it to be just an abundance of caution. I've addressed all feedback now.

danrbailey:
Thanks a lot for reviewing @Idclip!

PR #596, opened 2019-12-17
OVDB-136: Introduce StringMetaCache

danrbailey:
Currently all string->index logic for string attribute metadata is stored in source files and unnamed namespaces only. This PR introduces a StringMetaCache that exposes this as part of the public API and consolidates all existing logic to use this new API.

Performance wasn't a focus of these changes, but there are some minor optimizations included (around 5% -> 20%).


openvdb/points/AttributeArrayString.h
```diff
@@ -25,20 +25,20 @@ namespace points {
 ////////////////////////////////////////
 
 
-using StringIndexType = uint32_t;
+using StringIndexType = Index;
```
Idclip: Why change this then remove its usage below and in the .cc?

openvdb/points/AttributeArrayString.h
```diff
@@ -49,27 +49,68 @@ struct StringCodec
...
-    void insert(const Name& name);
+    /// Insert the string into the metadata, hint is used if non-zero
+    /// Returns the chosen index which will match hint if not in use
+    Index insert(const Name& name, Index hint = Index(0));
 
     /// Reset the cache from the metadata
     void resetCache();
 
 private:
+    using IndexPairArray = std::deque<std::pair<Index, Index>>;
```
Idclip: why deque? I'd imagine in the general case a vector is probably more efficient? Unless your indices vary widely for a large number of points

openvdb/points/AttributeArrayString.h
```diff
@@ -185,10 +226,8 @@ class OPENVDB_API StringAttributeWriteHandle : public StringAttributeHandle
     /// @note throws if name does not exist in cache
     Index getIndex(const Name& name) const;
 
-    using ValueMap = std::unordered_map<std::string, Index>;
-
-    ValueMap                                                    mCache;
-    AttributeWriteHandle<StringIndexType, StringCodec<false>>   mWriteHandle;
+    StringMetaCache                                             mCache;
+    AttributeWriteHandle<Index, StringCodec<false>>   mWriteHandle;
```
Idclip: some weird w/s with these members and the above

openvdb/points/AttributeArrayString.h
```diff
@@ -49,27 +49,68 @@ struct StringCodec
...
+    /// Returns @c true if key exists
+    bool hasKey(const Name& key) const;
+    /// Returns @c true if index exists
+    bool hasIndex(Index index) const;
 
-    /// Insert the string into the metadata
-    void insert(const Name& name);
+    /// Insert the string into the metadata, hint is used if non-zero
+    /// Returns the chosen index which will match hint if not in use
+    Index insert(const Name& name, Index hint = Index(0));
```
Idclip: I think we should try to get in the habbit of proper doxygen with any function changes

openvdb/points/AttributeArrayString.h
Idclip: Oh nvm I forgot my own algorithm... I agree with deque

openvdb/points/AttributeArrayString.h
Idclip: And perhaps a bit more of a description about the intended usage of hint and how it's expected to map the metadata name index

openvdb/points/AttributeArrayString.cc
```diff
@@ -20,86 +20,144 @@ namespace points {
...
+
+    Name hintKey;
+    bool canUseHint = false;
+
+    // hint must be non-zero to have been requested
+
+    if (hint > Index(0)) {
+        hintKey = getStringKey(hint);
+        // check if hint is already in use
+        if (!bool(mMetadata[hintKey])) {
```
Idclip: As this requires a map lookup is the usage of hint really more efficient?

openvdb/points/AttributeArrayString.h
danrbailey: I thought the code was simpler using the Index typedef directly as I'm not sure we will ever change this. I didn't want to remove it directly, but I realize I can mark this deprecated:

`using StringIndexType OPENVDB_DEPRECATED = Index;`

?

openvdb/points/AttributeArrayString.cc
danrbailey: It's not so much for efficiency here, it's more that you can insert using non-sequential indices so that you don't need to move any AttributeArray data if you can align the indices which _is_ much more efficient. You can do it without this method by creating a temporary data structure yourself, but it's a convenience to be able to hint at a preferred index.

openvdb/points/AttributeArrayString.h
danrbailey: I'll fix this.

openvdb/points/AttributeArrayString.h
danrbailey: I'll add a bit more description about the hint. I'm not going to fix all the doxygen for this class though as it's not one that I expect users to interact with directly. I don't mind if you want to make a "good first issue" in our GitHub issues for fixing some of these up though.

Idclip: I don't agree with that mentality. New users should not be writing doxygen, especially at this level - that's our job. We should get in the habit of writing good documentation whenever we introduce a new method as we're most likely best situated to explain it. And in regards to "not one that I expect users to interact with directly" - I'm a user and it wasn't clear to me without looking at the implementation! I wasn't suggesting you you update this entire class now, but IMO we should keep on top of doxygen with new methods.

danrbailey: Sorry, that's not what I meant. If the comments in the code are not sufficiently clear, that's totally valid (and appreciated) feedback. I just mean that I don't agree with a policy that every piece of documentation should be formatted for Doxygen. For important public API, sure. However, parameter names still show up in Doxygen, so provided that the description is clear and unambiguous, I don't think it should be a strict requirement to ensure correct use of `@param` for instance as I don't think that adds value to the project.

openvdb/points/AttributeArrayString.h
danrbailey: Edit - Turns out this doesn't work on Windows, so had to remove the deprecation there.

openvdb/points/AttributeArrayString.h
Idclip: Okay that makes sense to me now

openvdb/points/AttributeArrayString.cc
Idclip: Right, so as i understand it, you shouldn't then need to run the loop which checks for the next valid index? i.e., instead of

```
    IterT iter = mIdBlocks.begin();
    for (; iter != mIdBlocks.end(); ++iter) {
        const Index start = iter->first;
        const Index end = start + iter->second;

        if (index < start || index >= end) break;
        if (!canUseHint)    index = end;
    }
```
it's
```
if (!canUseHint) {
    IterT iter = mIdBlocks.begin();
    for (; iter != mIdBlocks.end(); ++iter) {
        const Index start = iter->first;
        const Index end = start + iter->second;

        if (index < start || index >= end) break;
        index = end;
    }
}
```

PR #605, opened 2020-01-09
OVDB-126: Add Grid::isTreeUnique

danrbailey:


jmlait:
Reading code I much prefer a isUnique() to use_count() == 1.  It makes the purpose of the test clear without having to think as much.

danrbailey:
> Is this method really necessary? Why not simply have
> "grid.treePtr().use_count() == 1" in the client code?

I think so. Your logic is wrong and would always return false. This is because treePtr() returns a _new_ shared ptr, so the use count is expected to be **two** if it's unique (one contained in the grid and the new one that has now been returned). 

That aside, I agree with Jeff, I think the dedicated function improves readability...

PR #616, opened 2020-01-23
Fixes to the Houdini source files bundled with OpenVDB

Idclip:

### DIFF
--- a/openvdb_houdini/CMakeLists.txt
+++ b/openvdb_houdini/CMakeLists.txt
@@ -47,9 +47,6 @@ houdini_get_default_install_dir():
   Mac: $ENV{HOME}/Library/Preferences/houdini/X.X
   Windows: $ENV{HOMEDRIVE}$ENV{HOMEPATH}\Documents\houdiniX.X
 If OFF, CMAKE_INSTALL_PREFIX is used.]=] ON)
-option(USE_SESI_OPENVDB_PRIM [=[
-Use the bundled Houdini source files to build the Houdini GU Primitive
-type. This is typically only for developer use.]=] OFF)
 
 set(_OPENVDB_HOUDINI_OPHIDE_POLICY_OPTIONS none aswf native)
 if(NOT OPENVDB_HOUDINI_OPHIDE_POLICY)
@@ -75,8 +72,6 @@ if(OPENVDB_HOUDINI_ENABLE_SESI_TAG)
   message(DEPRECATION "The OPENVDB_HOUDINI_ENABLE_SESI_TAG option is deprecated "
     "and will be removed.")
 endif()
-
-mark_as_advanced(USE_SESI_OPENVDB_PRIM)
 
 #########################################################################
 
@@ -187,11 +182,6 @@ target_link_libraries(openvdb_houdini PUBLIC
 )
 
 target_compile_definitions(openvdb_houdini PRIVATE "-DOPENVDB_HOUDINI_PRIVATE")
-
-if(USE_SESI_OPENVDB_PRIM)
-  message(STATUS "Configuring OpenVDB Houdini to use local SESI headers.")
-  target_compile_definitions(openvdb_houdini PUBLIC -DSESI_OPENVDB_PRIM)
-endif()
 
 if(OPENVDB_ENABLE_RPATH)
   # @todo There is probably a better way to do this for imported targets
--- a/openvdb_houdini/CMakeLists.txt
+++ b/openvdb_houdini/CMakeLists.txt
@@ -47,6 +47,9 @@ houdini_get_default_install_dir():
   Mac: $ENV{HOME}/Library/Preferences/houdini/X.X
   Windows: $ENV{HOMEDRIVE}$ENV{HOMEPATH}\Documents\houdiniX.X
 If OFF, CMAKE_INSTALL_PREFIX is used.]=] ON)
+option(USE_SESI_OPENVDB_PRIM [=[
+Use the bundled Houdini source files to build the Houdini GU Primitive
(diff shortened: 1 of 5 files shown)


Idclip:
@e4lam tagging for visibility

openvdb_houdini/GU_PrimVDB.cc
```diff
@@ -484,21 +484,37 @@ GU_PrimVDB::expandBorderFromPrimVolume(const GEO_PrimVolume &vol, int pad)
...
+                    *pprims = new GU_PrimVDB(static_cast<GU_Detail *>(&gdp), primoff);
+            });
+    }
+    else
+    {
+        // Allocate them serially if we're only allocating a few.
+        GA_Offset endprimoff(start_offset + nprimitives);
+        for (GA_Offset primoff(start_offset); primoff != endprimoff; ++primoff, ++new_prims)
+            *new_prims =  new GU_PrimVDB(static_cast<GU_Detail *>(&gdp), primoff);
+    }
```
Idclip: Copied from https://www.sidefx.com/docs/hdk/tetprim_2_g_e_o__prim_tetra_8_c-example.html#a116

openvdb_houdini/CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ houdini_get_default_install_dir():
   Mac: $ENV{HOME}/Library/Preferences/houdini/X.X
   Windows: $ENV{HOMEDRIVE}$ENV{HOMEPATH}\Documents\houdiniX.X
 If OFF, CMAKE_INSTALL_PREFIX is used.]=] ON)
+option(USE_SESI_OPENVDB_PRIM [=[
+Use the bundled Houdini source files to build the Houdini GU Primitive
+type. This is typically only for developer use.]=] OFF)
```
danrbailey: I don't think we should expose this at all in the CMake. This is such a rare use case and always risks being out-of-date with the Houdini version you're building against - in fact any time I have tried to build this, it's been out-of-date. As far as I'm concerned the effort of making sure that the GEO/GU_PrimVDB classes are working correctly falls entirely to SideFX and should not be an option in our build system that a user can accidentally switch on when trying to figure out how to build the open-source Houdini plugin.

The solution we went for with the mako scripts was to add a define macro that was commented out for this type of use case:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb_houdini/SOP_NodeVDB.cc#L58

I think we should follow convention there.

Idclip: It then becomes tricky to actually enable this - you either need to provide the define to the subsequent command (make) or change the cmake yourself. As per the current Windows issues with the openvdb houdini plugin, we actually have a valid use case for a user to potentially enable these. IMO we should either support them or we should remove them.

On another note, this exactly what cmake advanced commands are for. These should support the developers as well as the users. I would argue the mako define should also be a cmake define.

danrbailey: I don't think it should be necessary to touch the build system or the build commands to actually enable this though. You should just be able to uncomment a line like:

`// #define SESI_OPENVDB_PRIM`

With some very minor header reordering, I believe it should be possible to put that macro in just a single file (UT_VDBUtils.h) and pick it up from all the others.

I've been keeping up with the recent VDB Windows discussions and I don't agree this is a valid use case as is evident by the fact it was out-of-date when you tried to use it. The other aspect to consider is that the Houdini version doesn't have guards like `#if (UT_VERSION_INT > xx)` so we'd have to be responsible for maintaining compatibility across multiple versions of Houdini which we don't want to be doing. The right approach here is to report an issue to SideFX and then pick up fixes in the native VDB primitive.

All that being said, we had the same discussion with the mako define and it was ultimately decided that the code itself served a useful purpose purely as a reference implementation. Without this I would have little idea of what the code looks like in GEO_PrimVDB.cc / GU_PrimVDB.cc and sometimes that's useful.

Idclip: > I don't think it should be necessary to touch the build system or the build commands to actually enable this though. You should just be able to uncomment a line 

I don't think you should ever have to make any changed to the codebase to enable/disable functionality. You should be able to control all behaviour through the supported build system - that's what it's there for... I don't like floating defines which aren't represented, it's very messy and hard to keep track of. This also goes for the profile options in the unit tests and any other define that may exist in VDB currently. 

> The right approach here is to report an issue to SideFX and then pick up fixes in the native VDB primitive.

Except this isn't going to happen if we don't regularly support and build these files. The "sometimes that's useful" bit doesn't work if these files are not kept up to date until the moment we need them. 

If we just want the mako and GU files as a reference, they shouldn't be in the codebase and should instead exist purely as documentation. It can be very confusing seeing the Houdini headers as part of repository. 

I will, however, make this change for now so that at least this work is closed off - if we can't resolve here, lets discuss in TSC.

danrbailey: I definitely see where you're coming from - it does feel like pros and cons for each approach so think it would be good to discuss in the TSC at some point.

jmlait: I'm not sure where they'd go in "documentation"....  If they aren't in the code base, it would be even harder to keep them in sync than it is now.

Maybe we should have a subdirectory to openvdb_houdini for reference code that is provided but is not expected to be compiled?

Idclip: A sub-directory would be an improvement, but I was speculating we could go even further and move them into the doc/ folder or maybe even have them embedded into some of the doxygen files so that it removes all doubt what their intention is. This obviously assumes we all decide on not supporting them for local builds going forward.

PR #623, opened 2020-01-28
VDB Merge SOP UI Prototype

danrbailey:
DRAFT PR - not intended to be merged at this stage.

First attempt to collaborate on designing a SOP UI through code review. This new SOP doesn't depend on the OpenVDB core library or Houdini library so can be built independently (it uses an embedded ds file though I'm not proposing we use this in the final version). No implementation at this stage.

Specifically, I'm looking for any feedback on:

* parameter names and values
* UI layout
* default values
* disablewhen functionality

Not sharing any help documentation in this phase. It's all pretty self-explanatory except for the collation parameter so I'll briefly elaborate with a set of example input VDBs:

1) float "distance" SDF VDB
2) double "distance" SDF VDB
3) float "phi" SDF VDB

Would result in the following:

```
* Grid Name, Grid Class and Value Type => (1), (2), (3)     // no VDBs merged
* Grid Name and Grid Class => (1) + (2), (3)                // (2) down-sampled to float
* Grid Class and Value Type => (1) + (3), (2)               // "phi" + "distance" merged
```

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_Merge.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Merge.cc
@@ -1,0 +1,296 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+
+/// @file SOP_OpenVDB_Merge.cc
+///
+/// @author Dan Bailey
+///
+/// @brief Merge OpenVDB grids.
+
+#include "SOP_OpenVDB_Merge.proto.h"
+
+#include <PRM/PRM_TemplateBuilder.h>
+#include <UT/UT_DSOVersion.h>
+
+
+class SOP_OpenVDB_Merge : public SOP_Node
+{
+public:
+    static PRM_Template *buildTemplates();
+    static OP_Node *myConstructor(OP_Network *net, const char *name, OP_Operator *op)
+    {
+        return new SOP_OpenVDB_Merge(net, name, op);
+    }
+
+    static const UT_StringHolder theSOPTypeName;
+
+    virtual const SOP_NodeVerb *cookVerb() const final;
+
+protected:
+    SOP_OpenVDB_Merge(OP_Network *net, const char *name, OP_Operator *op)
+        : SOP_Node(net, name, op) { }
+
+    virtual ~SOP_OpenVDB_Merge() {}
+
+    virtual OP_ERROR cookMySop(OP_Context &context) final
+    {
+        return cookMyselfAsVerb(context);
+    }
+};
+
+
+const UT_StringHolder SOP_OpenVDB_Merge::theSOPTypeName("DW_OpenVDBMerge");
+
+
+void
+newSopOperator(OP_OperatorTable *table)
+{
+    table->addOperator(new OP_Operator(
+        SOP_OpenVDB_Merge::theSOPTypeName,
+        "VDB Merge",
+        SOP_OpenVDB_Merge::myConstructor,
+        SOP_OpenVDB_Merge::buildTemplates(),
+        1,
+        OP_MULTI_INPUT_MAX,
+        nullptr,
+        0));
+}
+
+
+static const char *theDsFile = R"THEDSFILE(
+{
+    name        parameters
+
+    inputlabel  1   "VDBs To Merge"
+
+    parm {
+        name    "group"
+        label   "Group"
+        type    string
+        default { "" }
+    }
+    parm {
+        name    "vdbpointsgroup"
+        label   "VDB Points Group"
+        type    string
+        default { "" }
+    }
+    parm {
+        name    "collation"
+        label   "Collation"
+        type    ordinal
+        default { "nameclassandtype" }
+        menu {
+            "nameclassandtype"  "Grid Name, Grid Class and Value Type"
+            "nameandclass"      "Grid Name and Grid Class"
+            "classandtype"      "Grid Class and Value Type"
+        }
+    }
+    parm {
+        name    "usereferencevdb"
+        label   "Use First Grid as Reference VDB Only"
+        type    toggle
+        default { "0" }
+    }
+    parm {
+        name    "resample"
+        label   "Resample Grids"
+        type    ordinal
+        default { "first" }
+        menu {
+            "first"         "To Match First Grid"
+            "highestres"    "To Match Highest-Res Grid"
+            "lowestres"     "To Match Lowest-Res Grid"
+        }
+        joinnext
+    }
+    parm {
+        name    "resampleinterp"
+        label   "Interpolation"
+        type    ordinal
+        default { "linear" }
+        menu {
+            "point"     "Nearest"
+            "linear"    "Linear"
+            "quadratic" "Quadratic"
+        }
+    }
+    groupsimple {
+        name    "mergemodegroup"
+        label   "Merge Mode"
+
+        parm {
+            name    "fogmode"
+            label   "Fog Mode"
+            type    ordinal
+            default { "add" }
+            menu {
+                "none"          "None"
+                "add"           "Add"
+                "multiply"      "Multiply"
+                "maximum"       "Maximum"
+                "minimum"       "Minimum"
+                "topounion"     "Activity Union"
+                "topointersect" "Activity Intersection"
+            }
+            joinnext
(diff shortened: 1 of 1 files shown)


danrbailey:
vdbmergesopui

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+    parm {
+        name    "collation"
+        label   "Collation"
+        type    ordinal
+        default { "nameclassandtype" }
+        menu {
+            "nameclassandtype"  "Grid Name, Grid Class and Value Type"
+            "nameandclass"      "Grid Name and Grid Class"
+            "classandtype"      "Grid Class and Value Type"
+        }
```
jmlait: Should use name attribute, not the intrinsic vdb name (which naturally should be the same :>)  Thus "grid" should be removed.

Grid is usually called VDB.

Thus
"Name, VDB Class and Value Type"
etc

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+    }
+    parm {
+        name    "usereferencevdb"
+        label   "Use First Grid as Reference VDB Only"
```
jmlait: First VDB Is Only for Reference
(The left hand is only for support!)
Grid -> VDB
Imply the Use with the toggle.

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+        default { "linear" }
+        menu {
+            "point"     "Nearest"
+            "linear"    "Linear"
+            "quadratic" "Quadratic"
+        }
+    }
+    groupsimple {
+        name    "mergemodegroup"
+        label   "Merge Mode"
```
jmlait: This isn't a mode, but an operation done to merge.  So I think
Merge Operation
is better for this group

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+
+        parm {
+            name    "fogmode"
```
jmlait: op_fog
The time to use underscores is when you have a bunch of parameters logically grouped with a common prefix.  Here we have a bunch of operation parms grouped in the Merge Operation group, so we can prefix them all with op_

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+            label   "Fog Mode"
```
jmlait: Fog VDBs
Mode is redundant in old system as we are in the group called Merge Mode.
There isn't a need to repeat operation for the same reason.
But as a naive person, I wouldn't know what Fog and Scalar are referring to, they are referring to types of VDBs.  Thus I suggest we clarify that for all of these parms.

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+                "maximum"       "Maximum"
+                "minimum"       "Minimum"
+                "topounion"     "Activity Union"
+                "topointersect" "Activity Intersection"
+            }
+        }
+        parm {
+            name    "distancemode"
+            label   "Distance Mode"
+            type    ordinal
```
jmlait: op_sdf
SDF VDBs
Levelset VDBs tend to be called SDF VDBs.  Yeah, random TLAs, but it is consistent :>

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+                "none"          "None"
+                "add"           "Add"
+                "multiply"      "Multiply"
+                "topounion"     "Activity Union"
+                "topointersect" "Activity Intersection"
+            }
+        }
+        parm {
+            name    "boolmode"
+            label   "Bool Mode"
```
jmlait: I prefer Boolean VDBs
op_bool vs op_boolean
should probably be determined by what the VDB class is referred to as a token.

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+            menu {
+                "none"          "None"
+                "topounion"     "Activity Union"
+                "topointersect" "Activity Intersection"
+            }
+            joinnext
+        }
+        parm {
+            name    "pointsmode"
+            label   "Points Mode"
```
jmlait: VDB Points
This is the one that breaks the Foo VDBs pattern as I think we call these VDB Points everywhere.

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+            menu {
+                "none"          "None"
+                "pointmerge"    "Merge"
+                "topounion"     "Activity Union"
+                "topointersect" "Activity Intersection"
+            }
+        }
+        parm {
+            name        "useallgridsfortopo"
+            label       "Use all Grids for Activity Modes"
```
jmlait: Should be
Use All VDBs for Activity Operations
Note that All is capitalized in Title Case.  Grids -> VDBs and Modes -> Operations.
I have no idea what this does, though, so I think it needs better text :>

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+            type        toggle
+            default     { "0" }
+            disablewhen "{ fogmode != topounion fogmode != topointersect scalarmode != topounion scalarmode != topointersect distancemode != topounion distancemode != topointersect vectormode != topounion vectormode != topointersect boolmode != topounion boolmode != topointersect pointsmode != topounion pointsmode != topointersect }"
```
jmlait: Such complicated disablewhens scare me....  I'm not going to know how to activate this!

It sounds almost, if I guess to what this does, you instead want a default Activity operation to be provided or something?

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+    groupsimple {
+        name    "postprocessgroup"
+        label   "Post Process"
```
jmlait: Post-Process

openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,296 @@
...
+        type    toggle
+        default { "0" }
+    }
+    parm {
+        name    "resample"
+        label   "Resample Grids"
+        type    ordinal
+        default { "first" }
+        menu {
+            "first"         "To Match First Grid"
```
jmlait: Grid -> VDB

openvdb_houdini/SOP_OpenVDB_Merge.cc
danrbailey: The "collation" modes groups the grids you can merge (type and class, etc). That's mainly because there's no obvious or useful way to value merge a vector grid with a level set grid for instance. However, for topology modes, it's valid to be able to "activity union" a set of grids of all different types and classes, so that feels like a natural extension to want to include.

The alternative workflow here is to create a temporary bool grid, use the VDB Activate SOP to activity union all input grids, then use another VDB Activate SOP to activity union all (or specific) input grids against this temporary grid, which is somewhat convoluted.

The more I think about this, the more I think perhaps the most intuitive way of offering this could be to add additional drop-down options:

* Activity Union
* Activity Intersect
* Activity Union (All Grids)
* Activity Intersect (All Grids)

danrbailey:
> Thank you for providing this. Hopefully these comments make sense!

Great feedback, exactly what I was after, thank you! I'll work on updating with a revised version...

Idclip:
@danrbailey I realise that #1031 only implemented part of this UI but can this PR now be closed? We can continue to discuss it as a remote branch when work is actioned

PR #626, opened 2020-01-30
OVDB-118: Add a Houdini ABI Test

danrbailey:
I had a go at this - it was a bit trickier to achieve than I expected but I have it working.

This adds a new openvdb_houdini/abitest directory which builds two source objects where each source objects pulls the OpenVDB headers from a different place - one from the local repo and one from the Houdini-shipped version of OpenVDB and then the resulting binary links against libopenvdb and libopenvdb_sesi. 

This test mirrors what happens in the VDB Primitive code in Houdini by passing around void pointers and static casting them. I tested this by reverting this commit (https://github.com/AcademySoftwareFoundation/openvdb/commit/fc3ccebd7d0cbf4d3c7f642933407de97076e91f) and I reliably get a seg fault or bus error, where it otherwise succeeds.

This is not comprehensively testing everything, but it's a starting point. I also have to update the change logs and the build documentation, so just want to get some thoughts on the approach and I'll complete this.


openvdb_houdini/abitest/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+# This binary is composed of two object files:
+# - TestABI uses headers from this version of OpenVDB
+# - main uses headers from the version of OpenVDB shipped with Houdini
+
+# Disable some compiler warnings in TBB and Half
+
+add_compile_options(
+  -Wno-conversion
+  -Wno-shorten-64-to-32
+)
```
Idclip: Hmm you shouldn't be getting these, cmake should treat these projects as system headers?

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -299,6 +299,25 @@ if(NOT ILMBASE_LIBRARYDIR)
...
+    ${_FIND_HOUDINI_ADDITIONAL_OPTIONS}
+    PATHS ${_HOUDINI_LIB_DIR}
+  )
+  if(NOT EXISTS ${OPENVDB_SESI_LIB})
+    message(WARNING "The OpenVDB Houdini CMake setup is unable to locate libopenvdb_sesi "
+      "within the Houdini installation at: ${_HOUDINI_LIB_DIR}. "
+      "This is needed for building the ABI tests."
+    )
+  endif()
+endif()
```
Idclip: This should probably be specific for the ABI tests? Otherwise I can see confusion with the value of `OPENVDB_INCLUDEDIR`. Perhaps just allow the values of `_HOUDINI_INCLUDE_DIR` and `_HOUDINI_LIB_DIR` to persist once `OpenVDBHoudiniSetup` has been called?

Also, I'd try initialising the openvdb_sesi library with `_houdini_create_libraries`

openvdb_houdini/abitest/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+
+# Create an object library that uses the Houdini OpenVDB headers
+
+add_library(TestABI OBJECT TestABI.cc)
+target_include_directories(TestABI PUBLIC ${OPENVDB_INCLUDEDIR})
```
Idclip: as sesis symbols are namespaced, couldn't we instead use them directly without this initial step? For example, to query the abi string:
```
std::string houabi = openvdb::v5_2_sesi::getLibraryAbiVersionString();
std::string abi = openvdb::getLibraryAbiVersionString();
```

etc etc

cmake/OpenVDBHoudiniSetup.cmake
danrbailey: Yeah, that's fair, I can lift this into the CMake for the abi test itself. I'm a bit confused about your comment regarding `_houdini_create_libraries` though - doesn't this get called by `OpenVDBHoudiniSetup.cmake`?

openvdb_houdini/abitest/CMakeLists.txt
danrbailey: I'll try and dig into what's going on here a bit.

openvdb_houdini/abitest/CMakeLists.txt
danrbailey: Right now there's no knowledge about which version of OpenVDB is being used for the ABI and Main layers. What you're suggesting sounds like you'd need to parse and extract the namespace from version.h that lives within `OPENVDB_INCLUDEDIR` to pass into the compilation? Wouldn't that be more complicated or am I misunderstanding you here?

cmake/OpenVDBHoudiniSetup.cmake
danrbailey: I've moved the logic regarding the openvdb include and libraries into the CMake for the ABI test now.

openvdb_houdini/abitest/CMakeLists.txt
danrbailey: The `target_include_directories` was missing the SYSTEM keyword. Fixed.

openvdb_houdini/abitest/CMakeLists.txt
Idclip: I'm suggesting you have a sing main and a single source which gets duplicated and is compiled once with our headers and once with Houdini's headers, but the executable is linked against both openvdb and openvdb_sesi. I've quickly tried the following out as an example:

main.cpp
```
#include <iostream>

const char* tmp();
namespace houdini { const char* tmp(); }

int main(int , char* [])
{
    std::cout << tmp() << std::endl;
    std::cout << houdini::tmp() << std::endl;
    return 0;
}
```

tmp/vdb_source.cpp
```
#ifdef HOUDINI
#include </toolkit/include/openvdb/version.h>
namespace houdini {
#else
#include <openvdb/version.h>
#endif

const char* tmp() {
    return openvdb::getLibraryVersionString();
}

#ifdef HOUDINI
}
#endif
```

CMake
```
file(COPY tmp/vdb_source.cpp DESTINATION ${CMAKE_CURRENT_LIST_DIR}/tmp2)

set(SOURCE_FILES
  main.cpp
  tmp/vdb_source.cpp
  tmp2/vdb_source.cpp)

set_property(SOURCE tmp2/vdb_source.cpp APPEND PROPERTY COMPILE_DEFINITIONS "HOUDINI;")

add_executable(sandbox_noddy ${SOURCE_FILES})
target_link_libraries(sandbox_noddy libopenvdb.so)
target_link_libraries(sandbox_noddy libopenvdb_sesi.so)
```

Your solution seems to work so I'm not saying you should do it this way, this is just to provide an alternative

Idclip: Instead of
```
#ifdef HOUDINI
#include </toolkit/include/openvdb/version.h>
namespace houdini {
#else
#include <openvdb/version.h>
#endif
```

You could also set the include path for one of the duplicated sources to point to the houdini toolkit and the other to the vdb installation, resulting simply in:
```
#include <openvdb/version.h>
#ifdef HOUDINI
namespace houdini {
#endif
```

danrbailey: Hmm, that doesn't seem too different to what I'm doing here to be honest. The main issue I see with your approach (if I've understood correctly) is that you have now put the declaration(s) of tmp() in one file (main.cpp) and the definition somewhere else (tmp/vdb_source.cpp) and they'd obviously still need to be synced up for this to work - which would likely mean turning to the preprocessor to achieve that.

I think a namespace is probably a little nicer than these extern methods, but I'm not sure there's much in it.

danrbailey: @Idclip - I just added some changes to implement what you're suggesting here, it makes the CMake a bit more complicated, but ultimately I think it's better because there's less use of the preprocessor so makes it's all more readable.

Note - I made a few tweaks to your suggestion, rather than copying the file into another destination in the source directory (which we'd have to be careful not to add to the repo), it generates it into the build directory and correctly sets up the dependencies. I think that makes more sense. Let me know what you think.

CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ option(OPENVDB_BUILD_PYTHON_MODULE "Build the pyopenvdb Python module" OFF)
 option(OPENVDB_BUILD_UNITTESTS "Build the OpenVDB unit tests" OFF)
 option(OPENVDB_BUILD_DOCS "Build the OpenVDB documentation" OFF)
 option(OPENVDB_BUILD_HOUDINI_PLUGIN "Build the Houdini plugin" OFF)
+option(OPENVDB_BUILD_HOUDINI_ABITESTS "Build the Houdini ABI tests" OFF)
```
Idclip: should probably be put in `mark_as_advanced`

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -302,9 +302,6 @@ endif()
 # Boost - currently must be provided as VDB is not fully configured to
 # use Houdini's namespaced hboost
 
-unset(_HOUDINI_INCLUDE_DIR)
-unset(_HOUDINI_LIB_DIR)
```
Idclip: Perhaps lets rename these to `HOUDINI_INCLUDE_DIR` and `HOUDINI_LIB_DIR` and add them to the comment at the top of this file ad advertised variables which can be used

openvdb_houdini/abitest/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+# Create an object library from the generated TestABIHoudini.cc file
+# and set the include directories and add a HOUDINI variable
+
+add_library(TestABIHoudini OBJECT ${PROJECT_BINARY_DIR}/TestABIHoudini.cc)
+target_include_directories(TestABIHoudini SYSTEM PUBLIC ${_HOUDINI_INCLUDE_DIR})
+
+set_property(
+  SOURCE ${PROJECT_BINARY_DIR}/TestABIHoudini.cc
+  APPEND PROPERTY COMPILE_DEFINITIONS "HOUDINI;"
+)
```
Idclip: can you instead use `target_compile_options(TestABIHoudini PRIVATE -DHOUDINI)`?

Idclip:
Minor comments - main one is to advertise the houdini vars from the setup cmake. Otherwise lgtm, thanks a lot for doing this

danrbailey:
@Idclip - can you look over this one last time? I believe I've addressed your feedback now.

Idclip:
I've already approved this - still looks good

danrbailey:
> I've already approved this - still looks good

Right you are, thank you!

PR #629, opened 2020-02-12
Add minimum MSVC version to CMake

danrbailey:
As discussed, add a CMake check to ensure that at least the minimum supported version of MSVC is being used.

(@e4lam)


e4lam:
Thanks!

PR #630, opened 2020-02-12
Fixed Point Optimization

danrbailey:
Rounding from float -> int can be achieved with a simple cast *provided* that the floating-point value is always positive. As we're independently handling the case where the value is not positive, we can remove the redundant floor operation. This improved performance of this method by ~10% in my testing.

This method is used a lot so I tested all possible 32-bit floating-point values to be sure I wasn't missing something.


Idclip:
Nice find!

PR #631, opened 2020-02-13
OVDB-136: Descriptor Group Improvements

danrbailey:
Some small descriptor group-related improvements:

* Renamed Descriptor::nextUnusedGroupOffset() => Descriptor::unusedGroupOffset(size_t hint). This now has a hint parameter, if the hint is valid (within the range of available group offsets) and not currently in use, it will be returned, otherwise it resorts to choosing the next unused group offset. A size_t max value is now returned if there are no unused group offsets.

* Added Descriptor::groupIndexCollision(Descriptor&). This method returns true/false whether there are any collisions in the group indices. An example is if "group1" is mapped to index 1 in one descriptor and index 2 in the other. This a convenient way of determining if any group index remapping needs to be done when merging two or more data sets.

(@richhones)


openvdb/points/AttributeSet.h
```diff
@@ -440,8 +440,10 @@ class OPENVDB_API AttributeSet::Descriptor
     /// Return @c true if there are sufficient empty slots to allow compacting
     bool canCompactGroups() const;
 
-    /// Return the next empty group slot
-    size_t nextUnusedGroupOffset() const;
```
Idclip, line 444: can we add proper deprecation support for this function

Idclip:
Release notes

danrbailey:
> Release notes

Hence why it was left un-merged for the time being. :)

PR #634, opened 2020-02-15
OVDB-136: Attribute Stealing

danrbailey:
Two new additions related to stealing data from a Point Data Grid:

* PointDataLeafNode::stealAttributeSet() - releases ownership of the AttributeSet and replaces it with a new empty AttributeSet.

* AttributeSet::removeAttribute(...) - releases the shared_ptr reference to a specific AttributeArray by index or name and returns it. There is also an "unsafe" variant that doesn't attempt to put the AttributeSet or Descriptor back into a valid state which is useful when you know the AttributeSet is being destroyed and want the extra performance.

### DIFF
--- a/openvdb/points/AttributeSet.cc
+++ b/openvdb/points/AttributeSet.cc
@@ -379,9 +379,10 @@ AttributeSet::removeAttribute(const size_t pos)
 {
     if (pos >= mAttrs.size())     return AttributeArray::Ptr();
 
-    AttributeArray::Ptr array = mAttrs[pos];
+    assert(mAttrs[pos]);
+    AttributeArray::Ptr array;
+    std::swap(array, mAttrs[pos]);
     assert(array);
-    mAttrs[pos].reset();
 
     // safely drop the attribute and update the descriptor
     std::vector<size_t> toDrop{pos};
@@ -396,9 +397,10 @@ AttributeSet::removeAttributeUnsafe(const size_t pos)
 {
     if (pos >= mAttrs.size())     return AttributeArray::Ptr();
 
-    AttributeArray::Ptr array = mAttrs[pos];
+    assert(mAttrs[pos]);
+    AttributeArray::Ptr array;
+    std::swap(array, mAttrs[pos]);
     assert(array);
-    mAttrs[pos].reset();
 
     return array;
 }
--- a/openvdb/points/AttributeSet.cc
+++ b/openvdb/points/AttributeSet.cc
@@ -365,6 +365,45 @@ AttributeSet::appendAttribute(  const Descriptor& expected, DescriptorPtr& repla
 }
 
 
+AttributeArray::Ptr
+AttributeSet::removeAttribute(const Name& name)
+{
+    const size_t pos = this->find(name);
+    if (pos == INVALID_POS) return AttributeArray::Ptr();
+    return this->removeAttribute(pos);
+}
+
+
+AttributeArray::Ptr
+AttributeSet::removeAttribute(const size_t pos)
+{
+    if (pos >= mAttrs.size())     return AttributeArray::Ptr();
+
+    AttributeArray::Ptr array = mAttrs[pos];
+    assert(array);
+    mAttrs[pos].reset();
+
+    // safely drop the attribute and update the descriptor
+    std::vector<size_t> toDrop{pos};
+    this->dropAttributes(toDrop);
+
+    return array;
+}
+
+
+AttributeArray::Ptr
+AttributeSet::removeAttributeUnsafe(const size_t pos)
+{
+    if (pos >= mAttrs.size())     return AttributeArray::Ptr();
+
+    AttributeArray::Ptr array = mAttrs[pos];
+    assert(array);
+    mAttrs[pos].reset();
+
+    return array;
+}
+
+
 void
 AttributeSet::dropAttributes(const std::vector<size_t>& pos)
 {
--- a/openvdb/points/AttributeSet.cc
+++ b/openvdb/points/AttributeSet.cc
@@ -400,7 +400,6 @@ AttributeSet::removeAttributeUnsafe(const size_t pos)
     assert(mAttrs[pos]);
     AttributeArray::Ptr array;
     std::swap(array, mAttrs[pos]);
-    assert(array);
 
     return array;
 }
--- a/openvdb/points/AttributeSet.h
+++ b/openvdb/points/AttributeSet.h
@@ -201,6 +201,35 @@ public:
                                         const bool constantStride,
                                         const AttributeArray::ScopedRegistryLock* lock);
 
+    /// @brief Remove and return an attribute array by name
+    /// @param name the name of the attribute array to release
+    /// @details Detaches the attribute array from this attribute set and returns it.
+    /// This also updates the descriptor to remove the reference to the attribute array.
+    /// @note AttributeArrays are stored as shared pointers, so they are not guaranteed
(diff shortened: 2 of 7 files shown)


openvdb/points/PointDataGrid.h
```diff
@@ -745,6 +748,15 @@ class PointDataLeafNode : public tree::LeafNode<T, Log2Dim>, io::MultiPass {
 
 // PointDataLeafNode implementation
 
+template<typename T, Index Log2Dim>
+inline AttributeSet*
+PointDataLeafNode<T, Log2Dim>::stealAttributeSet()
+{
+    std::unique_ptr<AttributeSet> ptr(mAttributeSet.release());
+    mAttributeSet.reset(new AttributeSet);
+    return ptr.release();
```
Idclip: just return a unique_ptr?

openvdb/points/AttributeSet.h
```diff
@@ -201,6 +201,35 @@ class OPENVDB_API AttributeSet
...
+    /// to be unique. Check the reference count before blindly re-using in a new AttributeSet.
+    AttributeArray::Ptr removeAttribute(const Name& name);
+
+    /// @brief Remove and return an attribute array by index
+    /// @param pos the position index of the attribute to release
+    /// @details Detaches the attribute array from this attribute set and returns it.
+    /// This also updates the descriptor to remove the reference to the attribute array.
+    /// @note AttributeArrays are stored as shared pointers, so they are not guaranteed
+    /// to be unique. Check the reference count before blindly re-using in a new AttributeSet.
+    AttributeArray::Ptr removeAttribute(const size_t pos);
```
Idclip: Maybe comment in these that if pos or name is invalid, nullptr is returned

openvdb/points/AttributeSet.cc
```diff
@@ -365,6 +365,45 @@ AttributeSet::appendAttribute(  const Descriptor& expected, DescriptorPtr& repla
...
+
+
+AttributeArray::Ptr
+AttributeSet::removeAttribute(const size_t pos)
+{
+    if (pos >= mAttrs.size())     return AttributeArray::Ptr();
+
+    AttributeArray::Ptr array = mAttrs[pos];
+    assert(array);
+    mAttrs[pos].reset();
```
Idclip: perhaps std:swap would be more efficient

openvdb/points/AttributeSet.h
```diff
@@ -201,6 +201,35 @@ class OPENVDB_API AttributeSet
...
+    /// @details Detaches the attribute array from this attribute set and returns it.
+    /// In cases where the AttributeSet is due to be destroyed, a small performance
+    /// advantage can be gained by leaving the attribute array as a nullptr and not
+    /// updating the descriptor. However, this leaves the AttributeSet in an invalid
+    /// state making it unsafe to call any methods that implicitly derefence the attribute array.
+    /// @note AttributeArrays are stored as shared pointers, so they are not guaranteed
+    /// to be unique. Check the reference count before blindly re-using in a new AttributeSet.
+    /// @warning Only use this method if you're an expert and know the risks of not
+    /// updating the array of attributes or the descriptor.
+    AttributeArray::Ptr removeAttributeUnsafe(const size_t pos);
```
Idclip: Hmm not a huge fan of this 'unsafe' method but I don't see any way around it. I assume you're already using this for some of the point moving changes, is there no way this can be done manually? The only other thing I can suggest is calling `AttributeArray::collapse` (or `AttributeSet::replace`) instead where you're using this which is a single virtual call + the reallocation of a single element - I'd hope that the performance differences would be insignificant vs this

openvdb/points/PointDataGrid.h
danrbailey: Opted for returning a unique_ptr as you suggested. Think we should go through the effort of updating the others at some point too.

danrbailey:
@Idclip - addressed your feedback except for the "unsafe" method as I think this is the best solution here. Thanks.

openvdb/points/AttributeSet.h
Idclip: Okay I see what you mean. `Grid::getUnsafeAccessor` doesn't leave any objects in an "invalid" state which is the main thing I'm concerned about with this method. Do you have an example of what does the book-keeping looks like with the vector method?

danrbailey: > Okay I see what you mean. Grid::getUnsafeAccessor doesn't leave any objects in an "invalid" state which is the main thing I'm concerned about with this method.

I don't agree with this statement. If you delete a node from the tree that the "unsafe" accessor is referencing, the internal cache can still retain a pointer to the deleted node, so any attempt to dereference it will crash. The Accessor _is_ in an invalid state is it not?

> Do you have an example of what does the book-keeping looks like with the vector method?

Sure. If you were to do per-attribute iteration, let's say you wanted to steal all the "N" attribute arrays for a collection of leaf nodes, you'd not only need to extract all the attributes into one vector per-leaf, but you'd also need to keep a reference to each descriptor of each leaf node as it's not guaranteed that the "N" attribute maps to the same index for each leaf. Then you'd look up into this temporary set of vectors/descriptors to find each attribute array.

The alternative with the proposed method is to just extract each attribute array as you needed it, knowing that the leaf node would be destroyed at the end anyway.

Idclip: > I don't agree with this statement. If you delete a node from the tree that the "unsafe" accessor is referencing, the internal cache can still retain a pointer to the deleted node, so any attempt to dereference it will crash. The Accessor is in an invalid state is it not?

There is are differences between the guarantees of the unsafe accessor and this method, but the main point is the accessor does not invalidate the tree. When we talk about what's in an "invalid state", yes I guess you could argue that the accessor could be with your example (though that's no different to say `auto& t = g.tree(); g.newTree(); t.oops();` and actually it's a quirk of the "safe" accessors that this works...). However your method leaves the VDB **data structure** in an invalid state which is, I would argue, a significantly different problem.

> If you were to do per-attribute iteration, let's say you wanted to steal all the "N" attribute arrays for a collection of leaf nodes, you'd not only need to extract all the attributes into one vector per-leaf, but you'd also need to keep a reference to each descriptor of each leaf node as it's not guaranteed that the "N" attribute maps to the same index for each leaf.

Surely in this example you just store N separately or move it to the front/back or w/e, there's no need to keep a reference to the descriptors? I do appreciate it's slightly more complex. Is there any way this method could be made private with friend classes or some such? If not, perhaps this is one for the TSC?

danrbailey: In the case where you're stealing attribute arrays in multiple passes (and some of them might not exist), you'd still need to handle all the indexing yourself.

Thanks, I don't have a better solution to propose, so would be good if we can merge this as it is. I'd be happy to revisit when it's clearer how this will be used.

openvdb/points/AttributeSet.cc
```diff
@@ -365,6 +365,47 @@ AttributeSet::appendAttribute(  const Descriptor& expected, DescriptorPtr& repla
...
+
+AttributeArray::Ptr
+AttributeSet::removeAttributeUnsafe(const size_t pos)
+{
+    if (pos >= mAttrs.size())     return AttributeArray::Ptr();
+
+    assert(mAttrs[pos]);
+    AttributeArray::Ptr array;
+    std::swap(array, mAttrs[pos]);
+    assert(array);
```
Idclip: last minor comment, this assert isn't technically valid as this particular method can be called multiple times on the same index and may cause undesirable debug behaviour - i'd remove it

PR #637, opened 2020-02-20
GitHub Actions Workflows

danrbailey:
Making this a draft PR for now, until we get the right Houdini API credentials in place.

This does a number of things:

* Migrates the Cirrus CI checks over to GitHub Actions
* Fixes a few tabs and checks more of the codebase
* Introduces some helper CI scripts around docs and doxygen
* Introduces more GitHub Actions workflows

One key benefit of separating these out into separate workflows is that they can be more strictly defined. A PR that only adds TSC meeting notes will no longer run any _build_ CI checks, a PR that only touches documentation which only run _whitespace_ and _documentation_ CI checks, etc.

I've also created a "build" and "fullbuild" workflow, the former which runs on any branch push and doesn't error if Houdini credentials are not defined, the latter which is only run for a PR into the master branch and is required to have Houdini credentials.


danrbailey:
Closing this draft PR as this workflow has been rolled out now.

PR #640, opened 2020-02-21
Add gitignore file

### LINKED ISSUE #639: Add a gitignore file to the repo
Ideally based on https://www.gitignore.io/

piyush0411:
Added gitignore file to avoid build directory to be included in commits. 
Fixes #639


danrbailey:
I think we'll probably want to add more to this file, but this is better than having nothing at all, so I'll merge this.

Issue #638, opened 2020-02-21
Allow OpenVDB to build without OpenEXR/Half.h


BenFrantzDale:
As far as I can tell, the only reason the base OpenVDB requires OpenEXR is to allow loading and saving half-precision data. As one who has no interest in that feature, it's annoying to set up OpenEXR as a dependency for what feels like a niche feature.

If there's interest in this, I can make a PR. The approach I took is to replace all the `bool`s denoting half-precision reading and writing with
```
// Type-safe flag for reading/writing as half.
enum class StoredAsHalf {
    no,
#ifdef OPENVDB_WITH_OPENEXR_HALF
    yes
#endif
};
```
that way if compiled without Half, there's no way to specify that you'd like to read/write halfs. It has the added advantage of using an enum rather than a `bool`, for added type safety.

Is this of interest?

danrbailey:
What you're proposing in general sounds good to me. However, it's probably quite a challenging change to make in practice as we use the half data type _a lot_ in the codebase.

A few specifics about such an implementation:

* It should follow the convention for enabling/disabling other features through CMake such as USE_EXR (https://github.com/AcademySoftwareFoundation/openvdb/blob/develop/CMakeLists.txt#L85)
* It should default to ON to avoid changing existing behavior when building the library
* Attempts to read a Grid that has been written as half will now fail - this should be handled in the same vein as attempting to read a Grid with Blosc compression with a library that has been built without Blosc compression enabled (https://github.com/AcademySoftwareFoundation/openvdb/blob/develop/openvdb/io/Compression.cc#L167)
* Logic handling half not being available should be introduced in as few places as possible, here's one example (https://github.com/AcademySoftwareFoundation/openvdb/blob/develop/openvdb/io/Compression.h#L623)

I'd suggest starting out by building the library and the unit tests, then commenting out this header file to get an idea of just how daunting this might prove to be:

https://github.com/AcademySoftwareFoundation/openvdb/blob/develop/openvdb/Types.h#L9

Idclip:
+1 to this idea, though I agree with @danrbailey that it may require some serious thought to keep the API consistent. This is just a passing comment, but I've always wondered whether it'd be feasible to switch to a header only half implementation which would remove some of the dependency pains.

BenFrantzDale:
Here's a proof of concept: https://github.com/AcademySoftwareFoundation/openvdb/pull/641

danrbailey:
[...]

PR #641, opened 2020-02-21
Proof of concept: Remove OpenEXR (half) dependency.

BenFrantzDale:
Proof of concept for https://github.com/AcademySoftwareFoundation/openvdb/issues/638

The idea here is that I added an enum, `StoredAsHalf`, which, if you compile with half, is
`enum class StoredAsHalf { no, yes };`
but if you aren't, it's just 
`enum class StoredAsHalf { no };`
which is to say if you aren't compiling with it, there's no way to request things to be stored as half, so it's typesafe like that.

Additionally, when building without half, `StoredAsHalf` arguments that are typically not defaulted are defaulted to `StoredAsHalf::no`.

Right now it throws an exception when it reads a file that is stored as half but it doesn't know how to read it.

Even if you are compiling with half, this has the advantage of adding type safety to the stored-as-half arguments.

### DIFF
--- a/openvdb/Grid.cc
+++ b/openvdb/Grid.cc
@@ -338,21 +338,27 @@ GridBase::setCreator(const std::string& creator)
 ////////////////////////////////////////
 
 
-bool
+StoredAsHalf
 GridBase::saveFloatAsHalf() const
 {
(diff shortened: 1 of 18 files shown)


danrbailey:
Closing this PR, see discussion here:

https://github.com/AcademySoftwareFoundation/openvdb/issues/638#issuecomment-589806760.

PR #644, opened 2020-02-24
Minor improvements to the Windows build

Idclip:
- Set CMake project "LANGUAGES" to CXX - stops CMake trying to find and test a compatible C compiler. Valid for all platforms
 - Fixed README windows build instructions with minor improvements

Also fixed the following warnings. I'll be attempting to slowly work through these, targeting the "heavy hitters" (console spamming) ones first.

 - Various C4996 warning fixes (`getenv` `tmpnam`)
 - Various C4146 fixes and skips (unary minus on unsigned types)


openvdb/io/Archive.cc
```diff
@@ -1082,7 +1082,15 @@ Archive::connectInstance(const GridDescriptor& gd, const NamedGridMap& grids) co
 bool
 Archive::isDelayedLoadingEnabled()
 {
+#if defined(_MSC_VER)
+    char* value;
+    _dupenv_s(&value, nullptr, "OPENVDB_DISABLE_DELAYED_LOAD");
```
Idclip: choosing to ignore the error state here, could throw or return false

openvdb/io/File.cc
```diff
@@ -71,10 +71,20 @@ struct File::Impl
     static Index64 getDefaultCopyMaxBytes()
     {
         Index64 result = DEFAULT_COPY_MAX_BYTES;
+#if defined(_MSC_VER)
+        char* s;
+        _dupenv_s(&s, nullptr, "OPENVDB_DISABLE_DELAYED_LOAD");
```
danrbailey: Wrong env var here - should be OPENVDB_DELAYED_LOAD_COPY_MAX_BYTES

openvdb/io/File.cc
```diff
@@ -71,10 +71,20 @@ struct File::Impl
...
+        if (s) {
+            char* endptr = nullptr;
+            result = std::strtoul(s, &endptr, /*base=*/10);
```
danrbailey: Maybe adjust the logic so it's no longer required to have this line twice? I'd also recommend putting the heap data straight into a unique_ptr to avoid any potential memory leaks here.

openvdb/io/TempFile.cc
```diff
@@ -15,8 +15,8 @@
 #include <unistd.h> // for access()
 #else
 #include <fstream> // for std::filebuf
+#include <cstdio> // for std::tmpnam_s(), L_tmpnam_s
```
danrbailey: P_tmpdir is still being used

README.md
```diff
@@ -107,7 +119,6 @@ git clone git@github.com:AcademySoftwareFoundation/openvdb.git
 cd openvdb
 mkdir build
 cd build
-cmake -DCMAKE_TOOLCHAIN_FILE=<PATH_TO_VCPKG>\scripts\buildsystems\vcpkg.cmake ..
-make -j4
-make install
```
danrbailey: If we change one instance of `make -j4` to `cmake --build . --parallel 4` we should change them all IMO.

Unless there's a good reason for using the cmake wrapper, I'd prefer using make directly for simplicity.

openvdb/io/Compression.cc
```diff
@@ -82,7 +82,9 @@ zipToStream(std::ostream& os, const char* data, size_t numBytes)
         os.write(reinterpret_cast<char*>(zippedData.get()), outZippedBytes);
     } else {
         // Write the size of the uncompressed data.
-        Int64 negBytes = -numBytes;
+        // numBytes expected to be <= the max value + 1 of a signed int64
+        assert(numBytes <= size_t(std::numeric_limits<Int64>::max())+size_t(1));
```
danrbailey: Can you elaborate on the "+1" part of this?

openvdb/io/File.cc
Idclip: Yikes thanks

README.md
Idclip: Well make doesn't work out the box on Windows so you have to use cmake to build

Idclip:
> Recommend converting all of the std::getenv calls for Windows - you've missed the ones in `TempFileImpl::getTempDir()`.

Ah yes, I debated whether to upload this PR now or wait until I had more changes - but if I'm just targeting those warnings here (and more specifically getenv and tmpnam) I can go through and make sure I've got them all.

openvdb/io/Compression.cc
Idclip: Sure, essentially size_t is a uint64_t on a lot of platforms (that we're targeting) but we negate the value to represent uncompressed data. This won't work if the value being passed in is greater than the max value of  **signed** 64 bit int (+1 as the lowest negative value will be one more than this) as it *may* (read: probably will) wrap and become positive again. In reality I sincerely much doubt anyone is writing out more than `9223372036854775807` bytes per leaf... but regardless this is still a potential bug. We're quite blasé about castinh between signed/unsigned values of the same precision with no checks and I'd like to try and improve these cases, especially as Windows complains at every single one

Idclip:
> > Recommend converting all of the std::getenv calls for Windows - you've missed the ones in `TempFileImpl::getTempDir()`.
> 
> Ah yes, I debated whether to upload this PR now or wait until I had more changes - but if I'm just targeting those warnings here (and more specifically getenv and tmpnam) I can go through and make sure I've got them all.

Oh actually these are already encapsulated in `#ifndef _MSC_VER` do don't require changing

openvdb/io/Compression.cc
danrbailey: Thanks. I suggest we just make this assert stricter than necessary here for greater simplicity (`numBytes < numeric_limits<Int64>::max()` ). I doubt anyone will see false positives as you say.

README.md
danrbailey: Right I see, that's a good enough reason. Could we just update the other references in this documentation to use `cmake --build` too?

Perhaps the `--config Release` part could use CMAKE_BUILD_TYPE in the line above instead?

Idclip: >  Could we just update the other references in this documentation to use cmake --build too?

I think we need a dedicated "Building on Windows" separate from the UNIX instructions. Both the dependency and build pages need update. I'll look to doing these separately after I've work through all the build components.

> Perhaps the --config Release part could use CMAKE_BUILD_TYPE in the line above instead?

Actually this only for multi tool configurations which we don't currently support (you choose the build type when you initially run cmake) so I can remove this.

Idclip: > > Perhaps the --config Release part could use CMAKE_BUILD_TYPE in the line above instead?
> 
> Actually this only for multi tool configurations which we don't currently support (you choose the build type when you initially run cmake) so I can remove this.

Okay this isn't true, without `--config Release` CMake will try to build both release and debug versions of the library and these instructions are specifically for Release - so it needs to be `--config Release`

PR #646, opened 2020-02-26
CMake/Windows improvements to the Python module

Idclip:
- Added options to enable/disable the python test and excluded the interpreter requirement if disabled 
 - Improved the report/error logic if python/numpy is not found
 - Fixed Windows C4146/C4996 warnings


openvdb/python/CMakeLists.txt
```diff
@@ -49,8 +52,14 @@ set(OPENVDB_PYTHON_INCLUDES)
 # the minimum python version is MINIMUM_PYTHON_VERSION - however this cannot
 # be provided to find_package(Python) with differing major versions. e.g.
 # calls to find_package(Python 2.7) fails if python3 is found on the system.
-function(OPENVDB_CHECK_PYTHON_VERSION _PY_VERSION _PY_PATH)
-  if(_PY_VERSION VERSION_LESS MINIMUM_PYTHON_VERSION)
```
Idclip: Made these arguments optional to handled the case where `_PY_VERSION` isn't defined due to find failure

openvdb/tools/MeshToVolume.h
```diff
@@ -825,11 +826,13 @@ class SweepExteriorSign
 
     void operator()(const tbb::blocked_range<size_t>& range) const {
 
+        constexpr Int32 DIM = static_cast<Int32>(LeafNodeType::DIM);
```
Idclip: Re-worked this method to use signed arithmetic for `pos` and `step` variables as `step` can decrement

danrbailey:
Change looks good in general - did you consider using the combination of OPENVDB_BUILD_PYTHON_MODULE=ON and OPENVDB_BUILD_UNITTESTS=ON to drive building the python unit tests, as opposed to introducing the new build flag?

Also, I'd like to see the python unit tests being enabled in CI before we merge this.

Idclip:
> Change looks good in general - did you consider using the combination of OPENVDB_BUILD_PYTHON_MODULE=ON and OPENVDB_BUILD_UNITTESTS=ON to drive building the python unit tests, as opposed to introducing the new build flag?

I did, but I thought it limiting that you couldn't also control the inclusion of the python test. The new flag will follow whatever the value of `OPENVDB_BUILD_UNITTESTS` is so in the majority of cases won't be touched, but the ability to enable/disable is irrespective of this flag is handy in my opinion.

> Also, I'd like to see the python unit tests being enabled in CI before we merge this.

Yes I'll look to do this. In appveyor we're picking up the boost libs from a custom folder - Can you explain how/where this folder gets built/initialised from?

danrbailey:
> > Change looks good in general - did you consider using the combination of OPENVDB_BUILD_PYTHON_MODULE=ON and OPENVDB_BUILD_UNITTESTS=ON to drive building the python unit tests, as opposed to introducing the new build flag?
> 
> I did, but I thought it limiting that you couldn't also control the inclusion of the python test. The new flag will follow whatever the value of `OPENVDB_BUILD_UNITTESTS` is so in the majority of cases won't be touched, but the ability to enable/disable is irrespective of this flag is handy in my opinion.

Sure, makes sense.

> > Also, I'd like to see the python unit tests being enabled in CI before we merge this.
> 
> Yes I'll look to do this. In appveyor we're picking up the boost libs from a custom folder - Can you explain how/where this folder gets built/initialised from?

Yep, it's pre-installed:

https://www.appveyor.com/docs/windows-images-software/#boost

e4lam:
```
// Use Windows only methods (_strdup); no POSIX.
#if defined(_MSC_VER)
#define strdup _strdup
#endif
```

This is annoying. Why aren't we just adding `-D_CRT_NONSTDC_NO_WARNINGS` to the MSVC compiler defines? Case in point, I recently had to remove various `#define snprintf _snprintf` in our code base (and all third party libraries) because they had added `std::snprintf` to C++11. If you happened to include two headers, one which had the define and another which called `std::snprintf()`, it would get preprocessed to `std::_snprintf()`, which of course doesn't exist.

Idclip:
> This is annoying. Why aren't we just adding `-D_CRT_NONSTDC_NO_WARNINGS` to the MSVC compiler defines? Case in point, I recently had to remove various `#define snprintf _snprintf` in our code base (and all third party libraries) because they had added `std::snprintf` to C++11. If you happened to include two headers, one which had the define and another which called `std::snprintf()`, it would get preprocessed to `std::_snprintf()`, which of course doesn't exist.

Hmm I guess we could do this - I was getting confused between `_CRT_NONSTDC_NO_WARNINGS` and `_CRT_SECURE_NO_WARNINGS`, the latter of which we I don't believe we want enabled. The MS docs seem pretty relaxed about disabling POSIX warnings, with an emphasis on portability:

> To fix this issue, we usually recommend you change your code to use the suggested function names instead. However, the updated names are Microsoft-specific. If you need to use the existing function names for portability reasons, you can turn off these warnings. The functions are still available in the library under their original names.

I think I'll add this as you've suggested. Note though that the define solution in this case is relatively harmless as it's at the bottom of a .cc.

Idclip:
I'm currently unable to get this working for appveyor. The version of boost installed by default doesn't come with boost-python, so I'm installing that through vcpkg. This takes around ~15minutes so the builds are now timing out at 1hour. I haven't been able to configure appveyors cache to include the boost installs - it seems I don't have the required permissions. @danrbailey do you know how to do this?

e4lam:
@Idclip I'm woefully ignorant of all this CI stuff? But it's not just a matter of changing the appveyor.yml ? eg. https://github.com/boostorg/beast/blob/develop/appveyor.yml#L53-L54

Idclip:
Hey @e4lam  yeah I tried that, but it doesn't seem to pull any changes. The appveyor CI docs on the build cache (https://www.appveyor.com/docs/build-cache/) mention you can remove and re-add it to clear it - but then it also has this:

> Note: By default, saving cache is disabled in Pull Request builds. Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.

https://www.appveyor.com/docs/build-cache/

danrbailey:
> Hey @e4lam yeah I tried that, but it doesn't seem to pull any changes. The appveyor CI docs on the build cache (https://www.appveyor.com/docs/build-cache/) mention you can remove and re-add it to clear it - but then it also has this:
> 
> > Note: By default, saving cache is disabled in Pull Request builds. Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.

If it's not trivial to update AppVeyor, let's go ahead and merge this PR as-is. However, we should follow this with a PR (or two) that adds Windows CI to GitHub Actions so that we can transition away from AppVeyor and add the Python support to it.

ci/build.sh
```diff
@@ -17,7 +17,10 @@ cmake \
     -DOPENVDB_USE_DEPRECATED_ABI=ON \
     -DUSE_BLOSC=${BLOSC} \
     -DOPENVDB_CXX_STRICT=ON \
+    -DOPENVDB_BUILD_PYTHON_MODULE=ON \
     -DOPENVDB_BUILD_UNITTESTS=ON \
+    -DOPENVDB_BUILD_BINARIES=ON \
+    -DOPENVDB_BUILD_VDB_PRINT=ON \
```
danrbailey: Aren't these two on by default?

Idclip: Yeah but I like to be explicit!

e4lam:
> Approved, pending dropping the attempted AppVeyor changes and updating the release notes.
> 
> I defer to @e4lam for the Windows changes...

Looks good to me since it's exactly what I suggested. :)

e4lam:
@Idclip Oh wait, maybe I spoke too soon or didn't properly see all the changes. We shouldn't need any of those other MSVC specific "No Posix" change anymore either right? eg. the uses of `std::getenv()` doesn't need a special case. Nor the use of of `std::tmpnam()`

Idclip:
> @Idclip Oh wait, maybe I spoke too soon or didn't properly see all the changes. We shouldn't need any of those other MSVC specific "No Posix" changes anymore either right? eg. the uses of std::getenv() doesn't need a special case. Nor the use of of std::tmpnam()

No we still need them. This is what I was referring to with my initial confusion with `_CRT_NONSTDC_NO_WARNINGS` vs `_CRT_SECURE_NO_WARNINGS`. https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2019#posix-function-names

In summary:

> Most of the deprecated functions allow unchecked read or write access to buffers. Their misuse can lead to serious security issues. The compiler issues a deprecation warning for these functions, and suggests the preferred function... Carefully verify it's not possible for a buffer overwrite or overread to occur in your code. Then, you can turn off the warning.

I've opted to switch them for now.

e4lam:
@Idclip I believe we should defined `_CRT_SECURE_NO_WARNINGS`. The "secure" APIs are discouraged outside of Microsoft. For example, from http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1967.htm it says in the conclusion:

e4lam:
Also from http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1106.txt we have this:
> Linus Torvalds had this to say about a similar, earlier proposal:
>
>  The above code is slow, ugly, non-straightforward, unportable and
>  no more secure than the original code was.
>
>  In short, it is just stupid code.
>
>  But hey, if you want to advocate stupid code in public, that's
>  your prerogative.  But please don't be proud of it.
>
>  --
> <http://lists.nas.nasa.gov/archives/ext/linux-security-audit/2002/01/msg00073.html>

:)

Idclip:
So if I've understood correctly, the TL;DR version is that this is all a bit of a mess - the glibc folks decided not to implement the "Bounds Checking" interface and the MS versions, even though (as far as I understand it) instigated these changes, are incomplete and don't conform to the original technical report (http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1225.pdf). 

Note that I've already merged the changes to `getenv` and `tmpnam` which were part of another PR (https://github.com/AcademySoftwareFoundation/openvdb/pull/644). This PR can be merged as is as it doesn't make any additional changes to C4996 aside from adding the `_CRT_NONSTDC_NO_WARNINGS` and I'll create another set of changes to revert some of #644 - that being said, I think I'll raise this in the TSC before committing to the revert.

Idclip:
@e4lam PR to solve this here https://github.com/AcademySoftwareFoundation/openvdb/pull/658/files

e4lam:
Thanks for the further changes!

Issue #617, opened 2020-01-23
Add VDB Point Support to vdb_print


danrbailey:
The vdb_print binary is missing support for VDB Point grids. This would be super useful. Some recommended functionality:

* total/active/inactive point count
* attribute name, type, uniform, index, flags, compression information
* group name and count information
* whether the descriptor is shared (option to print out all descriptors which differ?)

I suggest that this is only implemented for ABI=6 and above as there are some new virtual functions that will make this easier.

kubaroth:
Just a heads-up, as Nick suggested. I'd like to give this one a try.

RonMordo:
Hi, can i work on the issue?

kmuseth:
Yes, please, go ahead :)Cheers,KenOn Feb 5, 2024, at 9:00 AM, Ron_Mordo ***@***.***> wrote:﻿
Hi, can i work on the issue?

aveljkovic:
Hi, I am a first-time contributor and I would love to take this on!

PR #650, opened 2020-02-26
Added VDB Point Support to vdb_print - #617

kubaroth:
Supported for ABI >= 6
- total/active/inactive point count
- attribute name, type, uniform, index, flags, compression
- group name and count information
- whether the descriptor is shared - needs to be verified if I got it right. :)

Signed-off-by: kuba <kuba456@gmail.com>

### DIFF
--- a/openvdb/cmd/openvdb_print.cc
+++ b/openvdb/cmd/openvdb_print.cc
@@ -11,6 +11,10 @@
 #include <openvdb/openvdb.h>
 #include <openvdb/util/logging.h>
 
+#include <openvdb/points/PointDataGrid.h>
+
+using namespace openvdb;
+using namespace openvdb::points;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


openvdb/cmd/openvdb_print.cc
```diff
@@ -141,7 +245,18 @@ printLongListing(const StringVec& filenames)
                 grid->print(std::cout, /*verboseLevel=*/11);
                 firstGrid = false;
             }
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+            // Check for grid type, inspect points if this is a point grid
+            const openvdb::GridBase::ConstPtr grid = *it;
```
Idclip: I'd move this to the start of the loop so the shared pointer is only copied once, immediately followed by a check

openvdb/cmd/openvdb_print.cc
```diff
@@ -141,7 +245,18 @@ printLongListing(const StringVec& filenames)
...
+            Name gridType = grid->type();
+            if (gridType.find("ptdata") != std::string::npos)
```
Idclip: Don't bother with this nametype check, you can call `GridBase::grid<T>` with the grid and check the resulting ptr

openvdb/cmd/openvdb_print.cc
```diff
@@ -11,13 +11,116 @@
...
+
+            auto attrArr = attrset.getConst(pa.name);
+            pa.codec = attrArr->codecType();
+            pa.isUniform = attrArr->isUniform();
+            pa.isHidden = attrArr->isHidden();
+            pa.isTransient = attrArr->isTransient();
+            pa.isStreaming = attrArr->isStreaming();
+            points.pointAttribs[pa.name] = pa;
+        }
+    }
```
Idclip: I'm not sure how useful this multi leaf implementation is. In most cases the attributes will be consistent across the VDB, and even if they are not, this doesn't provide a way to infer exactly how the attributes are spread. My vote would be to simply inspect the first leaf and infer the attributes from that. You could then check to see fit he descriptor is shared and, if it isn't, simply print a warning that the attribute layouts are not consistent.

Originally comments on this are here https://jira.aswf.io/browse/OVDB-114

@danrbailey

openvdb/cmd/openvdb_print.cc
```diff
@@ -11,13 +11,116 @@
...
+            std::cout << sINDENT << "isShared: " << attr.isShared << '\n';
+            std::cout << sINDENT << "isHidden: " << attr.isHidden << '\n';
+            std::cout << sINDENT << "isTransient: " << attr.isTransient << '\n';
+            std::cout << sINDENT << "isStreaming: " << attr.isStreaming << '\n';
+          }
+        std::cout << "Point groups:" << '\n';
+        for (auto it=pointStats.groups.begin(); it!=pointStats.groups.end(); ++it){
+            std::cout << it->first << " " << it->second <<std::endl;
+        }
+}
```
Idclip: Can you provide an example print output which we can comment on w.r.t formatting

openvdb/cmd/openvdb_print.cc
```diff
@@ -11,13 +11,116 @@
...
 
 namespace {
 
 using StringVec = std::vector<std::string>;
 
 const char* INDENT = "   ";
 const char* gProgName = "";
+static const std::string sINDENT(INDENT);
+
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
```
Idclip: I don't think you need to ABI switch this. I think the only methods you're using which are ABI6 only are `codecType` and `valueType`, both of which you can query from the descriptor using `valueType()` and `type()`

openvdb/cmd/openvdb_print.cc
```diff
@@ -141,7 +245,18 @@ printLongListing(const StringVec& filenames)
...
             }
+#if OPENVDB_ABI_VERSION_NUMBER >= 6
+            // Check for grid type, inspect points if this is a point grid
+            const openvdb::GridBase::ConstPtr grid = *it;
+            Name gridType = grid->type();
+            if (gridType.find("ptdata") != std::string::npos)
+                PointStats::inspectPoints(grid, pointStats);
+        }
+        // Print out point stats only if there are any points
+        if (pointStats.total != 0) PointStats::printPointStats(pointStats);
```
Idclip: We would still want to print some information to show the grid is empty, as it may contain topology but no points

openvdb/cmd/openvdb_print.cc
```diff
@@ -11,13 +11,116 @@
 #include <openvdb/openvdb.h>
 #include <openvdb/util/logging.h>
 
+#include <openvdb/points/PointDataGrid.h>
+
+using namespace openvdb;
+using namespace openvdb::points;
```
Idclip: If we're to do this, I'd remove all usage of the `openvdb::` namespace - but as you've opted for auto in most places, I don't think we even need this and can use `openvdb::points` where necessary

danrbailey:
Thanks Kuba! :)

I actually have a new convenience class coming (AttributeSet::Info) that I hope will help to alleviate some of the extra work you've had to do here. It's an issue that comes up quite frequently - there's some metadata stored with the AttributeSet/Descriptor (order of arrays, groups, etc) and some with the AttributeArray itself (is the array uniform, hidden, etc) and to get a complete picture of the data being stored you have to deal with both layers. 

In attempting to introduce logic to cleanly merge multiple differing AttributeSets/Descriptors, a new convenience abstraction seemed to be a good place to put this. I'm aiming to share a version (without the merging functionality) before the end of the week so it would be good to see if this makes the code here simpler. 

(In the meantime, it would still be worth getting the CLA resolved)

kubaroth:
Thanks Dan, I will take a look at the new helper class and rework the PR. 
Regarding CLA looks I need to recommit with updated email as the previous one was not associated with my GitHub account. I'll force push to update this information.

kubaroth:
I was asked by EasyCLA support to close and reopen the PR to see if this has an effect on the failing check.

kubaroth:
I'm going to wait for OVDB-136 to be merged first.

openvdb/cmd/openvdb_print.cc
kubaroth: Perhaps I'm doing this wrong but using codecType() and valueType() gives me different output:
`
myAttributeArray.codecType() : fxpt16
myDescriptor->valueType() : vec3s
`

openvdb/cmd/openvdb_print.cc
kubaroth:
```
...
Transform:
  voxel size: 2.29
  index to world:
     [2.29, 0, 0, 0] 
     [0, 2.29, 0, 0] 
     [0, 0, 2.29, 0] 
     [0, 0, 0, 1] 

Total Point Count:
   total: 8
   active: 8
   inactive: 0
Point attributes:
  name: P
   type: vec3s
   codec: fxpt16
   uniform: 1
   shared: 1
   hidden: 0
   transient: 0
   group: 0
   string: 0
   constantStride: 1
  name: __group
   type: vec3s
   codec: grp
   uniform: 0
   shared: 1
   hidden: 0
   transient: 0
   group: 0
   string: 0
   constantStride: 1
Point groups:
   group1: 5
   group2: 6
```

kubaroth:
The update addressed all above comments and uses new AttributeSet::Info. Based on the [answer|https://github.com/AcademySoftwareFoundation/openvdb/pull/652#issuecomment-593223442] I will still need to rework PointAttrib helper struct.

openvdb/cmd/openvdb_print.cc
Idclip: Ah you're looking for `type()`. The value type and codec are stored in a pair of strings which is typedef'd to a `NamePair` in VDB - this is the return value of `type()`. i.e:

```
 desc->type().first == desc->valueType();
 desc->type().second == array->codecType();
```

kubaroth: Thanks, I will update that.

openvdb/cmd/openvdb_print.cc
danrbailey: I suggest we do a first pass of this where we don't include any details that require iterating over the AttributeArrays, but the ideal in my mind would be something like the following:

name: P
   type: vec3s
   codec: fxpt16
   stride: 1 (or stride: dynamic)
   hidden: 0/123 *
   transient: 5/123 *
   uniform: 100/123 *

The group and string fields should be evident from the value type / codec, so I'd drop those fields entirely. Also, I'd make "type" -> "valueType" just to eliminate confusion there.

The fields marked with asterisks I would probably only display if in verbose mode (but as I say let's just leave that out for now). These show the number of arrays that are uniform (ie 100) vs the total number of arrays across all leaf nodes (ie 123).

You'll note I've changed "constantStride" -> "stride". I think it's better to mark it as "dynamic" if it's not constant. Otherwise it's potentially confusing, does constantStride = 1 mean _true_ it's constant stride or that the stride is 1.

kubaroth: I just realized changes I've pushed in the meantime haven't made into this PR as the original PR was opened from a develop not a feature branch in my fork. Sorry about that.
@danrbailey I will add changes as suggested in your last comment. 
One thing which is not clear is to me is 
> we don't include any details that require iterating over the AttributeArrays,
I believe this refers to the new Info::Array

danrbailey: Well stride is available in Info::Array or by accessing the AttributeArray directly, I'm more talking about those fields such as uniform which are _not_ available in Info::Array. For now, I suggest we just ignore the starred ones in my list above.

kubaroth: Ok, I've addressed all comments, I hope. Below is the the current output:

```
...
Total Point Count:
   total: 8
   active: 8
   inactive: 0
Point attributes:
  name: P
   valueType: vec3s
   codec: fxpt16
   stride: 1
  name: __group
   valueType: vec3s
   codec: grp
   stride: 1
Point groups:
   group1: 5
   group2: 6
```

To double check, we still interested only to inspect the first leaf for attributes? (assuming there is consistency across all other leaves)

PR #651, opened 2020-02-26
Allow disabling half.h

BenFrantzDale:
Use new `OPENVDB_THROW_IF_HALF_NOT_SUPPORTED()` macro which, when `OPENVDB_USE_HALF` is `OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,9 @@ option(USE_EXR [=[
 Use OpenEXR while building the core openvdb library. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the located
 openvdb build configuration to decide on OpenEXR support. You may set this to on to force OpenEXR to be used if you
 know it to be required.]=] OFF)
+
+cmake_dependent_option(USE_HALF "Use the OpenEXR/half.h a half-precision floating-point library while building openvdb components. You may set this to off to remove the dependency on OpenEXR if you know you don't need float16 support." ON "USE_EXR" ON)
+
 cmake_dependent_option(OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING
   "Disable the implicit linking of Boost libraries on Windows" ON "WIN32" OFF)
 option(USE_CCACHE "Build using Ccache if found on the path" ON)
--- a/openvdb/Types.h
+++ b/openvdb/Types.h
@@ -6,7 +6,14 @@
 
 #include "version.h"
 #include "Platform.h"
+
+#ifdef OPENVDB_USE_HALF
 #include <OpenEXR/half.h>
+#define OPENVDB_THROW_IF_HALF_NOT_SUPPORTED() (void) // Nop if OPENVDB_USE_HALF is enabled.
+#else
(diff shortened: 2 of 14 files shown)


CMakeLists.txt
```diff
@@ -86,6 +86,9 @@ option(USE_EXR [=[
 Use OpenEXR while building the core openvdb library. If OPENVDB_BUILD_CORE is OFF, CMake attempts to query the located
 openvdb build configuration to decide on OpenEXR support. You may set this to on to force OpenEXR to be used if you
 know it to be required.]=] OFF)
+
+cmake_dependent_option(USE_HALF "Use the OpenEXR/half.h a half-precision floating-point library while building openvdb components. You may set this to off to remove the dependency on OpenEXR if you know you don't need float16 support." ON "USE_EXR" ON)
```
BenFrantzDale: I'm not sure if this is the right way to do this.

openvdb/Types.h
```diff
@@ -6,7 +6,14 @@
 
 #include "version.h"
 #include "Platform.h"
+
+#ifdef OPENVDB_USE_HALF
 #include <OpenEXR/half.h>
+#define OPENVDB_THROW_IF_HALF_NOT_SUPPORTED() (void) // Nop if OPENVDB_USE_HALF is enabled.
```
BenFrantzDale: I went with this approach to avoid lots of `#ifdef` lines and to have the error string in one place. Alternately this could be a function in the `openvdb` namespace, or I could go with just `OPENVDB_THROW(IoError, "OpenVDB has been compiled without Half support")` wherever it's needed.

openvdb/points/AttributeArray.h
```diff
@@ -463,7 +463,11 @@ struct AttributeArray::Accessor : public AttributeArray::AccessorBase
 namespace attribute_traits
 {
     template <typename T> struct TruncateTrait { };
+#ifdef OPENVDB_USE_HALF
     template <> struct TruncateTrait<float> { using Type = half; };
+#else
+    template <> struct TruncateTrait<float> { using Type = float; }; // If half is disabled, fall back to float->float.
```
BenFrantzDale: It was unclear to me if this is the correct behavior if we don't have `half`, or if there just shouldn't be a `TruncateTrait<float>` without `half`.

CMakeLists.txt
Idclip: This is where IlmBase/OpenEXR gets a bit weird. IIRC, half.h is part of IlmBase, *not* OpenEXR. So it's not actually dependent on OpenEXR at all. In other words, `USE_HALF` can be on or off irrespective of `USE_EXR` and vice-versa. The logic, for the core lib, would be:

- If  `USE_EXR` find both IlmBase and OpenEXR. However, if `USE_HALF` is off, disable VDB half support (no `OPENVDB_USE_HALF`)
- Otherwise if `USE_HALF`, find IlmBase only and set `OPENVDB_USE_HALF`
- Otherwise if not `USE_HALF`, do not find IlmBase or OpenEXR 

Probably worth noting that the *only* reason we depend on EXR at all outside of `half.h` (and vdb_render) is because of some functionality in `tools/RayTracer.h` :\

danrbailey:
Closing this PR, I have incorporated these changes into PR #819. Thanks!

PR #652, opened 2020-02-28
OVDB-136: Add AttributeSet::Info convenience class

danrbailey:
This new class resolves a number of issues:

* It collates information that lives in the attribute set descriptor (the names, types and order of attributes) with information that lives solely on the attribute arrays (flag state, strides, etc). This is a frequent source of complexity as it involves navigating both data structures simultaneously to fully understand the array data.
* Allows for turning the existing AttributeSet constructors into delegating constructors and adds a new constructor that contains all the logic for initializing the new AttributeSet so this lives in one place.
* Provides a way to steal attribute array data from an existing attribute set during construction, as an intentional design decision means that there are no methods for resizing attribute arrays once created. 
* Offers a clean mechanism for merging many AttributeSets into one - you create AttributeSet::Info classes out of each and merge those (as they contain all required information to do so), then create a new AttributesSet from the final AttributeSet::Info. @Idclip note - this method is purely a placeholder right now, it needs #631 to be merged to share this portion.

(@richhones @kubaroth)

### DIFF
--- a/openvdb/points/AttributeSet.cc
+++ b/openvdb/points/AttributeSet.cc
@@ -5,6 +5,7 @@
 
 #include "AttributeSet.h"
 #include "AttributeGroup.h"
+#include "AttributeArrayString.h"
 
 #include <algorithm> // std::equal
 #include <string>
@@ -57,67 +58,74 @@ namespace {
 // AttributeSet implementation
 
 
-AttributeSet::AttributeSet()
-    : mDescr(new Descriptor())
+AttributeSet::AttributeSet(const Info& info, Index arrayLength,
+    const AttributeArray::ScopedRegistryLock* lock,
+    AttributeSet* existingAttributeSet)
+    : mDescr(info.descriptorPtr())
+    , mAttrs(info.size())
 {
+    std::unique_ptr<AttributeArray::ScopedRegistryLock> localLock;
+    if (!lock) {
+        localLock.reset(new AttributeArray::ScopedRegistryLock);
+        lock = localLock.get();
+    }
+
+    const MetaMap& meta = mDescr->getMetadata();
+    bool hasMetadata = meta.metaCount();
+
+    for (const auto& namePos : mDescr->map()) {
+        const size_t& pos = namePos.second;
+        const Info::Array& arrayInfo = info.arrayInfo(pos);
+        Metadata::ConstPtr metadata;
+        if (hasMetadata)    metadata = meta["default:" + namePos.first];
+        bool constantStride = arrayInfo.constantStride;
+        const Index stride = arrayInfo.stride;
+
+        // attempt to steal array from the existingAttributeSet
+
+        size_t existingPos = existingAttributeSet ?
+            existingAttributeSet->find(namePos.first) : INVALID_POS;
+
+        // can only steal if array has been found and stride and size match
+
+        AttributeArray::Ptr array;
+        if (existingPos != INVALID_POS) {
+            array = existingAttributeSet->removeAttributeUnsafe(existingPos);
+            // if array stride or size does not match, need to create a new array
+            if (array->stride() != stride || array->size() != arrayLength) {
+                array.reset();
+            }
+        }
+
+        if (!array) {
+            array = AttributeArray::create(mDescr->type(pos), arrayLength,
+                stride, constantStride, metadata.get(), lock);
+        }
+
+        // transfer hidden and transient flags
+        if (arrayInfo.hidden)       array->setHidden(true);
+        if (arrayInfo.transient)    array->setTransient(true);
+
+        mAttrs[pos] = array;
+    }
+    // delete the contents of the existing attribute set
+    if (existingAttributeSet) {
+        existingAttributeSet->clear();
+    }
 }
 
 
 AttributeSet::AttributeSet(const AttributeSet& attrSet, Index arrayLength,
     const AttributeArray::ScopedRegistryLock* lock)
-    : mDescr(attrSet.descriptorPtr())
-    , mAttrs(attrSet.descriptor().size(), AttributeArray::Ptr())
+    : AttributeSet(Info(attrSet), arrayLength, lock)
 {
-    std::unique_ptr<AttributeArray::ScopedRegistryLock> localLock;
-    if (!lock) {
-        localLock.reset(new AttributeArray::ScopedRegistryLock);
-        lock = localLock.get();
-    }
-
-    const MetaMap& meta = mDescr->getMetadata();
-    bool hasMetadata = meta.metaCount();
-
-    for (const auto& namePos : mDescr->map()) {
-        const size_t& pos = namePos.second;
-        Metadata::ConstPtr metadata;
-        if (hasMetadata)    metadata = meta["default:" + namePos.first];
-        const AttributeArray* existingArray = attrSet.getConst(pos);
-        const bool constantStride = existingArray->hasConstantStride();
-        const Index stride = constantStride ? existingArray->stride() : existingArray->dataSize();
-
-        AttributeArray::Ptr array = AttributeArray::create(mDescr->type(pos), arrayLength,
-            stride, constantStride, metadata.get(), lock);
-
-        // transfer hidden and transient flags
-        if (existingArray->isHidden())      array->setHidden(true);
-        if (existingArray->isTransient())   array->setTransient(true);
-
-        mAttrs[pos] = array;
-    }
 }
 
 
 AttributeSet::AttributeSet(const DescriptorPtr& descr, Index arrayLength,
     const AttributeArray::ScopedRegistryLock* lock)
-    : mDescr(descr)
-    , mAttrs(descr->size(), AttributeArray::Ptr())
+    : AttributeSet(Info(descr), arrayLength, lock)
 {
-    std::unique_ptr<AttributeArray::ScopedRegistryLock> localLock;
-    if (!lock) {
-        localLock.reset(new AttributeArray::ScopedRegistryLock);
-        lock = localLock.get();
-    }
-
-    const MetaMap& meta = mDescr->getMetadata();
-    bool hasMetadata = meta.metaCount();
-
-    for (const auto& namePos : mDescr->map()) {
-        const size_t& pos = namePos.second;
-        Metadata::ConstPtr metadata;
-        if (hasMetadata)    metadata = meta["default:" + namePos.first];
-        mAttrs[pos] = AttributeArray::create(mDescr->type(pos), arrayLength,
-            /*stride=*/1, /*constantStride=*/true, metadata.get(), lock);
-    }
 }
 
 
@@ -125,6 +133,14 @@ AttributeSet::AttributeSet(const AttributeSet& rhs)
     : mDescr(rhs.mDescr)
     , mAttrs(rhs.mAttrs)
 {
+}
+
+
+void
+AttributeSet::clear()
+{
+    mDescr = std::make_shared<Descriptor>();
+    mAttrs.clear();
 }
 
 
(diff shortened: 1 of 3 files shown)


openvdb/points/AttributeSet.cc
```diff
@@ -1353,10 +1369,118 @@ AttributeSet::Descriptor::read(std::istream& is)
...
+        os << "\t\t\tGroup: " << (arrayInfo.group ? "yes" : "no") << std::endl;
+        os << "\t\t\tString: " << (arrayInfo.string ? "yes" : "no") << std::endl;
+        os << "\t\t\tHidden: " << (arrayInfo.hidden ? "yes" : "no") << std::endl;
+        os << "\t\t\tTransient: " << (arrayInfo.transient ? "yes" : "no") << std::endl;
+    }
+    os << "\tGroups: " << mDescriptor->groupMap().size() << " group(s)" << std::endl;
+    for (const auto& namePos : mDescriptor->groupMap()) {
+        os << "\t\t[" << namePos.second << "] " << namePos.first << std::endl;
+    }
+    os << "\tMetadata: " << mDescriptor->getMetadata().metaCount() << " metadata element(s)" << std::endl;
```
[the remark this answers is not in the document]
danrbailey: Not sure this is particularly performance-critical, but I fixed this anyway. Thanks.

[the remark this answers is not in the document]
Idclip: Doubt it, but why would we not!

openvdb/points/AttributeSet.cc
```diff
@@ -79,45 +75,57 @@ AttributeSet::AttributeSet(const AttributeSet& attrSet, Index arrayLength,
 
     for (const auto& namePos : mDescr->map()) {
         const size_t& pos = namePos.second;
+        const Info::Array& arrayInfo = info.arrayInfo(pos);
         Metadata::ConstPtr metadata;
         if (hasMetadata)    metadata = meta["default:" + namePos.first];
-        const AttributeArray* existingArray = attrSet.getConst(pos);
-        const bool constantStride = existingArray->hasConstantStride();
-        const Index stride = constantStride ? existingArray->stride() : existingArray->dataSize();
+        bool constantStride = arrayInfo.constantStride;
```
kubaroth: Was the `const` dropped on purpose in `bool constantStride`?

openvdb/points/AttributeSet.cc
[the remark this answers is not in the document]
kubaroth: According to https://en.cppreference.com/w/cpp/io/manip/endl

> In many implementations, standard output is line-buffered, and writing '\n' causes a flush anyway ... unnecessary endl only degrades the performance of file output, not standard output.

kubaroth:
Hey Dan, I'm taking another look at the list of members of the AttributeSet::Array struct and was wondering if adding following would require much work?
- type
- isUniform
- codec
- shared

danrbailey:
> Hey Dan, I'm taking another look at the list of members of the AttributeSet::Array struct and was wondering if adding following would require much work?
> 
> * type
> * isUniform
> * codec
> * shared
>   EDIT: Perhaps I'm missing that the above ones can be accessed with a descriptor?

The type and codec values can be retrieved from the descriptor - `const NamePair& Descriptor::type(size_t pos) const`. Think Nick already commented to that effect, but the first value is the value type, the second is the codec.

Whether the array is shared or not, you can retrieve from the attribute set - `AttributeSet::isShared(size_t pos)`. However, for the print binary, I'm not sure it's necessary to include this because it'll never be true - there's no concept of instancing when the grids are deserialized so an attribute array couldn't be shared across multiple leaf nodes or grids.

That leaves whether an array is uniform or not. I hesitated a little on whether to include this, but decided not to. The primary purpose of this new data structure is for attribute set construction and for this purpose it's redundant. Arrays should always be constructed as uniform and then expanded when first populated (either implicitly through attempting to set values or explicitly). Providing a new mechanism for producing a non-uniform array would be unnecessarily confusing I think. I don't think this feature is out of place in the print binary so that would mean having to iterate over all the arrays, apologies I didn't think of this when I first proposed that this class would alleviate this.

kubaroth:
Thanks for clarification.

openvdb/points/AttributeSet.cc
danrbailey: Thanks, fixed.

danrbailey:
I don't believe there to be any more feedback needing addressed for this PR. @Idclip - if you're happy can you approve please?

danrbailey:
> I don't believe there to be any more feedback needing addressed for this PR. @Idclip - if you're happy can you approve please?

@Idclip - can you have a last look over my changes please? I believe I've addressed your feedback now.

openvdb/points/AttributeSet.cc
```diff
@@ -1353,10 +1369,119 @@ AttributeSet::Descriptor::read(std::istream& is)
...
 
+////////////////////////////////////////
+
+// AttributeSet::Info implementation
+
+
+AttributeSet::Info::Info(const AttributeSet::Descriptor::Ptr& descriptor)
+    : mDescriptor(descriptor)
+    , mArrayInfo(descriptor->size(), Array())
+{
```
Idclip: What's the point on this constructor? AS far as I can tell it currently serves no purpose and makes the API more confusing when using this class.

openvdb/points/AttributeSet.h
```diff
@@ -540,6 +558,82 @@ class OPENVDB_API AttributeSet::Descriptor
...
+
+////////////////////////////////////////
+
+/// @brief A convenience class that collates the Descriptor with metadata that
+/// is only available on each AttributeArray, such as whether the array is
+/// uniform or not. This is useful for querying and during construction of new
+/// AttributeSets.
+/// @note For performance reasons, this class uses copy-on-write for storing
+/// the Descriptor so any modification of the source Descriptor also requires
+/// a deep-copy for the data in this class to remain valid.
```
Idclip: This comment doesn't make sense. Are you trying to say that if modify the attributes you need to rebuild an instance of this class?

openvdb/points/AttributeSet.h
```diff
@@ -540,6 +558,82 @@ class OPENVDB_API AttributeSet::Descriptor
...
+    const Descriptor& descriptor() const { return *mDescriptor; }
+    /// @brief Return a shared pointer to the descriptor
+    Descriptor::Ptr descriptorPtr() const { return mDescriptor; }
+    /// @brief Return the number of array attributes
+    size_t size() const { return mDescriptor->size(); }
+
+    /// @brief Return an Array metadata object by name
+    /// @param name The name of the array
+    /// @throws LookupError if name does not exist
+    Array& arrayInfo(const Name& name);
```
Idclip: What's the purpose of returning modifiable Array objects with any of these member methods?

openvdb/points/AttributeSet.h
```diff
@@ -540,6 +558,82 @@ class OPENVDB_API AttributeSet::Descriptor
...
+    /// @details This method extracts and stores Array metadata from the
+    /// AttributeArray objects stored within it
+    explicit Info(const AttributeSet& attributeSet);
+
+    /// @brief Return the descriptor
+    Descriptor& descriptor() { return *mDescriptor; }
+    /// @brief Return the descriptor
+    const Descriptor& descriptor() const { return *mDescriptor; }
+    /// @brief Return a shared pointer to the descriptor
+    Descriptor::Ptr descriptorPtr() const { return mDescriptor; }
```
Idclip: Do we really need all 3 of these? The one which returns a ptr provides all access types

openvdb/points/AttributeSet.h
```diff
@@ -69,7 +72,19 @@ class OPENVDB_API AttributeSet
...
+    /// from an existing attribute set.
+    /// @param info the attribute info
+    /// @param arrayLength the desired length of the arrays in the new AttributeSet
+    /// @param lock an optional scoped registry lock to avoid contention
+    /// @param attributeSet an existing attribute set that can be stolen from
+    /// @note if provided, the attribute set in this constructor will be
+    /// cleared on completion of this constructor
+    AttributeSet(const Info& info, Index arrayLength,
+        const AttributeArray::ScopedRegistryLock* lock = nullptr,
+        AttributeSet* attributeSet = nullptr);
```
Idclip: I don't understand the purpose of the info class here. It seems that the same behaviour could be achieved with a bool `steal` flag. I'm not sure the benefit of consolidating the code justifies how convoluted this makes it (i.e. an info class which is expected to hold a descriptor also matched by a provided attribute set). You could, for example, achieve the same consolidation with a different function.

openvdb/points/AttributeSet.h
```diff
@@ -540,6 +558,82 @@ class OPENVDB_API AttributeSet::Descriptor
...
+    /// @throws LookupError if idx out of range
+    Array& arrayInfo(size_t idx);
+    /// @brief Return an Array metadata object by index
+    /// @param idx The index of the array
+    /// @throws LookupError if idx out of range
+    const Array& arrayInfo(size_t idx) const;
+
+    /// Placeholder method for merging another Info object into this one
+    /// @throws NotImplementedError
+    void merge(const AttributeSet::Info&);
```
Idclip: Could just not introduce this now?

openvdb/points/AttributeSet.h
```diff
@@ -540,6 +558,82 @@ class OPENVDB_API AttributeSet::Descriptor
...
+
+    /// @brief Print information for debugging
+    /// @param os the output stream to write to
+    void print(std::ostream& os = std::cout) const;
+
+private:
+    Descriptor::Ptr mDescriptor = std::make_shared<Descriptor>();
```
Idclip: At first glance, this class seems to me to be a "viewer" class which helps consolidate data from both the attribute set and descriptor. I question why the descriptor that's stored here is modifiable through the member methods. This introduces another layer of complexity. It seems that methods like `merge` would do better to return new instances of the descriptor.

openvdb/points/AttributeSet.cc
danrbailey: There's an `AttributeSet(const Descriptor::Ptr& descriptor)` constructor. Introducing a matching constructor here means that all AttributeSet construction can be done from an Info instance which keeps the logic in one place.

I consider it an advanced option to build and manipulate an Info class yourself for AttributeSet construction. For most use cases, I expect users to continue to use the other AttributeSet constructors directly.

openvdb/points/AttributeSet.h
danrbailey: It cannot be achieved with a new `steal` flag - the current constructor takes a const reference which is what is expected of a copy constructor. Even making the reference non-const would be confusing. There's no simple solution here in my opinion and adding another constructor with largely duplicate logic from the other two is not ideal. I don't even think the status quo is a good solution - having two similar, but subtly different pieces of logic for AttributeSet construction is not ideal in my opinion.

I consider the `Info` instance to be an implementation detail of the `AttributeSet(const AttributeSet&)` and `AttributeSet(const DescriptorPtr& descriptor)`, it's not required to know the internal mechanism for construction when using this API. I consider the `AttributeSet(const Info&,...)` constructor to be the advanced constructor, which you'd only use in certain specific cases.

openvdb/points/AttributeSet.h
danrbailey: Yes exactly. I agree the wording can be improved here.

openvdb/points/AttributeSet.h
danrbailey: I don't see a problem with doing this. There's a (small) performance penalty with copying the shared ptr, you use the one that's most appropriate.

There's also precedence for exposing multiple different access methods elsewhere in the codebase (`Grid::baseTreePtr()`, `Grid::baseTree()`, `Grid::constBaseTree()` for example).

openvdb/points/AttributeSet.h
danrbailey: You can use it in combination with the `AttributeSet(const Info& ...)` constructor to compose exactly how you'd like to construct the desired AttributeSet. I should highlight that there's no link between this `Info` data and the member descriptor. It's purely transitory metadata that is passed down to the constructed AttributeArray.

openvdb/points/AttributeSet.h
danrbailey: If possible, I would prefer to make progress on agreeing on the public API, before we move on to discussing this quite complex method that has a lot of implementation detail in it. (I merely included it here to help illustrate the intention of this class).

openvdb/points/AttributeSet.h
danrbailey: It does increase the complexity, but it's primarily done like this for performance reasons. Doing what you're suggesting doesn't scale to wanting to merge hundreds or even thousands of Info classes or Descriptors.

This class is serving two different use cases - however I consider the "viewer" use case to be mainly a side benefit, the real purpose is a composable description of what you want the AttributeSet to contain. (I did consider two different classes, one for each use case, but I preferred the simplicity of putting everything in one class). 

I would be open to changing the name if you think we can make this primary use-case clearer, `Info` is a pretty generic name.

richhones:
I finally got round to looking at this following it being raised at the last TSC meeting. I think that similar to @kubaroth 's comments asking for the type and codec and @Idclip 's comments about storing the Descriptor::Ptr that I think this doesn't really go all the way of being properly convenient for the general case as this been designed specifically for the case of what you needed for merging. Either it should go further to simplify the AttributeSet/Descriptor stuff - my suggestion would be to remove the Descriptor from the Info class and store the type and codec information in the Array struct. Alternatively just put this internal to the merging code. This API is already pretty confusing and think that the inclusion of the Descriptor in this convenience class doesn't mean this class helps that very much. If you still have to know to query the Descriptor for types and codecs then what benefit does the rest of the Info class have over just using the AttributeSet itself.

I imagine in merging you want to merge Infos and so Descriptors need to be merged in the current setup so removing them from this struct would simplify that process completely no?

PR #653, opened 2020-02-28
Sync Develop

Idclip:


Idclip:
@danrbailey until this is automated, perhaps we should also be targeting the `develop` branch

danrbailey:
> @danrbailey until this is automated, perhaps we should also be targeting the `develop` branch

You don't need to make a PR, only the master branch is protected. The TSC have write permissions, so just push to the develop branch yourself as needed (as I just did).

PR #654, opened 2020-02-28
CMake improvements for Static builds

Idclip:
A fair amount of changes here, but they're mainly the same set of changes for each Find module. Key changes:

 - New `USE_STATIC_DEPENDENCIES` option to force CMake to only find static libraries.
 - New `USE_WINDOWS_STATIC_RUNTIME` option to always link to the static CRT on Windows. We should look to using https://cmake.org/cmake/help/v3.15/prop_tgt/MSVC_RUNTIME_LIBRARY.html in the future.
 - Homogenised FindModule logic for static libraries
 - FindTBB now adds the required `_NO_IMPLICIT_LINKAGE` defines for Windows.
 - FindModules now attempt to set the found library type for imported targets
 - FindModules all use `GNUInstallDirs`
 - Removed `OPENVDB_DLL` being applied to static builds
 - Fixed OpenEXR linking issues where `OPENVDB_DLL` was being defined in `PlatformConfig.h` by defining `OPENVDB_OPENEXR_STATICLIB` if OpenEXR is static. Note that this platform header can be removed with the removal of the makefiles.
 - Improved the support for linking against blosc if it has been built with external deps for lz4, zstd, zlib and snappy
 - Improved the support for FindOpenVDB on Windows and with a static build of OpenVDB


CMakeLists.txt
```diff
@@ -89,6 +89,14 @@ know it to be required.]=] OFF)
...
   "Disable the implicit linking of Boost libraries on Windows" ON "WIN32" OFF)
 option(USE_CCACHE "Build using Ccache if found on the path" ON)
+option(USE_STATIC_DEPENDENCIES [=[
+Only search for and use static libraries. If OFF the shared versions of requried libraries are prioritised, falling
+back to static libraries. Forcing individual static dependencies can be enabled by setting XXX_USE_STATIC_LIBS
+to ON, where XXX is the package name. On Windows, this has the important function of changing the default C Runtime
+Library (CRT) linkage to MT from MD for all projects.]=] OFF)
+cmake_dependent_option(USE_WINDOWS_STATIC_RUNTIME
+  "Causes all projects to use the multithread, static version of the C run-time library. Defaults to the value "
+  "of USE_STATIC_DEPENDENCIES." ${USE_STATIC_DEPENDENCIES} "WIN32" OFF)
```
Idclip: Turns out this doesn't work, though if `USE_STATIC_DEPENDENCIES` is ON, `USE_WINDOWS_STATIC_RUNTIME` should really be ON too. I'll update this PR to set `USE_WINDOWS_STATIC_RUNTIME` to ON, with an explicit message

Idclip:
Declining in favour of #706 and #708

PR #657, opened 2020-03-02
Fixes for building openvdb_houdini on Windows

Idclip:


Idclip:
@e4lam tagging for visibility

PR #660, opened 2020-03-03
VDB Combine in Flatten All mode shouldn't error with no second input

jmlait:
Before the collation re-write, the VDB Combine would silently do nothing if the second input was empty and Flatten All B into A was toggled.   The new drop down menu, however, errors in this case.  By testing the b iterator, we can avoid generating a pair in this case and restore the previous behaviour.

Bug: 103362


jmlait:
Okay, I clearly messed up this PR!

Issue #257, opened 2018-08-06
Remove unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [DensityOp](https://github.com/dreamworksanimation/openvdb/blob/18568814988362db3b6d661491a8179e6cec71de/openvdb_houdini/houdini/SOP_OpenVDB_Rasterize_Points.cc#L1547 "Destructor for DensityOp")
- [OpFactory::setObsoleteParms](https://github.com/dreamworksanimation/openvdb/blob/a7a1abde7c955d5a83acac2c82341497d73576c2/openvdb_houdini/houdini/ParmFactory.cc#L1232)
- [TestDenseSparseTools:tearDown](https://github.com/dreamworksanimation/openvdb/blob/a7a1abde7c955d5a83acac2c82341497d73576c2/openvdb/unittest/TestDenseSparseTools.cc#L44)

MichaelQuell:
You seem to be right!

Idclip:
Resolved by #672 #685

elfring:
Thanks for your small source code improvements.

* [Were any update candidates left over](https://github.com/AcademySoftwareFoundation/openvdb/blob/78b445dc8c86285539cccc512ebfbd4e96dcda26/openvdb/tools/LevelSetUtil.h#L223 "Update candidate: MaskInteriorVoxels::operator()()")?
* How do you think about to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability-delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

Idclip:
It seems that the original PR to address this (#672) took the examples you listed literally as the only cases of these in the VDB codebase. I'll re-open this issue.

To your second point, I agree that something like `clang-tidy` would be a great additional to our CI for things like this. We'd of course need to figure out which features/configurations we'd want to utilise. It's been discussed before but there's currently no tracked  issue or ticket. I'll raise this again in the next steering committee.

grvtgpt:
I would like you to assign this issue to me.

Luc35b:
I'll work on fixing this issue.

harika1314:
I wanted to work on this issue ,can you assign this to me

DevKartikBhardwaj:
[...]

PR #672, opened 2020-03-20
Remove unnecessary null pointer checks #257

kubaroth:
Removed checks according to cpp guidelines linked in the original post.

### DIFF
--- a/openvdb/unittest/TestDenseSparseTools.cc
+++ b/openvdb/unittest/TestDenseSparseTools.cc
@@ -14,7 +14,7 @@ class TestDenseSparseTools: public CppUnit::TestCase
 {
 public:
     virtual void setUp();
-    virtual void tearDown() { if (mDense) delete mDense;}
+    virtual void tearDown() { delete mDense; }
 
     CPPUNIT_TEST_SUITE(TestDenseSparseTools);
     CPPUNIT_TEST(testExtractSparseFloatTree);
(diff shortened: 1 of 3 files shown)


openvdb/unittest/TestDenseSparseTools.cc
```diff
@@ -14,7 +14,7 @@ class TestDenseSparseTools: public CppUnit::TestCase
 {
 public:
     virtual void setUp();
-    virtual void tearDown() { if (mDense) delete mDense;}
+    virtual void tearDown() {delete mDense;}
```
danrbailey: Looks good, thanks. One minor comment, I'd expect spaces around the statement:

```{ delete mDense; }``` not ```{delete mDense;}```

I haven't looked through the cpp guidelines to check that this is written there, but that feels more consistent with the rest of the codebase.

Don't worry about the Houdini debug build check failing, that's failing elsewhere, but the DCO will need to be fixed before we can accept.

kubaroth: Yeah, I wasn't sure about this and decided to wait for comments first. BTW has there been any discussion before to introduce a clang format tool etc to streamline that process?

danrbailey: I believe we talked about it at one point, but it's a fairly time-consuming thing to introduce and we don't get _a lot_ of external contributions. As a result, our focus regarding infrastructure has been on improving other higher priority things - CMake, windows builds, CI, etc.

danrbailey:
Merged, thanks!

PR #674, opened 2020-03-20
Migrate To GitHub Actions

danrbailey:
* Trigger GitHub Actions build workflow on pull_request event as well as push event
* Delete Azure Pipelines and Cirrus CI
* Skip the Houdini build if the secret key is not set (this will be temporary)


ci/build_houdini.sh
```diff
@@ -6,26 +6,28 @@ COMPILER="$1"
 RELEASE="$2"
 EXTRAS="$3"
 
-cd hou
-source houdini_setup_bash
-cd -
+if [ -d "hou" ]; then
```
Idclip: This is necessary to skip this step if the secret is missing? Can we document this here? Also is there a way to force this to fail in our repo or wherever we expect the Houdini secret to always exist? I don't want this silently failing on us

danrbailey: It will fail silently for *all* PRs as secrets cannot be passed to forks. This is only going to be temporary, but I need to change this piecemeal as there are quite a few moving parts here.

danrbailey:
Note - Windows failure is due to a recent vcpkg change that has been addressed here (https://github.com/microsoft/vcpkg/pull/10456), but not merged yet.

ci/build_houdini.sh
Idclip: Okay, can we update this with the same logic though to make things more consistent? I.e. `if [ "$HOUPASS" == "" ]; then` rather than checking the dir

Idclip: > It will fail silently for all PRs as secrets cannot be passed to forks
Should we be removing the azure pipeline Houdini builds until this is fixed?

danrbailey: > Okay, can we update this with the same logic though to make things more consistent?

Why does it need to be consistent? The Houdini secret is not currently passed into build_houdini as a variable. Trying to do so complicates migration to the new Houdini download API (as they use different secret variables).

> Should we be removing the azure pipeline Houdini builds until this is fixed?

I'd like to, yes. It's time-consuming maintaining both CI platforms and trying to use the same builds scripts for both while in the process of migrating. Note that this just applies to PRs, not regular builds.

Idclip: > Why does it need to be consistent?

Because it's weird that it depends on the creation of a directory that could be changed unintentionally

> I'd like to, yes. It's time-consuming maintaining both CI platforms and trying to use the same builds scripts for both while in the process of migrating. Note that this just applies to PRs, not regular builds.

Sure, but it's perhaps just as time consuming to work through a set of changes and commit them only to realise issues with the Houdini build. I'll refer to the TSC.

danrbailey: > > Why does it need to be consistent?
> 
> Because it's weird that it depends on the creation of a directory that could be changed unintentionally

I don't find it weird. I think this is acceptable for a temporary solution in code that is not part of our shipped software, so I'm not considering changing this.

> > I'd like to, yes. It's time-consuming maintaining both CI platforms and trying to use the same builds scripts for both while in the process of migrating. Note that this just applies to PRs, not regular builds.
> 
> Sure, but it's perhaps just as time consuming to work through a set of changes and commit them only to realise issues with the Houdini build. I'll refer to the TSC.

I have added a single H18 azure build back in and _temporarily_ duplicated the ci scripts so that I can make changes to the way this works in GitHub Actions without risking breaking Azure.

Idclip: > I have added a single H18 azure build back in and temporarily duplicated the ci scripts so that I can make changes to the way this works in GitHub Actions without risking breaking Azure.

This is great - the azure build doesn't need to necessarily block the merge but having some level of confidence of the Houdini build before merge is a must.

> I don't find it weird. I think this is acceptable for a temporary solution in code that is not part of our shipped software, so I'm not considering changing this.

I've probably misunderstood, but can you not pass ` ${{ secrets.HOUPASS }}` through to build_houdini? It's weird that *any* test depends on the existence of a directory (that fact that its temporary doesn't matter) - but because it's temporary I agree its not an issue if there's not easy alternative. I don't understand what something being "part of our shipped software" has to do with how our CI works.

danrbailey: > > I don't find it weird. I think this is acceptable for a temporary solution in code that is not part of our shipped software, so I'm not considering changing this.
> 
> I've probably misunderstood, but can you not pass ` ${{ secrets.HOUPASS }}` through to build_houdini? It's weird that _any_ test depends on the existence of a directory (that fact that its temporary doesn't matter) - but because it's temporary I agree its not an issue if there's no easy alternative. I don't understand what something being "part of our shipped software" has to do with how our CI works.

I'm changing this to use the new Houdini download API so all these keys will have to change as well as how they interact with all these scripts (see this draft PR for more detail https://github.com/AcademySoftwareFoundation/openvdb/pull/637). Can you please approve as-is.

Idclip:
> Note - Windows failure is due to a recent vcpkg change that has been addressed here ([microsoft/vcpkg#10456](https://github.com/microsoft/vcpkg/pull/10456)), but not merged yet.

This seems to have been merged, any chance we can get the windows build to pass on this PR?

danrbailey:
> > Note - Windows failure is due to a recent vcpkg change that has been addressed here ([microsoft/vcpkg#10456](https://github.com/microsoft/vcpkg/pull/10456)), but not merged yet.
> 
> This seems to have been merged, any chance we can get the windows build to pass on this PR?

It has been merged, but the version of vcpkg in GitHub Actions has not been updated yet. I would have to dig deeper about figuring out how to update this manually.

Idclip:
Okay, seeing as the appveyor build isn't even running any more I guess this is okay go to in as is?

PR #675, opened 2020-03-25
Improved Morphology

Idclip:
This PR demonstrates some proposed improvements to the Morphology class in Morphology.h. To avoid a complicated diff, I've pushed this up as a draft PR which contains all changes in a new Header and an example binary with timings which can be built with the old and new methods. Note that there are a few areas that still need investigating, mainly:

 - How to handle active tiles for erosion and dilation. In most cases (but not all), performing the dilation/erosion over a voxelized mask grid is more efficient.
 - Erosion 18/26 neighbours, using "inverse dilation" can be faster than erosion 6 in some cases.
 - Temporary support for an erosion/dilation mask, aimed to improve performance but I may remove.

Important changes:

 - dialteVoxels/erodeVoxel both only operate over leaf level values and ignore active tiles. Topology should be voxelized before creating this class.
 - The LeafCache scatter/methods have entirely been copied, but the class has been renamed to NodeMaskOp and can perform dilation/erosion on a single leaf node (expected to exist within the given tree).
 - Both erosion and dilate are able to consolidate dense/empty nodes on the fly to speed up subsequent iterations. This is primarily effective for larger dilation iterations.
 - Various unused subclasses removed

Timings here:

https://docs.google.com/spreadsheets/d/1hlgC5NPPWybjJjZ5F1GyWHxePXnhJVdWuUiwcNiXYTI/edit?usp=sharing

### DIFF
--- a/openvdb/cmd/CMakeLists.txt
+++ b/openvdb/cmd/CMakeLists.txt
@@ -81,6 +81,13 @@ endif()
 
 ##### VDB binaries
 
+add_executable(morph morph.cpp)
+target_link_libraries(morph ${OPENVDB_BINARIES_DEPENDENT_LIBS})
+if(OPENVDB_ENABLE_RPATH)
+  set_target_properties(morph PROPERTIES INSTALL_RPATH "${RPATHS}")
+endif()
+install(TARGETS morph DESTINATION bin)
+
 #### vdb_print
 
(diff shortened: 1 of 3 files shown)


Idclip:
Declining this - I will be up-streaming a new version next week

PR #678, opened 2020-03-28
resolves #613 : Added CMake option to disable pkg-config

### LINKED ISSUE #613: Add CMake Option to Disable Pkg-Config
Add a USE_PKGCONFIG option (that defaults to ON) to the root level CMakeLists.txt which globally disables the use of pkg-config.

As an example use case, vcpkg apply a patch that removes pkg-config in OpenVDB which would be better to be a CMake option (https://github.com/microsoft/vcpkg/blob/master/ports/openvdb/0001-remove-pkgconfig.patch).

(Bonus points for submitting a PR to vcpkg to upgrade OpenVDB to 7.0.0 and remove this patch in the process).

KartikShrivastava:
Added CMake option `USE_PKGCONFIG` to the root level CMakeLists.txt, followed by checks inside `openvdb/cmake/*.cmake` files according to patch mentioned in #613 .

Signed-off-by: KartikShrivastava <shrivastavakartik19@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,7 @@ list will be searched.]=] OFF)
 option(OPENVDB_USE_DEPRECATED_ABI "Bypass minimum ABI requirement checks" OFF)
 option(OPENVDB_FUTURE_DEPRECATION "Generate messages for upcoming deprecation" OFF)
 option(USE_COLORED_OUTPUT "Always produce ANSI-colored output (GNU/Clang only)." OFF)
+option(USE_PKGCONFIG "Globally disable/enable the use of pkg-config" ON)
 
 set(SYSTEM_LIBRARY_PATHS "" CACHE STRING [=[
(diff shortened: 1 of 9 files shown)


KartikShrivastava:
Hi, sorry for commenting this here but is there is place where I can discuss few things about my gsoc proposal with openvdb mentors? 
I'm looking forward to write a proposal and wondering if a mentor can review it before submission.

danrbailey:
> Hi, sorry for commenting this here but is there a place where I can discuss few things about my gsoc proposal with openvdb mentors?
> I'm looking forward to write a proposal and wondering if a mentor can review it before submission.

I think I'd recommend just sending a mail to openvdb-dev@lists.aswf.io. You can submit a draft proposal and point us to look at that as well.

Idclip:
lgtm

danrbailey:
Merged to new disablepkgconfig branch, thanks!

PR #679, opened 2020-03-30
Use the single-character search rather than string search.

jmlait:
The findfirstof can take a single char, which will be better than using a string of one character.


Idclip:
Not sure if this warrants a changelog update or not...

jmlait:
Yeah, not sure either, but considering the amount of effort it has taken to move this trivial 2-character change to this point, I'm loathe to put even more work into it!

PR #680, opened 2020-03-30
CMake Disable Pkg Config

danrbailey:


Idclip:
Main changelog also needs updating

danrbailey:
> Main changelog also needs updating

Missed adding that file, added now.

PR #690, opened 2020-04-02
Update CpuTimer to use C++11 chrono

danrbailey:
Re-enabling the Windows build has caused an issue in the CpuTimer unit test. On reading up on this, it appears that the sleep() mechanism offered by Windows (in particular) is not very accurate and merely ensures a sleep of at least a minimum duration and it can often be much larger. I believe this is showing up as a result of the Windows build using the latest TBB 2020 build where many tbb features have been deprecated - I haven't looked into it, but I suspect this is now falling through to std::this_thread::sleep_for().

I have addressed this using a non-asynchoronous while loop and updated the CpuTimer to use chrono instead of tbb's own time methods to remove these deprecated warnings as we look to support TBB 2020 in the future.


danrbailey:
I've had another look at this PR. std::chrono::time_point is ***not*** thread-safe, so looking at the implemention for tbb's tick_count which ***is*** thread-safe, I've done something similar, by simply storing the number of microseconds since the epoch instead of the time_point itself. With this change, the CpuTimer can now be read and written to from multiple threads safely. As a side-note, I couldn't actually find any examples where we do this, but it's a fairly trivial thing to support given that this class offers very limited functionality. This now can only represent microseconds, so I've clamped the value to a 64-bit signed integer (instead of a long long as used in TBB).

Finally, this implementation now also uses `std::chrono::steady_clock` instead of `std::chrono::high_resolution_clock` as it appears that high_resolution_clock is either aliased to steady_clock or system_clock depending on the platform, so we use the steady_clock directly as it's monotonic and designed for time durations.

openvdb/util/CpuTimer.h
```diff
@@ -162,8 +169,22 @@ class CpuTimer
...
+        // steady_clock is a monotonically increasing clock designed for timing duration
+        // note that high_resolution_clock is aliased to either steady_clock or system_clock
+        // depending on the platform, so it is preferrable to use steady_clock
+        const auto time_since_epoch =
+            std::chrono::steady_clock::now().time_since_epoch();
+        // cast time since epoch into microseconds (1 / 1000000 seconds)
+        const auto microseconds =
+            std::chrono::duration_cast<std::chrono::microseconds>(time_since_epoch).count();
+        // cast to a a 64-bit signed integer as this will overflow in 2262!
+        return static_cast<int64_t>(microseconds);
```
Idclip, line 183: Took me a second to realise 2262 refereed to the year 😄

Idclip (approved):
notes but lgtm

Issue #615, opened 2020-01-23
Use GEO_Detail::getBBox Instead of GEO_Detail::computeQuickBounds


danrbailey:
Various SOPs use computeQuickBounds() instead of getBBox(). Convert them over to use the correct method to fix bounding issues. SOPs affected include Clip, Fill, Points Group, Rasterize Points, Read, Remove Divergence.

Full description of the issue by Jeff:

> GEO_Detail::computeQuickBounds is including the point of the VDB in the bounding box. computeQuickBounds, as its name suggests, tries to be quick. So it uses ALL points as its bounds and then enlarges by all primitives.
> 
> The GEO_Detail::getBBox(), by comparison, will call into computeQuickBounds if it is safe. But if there are any odd primitives, like packed primitives and VDB primitives, it will instead compute the bounding box of all the primiitves; and then add any unused points. Since the origin of the VDB isn't unused, it won't be contained.
> 
> Long story short: use getBBox(). It should be just as fast for simple cases, and will handle VDBs/Packed Prims properly.

saepenumero:
Hi, may I work on this issue?

Idclip:
Currently in progress by @kubaroth

Idclip:
Resolved by #702

PR #697, opened 2020-04-09
Replace use of computeQuickBounds with getBBox.

kubaroth:
Addresses issue: Use GEO_Detail::getBBox Instead of GEO_Detail::computeQuickBounds #615
Signed-off-by: kubaroth <kubaroth@gmail.com>

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_Clip.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Clip.cc
@@ -576,7 +576,7 @@ SOP_OpenVDB_Clip::Cache::cookVDBSop(OP_Context& context)
                 }
             } else {
                 UT_BoundingBox box;
-                maskGeo->computeQuickBounds(box);
+                maskGeo->getBBox(&box);
 
                 clipBox.min()[0] = box.xmin();
                 clipBox.min()[1] = box.ymin();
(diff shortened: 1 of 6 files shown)


Idclip:
Looks good, I take it we're not concerned with breaking existing behaviour as its classified as a bug? @danrbailey any idea why the tests are failing?

danrbailey:
> Looks good, I take it we're not concerned with breaking existing behaviour as its classified as a bug? @danrbailey any idea why the tests are failing?

I think the PR just got made at a moment when the CI was broken. @kubaroth - everything should be working now, so I'd recommend git merging the main master branch into this one and then pushing that commit and seeing if the tests resolve themselves. (I attempted to kick it a few days ago by switching the branch, but that didn't help). Change looks good to me otherwise.

kubaroth:
Thanks for the comments. I've rebased onto master and updated PR. 
Btw should I use develop or master in the future? Looks like develop has been a bit behind.

danrbailey:
> Thanks for the comments. I've rebased onto master and updated PR.
> Btw should I use develop or master in the future? Looks like develop has been a bit behind.

Great stuff, looks like that has fixed it! Use master - we experimented with changing our default branch to develop for a short period, but decided the extra process overhead wasn't worth it and discovered we can adjust the target branch instead which is a good alternative. Thanks for the reminder, I'll double-check no outstanding PRs have been made into develop, then delete that branch to avoid confusion.

danrbailey:
Thanks @kubaroth - I've just merged and submitted to be upstreamed (#702).

kubaroth:
Super! Thanks @danrbailey.

PR #701, opened 2020-04-17
Added a basic MacOS CI build and test

Idclip:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: build
-      run: ./ci/build.sh clang++ Release 7 ON None
+      run: ./ci/build.sh clang++ Release 7 ON None -DOPENVDB_CXX_STRICT=ON
     - name: test
       run: ./ci/test.sh
 
@@ -22,7 +22,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: build
-      run: ./ci/build.sh clang++ Release 6 ON None
+      run: ./ci/build.sh clang++ Release 6 ON None -DOPENVDB_CXX_STRICT=ON
     - name: test
       run: ./ci/test.sh
 
@@ -33,7 +33,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: build
-      run: ./ci/build.sh clang++ Release 5 ON None
+      run: ./ci/build.sh clang++ Release 5 ON None -DOPENVDB_CXX_STRICT=ON
     - name: test
       run: ./ci/test.sh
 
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -180,3 +180,17 @@ jobs:
...
+  testmacos1015:
+    runs-on: macos-10.15
+    steps:
+    - uses: actions/checkout@v2
+    - name: install
+      shell: bash
+      run: ./ci/install_macos.sh
+    - name: build
+      shell: bash
+      run: ./ci/build.sh clang++ Release 7 ON SSE42 -DPython_ROOT_DIR=/usr/local/opt/python@3.8
```
Idclip, line 193: Annoying brew python installs 3.7 and boost_python install 3.8... it's _probably_ valid for to built against slightly different minor versions (https://stackoverflow.com/questions/28830653/build-boost-with-multiple-python-versions) but this is more correct. Also there's issues in the boost-python CMake which make this trickier. Yeah...

ci/build.sh
```diff
@@ -2,11 +2,18 @@
...
-RELEASE="$2"
-ABI="$3"
-BLOSC="$4"
-SIMD="$5"
+COMPILER="$1"; shift
+RELEASE="$1"; shift
+ABI="$1"; shift
+BLOSC="$1"; shift
+SIMD="$1"; shift
+CMAKE_EXTRA="$@"
```
Idclip, line 10: Changed this slightly to allow for arbitrary CMake commands after this positional args

ci/build.sh
```diff
@@ -2,11 +2,18 @@
...
+
+CXX_STRICT=ON
+if [ $(uname -s) == "Darwin" ]; then
+    echo "Disabling compiler warnings on MacOS"
+    CXX_STRICT=OFF
+fi
```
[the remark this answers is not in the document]
Idclip: I did this as `CXX_STRICT` should always be on for our CI, so I was working under the assumption that we'd fix this then remove this logic completely. But I can do as you've suggested, it just adds an extra argument to all builds.

.github/workflows/build.yml
Idclip: Done

ci/build_windows.sh
```diff
@@ -15,5 +20,9 @@ cmake \
   -DOPENVDB_BUILD_VDB_RENDER=ON \
   -DOPENVDB_BUILD_VDB_VIEW=ON \
   -DOPENVDB_BUILD_UNITTESTS=ON \
+  -DBOOST_ROOT="${BOOST_ROOT_1_72_0}" \
+  -DBOOST_INCLUDEDIR="${BOOST_ROOT_1_72_0}\boost\include" \
+  -DBOOST_LIBRARYDIR="${BOOST_ROOT_1_72_0}\lib" \
```
Idclip, line 25: Boost has changed on the windows 2019 serve. See the above links.

danrbailey:
Thanks @Idclip, I approved and merged this for you because I wanted to submit this PR that builds on this one - #705. We can make any subsequent changes in a new PR.

PR #703, opened 2020-04-20
Fix AttributeArray Default Value Bug

danrbailey:
This is bug that's been in the codebase for some time - setting the uniform value in a new AttributeArray was skipped if it was equal to the default value (to save doing this redundant operation), however it should have been checking against the default value for this AttributeArray instance _not_ the generic default value for this value type.

This didn't show up until recently because default metadata wasn't being supported properly but that was resolved in #590. Thanks to Nick for reporting this one.

### DIFF
--- a/openvdb/points/PointAttribute.h
+++ b/openvdb/points/PointAttribute.h
@@ -321,7 +321,13 @@ inline void appendAttribute(PointDataTreeT& tree,
     appendAttribute(tree, name, AttributeTypeConversion<ValueType, CodecType>::type(),
         strideOrTotalSize, constantStride, defaultValue, hidden, transient);
 
-    if (!defaultValue || !math::isExactlyEqual(uniformValue, defaultValue->value())) {
+    // if the uniform value is equal to either the default value provided
+    // through the metadata argument or the default value for this value type,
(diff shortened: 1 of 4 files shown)


openvdb/points/PointAttribute.h
```diff
@@ -321,7 +321,7 @@ inline void appendAttribute(PointDataTreeT& tree,
     appendAttribute(tree, name, AttributeTypeConversion<ValueType, CodecType>::type(),
         strideOrTotalSize, constantStride, defaultValue, hidden, transient);
 
-    if (!math::isExactlyEqual(uniformValue, Default<ValueType>::value())) {
+    if (!defaultValue || !math::isExactlyEqual(uniformValue, defaultValue->value())) {
```
Idclip, line 324: You shouldn't need to run this if `!defaultValue` and `math::isExactlyEqual(uniformValue, Default<ValueType>::value())` (i.e. uniform is 0). Something like:

```
 if (!defaultValue && math::isExactlyEqual(uniformValue, Default<ValueType>::value())) return;
 if (defaultValue  && math::isExactlyEqual(uniformValue, defaultValue->value())) return;
// collapse with uniformValue
```

PR #705, opened 2020-04-23
Use Debug Builds without Symbols to Fix CI

danrbailey:
Adds a new `DebugNoInfo` CMake build type which is kind of the opposite of `RelWithInfo`.

Switches the ABI7 debug build to use this build type (and switches the SSE42 build to release, not sure why it was ever set to debug).

I still can't switch on the Houdini debug build due to disk space issues. This can either be fixed in the short term by GitHub Actions resolving their current disk space issues (https://github.com/actions/virtual-environments/issues/709) or by stripping out some of the data in the Houdini install that we don't use (it is currently 4GB when uncompressed).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,7 +138,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: build
-      run: ./ci/build.sh clang++ Debug 7 ON None -DOPENVDB_CXX_STRICT=ON
+      run: ./ci/build.sh clang++ DebugNoInfo 7 ON None -DOPENVDB_CXX_STRICT=ON
 
   testabi7sse:
     runs-on: ubuntu-16.04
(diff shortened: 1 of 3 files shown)


Idclip:
Should this be supported in the root CMake instead of manually providing no CXX_FLAGS?
i.e. `if(CMAKE_BUILD_TYPE STREQUAL DebugNoInfo) set(CXX_FLAGS "") endif()` Does that even work? Looks good otherwise

danrbailey:
> Should this be supported in the root CMake instead of manually providing no CXX_FLAGS?
> i.e. `if(CMAKE_BUILD_TYPE STREQUAL DebugNoInfo) set(CXX_FLAGS "") endif()` Does that even work? Looks good otherwise

I can't think of any other use case for this mode besides this. That and the fact that it's not one of the CMake defaults suggests we should probably just keep this as a CI-only option for now?

Can you approve then I can update some of the other failing PRs?

Idclip:
> That and the fact that it's not one of the CMake defaults suggests we should probably just keep this as a CI-only option for now?

Sure, just concerned someone would see it and think it might be a supported CMake mode - if it's entirely for the CI, perhaps a more unique/specific name would be better? Even just "None" 
`-DCMAKE_CXX_FLAGS_None` or something. Will approve anyway.

danrbailey:
> > That and the fact that it's not one of the CMake defaults suggests we should probably just keep this as a CI-only option for now?
> 
> Sure, just concerned someone would see it and think it might be a supported CMake mode - if it's entirely for the CI, perhaps a more unique/specific name would be better? Even just "None"
> `-DCMAKE_CXX_FLAGS_None` or something. Will approve anyway.

Thanks! None would have been ok too. Think I'll just merge as-is though, not sure it's that important.

PR #706, opened 2020-04-23
Various improvements to CMake find modules

Idclip:
This PR is a partial implementation of #654 

 - Removing unnecessary DLL methods
 - Adding better lib type detection on UNIX
 - Added GNU path support
 - Added required defines for static TBB on Windows
 - Added required defines for static IlmBase and OpenEXR on Windows
 - Improved handling of finding static builds of OpenVDB

Signed-off-by: Nick Avramoussis <4256455+Idclip@users.noreply.github.com>

PR #708, opened 2020-04-27
Windows Static Builds + MSVC Runtime Options

Idclip:
This PR adds support for building with any MSVC Runtime Library configuration (/MD /MT etc). It is the second part of #654 (the first part being #706) but also includes a number of improvements to the original PR. Each commit is fairly isolated but as they are related I thought it best to include in a single PR. Reviewing the commits individually should provide clear differences.

 - Added support to FindBlosc for locating external blosc dependencies
 - Improved various CMake defines passed through the OpenVDB core lib
 - Improved installation directories for OpenVDB
 - Added an option to force CMake to only use static dependencies
 - Print CMake version through CI
 - Added support for CMAKE_MSVC_RUNTIME_LIBRARY on Windows
 - Added a Windows CI test for /MT

Release notes to be added.


.github/workflows/build.yml
```diff
@@ -164,12 +164,7 @@ jobs:
         echo "::add-path::${{github.workspace}}\build\openvdb\Release"
     - name: update_vcpkg
       shell: bash
-      run: |
-        # temporarily update vcpkg to fix tbb port
-        cd C:\vcpkg
-        git fetch
-        git checkout 97302ca1729eac9917c8a8977667601453795c90
-        bootstrap-vcpkg.bat
+      run: cd C:\vcpkg && git pull
```
Idclip, line 167: always update available ports (this is actually needed for the blosc external deps where they've recently changed the zstd library name)

CMakeLists.txt
```diff
@@ -90,6 +97,11 @@ know it to be required.]=] OFF)
 cmake_dependent_option(OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING
   "Disable the implicit linking of Boost libraries on Windows" ON "WIN32" OFF)
 option(USE_CCACHE "Build using Ccache if found on the path" ON)
+option(USE_STATIC_DEPENDENCIES [=[
+Only search for and use static libraries. If OFF the shared versions of requried libraries are prioritised, falling
+back to static libraries. Forcing individual static dependencies can be enabled by setting XXX_USE_STATIC_LIBS
+to ON, where XXX is the package name. On Windows this behaviour is less strict, with any located libraries assumed
+to be static. ]=] OFF)
```
Idclip, line 104: It's important to define `USE_STATIC_DEPENDENCIES` on windows even if you only have static libs installed as it's not possible to infer library types from the .libs alone. This will get better as we start to deprecate our Find modules in favour of 3rd party configuration files - but as we still need to support older versions and different pack-manager solutions, there's no single available fix.

Note that this does not impact C Runtime selection but is necessary for OpenEXR (where `OpenEXR_DLL` must *not* be defined) and Boost (where both shared and static libs are installed on the CI server).

CMakeLists.txt
```diff
@@ -269,6 +281,37 @@ elseif(CMAKE_CXX_STANDARD LESS ${MINIMUM_CXX_STANDARD})
...
+    endif()
+  endif()
+  if(NOT Boost_USE_DEBUG_RUNTIME)
+    set(Boost_USE_DEBUG_RUNTIME OFF)
+    if(CMAKE_MSVC_RUNTIME_LIBRARY STREQUAL MultiThreadedDebugDLL OR
+       CMAKE_MSVC_RUNTIME_LIBRARY STREQUAL MultiThreadedDebug)
+      set(Boost_USE_DEBUG_RUNTIME ON)
+    endif()
+  endif()
+endif()
```
Idclip, line 312: Boost should really do this itself

cmake/FindBlosc.cmake
```diff
@@ -231,6 +234,19 @@ if(Blosc_FOUND)
       INTERFACE_COMPILE_OPTIONS "${PC_Blosc_CFLAGS_OTHER}"
       INTERFACE_INCLUDE_DIRECTORIES "${Blosc_INCLUDE_DIRS}"
     )
+
+    # Blosc may optionally be compiled with external sources for
+    # lz4, snappy, zlib and zstd. Add them as interface libs if
+    # requested (there doesn't seem to be a way to figure this
+    # out automatically). Note that it's assumed these deps have
+    # been statically built if BLOSC_USE_STATIC_LIBS is ON so that
+    # zstd infers the correct name
```
Idclip, line 243: Second part of the comment here related to old code, will remove

cmake/FindBlosc.cmake
```diff
@@ -231,6 +234,19 @@ if(Blosc_FOUND)
...
+    if(BLOSC_USE_EXTERNAL_SOURCES)
+      set_target_properties(Blosc::blosc PROPERTIES
+        INTERFACE_LINK_DIRECTORIES "${Blosc_LIBRARY_DIRS}"
+        INTERFACE_LINK_LIBRARIES "lz4;snappy;zlib;zstd"
+      )
+    endif()
```
Idclip, line 249: No easy fix for this - vcpkg for some reason builds blosc against external deps. blosc itself can be built with any configuration of internal/external sources. afaik, blosc 1.5 does not install a config file. This is only required for vcpkg x64-windows-static variants but must be manually provided (see the new CI test)

ci/build_windows.sh
```diff
@@ -11,8 +12,13 @@ cd build
 #  https://github.com/actions/virtual-environments/issues/687
 #  https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md
 
+# print version
+cmake --version
+
 cmake \
   -G "Visual Studio 16 2019" -A x64 \
+  -DVCPKG_TARGET_TRIPLET="${VCPKG_DEFAULT_TRIPLET}" \
+  -DCMAKE_BUILD_TYPE=Release \
```
danrbailey, line 21: Isn't this set with `--config Release` below? Why does this need to be defined twice?

openvdb/CMakeLists.txt
```diff
@@ -145,6 +145,14 @@ set(OPENVDB_CORE_DEPENDENT_LIBS
   IlmBase::Half
 )
 
+if(WIN32)
+  if(OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING)
+      list(APPEND OPENVDB_CORE_DEPENDENT_LIBS
+        Boost::disable_autolinking  # add -DBOOST_ALL_NO_LIB
```
danrbailey, line 151: I wouldn't mind a quick comment explaining what this is for and why we need it?

.github/workflows/build.yml
danrbailey: Is this always needed or is this needed just until that change makes it way into the default vcpkg? I'd prefer this doesn't run from a different commit each time so that it's easier to reproduce and debug issues.

Idclip: It was necessary but I would anticipate I can now remove it - as you say, updating on each commit isn't ideal

ci/build_windows.sh
Idclip: I was confused by this as well, but they mean different things. Providing it as part of the cmake call tells CMake to generate a specific Makefile (or otherwise platform build system) configuration and exclude others. Providing it as an argument to --build tells CMake to only build that configuration. i.e. `cmake -DCMAKE_BUILD_TYPE=Release` `cmake --build --config Debug` would fail. The Windows Debug builds still need improving.

openvdb/CMakeLists.txt
Idclip: Boost headers contain #pragma commands on Windows which causes Boost libraries to be linked in automatically when including the appropriate header. Depending on how you're building boost, these naming conventions don't always match causing linker errors. This target turns that feature off. I'll add this comment.

PR #709, opened 2020-04-28
Fix move constructor compiler warning

danrbailey:
A new Clang compiler warning is detecting that the AttributeArray move constructor is being implicitly deleted due to the member mutex, despite it being marked as default. This shows up when using Apple Clang 11.0 (based off LLVM 8.0) that was introduced with Mac Catalina, and errors when OPENVDB_CXX_STRICT is enabled.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -91,6 +91,10 @@ Version 7.1.0 - In Development
     - Added missing TBB, OpenEXR and IlmBase defines for static builds on
       Windows through the relevant FindPackage modules.
     - Improved the logic in FindOpenVDB for static builds.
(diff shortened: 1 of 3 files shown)


Idclip:
Thanks @danrbailey  - does this mean the macos ci can be built with strict flags now? There may have been other issues but I can't remember. Approved regardless

danrbailey:
> Thanks @danrbailey - does this mean the macos ci can be built with strict flags now? There may have been other issues but I can't remember. Approved regardless

I tried to switch this strict warning back on, but ran into some warnings over gl deprecations in the vdbview binary. There may be others, but I'm going to merge this one as-is for now and we can submit new PRs to keep moving towards this goal.

PR #710, opened 2020-04-30
TSC Notes for April 30th, 2020.

jmlait:
Notes for April 30th.


Idclip:
#708 should also fix the windows CI issues we're seeing here

danrbailey:
> #708 should also fix the windows CI issues we're seeing here

That's quite a big PR - any chance you could just submit a PR that contains the fix for this by itself (assuming that will work)? Then we can keep things building and Jeff can merge master into this one to fix the build.

PR #711, opened 2020-05-13
change std::unexpected() to std::terminate()

marksisson:
Signed-off-by: Mark Sisson <5761292+marksisson@users.noreply.github.com>

Dynamic exception specification was deprecated in C++11, and removed in C++17.  This fixes compiler error when using C++17 compile feature by changing std::unexpected() to std::terminate().

### DIFF
--- a/openvdb/cmd/openvdb_lod.cc
+++ b/openvdb/cmd/openvdb_lod.cc
@@ -351,7 +351,7 @@ main(int argc, char *argv[])
     }
     catch (...) {
         OPENVDB_LOG_FATAL("Exception caught (unexpected type)");
-        std::unexpected();
+        std::terminate();
     }
 
     return exitStatus;


danrbailey:
> I don't mind this change, but I wonder why we don't just set EXIT_FAILURE and fall through? std::terminate doesn't define return value...
> 
> Mind you, I don't like EXIT_SUCCESS and EXIT_FAILURE either for a similar reason. I'd want the tools to be have the same on all platforms, and 0 to always be success.

It seems we're a bit inconsistent in this regard. vdb_print and vdb_render handle this case, but don't set the return code, vdb_view doesn't even handle it. I'll clean up the others too - from reading up on this, std::terminate seems to be _more_ correct, but I wouldn't mind if we settled on using EXIT_FAILURE instead here.

PR #718, opened 2020-05-22
GitHub Actions CI for Docs and Whitespace

danrbailey:
* Add whitespace CI check (and fix some whitespace issues)
* Add docs CI check (and fix some doxygen issues)
* Adjust rules for build workflow to only build on specific files

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,27 @@
 
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - '*'
+    paths:
+      - '.github/**'
+      - 'CMakeLists.txt'
+      - 'ci/**'
+      - 'cmake/**'
+      - 'openvdb/**'
+      - 'openvdb_houdini/**'
+  pull_request:
+    branches:
+      - '*'
+    paths:
+      - '.github/**'
+      - 'CMakeLists.txt'
+      - 'ci/**'
+      - 'cmake/**'
+      - 'openvdb/**'
+      - 'openvdb_houdini/**'
 
 jobs:
   testabi7:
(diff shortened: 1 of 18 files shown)


doc/changes.txt
```diff
@@ -38,9 +38,8 @@ New features:
   @vdblink::points::AttributeSet::removeAttribute()
   AttributeSet::removeAttribute@endlink for releasing ownership of attribute
   data.
-- Added @vdblink::points::AttributeSet::Descriptor::groupIndexCollision()
-  AttributeSet::Descriptor::groupIndexCollision@endlink for detecting group
-  index collisions when attempting to merge two Descriptors.
+- Added AttributeSet::Descriptor::groupIndexCollision for detecting group index
+  collisions when attempting to merge two Descriptors.
```
danrbailey, line 42: It looks like the @vdblink mechanism doesn't work for functions of a nested class, so I've removed them here.

.github/workflows/build.yml
```diff
@@ -1,7 +1,27 @@
...
+  pull_request:
+    branches:
+      - '*'
+    paths:
+      - '.github/**'
+      - 'CMakeLists.txt'
+      - 'ci/**'
+      - 'cmake/**'
+      - 'openvdb/**'
+      - 'openvdb_houdini/**'
```
Idclip, line 24: it would probably be better to use something like `paths-ignore: - '**.rst'  - '**.md`?

danrbailey: That's where I started out, but there are exceptions that would still force a build (openvdb/CHANGES, openvdb/doc/doxygen-config), so I figured it was better to do it like this.

Idclip: CHANGES should probably be a .md file and likewise we could add a .txt extension to the dox file? Probably be a nice solution that the directory masking - better to build very frequently when unnecessary than not at all

danrbailey: I'm going to just merge this for now so that we have the workflow in place, otherwise I'm playing catch up trying to keep this PR in sync with anything that gets merged in the meantime.

I don't disagree that we can do some additional reorganisation and tidy up these paths, but let's just do that in a subsequent pass if that's ok...

PR #719, opened 2020-05-23
Add Houdini CI caching workflow

danrbailey:
This new Houdini GitHub Actions workflow runs once a week and does the following:

* Download the latest Houdini production build using the new SideFX download API and the stored secret keys
* Strip out the headers, libraries and binaries required for building the OpenVDB Houdini plugin
* Try and build the OpenVDB Houdini plugin
* If successful, insert it into the GitHub Actions cache

With this PR alone, nothing is actually using the cache - using it will come in a subsequent PR.

I set this up to run on Wednesdays prior to our TSC meetings in case issues come up that need discussing. The intention is to deprecate the old Houdini download and install scripts, but for the purposes of migration I've introduced new ones (ci/download_houdini2.py, etc). For now, I'm just adding Houdini 18.0.

### DIFF
--- a/.github/workflows/houdini.yml
+++ b/.github/workflows/houdini.yml
@@ -1,0 +1,30 @@
+
+name: Houdini
+
+on:
+  schedule:
+    # run this workflow every wednesday at noon UTC
+    - cron:  '0 12 * * 3'
+
+# download the latest production version of Houdini, strip out headers,
+# libraries and binaries required for building OpenVDB and if the build
+# succeeds, put it into the GitHub Actions cache
+
+jobs:
+  houdini18_0:
+    runs-on: ubuntu-16.04
+    container:
+      image: aswf/ci-base:2019
+    steps:
+    - uses: actions/checkout@v2
+    - name: download_houdini
+      run: ./ci/download_houdini2.sh 18.0 ON ON ${{ secrets.HOUDINI_CLIENT_ID }} ${{ secrets.HOUDINI_SECRET_KEY }}
+    - name: install_houdini
+      run: ./ci/install_houdini2.sh
+    - name: build
+      run: ./ci/build_houdini.sh clang++ Release ON
+    - name: write_houdini_cache
+      uses: actions/cache@v1
+      with:
+        path: hou
+        key: vdb1-houdini18_0-${{ hashFiles('hou/hou.tar.gz') }}
--- a/ci/download_houdini2.py
+++ b/ci/download_houdini2.py
@@ -1,0 +1,169 @@
+#!/usr/local/bin/python
+#
+# Copyright Contributors to the OpenVDB Project
+# SPDX-License-Identifier: MPL-2.0
+#
+# Python script to download the latest Houdini builds
+# using the SideFX download API:
+#
+# https://www.sidefx.com/docs/api/download/index.html
+#
+# Authors: Dan Bailey, SideFX
+
+import time
+import sys
+import re
+import shutil
+import json
+import base64
+import requests
+import hashlib
+
+# this argument is for the major.minor version of Houdini to download (such as 15.0, 15.5, 16.0)
+version = sys.argv[1]
+only_production = True if sys.argv[2] == 'ON' else False
+user_client_id = sys.argv[3]
+user_client_secret_key = sys.argv[4]
+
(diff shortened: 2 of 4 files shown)


.github/workflows/houdini.yml
```diff
@@ -0,0 +1,30 @@
...
+      run: ./ci/download_houdini2.sh 18.0 ON ON ${{ secrets.HOUDINI_CLIENT_ID }} ${{ secrets.HOUDINI_SECRET_KEY }}
+    - name: install_houdini
+      run: ./ci/install_houdini2.sh
+    - name: build
+      run: ./ci/build_houdini.sh clang++ Release ON
+    - name: write_houdini_cache
+      uses: actions/cache@v1
+      with:
+        path: hou
+        key: vdb1-houdini18_0-${{ hashFiles('hou/hou.tar.gz') }}
```
Idclip, line 30: This is only building the houdini plugin every wednesday and never on Push/PR? Or is that being handled already? If not, I think the Houdini plugin should always be built on code change, or at the very least if `openvdb_houdini/**` changes. IMO this workflows only job should be to download and cache the houdini build. Something like:

```
      - name: houdini_cache
        id: houdini_cache
        uses: actions/cache@v1
        with:
          key: -${{ get_current_hash() }}  # Get the current houdini build hash from a python script (https://www.sidefx.com/docs/api/download/index.html)
          path: /location/of/houdini/install
      - name: download_houdini
        if: steps.houdini_cache.outputs.cache-hit != 'true'
        run: ./ci/download_houdini2.sh 
      - name: install_houdini
        if: steps.houdini_cache.outputs.cache-hit != 'true'
        run: ./ci/install_houdini2.sh 
```

danrbailey: > This is only building the houdini plugin every wednesday and never on Push/PR? Or is that being handled already? If not, I think the Houdini plugin should always be built on code change, or at the very least if `openvdb_houdini/**` changes. IMO this workflows only job should be to download and cache the houdini build. Something like:
> 
> ```
>       - name: houdini_cache
>         id: houdini_cache
>         uses: actions/cache@v1
>         with:
>           key: -${{ get_current_hash() }}  # Get the current houdini build hash from a python script (https://www.sidefx.com/docs/api/download/index.html)
>           path: /location/of/houdini/install
>       - name: download_houdini
>         if: steps.houdini_cache.outputs.cache-hit != 'true'
>         run: ./ci/download_houdini2.sh 
>       - name: install_houdini
>         if: steps.houdini_cache.outputs.cache-hit != 'true'
>         run: ./ci/install_houdini2.sh 
> ```

The intention of this PR is to update the Houdini install that's available in the GitHub Actions cache once a week (after checking OpenVDB builds successfully). The next step is modifying the existing build steps (in build.yml) to not download from the net, but from the GitHub Actions cache, this resolves the long-standing issue of secrets not being available from forks as the key matching rules allow retrieving data from the cache for branches and forks. 

I didn't submit the next phase of this (yet), because Houdini needs to be in the cache for that to work.

Idclip: Can this not be done with this PR? Disable the cron job (build on every push/pull) then re-enable once everything is working

danrbailey: No, it can't be done in one PR. 

Once this PR is merged in, it will kick off the `houdini` and `build` workflows concurrently on the master branch and the `build` one won't be able to access the cache until the `houdini` one completes, so at least for the first run, it is guaranteed to fail. Note that the key matching rules mean that a build on `master` can't read a cache generated from a "downstream" branch or PR.

A more sensible strategy is simply to do two PRs - the first that generates and writes the cache (which is what this is) and a second one that then reads it. We can even leave the workflow that generates the cache running for a week or two to assess stability and that the cron job is working correctly before we try and actually use the cache itself.

Re-reading your previous comment, I just wanted to clarify that the goal of this is to make sure that any job that is spawned from a branch or PR does not under _any_ circumstance download Houdini. That is the root of the problem with secret keys. So we _never_ test a PR against the very latest Houdini build, at most we test against a Houdini build that is one week old. If there is a new compatibility problem with a Houdini build, it's picked up by the `houdini` workflow _not_ by the `build` workflow, so it's abundantly clear where the problem lies and contributions are isolated from any issues relating to new builds of Houdini.

PR #720, opened 2020-05-26
Remove ABI=4 support

danrbailey:
As noted in our deprecation strategy, drop support for ABI=4 with the first minor release after a major release.

I hesitated about marking ABI=5 as deprecated now. I think that's a bit premature to do a whole year early though, as that would mean that only one version of Houdini (18.0) could be built without enabling OPENVDB_USE_DEPRECATED_ABI.


openvdb/version.h
```diff
@@ -58,27 +58,19 @@
...
-        PRAGMA(message("NOTE: ABI = 4 is deprecated, define OPENVDB_USE_DEPRECATED_ABI "
-            "to suppress this message"))
-    #endif
-#endif
+// #ifndef OPENVDB_USE_DEPRECATED_ABI
+//     #if OPENVDB_ABI_VERSION_NUMBER == 5
+//         PRAGMA(message("NOTE: ABI = 5 is deprecated, define OPENVDB_USE_DEPRECATED_ABI "
+//             "to suppress this message"))
+//     #endif
+// #endif
```
Idclip, line 73: Why comment this out?

danrbailey: Because it’s a bit premature to deprecate ABI=5, we put this back when we release 8.0?

Idclip: Ah understood -  I would actually advocate marking ABI 5 as deprecated in the same set of changes as this, but looks like this might need some discussion? Lets discuss in the TSC tomorrow before merging this

danrbailey: Sure

danrbailey:
As discussed, I've made a change to this PR to add the version number to the macro - OPENVDB_USE_DEPRECATED_ABI_5. Looking for any remaining feedback before I merge this.

Idclip:
Any reason you opted for a complete string over `OPENVDB_USE_DEPRECATED_ABI=5` ?

danrbailey:
> Any reason you opted for a complete string over `OPENVDB_USE_DEPRECATED_ABI=5` ?

Mainly that a string is ever so slightly more challenging to script, so for instance you couldn't add a line like the following to the code:

#define OPENVDB_USE_DEPRECATED_ABI OPENVDB_ABI_VERSION_NUMBER

danrbailey:
> > Any reason you opted for a complete string over `OPENVDB_USE_DEPRECATED_ABI=5` ?
> 
> Mainly that a string is ever so slightly more challenging to script, so for instance you couldn't add a line like the following to the code:
> 
> #define OPENVDB_USE_DEPRECATED_ABI OPENVDB_ABI_VERSION_NUMBER

And by having to change the ifndef line in the code every time we bump this version, we'd also cause a conflict in any attempt to comment out that check.

Idclip:
Understood, I'm fine with this

PR #722, opened 2020-05-27
Updated FindOpenVDB to support pyopenvdb and openvdb_houdini

Idclip:

### DIFF
--- a/cmake/FindOpenVDB.cmake
+++ b/cmake/FindOpenVDB.cmake
@@ -14,7 +14,9 @@ Use this module by invoking find_package with the form::
     [version] [EXACT]      # Minimum or EXACT version
     [REQUIRED]             # Fail with error if OpenVDB is not found
     [COMPONENTS <libs>...] # OpenVDB libraries by their canonical name
-                           # e.g. "openvdb" for "libopenvdb"
+                           # e.g. "openvdb" for "libopenvdb",
+                           # "pyopenvdb" for the python plugin
+                           # "openvdb_houdini" for the houdini plugin
     )
 
 IMPORTED Targets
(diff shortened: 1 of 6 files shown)


cmake/FindOpenVDB.cmake
```diff
@@ -224,6 +287,28 @@ list(APPEND _OPENVDB_LIBRARYDIR_SEARCH_DIRS
...
+set(OPENVDB_PYTHON_PATH_SUFFIXES
+  ${CMAKE_INSTALL_LIBDIR}/python
+  ${CMAKE_INSTALL_LIBDIR}/python2.7
+  ${CMAKE_INSTALL_LIBDIR}/python3
+  lib64/python
+  lib64/python2.7
+  lib64/python3
+  lib/python
+  lib/python2.7
+  lib/python3
```
danrbailey, line 299: In theory a user could have installed multiple versions of OpenVDB - say one against Python2 and one against Python3. Is it worth providing a mechanism for them to pick the specific paths they want here?

Idclip:
> In general these changes look fine to me - is there any way we could actually test this module through CI? A simple binary application that tried to link against OpenVDB using this mechanism?

Yes absolutely - I can do this as part of this PR if you like. Though this is 1/2 PRs that AX depends on, so it'll block that.

cmake/FindOpenVDB.cmake
Idclip: Yeah the variable ` ${OPENVDB_${COMPONENT}_ROOT}` and ` ${OPENVDB_${COMPONENT}_LIBRARYDIR}` should do what you're suggesting here. i.e:
`-DOPENVDB_pyopenvdb_ROOT="/path/to/custom/python/install"`

Idclip:
bump

Idclip:
I will update the changelogs and merge this

PR #724, opened 2020-06-03
Bump Actions cache to v2

danrbailey:
The scheduled action ran as expected today and succeeded, however GitHub Actions cache v1 does not support reading or writing caches when using a "schedule" trigger. Bumping to v2 which does support this.


danrbailey:
Fixed the failing trailing space check here - #725

PR #728, opened 2020-06-07
Modernize to use override to avoid gcc's -Wsuggest-override warning

e4lam:


e4lam:
> Looks good! Even better would have been using final instead of override. :)

Yeah, well, that's a sweep for another day.

PR #729, opened 2020-06-08
Removed boost/constant and numeric methods

Idclip:


openvdb/tools/PointPartitioner.h
```diff
@@ -84,7 +82,11 @@ class PointPartitioner
     using ConstPtr = SharedPtr<const PointPartitioner>;
 
     using IndexType = PointIndexType;
-    using VoxelOffsetType = typename boost::int_t<1 + (3 * BucketLog2Dim)>::least;
+
+    static constexpr Index bits = 1 + (3 * BucketLog2Dim);
+    using VoxelOffsetType = typename std::conditional<(bits <= 16),
+        int16_t, typename std::conditional<(bits <= 32), int32_t, int64_t>::type>::type;
```
Idclip, line 88: this is signed, should actually be `<` not `<=`

openvdb/math/FiniteDifference.h
```diff
@@ -24,6 +24,12 @@ namespace OPENVDB_VERSION_NAME {
 namespace math {
 
 
+/// @brief  Pi constant taken from Boost to match old behaviour
+template <typename T> inline constexpr T pi() { return 3.141592653589793238462643383279502884e+00; }
+template <> inline constexpr float pi() { return 3.141592653589793238462643383279502884e+00F; }
+template <> inline constexpr double pi() { return 3.141592653589793238462643383279502884e+00; }
+template <> inline constexpr long double pi() { return 3.141592653589793238462643383279502884e+00L; }
```
danrbailey, line 31: I think this a good idea, I considered proposing this before. I would expect this constant to be defined in math/Math.h though as it's a bit surprising it would be in the FiniteDifference header.

Also, probably worth putting a brief comment for the future to mention that pi will be available as a constant in C++20 (std::numbers::pi).

Idclip: done

openvdb/tools/PointPartitioner.h
Idclip: fixed

PR #730, opened 2020-06-08
Removed boost/mpl methods in favour of new methods on openvdb::TypeList

Idclip:

One important consideration in this PR is compile times - my implementation probably differs from boost so this may have been impacted (could have been improved as well as impaired). Need to test.


openvdb/Types.h
```diff
@@ -316,6 +316,70 @@ namespace internal {
...
+//     using type = T;
+// };
+
+// template <typename... Ts, typename U, typename... Us>
+// struct TSMakeUniqueImpl<TypeList<Ts...>, U, Us...> {
+//     using type =
+//         typename std::conditional<(std::is_same<U, Ts>::value || ...),
+//            typename TSMakeUniqueImpl<TypeList<Ts...>, Us...>::type,
+//            typename TSMakeUniqueImpl<TypeList<Ts..., U>, Us...>::type >::type;
+// };
```
Idclip, line 380: `(...)` requires c++17 - we don't actually use `TypeList::Unique` anywhere so this is unnecessary. But a c++14 solution should be fairly straightforward.

danrbailey:
Very cool stuff, nice one! I like the readability of the new methods much better than the old ones.

Would be good to get the `TestTypes::testTypeList()` unit test updated to test this new functionality, otherwise I'm happy with this.

openvdb/Types.h
Idclip: Fixed for c++14

Idclip:
I've added a set of static asserts to `TestTypes::testTypeList()` to test these new methods

danrbailey:
This is great. In case you weren't aware, the Houdini plugin is failing at the moment and will likely need updating to use this (see the Azure job).

Idclip:
> In case you weren't aware, the Houdini plugin is failing at the moment and will likely need updating to use this (see the Azure job).

Oops thanks, will fix this

Idclip:
> I'd quite like @kmuseth to look over this change before we merge it in.

I won't merge any of these boost PRs until we've got consensus in the tsc

kmuseth:
I'll review it today!

openvdb/Types.h
```diff
@@ -316,6 +316,71 @@ namespace internal {
 
 // Implementation details of TypeList
 
+struct NullType {};
```
kmuseth, line 319: Please add a comment describing what this "dummy" type is used for

kmuseth: this comment applies to more then the struct listed above (please see review summary)

Idclip:
@kmuseth I've added documentation to all the type internal meta-programming structures - I think I went a bit overboard, I added doxygen documentation but didn't realise that there exists a `@cond OPENVDB_TYPES_INTERNAL` doxygen rule which disabled this documentation generation by default. Oh well, it's there should we ever want it in the future! 

Regarding the comment on compile times, this was something I wanted to test and hadn't yet. I didn't anticipate any adverse effects but I wanted to make sure that this solution performed at least as fast as the boost solution. I've tested with `gcc` and am getting similar results (across 32 threads, building the .so, the binaries and the tests, obviously no `ccache`):

before: real 4m52.941s
after: real 4m53.310s

kmuseth:
great, thanks Nick

PR #731, opened 2020-06-08
Removed boost/string methods in favour of utility methods in util/Name.h

Idclip:

### DIFF
--- a/openvdb/openvdb/Grid.cc
+++ b/openvdb/openvdb/Grid.cc
@@ -2,10 +2,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
 #include "Grid.h"
-
 #include "Metadata.h"
-#include <boost/algorithm/string/case_conv.hpp>
-#include <boost/algorithm/string/trim.hpp>
+#include "util/Name.h"
 #include <mutex>
 
 
@@ -124,8 +122,8 @@ GridBase::stringToGridClass(const std::string& s)
 {
     GridClass ret = GRID_UNKNOWN;
     std::string str = s;
-    boost::trim(str);
-    boost::to_lower(str);
+    openvdb::string::trim(str);
+    openvdb::string::to_lower(str);
     if (str == gridClassToString(GRID_LEVEL_SET)) {
         ret = GRID_LEVEL_SET;
     } else if (str == gridClassToString(GRID_FOG_VOLUME)) {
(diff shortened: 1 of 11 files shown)


openvdb/io/Compression.cc
```diff
@@ -22,11 +21,12 @@ compressionToString(uint32_t flags)
...
-    std::vector<std::string> words;
-    if (flags & COMPRESS_ZIP) words.push_back("zip");
-    if (flags & COMPRESS_BLOSC) words.push_back("blosc");
-    if (flags & COMPRESS_ACTIVE_MASK) words.push_back("active values");
-    return boost::join(words, " + ");
+    std::string descr;
+    if (flags & COMPRESS_ZIP) descr += "zip + ";
+    if (flags & COMPRESS_BLOSC) descr += "blosc + ";
+    if (flags & COMPRESS_ACTIVE_MASK) descr += "active values + ";
+    if (!descr.empty()) descr.resize(descr.size() - 3);
```
danrbailey, line 28: This is pretty clunky. You can't have ZIP and BLOSC defined at the same time, so I think I'd prefer this to just be explicit for readability:

```
if (flags & COMPRESS_ZIP) {
    if (flags & COMPRESS_ACTIVE_MASK)    return "zip + active values";
    else                                 return "zip";
}
// etc
```

openvdb/util/Name.h
```diff
@@ -35,6 +37,77 @@ writeString(std::ostream& os, const Name& name)
     os.write(&name[0], size);
 }
 
+namespace util {
```
danrbailey, line 40: I think I'd be tempted to put these in a new namespace for extra readability - openvdb::util::split() doesn't tell you anything about what it is you're actually splitting for example. Perhaps openvdb::string::split() would be clearer?

There are other non-util namespaces defined in this util sub-directory (such as openvdb::logging) so it wouldn't be breaking with precedence there.

openvdb/util/Name.h
```diff
@@ -35,6 +37,77 @@ writeString(std::ostream& os, const Name& name)
...
+{
+    ltrim(s);
+    rtrim(s);
+}
+
+inline void to_lower(std::string& s)
+{
+    std::transform(s.begin(), s.end(), s.begin(),
+        [](auto c){ return std::tolower(c); });
+}
```
danrbailey, line 107: Would be worth adding unit tests for these new methods.

openvdb/util/Name.h
Idclip: Happy to go with `openvdb::string` for now

openvdb/io/Compression.cc
Idclip: Technically, as far as this function is concerned, blosc and zip flags can exist at the same time as they're different bit toggles i.e. compressionToString(7) == "zip + blosc + active values". I realise that it shouldn't be (isn't) possible for them to both to exist, but seems that this method works without that context

danrbailey: Perhaps we error or assert that both are not supplied at once, I still prefer the explicit solution here for readability.

danrbailey:
Looks good to me. If you just make the switch from openvdb::util to openvdb::string and maybe add a line of two of comments to each of these new methods as way of documentation, I'm happy to approve this.

danrbailey:
@Idclip - can this one be merged now?

PR #733, opened 2020-06-10
Install blosc 1.5.0 for macos CI

Idclip:


Idclip:
> Thanks Nick, we should really put some effort into figuring out what we do about later versions of Blosc.

Agreed

PR #736, opened 2020-06-12
Upstream #728 (Fix suggest-override warnings)

danrbailey:


danrbailey:
> Can you also delete other temporary branches that exist in the vdb repo that have been made from this workflow

I just updated our branch protection rules so that this is possible (see email).

PR #738, opened 2020-06-12
Lock H18.0 to ABI=6, otherwise use ABI=7

danrbailey:


cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -328,8 +328,10 @@ if(Houdini_VERSION VERSION_LESS 17)
   set(OPENVDB_HOUDINI_ABI 4)
 elseif(Houdini_VERSION VERSION_LESS 18)
   set(OPENVDB_HOUDINI_ABI 5)
-else()
+elseif(Houdini_VERSION VERSION_LESS 18.5)
   set(OPENVDB_HOUDINI_ABI 6)
+else()
+  set(OPENVDB_HOUDINI_ABI 7)
```
Idclip, line 334: H18.5 guaranteed to use abi 7?

danrbailey: I'm not sure we can talk about unreleased software, so all this is saying is that H18.0 is the last major/minor release of Houdini that uses ABI=6. :)

Idclip: Has Houdini ever bumped the VDB ABI in-between major versions? Assuming it follow previous patterns, it will still be on ABI 6? So what you're proposing here will need changing again and may break releases of OpenVDB 7-8 with 18.5? Would it be better to predict ABI 6 for 18.5 and change this to `VERSION_LESS 19`? Otherwise I don't see the point of changing this yet until we have more information?

danrbailey: The next version of Houdini (whatever that might be) will be ABI=7:

https://github.com/AcademySoftwareFoundation/openvdb/blame/master/tsc/meetings/2020-04-02.md#L74

PR #739, opened 2020-06-12
Remove PRM_CHOICELIST_EXCLUSIVE from VDB from Polygons/Particles

danrbailey:
Added "v" and "point.v" as first option to the drop-down with a separator. Any others we should include here?


Idclip:
To clarify, there's no breaking changes here for existing setups?

danrbailey:
We didn't quite get to talking about this one. To answer your question Nick, there are no breaking changes for existing setups, it is just about what the drop-down looks like. I wanted to get some clarification on including just "point.v" as an explicit option? Any others we should consider? "point.N", "point.Cd"? It feels a little odd to me to only include one attribute in this list?

PR #740, opened 2020-06-16
TSC Minutes June 4th

Idclip:
Sorry for delay with these

### DIFF
--- a/tsc/meetings/2020-06-04.md
+++ b/tsc/meetings/2020-06-04.md
@@ -1,0 +1,72 @@
+Minutes from 51st OpenVDB TSC meeting, June 4th, 2020, (GMT)
+
+Attendees: *Nick* A., *Jeff* L., *Ken* M., *Dan* B.
+
+Additional Attendees: Johannes Meng (Intel), JT Nelson (Blender),
+Peter Cheng (DW), Andre Pradhana (DW), Bruce Chernia (Intel),
+Ibrahim Sani Kache (DW)
+
+Regrets: *Peter* C.
+
+Agenda:
+
+1) Confirm quorum
+2) Secretary
+3) Fast Sweeping
+4) Python 3
+5) Other
+6) Next meeting
+
+
+1) Quorum was confirmed.
+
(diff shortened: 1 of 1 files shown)


tsc/meetings/2020-06-04.md
```diff
@@ -0,0 +1,72 @@
...
+
+5) Other
+
+Discussion into OpenVDB's dependencies. Ken, one of the biggest complaints he
+hears is VDB's dependency list. Having some of these dependencies as VFX
+platform candidates could help - mainly referring to blosc. Jeff, LLVM (as a
+requirement of AX) should not really be a part of the VFX platform (due to its
+volatility w.r.t API/ABI), with downstream software instead opting for always
+hiding LLVM symbols. Boost python is heavily relied on - pybind would be a
+reasonable alternative but that also does not appear as part of the VFX platform.
```
meshula, line 60: The reason pybind11 doesn't appear as part of the VFX platform, is that the platform biases to specifying only those dependencies that cause versioning problems. boost python falls into that category, but pybind11 doesn't.

Our plan for Imath is to drop boost python and port it to pybind11.

tsc/meetings/2020-06-04.md
```diff
@@ -0,0 +1,72 @@
...
+VFX 2021 container has now be drafted.
+https://vfxplatform.com/
+
+5) Other
+
+Discussion into OpenVDB's dependencies. Ken, one of the biggest complaints he
+hears is VDB's dependency list. Having some of these dependencies as VFX
+platform candidates could help - mainly referring to blosc. Jeff, LLVM (as a
+requirement of AX) should not really be a part of the VFX platform (due to its
+volatility w.r.t API/ABI), with downstream software instead opting for always
```
meshula, line 58: One option we are considering is ASWF specific forks of problematic libraries, but we haven't taken any decisions or actions on that. LLVM was discussed, but our preliminary thoughts, in the context of OSL, were that we felt it was too volatile, as mentioned here, to take on attempting to track it.

PR #743, opened 2020-06-18
Fetch Houdini from cache

danrbailey:
This pulls the Houdini install from the GitHub Actions cache rather than attempting to download it directly. This side-steps all the issues with secret keys - once a week the Houdini install is updated in a regular scheduled job that _does_ have access to secret keys.

This is only for one job for now (the main Houdini 18.0 job). Once this has run successfully for a little while, I'll tidy up some of the old scripts, extend this to the other Houdini jobs and then finally deprecate Azure.

### DIFF
--- a/ci/download_houdini2.sh
+++ b/ci/download_houdini2.sh
@@ -9,11 +9,11 @@ HOUDINI_SECRET_KEY="$5"
 
 if [ "$HOUDINI_CLIENT_ID" == "" ]; then
     echo "HOUDINI_CLIENT_ID GitHub Action Secret needs to be set to install Houdini builds"
-    exit 1
+    exit 0
 fi
 if [ "$HOUDINI_SECRET_KEY" == "" ]; then
     echo "HOUDINI_SECRET_KEY GitHub Action Secret needs to be set to install Houdini builds"
-    exit 1
+    exit 0
 fi
 
 pip install --user requests
--- a/ci/install_houdini2.sh
+++ b/ci/install_houdini2.sh
@@ -2,6 +2,8 @@
 
 set -e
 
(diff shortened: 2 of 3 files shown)


ci/download_houdini2.sh
```diff
@@ -9,11 +9,11 @@ HOUDINI_SECRET_KEY="$5"
 
 if [ "$HOUDINI_CLIENT_ID" == "" ]; then
     echo "HOUDINI_CLIENT_ID GitHub Action Secret needs to be set to install Houdini builds"
-    exit 1
+    exit 0
```
danrbailey, line 12: This script is only run in the weekly Houdini workflow. Note that this change has been primarily made for repos that are forks of OpenVDB. Right now, if you fork OpenVDB and create a fork/master that mirrors the parent/master, you'll introduce a recurring GitHub Action workflow once a week that fails unless you add the secret keys or otherwise disable GitHub Actions. This makes it silently fail if it doesn't find the secret key.

ci/install_houdini2.sh
```diff
@@ -2,6 +2,8 @@
...
 set -e
 
-# move hou tarball into top-level and untar
-cp hou/hou.tar.gz .
-tar -xzf hou.tar.gz
+if [ -d "hou" ]; then
+    # move hou tarball into top-level and untar
+    cp hou/hou.tar.gz .
+    tar -xzf hou.tar.gz
+fi
```
danrbailey, line 9: This is also for fork repos, where the job will no longer fail if Houdini has not been downloaded, instead it'll silently succeed. It should be possible to create a fork of OpenVDB and make a change to the core library with successful GitHub Actions workflows running without needing to first setup Houdini installs.

PR #746, opened 2020-06-18
Ovdb 133 fast sweeping

kmuseth:


openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///
+/// @note Strictly speaking a fog volume is normalized to the range [0,1] but this
+///       method accepts a scalar volume with an arbitary range, as long as the it
+///       includes the @a isoValue.
+///
+/// @details Topology of output grid is identical to that of the input grid, except
+///          active tiles in the input grid will be converted to active voxels
+///          in the output grid!
+///
+/// @throw  RuntimeError if the iso-surface does not intersected any active
```
jmlait, line 91: This does not sound like an exceptional situation.  The result of that should just be an empty SDF grid?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///         voxels or if it intersects any active tiles in @a fogGrid.
```
jmlait, line 92: ....  I can see why this is the case, but it reminds me of the problems with volume blur.  Both can be fixed in the same manner, so I don't think this should delay this PR at all; but it does add more urgency to getting a system to densify boundaries.

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+/// @param sdfGrid  An approximate signed distance field to the specified iso-surface.
+///
+/// @param isoValue A value which defines a smooth iso-surface that
+///                 intersects active voxels in @a sdfGrid.
+///
+/// @param nIter    Number of iterations of the fast sweeping algorithm.
+///                 each performing 2^3 = 8 sweeps.
+///
+/// @note The only difference between this method and fogToSdf, defined above, is the
+///       convention of the sign of the output distance field.
```
jmlait, line 116: With this in mind, why do have fog and sdf as aliases for positive and negative?

To answer my question, I like having these methods for clear legibility.  But I'm wondering if they can't all just invoke a common exposed method that takes an isosurface and direction parameter?  This would also make it easier to expose this on an Extrapolate SOP.

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///         of the input fog volume.
+///
+/// @param fogGrid  Scalar (floating-point) volume from which an
+///                 iso-surface can be defined.
+///
+/// @param isoValue A value which defines a smooth iso-surface that
+///                 intersects active voxels in @a fogGrid.
+///
+/// @param nIter    Number of iterations of the fast sweeping algorithm.
+///                 each performing 2^3 = 8 sweeps.
```
jmlait, line 81: I think I mentioned in the meeting that as future work it would be cool to have a tolerance for convergence...

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///
+/// @note The only difference between this method and fogToSdf, defined above, is the
+///       convention of the sign of the output distance field.
+///
+/// @details Topology of output grid is identical to that of the input grid, except
+///          active tiles in the input grid will be converted to active voxels
+///          in the output grid!
```
jmlait, line 120: Isn't that implied because we are building an SDF, so the values must vary, so naturally constant tiles will densify?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///
+/// @note Strictly speaking a fog volume is normalized to the range [0,1] but this
+///       method accepts a scalar volume with an arbitary range, as long as the it
+///       includes the @a isoValue.
+///
+/// @details Topology of output grid is identical to that of the input grid, except
+///          active tiles in the input grid will be converted to active voxels
+///          in the output grid!
+///
+/// @throw  RuntimeError if the iso-surface does not intersected any active
```
jmlait, line 157: In this case the grid should be set to the background value and not be a runtime error.

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+/// @param sdfGrid  Input signed distance field to to be dilated.
+///
+/// @param dilation Numer of voxels that the input SDF will be dilated.
+///
+/// @param nn       Stencil-pattern used for dilation
+///
+/// @param nIter    Number of iterations of the fast sweeping algorithm.
+///                 each performing 2^3 = 8 sweeps.
+///
+/// @details Topology will change as a result of the dilation
```
jmlait, line 283: Details I want: Is this re-initailizating from the zero-crossing; or is it just updating from the boundary of the old bandwidth?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+template<typename GridT>
+typename GridT::Ptr
+dilateSdf(const GridT &sdfGrid,
+          int dilation,
+          NearestNeighbors nn = NN_FACE,
+          int nIter = 1);
+
+/// @brief Expand an existing signed distance fild into a mask
```
jmlait, line 291: Maybe:

Fills mask by extending an existing signed distance field into it.

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///       includes the @a isoValue.
+///
+/// @details Topology of output grids are identical to that of the input grid, except
+///          active tiles in the input grid will be converted to active voxels
+///          in the output grids!
+///
+/// @throw  RuntimeError if the iso-surface does not intersected any active
+///         voxels or if it intersects any active tiles in @a fogGrid.
+template<typename GridT, typename OpT>
+std::array<typename GridT::Ptr, 2>
```
Idclip, line 229: Perhaps return a std::pair or GridPtrVec instead?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+    typename GridT::Ptr sdfGrid() { return mGrid1; }
+    typename GridT::Ptr extGrid() { return mGrid2; }
+
+    void initSdf(const GridT &fogGrid, ValueT isoValue, bool isInputSdf);
+
+    template <typename OpT>
+    void initExt(const GridT &fogGrid, const OpT &op, ValueT isoValue, bool isInputSdf);
+
+    // use tools::NN_FACE_EDGE for improved dilation
+    void initDilate(const GridT &sdfGrid, int dilation, NearestNeighbors nn = NN_FACE);
```
Idclip, line 381: What does the above comment refer to, is this a todo? What's the advantage over performing face edge dilation instead of face or face edge vertex in context of these algorithms

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+    // Private classes to initialize the grid and construct
+    // mPagedArray with voxel coordinates.
+    template<typename>
+    struct MaskKernel;//   initialization to extand a SDF into a mask
+    template<typename>
+    struct InitExt;
+    struct InitSdf;
+    struct DilateKernel;// initialization to dilate a SDF
+    struct MinMaxKernel;
+    struct SweepingKernel;// Private class to perform the actual concurrent sparse fast sweeping
```
Idclip, line 407: This is a broader design question and I'm not asking or implying we should meke any changes to the internals now - but what's the benefit of these private sub structs? I'd instead propose these were inline lambdas within the forward api methods i.e. `initDilate` doesn't construct a `DilateKernel` but instead implements `DilateKernel::run` itself

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+///
+/// @details The algorithm used below for parallel fast sweeping was first publised in:
+///          "New Algorithm for Sparse and Parallel Fast Sweeping: Efficient
+///          Computation of Sparse Distance Fields" by K. Museth, ACM SIGGRAPH Talk,
+///          2017, http://www.museth.org/Ken/Publications_files/Museth_SIG17.pdf
+
+#ifndef OPENVDB_TOOLS_FASTSWEEPING_HAS_BEEN_INCLUDED
+#define OPENVDB_TOOLS_FASTSWEEPING_HAS_BEEN_INCLUDED
+
+//#define BENCHMARK_FAST_SWEEPING
```
Idclip, line 35: Perhaps a better solution here would be to allow the enabling of the `CpuTimer` wholescale. i.e. build openvdb with `-DENABLE_PROFILING` or some such and, in `CpuTimer`, you do something like:

```
#ifdef ENABLE_PROFILING
// normal implementation of timer
#else 
struct CpuTimer { CpuTimer(std::string&) { /*noop*/} ... };
#endif

This would help clear up the code somewhat

File: openvdb/tools/FastSweeping.h [hunk 13]
```diff
@@ -0,0 +1,1432 @@
...
+        changeLevelSetBackground(mgr, Unknown);//multi-threaded
+
+ #ifdef BENCHMARK_FAST_SWEEPING
+        timer.restart("Dilating and updating mgr (parallel)");
+        //timer.restart("Dilating and updating mgr (serial)");
+#endif
+
+        const int delta = 5;
+        for (int i=0, d = dilation/delta; i<d; ++i) dilateActiveValues(mgr, delta, nn, IGNORE_TILES);
+        dilateActiveValues(mgr, dilation % delta, nn, IGNORE_TILES);
```
Idclip, line 620: Why have these multiple calls to `dilateActiveValues`? Is this the same as:
```
Why is this call to `dilateActiveValues` in a loop? Could you do:
```
const int iter = delta * (dilation/delta) + (dilation % delta);
dilateActiveValues(mgr, iter , nn, IGNORE_TILES);
```

File: openvdb/tools/FastSweeping.h [hunk 14]
```diff
@@ -0,0 +1,1432 @@
...
+#endif
+        static const ValueT Unknown = std::numeric_limits<ValueT>::max();
+        changeLevelSetBackground(mgr, Unknown);//multi-threaded
+
+ #ifdef BENCHMARK_FAST_SWEEPING
+        timer.restart("Dilating and updating mgr (parallel)");
+        //timer.restart("Dilating and updating mgr (serial)");
+#endif
+
+        const int delta = 5;
```
Idclip, line 618: What does this delta refer too?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+        //for (int i=0, n=5, d=dilation/n; i<d; ++i) dilateActiveValues(mgr, n, nn, IGNORE_TILES);
+        //dilateVoxels(mgr, dilation, nn);
+
+#ifdef BENCHMARK_FAST_SWEEPING
+        timer.restart("Initializing grid and coords");
+#endif
+        BufferT prototype(mParent->mPagedArray);//exemplar used for initialization
+        PoolT pool(prototype);//thread local storage pool of ValueBuffers
+        mPool = &pool;
+        tbb::parallel_for(mgr.leafRange(32), *this);//multi-threaded
```
Idclip, line 630: perhaps use `tbb::task_scheduler_init::default_num_threads()` ?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+        }// loop over leaf nodes
+        mParent->mBoundaryCount += boundaryCount;//reduces pressure on the atomic
+    }// operator(const LeafRange& r)
+    template<typename RootOrInternalNodeT>
+    void operator()(const RootOrInternalNodeT& node) const
+    {
+        const ValueT isoValue = mIsoValue, above = mAboveSign*std::numeric_limits<ValueT>::max();
+        for (auto it = node.cbeginValueAll(); it; ++it) {
+          ValueT& v = const_cast<ValueT&>(*it);
+          v = v > isoValue ? above : -above;
```
Idclip, line 774: assigning to the result of a const_cast is undefined - is there a way to make this non-const? The NodeManager seems to iterate on a non-const tree, so perhaps the argument to this function can simply be non-const too?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+    }
+
+    // Coord has a non-empty default constructor so we use the following trick to
+    // avoid initialization of the large array of Coords. The resulting
+    // allocation is virtually instantaneous. Note, we use mCoords as
+    // temporal storage of mPagedArray since this allows us to avoid
+    // performing multiple sorts directly on mPagedArray, which has poor
+    // performance due to the fact that the sorts are done with
+    // respect to changing hash functions (resulting in worst-case inputs).
+    mCoords.reset(reinterpret_cast<Coord*>(new char[sizeof(Coord)*mPagedArray.size()]));
```
Idclip, line 522: I like this technique 👍

openvdb/tools/FastSweeping.h
Idclip: Ah actually I see you've used `std::thread::hardware_concurrency()` in some cases - I think would be fine instead of the tbb implementation too?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+        };
+        planes.push_back(0);
+        tbb::parallel_for(tbb::blocked_range<size_t>(1, mParent->voxelCount(), 1<<12), func);// 4096
+        for (auto iter=pool.begin(); iter!=pool.end(); ++iter) iter->flush();
+        planes.push_back(mParent->voxelCount());
+        planes.sort();
+    }
+    /// @brief Locally solves @f$|\nabla \phi|^2 = 1 @f$ by means
+    /// of Godunov's upwind finite difference scheme
+#if 0
```
Idclip, line 1134: Can you elaborate on this disabled block?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+        auto tmp = std::make_unique<uint32_t[]>(count);
+
+        tbb::parallel_for(tbb::blocked_range<uint32_t>(0, count, 64),
+                          [&](const tbb::blocked_range<uint32_t>& r){auto *p=&tmp[r.begin()]; for (uint32_t i = r.begin(); i < r.end(); ++i) *p++=i;});
+        //if (tmp[134] != 134) std::cerr << "ERROR" << std::endl;
+        auto hashComp2 = [&](uint32_t a, uint32_t b){return hash(coords[a]) < hash(coords[b]);};
+#ifdef BENCHMARK_FAST_SWEEPING
+        timer.restart("Alternative sort");
+#endif
+        tbb::parallel_sort(&tmp[0], &tmp[0] + mParent->voxelCount(), hashComp2);
```
Idclip, line 1073: Do you need to remove/comment this alternative sorting implementation out for a final merge?

openvdb/tools/FastSweeping.h
jmlait: Ideally it should scale with threads in your group in tbb; not based on core counts.  Might be running in a -j4 environment, for example.

openvdb/tools/FastSweeping.h
Idclip: What do you mean by a direction parameter? imo, it would be simpler if these were consolidated into a single method which checked the grid metadata to infer its type and handle it accordingly. Something like `sweepToLevelSet`

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+/// @details Topology of output grid is identical to that of the input grid, except
+///          active tiles in the input grid will be converted to active voxels
+///          in the output grid!
+///
+/// @throw  RuntimeError if the iso-surface does not intersected any active
+///         voxels or if it intersects any active tiles in @a sdfGrid.
+template<typename GridT>
+typename GridT::Ptr
+sdfToSdf(const GridT &sdfGrid,
+         typename GridT::ValueType isoValue = 0,
```
Idclip, line 127: compiler errors when trying to instantiate this with a non scalar grid. Can we use `zeroVal` here? the static_assert in `FastSweeping` should then make it clearer

openvdb/unittest/TestFastSweeping.cc
```diff
@@ -0,0 +1,635 @@
...
+
+    // Define a simple lambda function to write a grid to disk
+    // This is useful for visual inspections in Houdini!
+    auto writeFile = [](std::string name, FloatGrid::Ptr grid){
+        io::File file(name);
+        file.setCompression(io::COMPRESS_NONE);
+        GridPtrVec grids;
+        grids.push_back(grid);
+        file.write(grids);
+    };
```
Idclip, line 154: lots of writing in these unit tests

openvdb/unittest/TestFastSweeping.cc
```diff
@@ -0,0 +1,635 @@
...
+             CPPUNIT_ASSERT(percent < 4.5);// crossing characteristics!
+             //std::cout << "\nOutput 1: " << message << std::endl;
+         }
+     }
+#endif
+}// testMaskSdf
+
+void
+TestFastSweeping::testFogToSdf()
+{
```
Idclip, line 325: This test never actually calls `fogToSdf`. should probably call it at least once to make sure it can be invoked

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1432 @@
...
+template <typename GridT>
+void FastSweeping<GridT>::sweep(int nIter, bool finalize)
+{
+    if (!mGrid1) {
+      OPENVDB_THROW(RuntimeError, "FastSweeping::sweep called before initialization");
+    }
+    if (this->boundaryCount() == 0) {
+        OPENVDB_THROW(RuntimeError, "FastSweeping: No boundary voxels found!");
+    } else if (this->voxelCount() == 0) {
+        OPENVDB_THROW(RuntimeError, "FastSweeping: No computing voxels found!");
```
Idclip, line 512: I'm running into this runtime error consistently with different inputs. Although I've figured out what it means it's not immediately clear. For example, calling `dilateSdf` with a narrow band level set or even with a filled but small interior sdf will cause this check to fire. Is there a way we can handle this case more garcefully?

openvdb/tools/FastSweeping.h
kmuseth: when attempting to create an SDF from a volume at a non-existing iso-value, I think it's misleading to return an empty grid. First, an empty grid does exist unless of course, you mean a shared point to NULL? If not, what should be background value be (recall it's supposed to be an SDF)? Returning a null smart pointer is not much differet from throwing .. is it.

openvdb/tools/FastSweeping.h
kmuseth: agree on both points

openvdb/tools/FastSweeping.h
kmuseth: this parameter is only used for "mAboveSign  = isInputSdf ? 1 : -1;". Relying on the grid-class metadata is something we avoid if we can (as in this case where the name of the method clearly states that the input is either an SDF or FOG volume),

openvdb/tools/FastSweeping.h
kmuseth: I agree, but the problem is to come up with a metric or tolerance. The norm of the gradient would seem like a good starting point but the first-order nature of this fast sweeping method implies that there's a limit to the accuracy and the existence of singularities (critical points) might complicated things as well. In other words, it's far from trivial.

openvdb/tools/FastSweeping.h
kmuseth: sure, but it doesn't hurt to explain it does it?

openvdb/tools/FastSweeping.h
kmuseth: what background value? This method takes in a density, and return a scalar field, with values defined by the functor, that is extended from an iso-surface of the density. If not iso-surface exists then there is no meaningful output in my opinion. For instance, what should the background value me? The only alternative to throwing is to return a null shared pointer. Would you prefer that?

openvdb/tools/FastSweeping.h
kmuseth: It is propagated from the iso-surface of the input SDF. I'll update the documentation to reflect this

kmuseth: Actually it currently says "Computes the signed distance field and the extension of a scalar field,
defined by the specified functor, off an iso-surface from an input SDF volume." Is that not enough?

openvdb/tools/FastSweeping.h
kmuseth: I'll adopt your version

openvdb/tools/FastSweeping.h
kmuseth: sure, why do you prefer that over a fixed-length array?

openvdb/tools/FastSweeping.h
kmuseth: just that - dilation with higher connectivity is obviously slower but arguable also more accurate (depending on the application ). I can remove this comment if it bothers you

openvdb/tools/FastSweeping.h
kmuseth: they are used multiple times and each of them have multiple member methods

openvdb/tools/FastSweeping.h
kmuseth: I'm planning to remove all the timing for the final commit - it is, however, useful as long as I'm making changes to the code.

openvdb/tools/FastSweeping.h
kmuseth: I believe I found this to be faster. Maybe that's no longer the case with Nick's changes?!

openvdb/tools/FastSweeping.h
kmuseth: the value five :-D This is the sub-division that seemed to result in the best performance

openvdb/tools/FastSweeping.h
kmuseth: an old block of code that will be removed (as replaced by the other fork of the #else

openvdb/tools/FastSweeping.h
kmuseth: I'm not sure that applies here since the value that the reference is pointing to in the leaf-node is not declared as const. Changing it is tricky since the ValueAccessor (or tree) does not have any access methods that return a non-const reference. Only alternative I can think of is to get a const pointer to the buffer and cast that but I'm actually not convinced this is an issue to begin with

openvdb/tools/FastSweeping.h
kmuseth: good catch - code from an old benchmark test that I forgot to remove.

openvdb/tools/FastSweeping.h
kmuseth: this whole tool currently only work on scalar floating point values: static_assert(std::is_floating_point<ValueT>::value,
                  "FastSweeping requires a grid with floating-point values");

openvdb/unittest/TestFastSweeping.cc
kmuseth: yes not ideal but it was the easiest way to validate the output. I can remove the writes

openvdb/unittest/TestFastSweeping.cc
kmuseth: this unit-test should be renamed testSdfToFog :) Additionally, I should add a testFogToSdf

openvdb/tools/FastSweeping.h
kmuseth: yes probably but we should discuss what behaviour you would expect/refer since it's not intuitive (at least not to me)

openvdb/tools/FastSweeping.h
Idclip: I absolutely agree - which is why it would actually be nice if, in theory, these could be left in. I'm basically proposing that we introduce a system where it's encourages to add timings around code and enable/disable these metrics wholesale at compile time across VDB. It's a passing thought which, so long as these are being removed, doesn't apply to this PR.

openvdb/tools/FastSweeping.h
Idclip: Well the prior, in my mind, implies a direct relationship between the two elements. The latter simply because it's already a VDB typedef. This is minor though

openvdb/tools/FastSweeping.h
Idclip: Ah okay understood! If it remains a comment would be good, but it sounds like we agreed on replacing the below with a single call to `dilateActiveValues`

openvdb/tools/FastSweeping.h
Idclip: Think this still needs to be fixed in a number of placed

openvdb/tools/FastSweeping.h
jmlait: I think these routines should all maintain the grid topology.  So return the original grid with the same activity mask; but all the values set to the outside background value.

openvdb/tools/FastSweeping.h
jmlait: Agreed, I'd like to use metaclass data at all to change behaviour.  I like the fogToSDF and sdfToFog namings as they are quite clear when reading code.  While "1" and "-1" will make me very confused all the time which way it is working.  But both fancily named ones should call something with an explicit mAboveSign parameter.

openvdb/tools/FastSweeping.h
jmlait: Yeah, the gradient is unlikely to converge.  I was thinking maybe just absolute change could be used.  When the largest voxel change update less than a certain percentage of voxel diameter?   Are the sweeps monotonic?  If so, this should at least always stop (even if it does so by stalling out far from the final values)

openvdb/tools/FastSweeping.h
jmlait: :>  Yeah, I think this is my bias to want transparent constant tiles :>

openvdb/tools/FastSweeping.h
jmlait: I'd like for the resturned grid to have the same topology as the fog grid.  That is a nice invariant for the caller to work with.

I think we should add a typename GridT::ValueType backgroundValue to answer exactly the question of what the extension is.   (Or maybe not GridT::ValueType, as that is always scalar, but derived from OpT?)

One thing this doesn't let you specify is the type of grid to fill...  OpT is an opaque functor of unspecified return type, so how can you use this for vector grids?   If we had a parameter that is the background value, that parameter can be the GridT of the generated grid?

Hmm...  I actually don't like that very much as it runs into the 0 problem.

Probably best to prefix an explicit template on the extrap-gridtype.

template<typename ResultT, typename GridT, typename OpT>
￼typename Grid<ResultT>::Ptr
￼fogToExt(const GridT &fogGrid,
￼         const OpT &op,
￼         typename GridT::ValueType isoValue,
         ResultT backgroundValue,
￼         int nIter = 1);

openvdb/tools/FastSweeping.h
jmlait: That isn't what dilateSdf() has as comment... If I have a SDF with a bandwidth of 3 voxels and call this with n-voxels set to 3, do I get a 3 voxel bandwidth or a 6 voxel?

I think the fact there is no isovalue answers my question, it is 6 as it adds 3 extra voxels.

kmuseth:
I've made several improvements (not notably support for vector grids) and addressed the review comments. Please approve or give additional feedback - the deadline is Monday next week

openvdb/tools/FastSweeping.h
Idclip: Did we decide on a plan for this runtime error? Are we going to solve this at the SOP level? My concern is what users will see in Houdini when this gets hit as this won't mean anything to most people!

openvdb/tools/FastSweeping.h
Idclip: To clarify previous discussion, we'vs decided to leave this as is and update it once #754 is accepted

openvdb/tools/FastSweeping.h
Idclip: I believe you are correct, I was thrown off by the fact this operator takes a const node. So my follow up question is, why do these operators accept const nodes? Can we simply remove the const from the input node argument?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1577 @@
...
+        using LeafT = typename SweepMaskTreeT::LeafNodeType;
+        LeafManagerT leafManager(maskTree);
+
+        // compute the leaf node slices that have active voxels in them
+        // the sliding window of the has keys is -14 to 21 (based on an 8x8x8 leaf node
+        // and the extrema hash values i-j-k and  i+j+k), but we use a larger mask window here to
+        // easily accomodate any leaf dimension. The mask offset is used to be able to
+        // store this in a fixed-size byte array
+        constexpr int maskOffset = LeafT::DIM * 3;
+        constexpr int maskRange = maskOffset * 2;
```
Idclip, line 1180: Shouldn't the maskOffset be `(LeafT::DIM-1) * 2` and the maskRange be `maskOffset + ((LeafT::DIM-1) * 3) + 1`? i.e., the lowest possible value is `(LeafT::DIM-1)  * 2` and the largest possible id is `(LeafT::DIM-1)  *  2 + max` where `max` is simply `(LeafT::DIM-1) + (LeafT::DIM-1) + (LeafT::DIM-1)`.

also spl "hash"

openvdb/tools/FastSweeping.h
Idclip: Yes, it was simply that this assignment causes other compiler errors to be produced which can hide the static assertion. But it's a minor thing

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1577 @@
...
+
+#ifdef BENCHMARK_FAST_SWEEPING
+        util::CpuTimer  timer("Initialize voxels");
+#endif
+        mParent->mSweepMask.clear();
+        mParent->mSweepMask.topologyUnion(mParent->mSdfGrid->constTree());
+
+        {// Process all voxels
+          tree::LeafManager<SdfTreeT> mgr(tree, 1);// we need one auxiliary buffer
+          tbb::parallel_for(mgr.leafRange(32), *this);//multi-threaded
```
Idclip, line 846: Still manually specifying a leaf range here of 32 per thread. Is this intentional?

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1577 @@
...
+        mParent->mSweepMask.clear();
+        mParent->mSweepMask.topologyUnion(mParent->mSdfGrid->constTree());
+
+        {// Process all voxels
+          // Define thread-local operators
+          OpPoolT opPool(opPrototype);
+          mOpPool = &opPool;
+
+          tree::LeafManager<SdfTreeT> mgr(tree1, 1);// we need one auxiliary buffer
+          tbb::parallel_for(mgr.leafRange(32), *this);//multi-threaded
```
Idclip, line 967: Also here

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1577 @@
...
+    size_t mSweepingVoxelCount, mBoundaryVoxelCount;
+};// FastSweeping
+
+////////////////////////////////////////////////////////////////////////////////
+
+// Static member data initialization
+template <typename SdfGridT, typename ExtValueT>
+const Coord FastSweeping<SdfGridT, ExtValueT>::mOffset[6] = {{-1,0,0},{1,0,0},
+                                                             {0,-1,0},{0,1,0},
+                                                             {0,0,-1},{0,0,1}};
```
Idclip, line 530: Can you not use the offsets available in `openvdb/tools/Utils.h`

openvdb/tools/FastSweeping.h
kmuseth: The throw has been removed and instead empty shared pointers are returned. If would be east to instead return empty grids (with the provided background values), if you still prefer that?

openvdb/tools/FastSweeping.h
kmuseth: Not only will this make make naming scheme (what would be the name to replace sdfToSdf and fogToSdf?) but in the past we had refrained from relying (100%) on the meta-class tags. It's obviously a simple change to make but I don't see the advantage over the current more explicit naming

openvdb/tools/FastSweeping.h
kmuseth: this sill be added later

openvdb/tools/FastSweeping.h
kmuseth: see comment above - I'm no longer throwing

openvdb/tools/FastSweeping.h
kmuseth: improved the documentation

openvdb/tools/FastSweeping.h
kmuseth: 32 is the grain-size, i.e. the smallest number of leaf nodes to be part of a tbb task. it's not the thread-point. 32 has proved to perform well on different platforms and certainly much better than the default value of 1. In other words I'm keeping it

openvdb/tools/FastSweeping.h
kmuseth: strictly you're right that they could take non-const nodes but it wouldn't change anything since the iterator still returns a const reference (a design limitation coming from the fact that LeafNode<bool> is template specialized with bit-arrays). I guess it would make the function signature more "honest". I'll change it ...

kmuseth: tried this and the compiler gets confused since it cannot make the correct distinction between 

template<typename RootOrInternalNodeT>
    void operator()(RootOrInternalNodeT& node) const

and

void operator()(const LeafRange& r) const

openvdb/unittest/TestFastSweeping.cc
kmuseth: removed them

openvdb/unittest/TestFastSweeping.cc
kmuseth: fixed this

openvdb/tools/FastSweeping.h
kmuseth: this is no longer and issue since all free-standing functions are careful never to call FastSweeping::sweep if and of the initialization methods failed

openvdb/tools/FastSweeping.h
kmuseth: Dan added this

openvdb/tools/FastSweeping.h
kmuseth: see comment above - this is indeed intentional

openvdb/tools/FastSweeping.h
kmuseth: not really - the ordering is different and I do a lot of index access like mOffset[i]. Changing the order would be a mess

openvdb/tools/FastSweeping.h
Idclip: Yes, so long as you don't need a const Leaf/NodeManager you should also be able to remove the const there too!

openvdb/tools/FastSweeping.h
Idclip: @danrbailey thoughts?

openvdb/tools/FastSweeping.h
Idclip: Oops, didn't realise the orders were different

openvdb/unittest/TestFastSweeping.cc
Idclip: I think I can still see a few uses of the `writeFile` lambda being invoked in this file?

openvdb/tools/FastSweeping.h
danrbailey: This is intentionally an over-estimate. I initially had some more complicated logic in there similar to what you're proposing, but I simplified it to reduce the risk that I'd missed an edge case or that it wouldn't work for a particular leaf node configuration. There's a much greater benefit from using this voxel slicing algorithm than what you would save in this small optimization so I opted for caution at this stage.

In fact, there are even better schemes than what you're proposing if you factor in the exact hash and you compute an offset and range per hash. On top of that, there are more complex but much more memory efficient schemes than storing a nodemask per slice. Finally, there's the improvements we discussed the other day of segmenting the active voxels into different regions that can be computed concurrently. All of these improvements and optimisations won't affect the API in any way, so they can always come later.

openvdb/tools/FastSweeping.h
```diff
@@ -0,0 +1,1577 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+//
+/// @file    FastSweeping.h
+///
+/// @author  Ken Museth
+///
+/// @brief   Defined the six functions {fog,sdf}To{Sdf,Ext,SdfAndExt} in
+///          addition to the two functions maskSdf and dilateSdf. Sdf denotes
+///          a signed-distance field (i.e. negative values are insdie), fog
```
anil100391, line 10: Typo here... insdie should be inside

openvdb/tools/FastSweeping.h
Idclip: Thanks @danrbailey, still a bit confused though - for my own understanding, are the values I've calculated correct and a better representation of the required sizes? You mention you've over estimated for a "particular leaf node config", can you give an example? This seems like a pretty arbitrary choice, no other algorithm works as such and expects a uniform leaf node dim. In fact this may make the situation worse by working for these other leaf node types and not others! We should at least be asserting here in that case - I don't like the idea that we're just randomly choosing an allocation size to make sure we haven't missed any phantom edge cases :\

To be clear, the only reason I chose to look at these lines in more detail is *because* of your decision to use this particular math. It didn't make any sense to me! It unfortunately only serves to obfuscate the intended implementation to the reader.

openvdb/tools/FastSweeping.h
jmlait: I'd prefer empty grids to null pointers.  Indeed, I think from my understanding they should be empty grids with the same topology as the input, but all active voxels set to the background value, because these functions are not supposed to affect topology?

What happens if a disconnected input grid has one connected component have no iso-surface crossing?  Does it properly still extend the components that have a crossing?

kmuseth: The problem is, the background value of an SDF has a very specific meaning, which is breaks down for empty grids. The background value for a truncated SDF is the maximum positive distance value, i.e. the value at the out-most level-set of the narrow band. However, when the fundamental assumptions of the fast sweeping algorithm as not satisfied (e.g. no boundary conditions), then obviously not distance values can be computed, which in turn means it's unclear what the background value should be of the "empty sdf with active voxels" that you want. I could set it to std::numerical_limits<T>::max() but why is that useful?

openvdb/tools/FastSweeping.h
kmuseth: thx

kmuseth:
@jmlait  I have changed all the return types so  instead of returning null shared pointers if fast sweeping was not performed it returns empty grids with the background values set to respectively infinity for SDFs and the provided background value for extension fields. I much prefer empty grids over grids with identical topology, both because it allows for a simple and reliable why to check is the algorithm failed (Grid::empty()) and because it avoids slow, memory heavy and unnecessary deep copies.

kmuseth:
@jmlait Regarding your question about, what happens when topologically disconnected FOG values are partially intersecting the iso-surface that represented the boundary condition? The answer is elegant and IMO exactly what you'd expect: Any values that are topologically connected to a boundary value will have a correct signed distance value, whereas all other values will be set to plus or minus infinity depending on wither their original value was above or below the specified iso-value.

kmuseth:
I believe I've now addressed all review comments.

Idclip:
@kmuseth This looks great! I still have two comments which were never fully addressed:

https://github.com/AcademySoftwareFoundation/openvdb/pull/746/files#r446127893
https://github.com/AcademySoftwareFoundation/openvdb/pull/746#discussion_r463046575

But yeah, lets just get this in now...

One last thing I'd like to comment on (and sorry to be picky), but I'm not a fan of how the unit tests look with so much commenting our, personal data paths and custom ifdefs. I've recently been refactoring the morphology and filter unit tests and have had to deal with a lot of code like this where it's not clear what the intention was. I don't think this code will ever really be cleaned up? Looks like all the file writes have been disabled by default, so for the interest of getting this in to 7.1 I'll approve - but I recommend we avoid doing this where possible.

Note that the windows build should be fixed with #775. However, @kmuseth, the DCO sign-off is failing for this PR? That needs to be fixed so we can merge

Idclip:
There's also no mention of the blosc CMake changes in the release notes

danrbailey:
Closing as has now been merged as #746

PR #747, opened 2020-06-22
Always set the vector UI to visible if we fail to get the input mesh.

jmlait:
If we fail to retrieve the input mesh, we want to make the vector UI visible as we don't know if it is needed or not.  Otherwise it maintains the last set values which may cause it to become stuck off even though it is affecting the output.

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_From_Polygons.cc
+++ b/openvdb_houdini/SOP_OpenVDB_From_Polygons.cc
@@ -684,7 +684,7 @@ SOP_OpenVDB_From_Polygons::updateParmsFlags()
     GA_ROAttributeRef attrRef;
     int attrClass = POINT_ATTR;
     const GU_Detail* meshGdp = this->getInputLastGeo(0, time);
-    for (int i = 1, N = evalInt("numattrib", 0, time); i <= N; ++i) {
+    for (int i = 1, N = static_cast<int>(evalInt("numattrib", 0, time)); i <= N; ++i) {
         bool isVector = true;
         if (meshGdp) {
             isVector = false;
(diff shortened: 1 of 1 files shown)


danrbailey:
This PR was a good test for the new actions workflow, the first commit succeeded, however the second one silently timed out after 6 minutes with a "[warning]Premature close" error. Looking into this deeper, it looks like there's an issue with reliability in the caching action right now and an upcoming improvement to break the cache download into 4MB chunks:

https://github.com/actions/cache/issues/267#issuecomment-646156084

I'll keep an eye out for when a new cache release becomes available and bump us to that. In the meantime we should always error when the action is being run with the ASWF repo as the target, I'll look into making that change.

Idclip:
This was merged with a failing build? Also no release notes!

danrbailey:
> This was merged with a failing build? Also no release notes!

The failing job in this PR was the Mac build which succeeded on the previous run - https://github.com/AcademySoftwareFoundation/openvdb/runs/796854115?check_suite_focus=true. As there were no code changes between these two runs, it must be a CI failure rather than a build failure so shouldn't hold up a PR. I can't see the failure log to tell what the issue might have been but we should look to address it if it happens again.

Idclip:
>  it must be a CI failure rather than a build failure so shouldn't hold up a PR

You're obviously right with regards to the cause, but I very much disagree with this. There is no rush here. Kicking off the build again is easy and would confirm this assumption. We should be trying to stick to the 3 day grace period as stated in our best practises.

Idclip:
I'm getting the following warning with these changes:

```
vdbfrompolygons exint OP_Parameters::evalIntT(const UT_StringRef&, int, fpreal, int) const: Invalid parameter name numattrib
```

@jmlait I think the `numattrib` param is wrong - it should be `attrList`? Also can we update the release notes please

Issue #688, opened 2020-04-01
Incorect TBB test in cmake scripts


yurivict:
```
CMake Error at /usr/local/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find TBB (missing: Tbb_LIB_COMPONENTS tbbmalloc_proxy) (found
  suitable version "2019.0", minimum required is "2017.0")
```

aflecken:
also having this issue with 2018.0

Idclip:
@yurivict / @aflecken what platforms are you running into this issue on, are you both on FreeBSD?

yurivict:
I am on FreeBSD.

Idclip:
Resolved by #749

PR #749, opened 2020-06-26
Disable jemalloc by default and allow for downstream jemalloc targets

Idclip:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,13 +120,13 @@ A global list of library paths to additionally use into when searching for depen
 
 set(_CONCURRENT_MALLOC_OPTIONS None Auto Jemalloc Tbbmalloc)
 if(NOT CONCURRENT_MALLOC)
-  set(CONCURRENT_MALLOC None CACHE STRING
-    "Explicitly link the Core OpenVDB builds against a particular concurrent malloc library.
-    Options are: None Auto Jemalloc Tbbmalloc. Note that this choice should usually be deffered
-    to your downstream application. Although not required, it is strongly recommended to use a
-    concurrent memory allocator. Has no effect if OPENVDB_BUILD_CORE is false. The 'Auto' option
-    implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable, in which case Tbbmalloc
-    is used." FORCE
+  set(CONCURRENT_MALLOC Auto CACHE STRING
+    "Explicitly link the OpenVDB executables against a particular concurrent malloc library.
+    Options are: None Auto Jemalloc Tbbmalloc. Although not required, it is strongly recommended
+    to use a concurrent memory allocator. Has no effect if OPENVDB_BUILD_BINARIES is false. Auto
+    is the default and implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable, in which
+    case Tbbmalloc is used. Note that this is not linked into library builds and defers this
+    choice to downstream applications via explicit CMake targets." FORCE
   )
 elseif(NOT ${CONCURRENT_MALLOC} IN_LIST _CONCURRENT_MALLOC_OPTIONS)
   message(WARNING "Unrecognized value for CONCURRENT_MALLOC, using Auto instead.")
@@ -589,11 +589,18 @@ unset(HAS_AVAILABLE_WARNINGS)
 # * On Windows, we follow SideFX's example in using Tbbmalloc due to the challenges
 #   of injecting into the Windows runtime to replace the system allocator.
 
-if(CONCURRENT_MALLOC STREQUAL "Auto")
+if(OPENVDB_BUILD_BINARIES AND CONCURRENT_MALLOC STREQUAL "Auto")
   if(WIN32 OR APPLE OR USE_MAYA)
     set(CONCURRENT_MALLOC "Tbbmalloc")
   else()
-    set(CONCURRENT_MALLOC "Jemalloc")
+    find_package(Jemalloc QUIET)
+    if(NOT TARGET Jemalloc::jemalloc)
+      message(WARNING "Unable to find Jemalloc, attempting to fall back to TBB malloc.
+        It is recommended to use Jemalloc for optimum performance.")
+      set(CONCURRENT_MALLOC "Tbbmalloc")
+    else()
+      set(CONCURRENT_MALLOC "Jemalloc")
+    endif()
   endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,10 +123,11 @@ if(NOT CONCURRENT_MALLOC)
   set(CONCURRENT_MALLOC Auto CACHE STRING
     "Explicitly link the OpenVDB executables against a particular concurrent malloc library.
     Options are: None Auto Jemalloc Tbbmalloc. Although not required, it is strongly recommended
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -120,11 +120,13 @@ A global list of library paths to additionally use into when searching for depen
...
 set(_CONCURRENT_MALLOC_OPTIONS None Auto Jemalloc Tbbmalloc)
 if(NOT CONCURRENT_MALLOC)
-  set(CONCURRENT_MALLOC Auto CACHE STRING
-    "Choose a concurrent malloc library, options are: None Auto Jemalloc Tbbmalloc.
-    Although not required, it is strongly recommended to use a concurrent memory allocator.
-    Auto is the default and implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable,
-    in which case Tbbmalloc is used." FORCE
+  set(CONCURRENT_MALLOC None CACHE STRING
+    "Explicitly link the Core OpenVDB builds against a particular concurrent malloc library.
+    Options are: None Auto Jemalloc Tbbmalloc. Note that this choice should usually be deffered
```
danrbailey, line 125: deferred

CMakeLists.txt
```diff
@@ -120,11 +120,13 @@ A global list of library paths to additionally use into when searching for depen
...
+    to your downstream application. Although not required, it is strongly recommended to use a
+    concurrent memory allocator. Has no effect if OPENVDB_BUILD_CORE is false. The 'Auto' option
+    implies Jemalloc, unless USE_MAYA is ON or Jemalloc is unavailable, in which case Tbbmalloc
```
danrbailey, line 128: Can we remove the 'Auto' option and require the user to make an explicit choice? It was mainly about having a sensible default, but if we're defaulting to 'None' think it makes sense to remove it entirely.

cmake/FindOpenVDB.cmake
```diff
@@ -24,6 +24,8 @@ IMPORTED Targets
 
 ``OpenVDB::openvdb``
   The core openvdb library target.
+``OpenVDB::openvdb_je``
+  The core openvdb library target with jemalloc.
```
lgritz, line 28: Nice. I like this approach.

Idclip:
> This looks fine to me. As discussed, we should update the binaries and the unit tests to build against the jemalloc version of OpenVDB though?

You can still do this by providing a value to `CONCURRENT_MALLOC`. However if you want to be able to disable a concurrent malloc at build time and still allow for the binaries to link against one then we'll need yet another CMake argument. Something like
`OPENVDB_BINARIES_CONCURRENT_MALLOC`. Gets a bit silly though. Perhaps `CONCURRENT_MALLOC` should only every apply to the final binaries and we could introduce a CMake var like `OPENVDB_LINKIN_CONCURRENT_MALLOC` which is on/off (off by default) and this is the variable that links in your selected allocator to **libraries**. It then makes sense for the auto option to stick around as the value of ` CONCURRENT_MALLOC` is always applied to the binaries.

danrbailey:
> > This looks fine to me. As discussed, we should update the binaries and the unit tests to build against the jemalloc version of OpenVDB though?
> 
> You can still do this by providing a value to `CONCURRENT_MALLOC`. However if you want to be able to disable a concurrent malloc at build time and still allow for the binaries to link against one then we'll need yet another CMake argument. Something like
> `OPENVDB_BINARIES_CONCURRENT_MALLOC`. Gets a bit silly though. Perhaps `CONCURRENT_MALLOC` should only every apply to the final binaries and we could introduce a CMake var like `OPENVDB_LINKIN_CONCURRENT_MALLOC` which is on/off (off by default) and this is the variable that links in your selected allocator to **libraries**. It then makes sense for the auto option to stick around as the value of ` CONCURRENT_MALLOC` is always applied to the binaries.
> 
> edit: Note you can also achieve this without any additional CMake argument in two build steps

Yes, I like the suggestion of keeping `CONCURRENT_MALLOC` around but make it apply only to binaries. In fact, I don't think it's even necessary to add the LINKIN variable, providing the two cmake targets (with and without jemalloc) should be sufficient. As we discussed it should be up to the client application to set this correctly, but we do at least want to set the example by using it in our own binaries. 

I don't mind if we change this build behaviour for 7.1 provided we include a specific mention at the head of the release notes and add the warning when calling `openvdb::initialize()` to indicate that a concurrent allocator is not being used.

cmake/FindOpenVDB.cmake
danrbailey: When we last looked at concurrent allocators which was a little while ago, there was an issue using jemalloc on Mac so our Auto mode chooses tbbmalloc there. How should we handle this - does it make sense to include a target with tbbmalloc?

Idclip:
> I don't mind if we change this build behaviour for 7.1 provided we include a specific mention at the head of the release notes and add the warning when calling openvdb::initialize() to indicate that a concurrent allocator is not being used.

Not sure there's a particularly reliable way to do this. See https://github.com/jemalloc/jemalloc/issues/388 It would be annoying if you're certain it's in use and this check fails, resulting in a print on every startup

danrbailey:
> > I don't mind if we change this build behaviour for 7.1 provided we include a specific mention at the head of the release notes and add the warning when calling openvdb::initialize() to indicate that a concurrent allocator is not being used.
> 
> Not sure there's a particularly reliable way to do this. See [jemalloc/jemalloc#388](https://github.com/jemalloc/jemalloc/issues/388) It would be annoying if you're certain it's in use and this check fails, resulting in a print on every startup

Agreed. I don't think it's as important to check that jemalloc is being used than it is to check that the glibc malloc _isn't_ being used. Pretty much every other memory allocator out there outperforms the default allocator. That being said, I can't see any simple way to do this either, so I think we'll just have to give up on this check for now.

Idclip:
> Yes, I like the suggestion of keeping CONCURRENT_MALLOC around but make it apply only to binaries. In fact, I don't think it's even necessary to add the LINKIN variable, providing the two cmake targets (with and without jemalloc) should be sufficient. As we discussed it should be up to the client application to set this correctly, but we do at least want to set the example by using it in our own binaries.

I've had a stab at this. I'm not using the `je` target so I'm able to handle both` je`/`tbb` and also to handle the case where both the core lib and binaries are being built (otherwise there's lots of cmake branching) but I've made a comment referring to the `je` target.

Slightly off topic, but do we need to explicitly find tbbmalloc_proxy? I don't think it serves any purpose on our centos builds and there's an issue about it here https://github.com/AcademySoftwareFoundation/openvdb/issues/688. I won't change this behaviour in this PR anyway, but thought it would be good to touch on here.

Idclip:
There's also a question as to what the unit tests should link against

danrbailey:
> There's also a question as to what the unit tests should link against

The unit tests should use the same concurrent allocator as the binaries - is that an issue?

Regarding the tbbmalloc_proxy issue, I think we just link against tbbmalloc and ignore the proxy lib.

I looked at your recent changes and they seem fine to me. We can refactor this CMake logic at a later point, but would be good to get this jemalloc change in before we release 7.1. Only thing missing from this PR is to switch on the concurrent allocator for the unit tests, then I'm happy for this to go in.

Idclip:
> The unit tests should use the same concurrent allocator as the binaries - is that an issue?

I think there's an argument to be made that they should use the same default allocator as the library - but I'll go ahead and change this.

Idclip:
@danrbailey Updated the unittest build and removed tbbmalloc_proxy

PR #750, opened 2020-06-26
Add a check that ensures that Houdini is available in the ASWF repo

danrbailey:
It should always have been downloaded from the GHA cache for the parent repo, but it is allowed to not exist on fork repos.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,7 @@ jobs:
         restore-keys: vdb1-houdini18_0-
     - name: validate_houdini
       if: github.repository_owner == 'AcademySoftwareFoundation'
-      run: ./ci/validate_houdini.sh
+      run: test -f "hou/hou.tar.gz"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -84,6 +84,9 @@ jobs:
         path: hou
         key: vdb1-houdini18_0-${{ hashFiles('hou/hou.tar.gz') }}
         restore-keys: vdb1-houdini18_0-
+    - name: validate_houdini
+      if: github.repository_owner == 'AcademySoftwareFoundation'
+      run: ./ci/validate_houdini.sh
```
Idclip, line 89: can you just do this  in one line? `test -f "hou/hou.tar.gz"`

danrbailey: I think so - I'll try that. It is simpler.

Idclip:
I re-ran the build for the sake of the green tick

PR #751, opened 2020-06-29
Added the option to preserve active tiles during topology unions

Idclip:
This adds functionality to topology unions to preserve active tiles when the second tree contains active child topology which overlaps with a parent active tile in the first tree. Currently a leaf level tile in the first tree will be expanded to an allocated but constant (all active, all values the same) leaf node if the second tree contains a corresponding leaf node regardless of the leaf's activity. This is the current behaviour and is reflected in the function description.

As an example, I ran into an issue with this with the recent attempts to improve morphological dilations (and this is thus somewhat of a prerequisite for those upcoming improvements). The original unthreaded `dilateVoxels` ignores tiles. If a leaf neighbours an adjacent active tile and dilates into it, the tile remains unmodified. In the **current** implementation of `dilateActiveValues` the multithreaded implementation calls `topologyUnion` on subtrees. If a subtree happens to dilate into overlapping active tiles, these tiles are forcibly voxelized. If called with `IGNORE_TILES`, this seems incorrect. Ultimately, if you all you care about is an activity mask, this new behaviour is far more efficient.

I've opted for a new boolean flag on the `topologyUnion` call which defaults to the current behaviour, but perhaps this should really be a new `activityUnion` function. However this PR **only solves the preservation of active tiles**. The preservation of **inactive** tiles is unfortunately harder due to the complexity with updating the internal nodemasks (e.g. the case where an inactive tile is voxelized due to overlapping child topology, but all the child topology is inactive). You can avoid this with a pre `pruneInactive` on the second tree.

### DIFF
--- a/openvdb/tree/Tree.h
+++ b/openvdb/tree/Tree.h
@@ -684,8 +684,11 @@ public:
     /// Specifically, active tiles and voxels in this tree are not changed, and
     /// tiles or voxels that were inactive in this tree but active in the other tree
     /// are marked as active in this tree but left with their original values.
+    ///
+    /// @note If preserveTiles is true, any active tile in this topology
+    /// will not be densified by overlapping child topology.
     template<typename OtherRootNodeType>
-    void topologyUnion(const Tree<OtherRootNodeType>& other);
+    void topologyUnion(const Tree<OtherRootNodeType>& other, const bool preserveTiles = false);
 
     /// @brief Intersects this tree's set of active values with the active values
     /// of the other tree, whose @c ValueType may be different.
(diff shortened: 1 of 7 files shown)


openvdb/tree/Tree.h
```diff
@@ -684,8 +684,11 @@ class Tree: public TreeBase
     /// Specifically, active tiles and voxels in this tree are not changed, and
     /// tiles or voxels that were inactive in this tree but active in the other tree
     /// are marked as active in this tree but left with their original values.
+    ///
+    /// @note If preserveTiles is true, any active tile in this topology
+    /// will not be densified by overlapping child topology.
     template<typename OtherRootNodeType>
-    void topologyUnion(const Tree<OtherRootNodeType>& other);
+    void topologyUnion(const Tree<OtherRootNodeType>& other, const bool preserveTiles = false);
```
Idclip, line 691: This should really be `preserveActiveTiles` and I need to update the function description more. See the PR description for details

danrbailey:
Could we perhaps pull this out into a new standalone tool and fix this behavior here, then mark this method as deprecated?

Idclip:
> Could we perhaps pull this out into a new standalone tool and fix this behaviour here, then mark this method as deprecated?

Yes I'd really like to do that - unfortunately this work spawned from the the morphology improvements I'm working on and creating a new tool and including handling the extra cases I've mentioned above is currently too out of scope for me. If we feel that this change isn't worth it on it's own merit then I can work around it (with regards to the dilation issues) and instead ticket this up into a new tool.

Idclip:
Merging this into the morphology improvements PR so I can complete those changes #754

PR #754, opened 2020-07-03
New Morphology methods and improvements

Idclip:
Aims of this PR:
 - Deprecate `dilateVoxels` and `erodeVoxels` in favour of `dilateActiveLeafValues` and `erodeActiveLeafValues`
 - Introduce `erodeActiveValues` to match `dialteActiveValues` to handle tile policies for erosion
 - Introduce multithreaded erosion of edge and vertex neighbours
 - Significantly improve the performance of large dilations
 - Remove unused code and significantly clean-up Morphology.h such that namespaced classes can be used by clients for more granular control

NOTE:
 - Incorporates PR #751
 - `erodeVoxels` calls `pruneLevelSet`, the new methods do **not** and leave such decisions either to the tile policy or client code.

As of this PR `dilateVoxels` simply called `dilateActiveLeafValues` (unthreaded) and `erodeVoxels` calls `erodeActiveLeafValues`, but follows with a call to `pruneLevelSet`. The behaviour of these old functions thus remains exactly the same.

Results for discussion:
https://docs.google.com/spreadsheets/d/13-6yTmmPdj_gLLDj-3hgaCUaDy_-hgs-aY1BoMQr6PQ/edit?usp=sharing


openvdb/tools/Activation.h
```diff
@@ -0,0 +1,153 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+//
+/// @file   Activation.h
+///
+/// @brief  Implementation of topological activation/deactivation
+///
+/// @author Ken Museth
+///
```
Idclip, line 9: This file is a copy/paste of the activation/deactivation methods out of Morphology.h and can be mostly be ignored

openvdb_houdini/SOP_OpenVDB_Activate.cc
```diff
@@ -493,14 +493,20 @@ template <typename GridType>
...
 sopErodeVoxels(GridType& grid, int count)
 {
-    openvdb::tools::erodeVoxels(grid.tree(), count);
+    openvdb::tools::erodeActiveLeafValues(grid.tree(), count);
+    if (grid.getGridClass() == openvdb::GRID_LEVEL_SET) {
+        openvdb::tools::pruneLevelSet(grid.tree());
+    }
+    else {
+        openvdb::tools::pruneInactive(grid.tree());
+    }
```
Idclip, line 509: @jmlait wasn't too sure what to do here. The VDB Activate SOP used to call `erodeVoxels` which always calls `pruneLevelSet`. If the grid is not marked as a level set (but really, it's if the grid has a negative background value), the VDB Activate SOP will fail with an error if you try to expand with negative values (i.e. erode). To solve this and maintain compatible behaviour, I've just branched on the grid type. However...this SOP also has a "Prune Tolerance" parameter which also calls "prune". I Could remove this branching in favour of that parameter, but it will change behaviour for anyone who is eroding and has the "Prune Tolerance" disabled.

Another question, expansion currently only expands leaf nodes (due to the old impl of dilateVoxels and erodeVoxels). I could update this to handle tiles with the new methods now? Again, it would change behaviour.

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +62,831 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
-                leaf = acc.probeLeaf(orig);
-                isOn = leaf ? false : acc.isValueOn(orig);
+            this->clear();
+            mNeighbours[0] = &(leaf.getValueMask());
+            this->setOrigin(leaf.origin());
+            switch (mOp) {
+                case NN_FACE_EDGE        : { this->dilate18(mask); return; }
+                case NN_FACE_EDGE_VERTEX : { this->dilate26(mask); return; }
+                case NN_FACE             : { this->dilate6(mask);  return; }
+                default                  : { this->dilate6(mask);  return; }
```
jmlait, line 335: Why not assert(false) and do nothing in this case?

jmlait (approved):

openvdb/openvdb/tools/Morphology.h
Idclip: Yeah valid point, will change

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
```diff
@@ -493,14 +493,20 @@ template <typename GridType>
...
 sopErodeVoxels(GridType& grid, int count)
 {
-    openvdb::tools::erodeVoxels(grid.tree(), count);
+    openvdb::tools::erodeActiveLeafValues(grid.tree(), count);
+    if (grid.getGridClass() == openvdb::GRID_LEVEL_SET) {
+        openvdb::tools::pruneLevelSet(grid.tree());
+    }
+    else {
+        openvdb::tools::pruneInactive(grid.tree());
+    }
```
Idclip, line 509: @jmlait just an FYI, I have not changed the behaviour of the activate SOP in this PR; it still only works on voxels

Idclip:
I have a single commit way back when that wasn't signed off which is extremely annoying. As we can't use the DCO signoff button, I'll have to remake this PR, but I will do that once it's been approve by 1 more person

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
 ///
 /// @param tree          tree to be dilated
 /// @param iterations    number of iterations to apply the dilation
 /// @param nn            connectivity pattern of the dilation: either
 ///     face-adjacent (6 nearest neighbors), face- and edge-adjacent
 ///     (18 nearest neighbors) or face-, edge- and vertex-adjacent (26
 ///     nearest neighbors).
+/// @param threaded      Whether to multi-thread execution
+template<typename TreeOrLeafManagerT>
+inline void dilateActiveLeafValues(TreeOrLeafManagerT& tree,
```
danrbailey, line 158: I remember talking about this but I can't remember the reason. I still find it rather odd that we have `dilateActiveValues(..., IGNORE_TILES)` and `dilateActiveLeafValues(...)` which are the same thing?

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
+    /// @param nn Connectivity pattern of the dilation
+    /// @param prune Whether to collapse constant leaf nodes after the dilation
+    /// @param preserveMaskLeafNodes When dilating mask trees, the default behaviour
+    ///    chooses to steal the mask nodes rather than copy them. Although faster,
+    ///    this means that leaf nodes may be re-allocated. Set this to true if you
+    ///    need the original topology pointers to be preserved.
+    void dilateVoxels(const size_t iter,
+        const NearestNeighbors nn,
+        const bool prune = false,
+        const bool preserveMaskLeafNodes = false);
```
danrbailey, line 245: Does this `preserveMaskLeafNodes` value actually get used anywhere? I can't seem to find any references to it being true.

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
-            if (init) {
-                init = false;
-                Coord orig = xyz.offsetBy(DX*LEAF_DIM, DY*LEAF_DIM, DZ*LEAF_DIM);
-                leaf = acc.probeLeaf(orig);
-                if ((leaf == nullptr) && !acc.isValueOn(orig)) leaf = acc.touchLeaf(orig);
-            }
-            static const int N = (LEAF_DIM - 1)*(DY + DX*LEAF_DIM);
-            if (leaf) leaf->getValueMask().template getWord<Word>(indx-N) |= mask;
+            // copy the mask
+            const MaskType cache = leaf.getValueMask();
```
danrbailey, line 307: It confused me this was called `cache` not `mask`.

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
+
+        // build cache objects
+        typename Morphology<TreeT>::NodeMaskOp cache(accessor, nn);
+        std::vector<MaskType> nodeMasks;
+        std::vector<std::unique_ptr<LeafT>> nodes;
+        const ValueT& bg = tree.background();
+        const bool steal = iter > 1;
+
+        for (size_t i = 0; i < iter; ++i) {
+            if (i > 0) manager.rebuild(!threaded);
```
danrbailey, line 638: Did you try this without using a LeafManager - ie just a linear list of nodes with a on/off toggle for each one? Feels like you might be spending some time extracting and re-inserting leaf nodes and rebuilding the leaf manager with this technique?

danrbailey (commented):
In general looks good. A few additional questions:

* Should there also be a unit test for the new activation header?
* You presented some nice benchmark graphs of the results of this PR - are these included in the unit tests now?
* Many of the parallel_for methods you're using do very simple operations that I imagine is only a handful of instructions, did you consider using a larger grain size than 1?
* I would expect the non-threaded mode to call a lambda directly rather than to call into tbb::parallel_for() with a very large grain size. I think we're fairly consistent about how we do that, would be nice to avoid the tbb initialization if it's not strictly necessary. I also find it tends to make it easier to track down an issue because the stracktraces from tbb can be a bit verbose.

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
+
+    /// @brief Copy the current node masks onto the provided vector. The vector
+    ///    is resized if necessary.
+    /// @param masks The vector of NodeMasks to copy onto
+    void copyMasks(std::vector<MaskType>& masks) const
+    {
+        if (masks.size() < mManager.leafCount()) {
+            masks.resize(mManager.leafCount());
+        }
+        // @note this is marginally faster than using leafRange or foreach
```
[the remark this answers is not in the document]
Idclip: >  Have you got data to back this up? 

Yes of course... I did anyway. I encourage you to run your own tests as it's only noticeable with smaller operations, most likely due to the very minor overhead of extra data on the internal leaf iterator sub class.

openvdb/openvdb/tools/Morphology.h
Idclip: I think this was to provide an easier 1:1 mapping from the old functions. I guess I could remove them?

openvdb/openvdb/tools/Morphology.h
Idclip: No it doesn't. This was introduced due a problem with the underlying implementation. This algorithm uses a mask topology  to perform the dilation. If the input is already a mask topology it just uses it directly. When dilating the mask topology, it steals the nodes and splits them across active threads. When it comes to re-combine the nodes, it's possible it'll favour newly allocated data over the input nodes, resulting in any held leaf pointers being invalidated. Users can use this internal option to retain the state for input mask dilation. It's impossible to know if this is required or not as the alternative is to deep copy the mask tree, a significant overhead. It's good to have this as an option (I think a problem with various bits of the VDB API is that it's sometimes clear what guarantees are provided from various functions w.r.t the state of the tree).

Idclip:
@danrbailey 

> * Should there also be a unit test for the new activation header?

Yes? But I'm not going to add it in this PR. I didn't touch this code, it's simply been moved out of Morphology.h.

> * You presented some nice benchmark graphs of the results of this PR - are these included in the unit tests now?

No, you can see on the spreadsheets that the VDBs in use were Houdini test assets.

> * Many of the parallel_for methods you're using do very simple operations that I imagine is only a handful of instructions, did you consider using a larger grain size than 1?

No, I can do another pass.

> * I would expect the non-threaded mode to call a lambda directly rather than to call into tbb::parallel_for() with a very large grain size.

As would I, I will change these.

openvdb/openvdb/tools/Morphology.h
Idclip: For iter>1 I always need to call `getNodes`. So unless `initLeafArray` is slower than `getNodes`, the leaf manager way is better.

openvdb/openvdb/tools/Morphology.h
Idclip: renamed

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
+
+        // compute the wavefront. If the leaf previously existed, compute the
+        // xor activity result which is guaranteed to be equal to but slightly
+        // faster than a subtraction
+        tbb::parallel_for(manager.getRange(grainSize),
+            [&](const tbb::blocked_range<size_t>& r){
+            for (size_t idx = r.begin(); idx < r.end(); ++idx) {
+                auto& leaf = manager.leaf(idx);
+                auto& nodemask = leaf.getValueMask();
+                if (auto original = tree.probeConstLeaf(leaf.origin())) {
```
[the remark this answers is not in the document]
Idclip: done

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
+        , mThreaded(true) {}
+
+    /// @brief  Return whether this class is using multi-threading.
+    bool getThreaded() const { return mThreaded; }
+    /// @brief Set whether to use multi-threading.
+    /// @note  The grain size is not exposed
+    inline void setThreaded(const bool threaded) { mThreaded = threaded; }
+
+    /// @brief  Return a const reference to the leaf manager
+    inline const tree::LeafManager<TreeType>& manager() const { return mManager; }
```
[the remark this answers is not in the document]
Idclip: done

openvdb/openvdb/tools/Morphology.h
```diff
@@ -3,35 +3,31 @@
...
-#include <openvdb/tree/TreeIterator.h>
-#include <openvdb/tree/ValueAccessor.h>
-#include <openvdb/tree/LeafManager.h>
-#include "Prune.h"// for pruneLevelSet
-#include "ValueTransformer.h" // for foreach()
-#include <tbb/tbb_thread.h>
+#include "Activation.h" // backwards compatibility
+#include "Prune.h"
+#include "ValueTransformer.h"
+#include "../Types.h"
```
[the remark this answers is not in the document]
Idclip: changed

CHANGES
```diff
@@ -3,6 +3,26 @@ OpenVDB Version History
...
+      tools::dilateActiveValues() and tools::dilateActiveLeafValues().
+      Performance gains will improve relative to increases in the inputs size
+      and dilation scheme.
+    - Added an optional argument to Tree::topologyUnion() which preserves
+      active tiles on the destination tree should they overlap leaf nodes
+      from the source tree.
+
+    API changes:
+    - Restructured the internals of Morphology.h and moved tools::activate()
+      and tools::deactivate() to a new header, Activation.h.
```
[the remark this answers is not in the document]
Idclip: renamed

openvdb/openvdb/tools/Morphology.h
```diff
@@ -63,540 +59,835 @@ enum NearestNeighbors { NN_FACE = 6, NN_FACE_EDGE = 18, NN_FACE_EDGE_VERTEX = 26
...
-                : isOn ? ~Word(0) : Word(0);
         }
-    };// Neighbor
 
-    struct LeafCache
-    {
-        LeafCache(size_t n, TreeType& tree) : size(n), leafs(new LeafType*[n]), acc(tree)
+        /// @brief Erode a single leaf node by the current spatial scheme
+        ///        stored on the instance of this NodeMaskOp.
+        /// @details  Unlike dialte, this method updates the provided mask
```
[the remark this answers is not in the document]
Idclip: fixed

openvdb/openvdb/tools/Morphology.h
danrbailey: I'm be in favor of removing `dilateActiveLeafValues`. Perhaps a little more work to migrate, but less API to support is good?

openvdb/openvdb/tools/Morphology.h
danrbailey: I don't mind leaving this flag in, but I think if we include it without ever actually using it ourselves, it should at least be evaluated in the unit tests, which I don't think it is currently?

danrbailey:
> @danrbailey
> 
> > * Should there also be a unit test for the new activation header?
> 
> Yes? But I'm not going to add it in this PR. I didn't touch this code, it's simply been moved out of Morphology.h.

Ok, that's fine. Perhaps I can add this when I look at switching out the implementation of the VDB Activate SOP.

> > * You presented some nice benchmark graphs of the results of this PR - are these included in the unit tests now?
> 
> No, you can see on the spreadsheets that the VDBs in use were Houdini test assets.

If you happen to have it still, would be nice to email around the hip file with these tests in it. If not, no worries.

> > * Many of the parallel_for methods you're using do very simple operations that I imagine is only a handful of instructions, did you consider using a larger grain size than 1?
> 
> No, I can do another pass.
> 
> > * I would expect the non-threaded mode to call a lambda directly rather than to call into tbb::parallel_for() with a very large grain size.
> 
> As would I, I will change these.

Great, thanks for the other changes.

openvdb/openvdb/tools/Morphology.h
Idclip: Removed both `dilateActiveLeafValues` and `erodeActiveLeafValues`

openvdb/openvdb/tools/Morphology.h
Idclip: Added a test

Idclip:
> If you happen to have it still, would be nice to email around the hip file with these tests in it. If not, no worries.

I have the scene and will e-mail it around soon.

I think I've addressed all feedback. A reminder that once this is approved I'll have to close and re-create these changes on a new PR because of a single commit missing dco signoff and being told not to use the nice dco button :|

danrbailey (approved):
This is great, thanks for making those changes and emailing around the scene, Nick. I'm happy to approve the new PR too.

danrbailey:
I think we should bump the next release to 8.1 once this PR gets merged as well as there are some API changes included in this PR.

Idclip:
Awesome. thanks all. I've remade the PR here: https://github.com/AcademySoftwareFoundation/openvdb/pull/973

Closing this due to DCO issues. PR https://github.com/AcademySoftwareFoundation/openvdb/pull/973 is completely identical but without the history.

PR #756, opened 2020-07-06
OVDB-101: Extended sharpening support to non-level-set volumes.

danrbailey:
Resubmission of Peter's sharpening work from #455. 

@Idclip - As noted below, this does unfortunately add a new boost dependency that wasn't there before, I couldn't see a trivial way of stripping this out. :(

- Added tools::Filter::sharpen().
- Moved sharpeningKernel() from LevelSetFilter.h to Stencils.h,
  for use by both Filter and LevelSetFilter.
- Changed BaseStencil::BufferType to boost::container::vector<T>
  to work around the absence of std::vector<bool>::data().
- Added a Sharpen option to the Filter SOP.

### DIFF
--- a/openvdb/openvdb/math/Stencils.h
+++ b/openvdb/openvdb/math/Stencils.h
@@ -16,6 +16,9 @@
 #include <algorithm>
 #include <vector>                          // for std::vector
 #include <bitset>                          // for std::bitset
+#include <type_traits>
+#include <boost/container/vector.hpp>
+#include <boost/multi_array.hpp>
 #include <openvdb/math/Math.h>             // for Pow2, needed by WENO and Godunov
 #include <openvdb/Types.h>                 // for Real
 #include <openvdb/math/Coord.h>            // for Coord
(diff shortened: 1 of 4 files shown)


openvdb/math/Stencils.h
```diff
@@ -34,32 +37,33 @@ template<typename DerivedType, typename GridT, bool IsSafe>
...
-    typedef tree::ValueAccessor<const TreeType, IsSafe> AccessorType;
-    typedef std::vector<ValueType>                      BufferType;
-    typedef typename BufferType::iterator               IterType;
+    using GridType = GridT;
+    using TreeType = typename GridT::TreeType;
+    using ValueType = typename GridT::ValueType;
+    using AccessorType = tree::ValueAccessor<const TreeType, IsSafe>;
+    // Use boost::container::vector instead of std::vector,
+    // because std::vector<bool> doesn't provide a data() method.
+    using BufferType = boost::container::vector<ValueType>;
```
danrbailey, line 46: @kmuseth - this is first of two new boost dependencies being introduced in this PR.

openvdb/math/Stencils.h
```diff
@@ -1794,26 +1798,197 @@ class DenseStencil: public BaseStencil<DenseStencil<GridT, IsSafe>, GridT, IsSaf
...
     const int mHalfWidth;
 };// DenseStencil class
 
 
+//////////////////////////////////////////////////////////////////////
+
+
+/// Three-dimensional array with which to specify the kernel for a ConvolutionStencil
+template<typename ValueT>
+using ConvolutionKernel = typename boost::multi_array<
```
danrbailey, line 1826: @kmuseth - and the second boost dependency.

PR #758, opened 2020-07-10
A bad merge in PR747 grabbed the Houdini name for this parameter,

jmlait:
Fixes the wrong parameter name being used.


Idclip:
@jmlait can you add release notes for the initial change from PR747

PR #760, opened 2020-07-14
First draft of SOP_OpenVDB_Extrapolate.

apradhana:
Signed-off-by: Andre Pradhana <andre.pradhana@dreamworks.com>

This is the first draft for SOP_OpenVDB_Extrapolate, which is the Houdini-facing interface for Fast Sweeping. At this point, I only exposed three tools from Fast Sweeping:

- maskSdf

- dilateSdf

- fogToSdf

We probably want to also add the other tools, namely:

- sdfToSdf

- fogToExt

- sdfToExt

- fogToSdfAndExt

- sdfToSdfAndExt


openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,311 @@
...
+            "    they are assumed to be INSIDE the uso-surface."));
+
+    parms.add(hutil::ParmFactory(PRM_STRING, "mask", "Mask VDB")
+        .setChoiceList(&hutil::PrimGroupMenuInput2)
+        .setTooltip("Specify a VDB volume whose active voxels are to be used as a mask.")
+        .setDocumentation(
+            "A VDB volume whose active voxels are to be used as a mask"
+            " (see [specifying volumes|/model/volumes#group])"));
+
+    parms.add(hutil::ParmFactory(PRM_TOGGLE, "tiles", "Ignore Active Tiles")
```
jmlait, line 94: Token: ignoretiles

openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,311 @@
...
+            "be densified, which could significantly increase memory usage.\n\n"
+            "Proper signed distance fields don't have active tiles.\n"));
+
+    parms.add(hutil::ParmFactory(PRM_INT_J, "dilate", "Dilation")
+        .setDefault(3)
+        .setRange(PRM_RANGE_RESTRICTED, 0, PRM_RANGE_UI, 10)
+        .setTooltip("The number of voxels by which to dilate the level set narrow band"));
+
+    // Dilation Pattern
+    parms.add(hutil::ParmFactory(PRM_STRING, "nn", "Dilation Pattern")
```
jmlait, line 110: Token: pattern

apradhana:
@jmlait: I have a couple of questions about the errors generated from the auto-check.
1. Do you know what needs to be done on my side so that this PR passes through EasyCLA?
2. Do you think that some of the errors (AZP openvdb and testhou180) are generated because FastSweeping.h has not been merged to the master branch? In that case, should I also add the version of FastSweeping that works with this SOP on this branch for the time being before taking it out before the final merge?
Thank you for the help Jeff!

jmlait:
Yep, those compile errors appear to be the missing FastSweeping.h.  I wouldn't worry about those until FastSweeping is in, then you can do an update to it to move it to latest and they should clean up.

The trailing spaces is for you to fix though!  :>

apradhana:
Right, right. I'll also do convergence test using this SOP and send the result to Ken et al. soon.

openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -230,29 +259,44 @@ SOP_OpenVDB_Extrapolate::Cache::process(
...
-        // TODO: Confirm that convert is meant to be used for fogToSdf. Previously, the API call
-        // is maskSDF(*outGrid, evalFloat("isovalue", 0, time), evalInt("tiles", 0, time), nSweeps);
-        outGrid = fogToSdf(*inGrid, evalFloat("isovalue", 0, time), nSweeps);
+        outGrid = fogToSdf(*inGrid, isoValue, nSweeps);
         lsPrim->setVisualization(GEO_VOLUMEVIS_ISO, lsPrim->getVisIso(), lsPrim->getVisDensity());
+    } else if (mode == "correct") {
+        outGrid = sdfToSdf(*inGrid, isoValue, nSweeps);
+    } else if (mode == "fogext") {
+        // TODO: op
+        // outGrid = fogToExt(*inGrid, op, isoValue, nSweeps);
```
apradhana, line 288: @jmlait: Jeff, do you have a suggestion on how to expose the functor to the users in Houdini? According to the documentation, the functor defines the Dirichlet boundary condition and its signature is [](const Vec3R &xyz)->float.

jmlait: I think the most common use case is the user provides a VDB to provide the values of the functor.  Having a parameter callback for the functor would likely be too slow, and a VDB lets you pre-populate the isosurface with your favorite functor in an efficient manner.

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,768 @@
...
+    // Dynamic grid menu
+    hutil::ParmList gridParms;
+    {
+        {   // Element type menu
+            std::vector<std::string> items;
+            for (int i = 0; i < NUM_DATA_TYPES; ++i) {
+                items.push_back(dataTypeToString(DataType(i))); // token
+                items.push_back(dataTypeToMenuItems(DataType(i))); // label
+            }
+            gridParms.add(hutil::ParmFactory(PRM_STRING, "elementType#", "Type")
```
jmlait, line 341: elementtype#

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,768 @@
...
+                .setChoiceListItems(PRM_CHOICELIST_SINGLE, items)
+                .setTooltip("The type of value stored at each voxel")
+                .setDocumentation(
+                    "The type of value stored at each voxel\n\n"
+                    "VDB volumes are able to store vector values, unlike Houdini volumes,\n"
+                    "which require one scalar volume for each vector component."));
+        }
+
+        // Optional grid name string
+        gridParms.add(hutil::ParmFactory(PRM_STRING, "gridName#", "Name")
```
jmlait, line 351: This isn't a name, but a group.
matchgroup#, "Match Group"
It should also show up before Type because the match group will filter out which things this multparm will generate from?

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,768 @@
...
+            .setTooltip(bgHelpStr)
+            .setDocumentation(nullptr));
+        gridParms.add(hutil::ParmFactory(PRM_FLT_J, "bgVec3f#", "Background Value")
+            .setVectorSize(3)
+            .setTooltip(bgHelpStr)
+            .setDocumentation(nullptr));
+        gridParms.add(hutil::ParmFactory(PRM_INT_J, "bgVec3i#", "Background Value")
+            .setVectorSize(3)
+            .setTooltip(bgHelpStr)
+            .setDocumentation(nullptr));
```
jmlait, line 372: Should be bgfoo#, no internal capitilization.

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,768 @@
...
+        // }
+    }
+
+    parms.add(hutil::ParmFactory(PRM_MULTITYPE_LIST, "extGridNum", "VDBs")
```
jmlait, line 376: numextvdb

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -0,0 +1,768 @@
...
+            }
+            else {
+                outGrid = sdfToExt(*fsGrid, op, background, fsIsoValue, parms.mNSweeps);
+            }
+
+            // Create GEO_PrimVDB*
+            std::string primName = parms.mExtPrimName + "Ext";
+            outGrid->setTransform(fsGrid->transform().copy());
+            outGrid->setName(primName);
+            hvdb::createVdbPrimitive(*gdp, outGrid, primName.c_str());
```
jmlait, line 600: If I understand properly, we are creatring a new grid called fooExt for the extension.  (And fooSdf down below).  Why not replace the incoming grid with the extension grid?

If I have a Cd grid I want an extended Cd grid out, not an unmodified Cd grid and a CdExt grid?

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Extrapolate.cc
```diff
@@ -348,32 +348,32 @@ newSopOperator(OP_OperatorTable* table)
         }
 
         // Optional grid name string
-        gridParms.add(hutil::ParmFactory(PRM_STRING, "gridName#", "Name")
+        gridParms.add(hutil::ParmFactory(PRM_STRING, "matchgroup#", "Match Group")
             .setTooltip("A name for this VDB")
             .setDocumentation("A value for the `name` attribute of this VDB primitive"));
```
jmlait: This help is from vdb create?   Seems out of date with what it does.

jmlait (commented):
Thanks for fixing the tokens and switching to replacement.

There seem to be many failures from float - double in the CI, though?  They seem to be in FastSweeping.
@kmuseth : Line 1414 of FastSweeping multiplies the SdfType of `w` with the ExtType of acc2->get().  This causes promotion of float -> double, so when you try and extend a Float Ext grid using a Double SDF grid, I think you get these errors.

I'm not sure it is as easy as casting `w` to ExtType because the latter might be vector?

apradhana (commented):
> Thanks for fixing the tokens and switching to replacement.
> 
> There seem to be many failures from float - double in the CI, though? They seem to be in FastSweeping.
> @kmuseth : Line 1414 of FastSweeping multiplies the SdfType of `w` with the ExtType of acc2->get(). This causes promotion of float -> double, so when you try and extend a Float Ext grid using a Double SDF grid, I think you get these errors.
> 
> I'm not sure it is as easy as casting `w` to ExtType because the latter might be vector?

It's resolved for clang if I'm casting the result of the whole operation of, e.g., w*(d1.v*acc2->getValue(d1(ijk))+d2.v*acc2->getValue(d2(ijk))) to be ExtValue. In the case of d1.v*acc2->getValue(d1(ijk)) being a vector, it should call the [operator * defined in Vec3.h](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/math/Vec3.h#L485). 

However, I'm seeing that it's failing with g++ when ExtValueT = int. This is because of conversion error casting acc2->getValue(d1(ijk)) from int to float. Do you think we need a template-specialization for this particular case (i.e. ExtValueT = int)?

jmlait:
> It's resolved for clang if I'm casting the result of the whole operation of, e.g., w*(d1.v_acc2->getValue(d1(ijk))+d2.v_acc2->getValue(d2(ijk))) to be ExtValue. In the case of d1.v*acc2->getValue(d1(ijk)) being a vector, it should call the [operator * defined in Vec3.h](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/math/Vec3.h#L485).
> 
> However, I'm seeing that it's failing with g++ when ExtValueT = int. This is because of conversion error casting acc2->getValue(d1(ijk)) from int to float. Do you think we need a template-specialization for this particular case (i.e. ExtValueT = int)?

Yuck.. .That actually is another problem.  When extrapolating integer grids, I'd strongly recommend *NOT* to do any weighting.  Instead, we should be doing a selection based on source voxel of highest weight.

An integer grid's values usually should never be interpolated.  If we wanted interpolation, we'd be using float or double :>  Integer grids usually are storing stuff like "Slice #" or "Region ID" which is meaningless to blend.  So we probably want specialization in extrapolate, but instead extrapolation like

setValue( d1.v > d2.v ? acc2->getValue(d1(ijk)) : acc2-> getValue(d2(ijk) );

openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -1296,6 +1310,24 @@ struct FastSweeping<SdfGridT, ExtValueT>::SweepingKernel
...
+
+    template<typename ExtT = ExtValueT, typename SdfT = SdfValueT, typename std::enable_if<!std::is_same<ExtT, int>::value, int>::type = 0>
+    ExtT twoNghbr(const NN& d1, const NN& d2, const SdfT& w, const ExtT& v1, const ExtT& v2) const { return ExtT(w*(d1.v*v1 + d2.v*v2)); }// non-int implementation
+
+    template<typename ExtT = ExtValueT, typename SdfT = SdfValueT, typename std::enable_if<std::is_same<ExtT, int>::value, int>::type = 0>
+    ExtT threeNghbr(const NN& d1, const NN& d2, const NN& d3, const SdfT& /* w */, const ExtT& v1, const ExtT& v2, const ExtT& v3) const {
+        ExtT tmp = d1.v < d2.v ? v1 : v2;
+        SdfT sdfTmp = d1.v < d2.v ? d1.v : d2.v;
+        return sdfTmp < d3.v ? tmp : v3;
+    }// int implementation
```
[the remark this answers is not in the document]
kmuseth: one more comment (though just a suggestion). if d1, d2, and d3 and v1, v2, and v3 were components of Vec3 vector, say d and v, then a faster (since it's branchless) and more elegant solution would simply read as:
 
return v[math::MinIndex(d)];

Not sure if that's possible without constructing temporary Vec3's, which might be costly.

kmuseth (approved):
Did you forget to push up your changes in response to my comments? Anyway, I don't want to block you anymore (in worst case we can always go back and add my suggestions for improvement) so I've approved your PR.

jmlait:
I note the use of hvdb::GU_PrimVDB, which is relying on the injection of GU_PrimVDB into the openvdb_houdini namespace by GU_PrimVDB.  I think we've moved away from using that qualification, at least I don't see it outside of the From_Particles code.  It showed up for me as it doesn't work on the Houdini side as we obviously have our own GU_PrimVDB and don't inject it.

But nothing blocking, I can also remove the hvdb:: prefix from GU_PrimVDB in some future sync PR.

jmlait:
I have it working, thank you for putting this together.

It seems however that if the field to extrapolate does not line up with the surface to extrapolate from, it does nothing silently?  Should we at least warn in this case?

I was trying to extrapolate a velocity field out from a surface..
1) It seems the two must be matching transforms, which is always a bit worrying in pipelines.  I thought the velocity field could be used as an oracle for the surface values to avoid that requirement?
2) I kind of expected to be able to do an "outwards only" extrapolation, rather than interpolation as well.  Ie, leave interior velocity values unchanged?  I can do this with a successive composite operation, of course....

If we can't have unalligned fields we should get a warning or something so people track that down.  But I don't see anything to hold off this PR, this is all stuff we can add as successive polish as well.

PR #761, opened 2020-07-16
Add DynamicNodeManager

danrbailey:
This PR adds a new DynamicNodeManager that builds the node caches lazily. The main use case is for calling foreachTopDown() with an operation that modifies the tree (below the current level) as it is recursing down it.

There are a few small refactoring changes to the NodeManager to make this possible and you can now use a non-const operator in a foreach() just as you can in a reduce() operation.


danrbailey:
Closing this PR in favour of #763 which also includes this DynamicNodeManger. As noted there, this workflow still needs filtering to improve performance.

PR #763, opened 2020-07-21
Level Set CSG Union Merging

danrbailey:
Draft PR to give an idea of where this merging work is heading. This includes:

* NodeManager.h - a new DynamicNodeManager that caches the nodes while traversing to allow topology-changing operations
* Merge.h - a CsgUnionMergeOp designed to work with the new DynamicNodeManager with accompanying unit test
* Composite.h - manipulating tools::csgUnion() to call the new NodeManager-based functionality

Benchmarking two 1.4 billion voxel spheres merging under different configurations:

* Overlapping (Old Method): 13.0s
* Overlapping (New Method): 0.71s
* Disjoint (Old Method): 0.05ms
* Disjoint (New Method): 282ms

This is marked as draft because filtering needs to be added to the DynamicNodeManager workflow to address the performance regression in merging disjoint level sets. Questions welcome at this stage of development!

### DIFF
--- a/openvdb/CMakeLists.txt
+++ b/openvdb/CMakeLists.txt
@@ -312,6 +312,7 @@ set(OPENVDB_LIBRARY_TOOLS_INCLUDE_FILES
   tools/LevelSetTracker.h
   tools/LevelSetUtil.h
   tools/Mask.h
+  tools/Merge.h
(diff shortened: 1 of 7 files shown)


danrbailey:
This work has now been completed and submitted for review as #785.

PR #764, opened 2020-07-21
Add childCount() to RootNode, InternalNode and LeafNode

danrbailey:
This follows the convention of using leafCount() to count the number of leaf nodes and addChild() to add an immediate child node. I added a unit test to cover the statistics methods in the leaf node too.

### DIFF
--- a/openvdb/tree/RootNode.h
+++ b/openvdb/tree/RootNode.h
@@ -479,6 +479,7 @@ public:
 
     Index32 leafCount() const;
     Index32 nonLeafCount() const;
+    Index32 childCount() const;
     Index64 onVoxelCount() const;
     Index64 offVoxelCount() const;
     Index64 onLeafVoxelCount() const;
@@ -1543,6 +1544,14 @@ RootNode<ChildT>::nonLeafCount() const
         }
     }
     return sum;
+}
+
+
+template<typename ChildT>
+inline Index32
+RootNode<ChildT>::childCount() const
+{
+    return this->getChildCount();
 }
 
 
--- a/CHANGES
+++ b/CHANGES
@@ -44,6 +44,9 @@ Version 7.1.0 - In Development
       data.
     - Added AttributeSet::Descriptor::groupIndexCollision() for detecting
       group index collisions when attempting to merge two Descriptors.
+    - Added RootNode::childCount(), InternalNode::childCount() and
+      LeafNode::childCount() to count the number of immediate child nodes that
+      exist below the current node.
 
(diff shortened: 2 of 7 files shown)


openvdb/tree/RootNode.h
```diff
@@ -1546,6 +1547,14 @@ RootNode<ChildT>::nonLeafCount() const
 }
 
 
+template<typename ChildT>
+inline Index32
+RootNode<ChildT>::childCount() const
+{
+    return this->getChildCount();
+}
```
Idclip, line 1555: As I understand it, this is required due to making private methods public changing ABI on Windows? With this, can we remove `getChildCount()` entirely?

danrbailey: Yes, it looks like it's only being called by "writeTopology()" so that's an easy change. 

Incidentally, I briefly looked at trying to unify the rather confusing set of count methods - for example:

* RootNode::getActiveTileCount() - counts all the active tiles in the root node and **not** the nodes below
* InternalNode::onTileCount() - counts all the active tiles in this internal node **and** all nodes below

It's not totally clear what to do here though, so I think I'll pass on attempting this at least until we introduce a tools/Count.h to put some threaded, recursive variations.

Idclip: As removing it will break ABI as well (*?) I suggest updated all usage and wrapping the old one in an ifdef ABI 8 clause

* https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts

Idclip: Actually sorry, this is fine so long as "they're never called from an inline method"

danrbailey: Ah yes, the writeTopology method is marked inline, so we can't safely remove it.

There's no ABI=8 defined yet and we agreed a while ago not to define ABI _beyond_ the current ABI. Instead, I've created a JIRA with the "ABI8" label - https://jira.aswf.io/issues/?jql=labels%20%3D%20ABI8

danrbailey: > Actually sorry, this is fine so long as "they're never called from an inline method"

It's not totally clear to me, but writeTopology _is_ inline, so think it's safest to just punt this change until later.

Idclip: Well if there's ever any doubt we'd need to wrap it in an compile time define between major versions? Could do that now?

PR #765, opened 2020-07-22
Add stricter version warning for Blosc

matthewdcong:
Recently, we've discovered that Blosc 1.5.4 segfaults when saving OpenVDB files with LZ4 compression on Linux/GCC. The recommended version, Blosc 1.5.0, works correctly.

The current Blosc version warning in the build system only checks for differences in the major and minor version. Since we suspect that 1.5.4 has issues, I think it would be a good idea to extend the warning to check for differences in the patch version too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,8 +215,9 @@ mark_as_advanced(
 # outside of the DISABLE_DEPENDENCY_VERSION_CHECKS catch
 
 # @note  Blosc version is currently treated as exception which must be adhered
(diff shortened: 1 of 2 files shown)


kmuseth:
I can take over this PR and submit it to address the CLA issue

matthewdcong:
Sounds good, thanks Ken!

kmuseth:
This PR is not merged into PR 133

danrbailey:
Closing, PR merged as part of #778

PR #767, opened 2020-07-24
Thread LeafManager and NodeManager construction

danrbailey:
The main motivation here was to eliminate the feature where the leaf nodes of the LeafManager could be used in the NodeManager construction to improve performance (as it was causing complexity in attempting to re-use the NodeList elsewhere). The simplest solution to achieving this was to just thread the construction of the LeafManager and the NodeManager. 

I also converted the NodeList and the LeafManager over to use unique ptr arrays instead of std::deque and raw ptr arrays (I expect these classes were written before C++11 features were available). Note that I've treated the NodeList as internal API due to it being described as such in the documentation, so made more invasive changes than I would otherwise do - it's non-trivial to try and also support backwards compatibility in this class. Finally, I've marked LeafManager::getNodes() methods as deprecated, they were only used in NodeManager construction to match the methods found in the tree hierarchy, so feels safe to mark them to be removed if we remove this code path. 

Here's construction times using the old implementation based on a 2.3 billion voxel sphere:

* LeafManager construction - 127ms
* NodeManager construction from Tree - 161ms
* NodeManager construction from LeafManager - 81ms

And this new threaded implementation:

* LeafManager construction - 18ms
* NodeManager construction from Tree - 19ms
* NodeManager construction from LeafManager - 20ms (now uses the same implementation as construction from Tree)

I also tested LeafManager construction using the data sets in the testActiveLeafVoxelCount to demonstrate scalability of this implementation down to cases with fewer leaf nodes and more unbalanced hierarchies. Old Method:

* Case 1 - 0.03ms
* Case 2 - 8.7ms
* Case 3 - 12.0ms
* Case 4 - 68.0ms

New Method:

* Case 1 - 0.04ms
* Case 2 - 3.7ms
* Case 3 - 2.8ms
* Case 4 - 12.8ms

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -35,6 +35,7 @@ Version 7.1.0 - In Development
     - Changed the tree::NodeChain implementation to use an openvdb::TypeList
       rather than a boost::mpl::vector and updated all usage of tree::NodeChain
       accordingly.
+    - Threaded the construction of tree::LeafManager and tree::NodeManager.
 
     New features:
     - Added Grid::isTreeUnique() to tell if the tree is shared with another
@@ -77,6 +78,8 @@ Version 7.1.0 - In Development
     - Renamed AttributeSet::Descriptor::nextUnusedGroupOffset() to
       AttributeSet::Descriptor::unusedGroupOffset() to allow for providing an
       optional group offset hint.
+    - LeafManager::getNodes() is now deprecated as this is no longer used in
+      NodeManager construction.
 
     Houdini:
     - Platonic SOP is now verbified.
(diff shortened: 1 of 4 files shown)


danrbailey:
> Haven't reviewed this fully but, due to the nature of the changes, I propose these don't go in to 7.1 (I'd rather not re-live [OVDB-117](https://jira.aswf.io/browse/OVDB-117))

Yeah I don't mind not including this for 7.1. Not going to introduce it right now, but might be worth having a new remote branch for this work while we're waiting on 7.1 to go out.

danrbailey:
Closing this PR, I've rolled these changes into #785.

PR #769, opened 2020-07-24
Building with GCC 9.1

danrbailey:
This adds a GCC 9.1 build to OpenVDB (we were previously only building from VFX 2021 using Clang 10.0) and fixes the following warnings:

* Swapped out std::random_shuffle for std::shuffle (was being used in one of the unit tests)
* Fix a number of memcpy operations that produced warnings with GCC's new class-memaccess warning
* Remove bool and mask grid support from openvdb_lod binary to fix int-in-bool-context
* Use various bool traits and refactoring to fix int-in-bool-context warnings


openvdb/tools/Interpolation.h
```diff
@@ -796,7 +798,30 @@ QuadraticSampler::triquadraticInterpolation(ValueT (&data)[N][N][N], const Vec3R
...
+{
+    static bool eval(const bool* value, double weight)
+    {
+        OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN
+        const bool
+            a = static_cast<bool>(0.5 && (value[0] + value[2]) - value[1]),
+            b = static_cast<bool>(0.5 && (value[2] - value[0])),
+            c = static_cast<bool>(value[1]);
+        const auto temp = weight && (weight && a + b) + c;
+        OPENVDB_NO_TYPE_CONVERSION_WARNING_END
```
jmlait, line 813: Not sure if this is actually used at all, but I'm having trouble figuring out the logic.  Is the 0.5 && to force a cast to float for the rest?

Giving values v0, v1, v2:
a = 0.5 v0 - v1 + 0.5 v2
b = - 0.5 v0 + 0.5 v2
c = v1
result = w * (w * a + b) + c;
= w^2 a + w b + c
= w^2 (0.5 v0 - v1 + 0.5 v2) + w (- 0.5 v0 + 0.5 v2) + v1
= (0.5 w^2 - 0.5 w) v0 + (- w^2 + 1) v1 + (0.5 w^2 + 0.5 w) v2
A sensible bool behaviour is to pick the term with highest weight.  So on the [-1..1] interval, the first transition is 
0.5 w^2 -0.5 w > -w^2 + 1
3 w^2 - w -2 > 0
(3w  + 2) (w - 1) > 0
3w + 2 = 0
w = -2/3

Symmetry argument gives w = 2/3 for positive cross over.

static bool eval(const bool *value, double weight)
{
   if (weight < -2./3.)
    return value[0];
  if (weight > 2./3.)
    return value[2];
  return value[1];
}

Not sure how this crosses over from voxel to voxel as I've not looked at the rest of the function.

But any interpolation of bool should be replaced by a selection of some sort keyed on the interpolant.

danrbailey: I tracked down where this implementation was being used and it was in the point sample code. Instead of trying to fix it here, I added conditional logic for bool and mask grids to resort to always use the PointSampler instead of the BoxSampler or QuadraticSampler. That seems to be the approach used elsewhere and that has fixed this warning now, so we don't need to figure out what the correct logic should be here.

PR #771, opened 2020-07-27
OpenVDB Points Convert SOP Fixes

Idclip:
- Fixed a bug where the auto voxel transform was being calculated on the packed geometry in the OpenVDB Points Convert SOP
- Fixed a bug where points representing packed geometry were also being converted in the OpenVDB Points Convert SOP

### DIFF
--- a/openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
@@ -828,7 +828,7 @@ SOP_OpenVDB_Points_Convert::Cache::cookVDBSop(OP_Context& context)
             // instead of mergePrims to make sure we merge points that are not
             // associated with any primitive
             GA_PointGroup pointsWithPackedPrims(*ptGeo);
-            pointsWithPackedPrims.GA_ElementGroup::operator|=(*packgroup);
+            pointsWithPackedPrims.combine(packgroup.get());
 
             // Merge everything except the geo associated with prims we've just unpacked
             nonConstDetail.mergePoints(*ptGeo, GA_Range(pointsWithPackedPrims, /*invert*/true));
--- a/openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
+++ b/openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
@@ -799,6 +799,45 @@ SOP_OpenVDB_Points_Convert::Cache::cookVDBSop(OP_Context& context)
         gdp->clearAndDestroy();
 
         const GU_Detail* ptGeo = inputGeo(0, context);
+        GU_Detail nonConstDetail; // tmp storage of unpacked geo
+        const GU_Detail* detail;  // ptr to geo to convert; either ptGeo or nonConstDetail
+
+        boss.start();
+
+        // Unpack any packed primitives
+
+        std::unique_ptr<GA_PrimitiveGroup> packgroup(ptGeo->newDetachedPrimitiveGroup());
+
+        for (GA_Iterator it(ptGeo->getPrimitiveRange()); !it.atEnd(); ++it) {
+            GA_Offset offset = *it;
+            const GA_Primitive* primitive = ptGeo->getPrimitive(offset);
+            if (!primitive || !GU_PrimPacked::isPackedPrimitive(*primitive)) continue;
+
+            const GU_PrimPacked* packedPrimitive = static_cast<const GU_PrimPacked*>(primitive);
+            packedPrimitive->unpack(nonConstDetail);
+            packgroup->addOffset(offset);
+        }
+
+        if (packgroup->entries() == 0) {
+            // If no packed geometry was converted, avoid the merge by
+            // simply using the original geometry
+            detail = ptGeo;
+        }
+        else {
+            // Convert the prim group to points - we have to use mergePoints
+            // instead of mergePrims to make sure we merge points that are not
+            // associated with any primitive
+            GA_PointGroup pointsWithPackedPrims(*ptGeo);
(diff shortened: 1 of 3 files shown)


openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
```diff
@@ -799,6 +799,45 @@ SOP_OpenVDB_Points_Convert::Cache::cookVDBSop(OP_Context& context)
...
+        }
+        else {
+            // Convert the prim group to points - we have to use mergePoints
+            // instead of mergePrims to make sure we merge points that are not
+            // associated with any primitive
+            GA_PointGroup pointsWithPackedPrims(*ptGeo);
+            pointsWithPackedPrims.GA_ElementGroup::operator|=(*packgroup);
+
+            // Merge everything except the geo associated with prims we've just unpacked
+            nonConstDetail.mergePoints(*ptGeo, GA_Range(pointsWithPackedPrims, /*invert*/true));
```
Idclip, line 834: @jmlait might be worth removing the operator overrides for `GA_PointGroup`, `GA_PrimitiveGroup` etc as they seem to just call through to `GA_ElementGroup` and obfuscate the input group type - assuming I've understood this behaviour correctly!

openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
```diff
@@ -827,40 +866,11 @@ SOP_OpenVDB_Points_Convert::Cache::cookVDBSop(OP_Context& context)
...
-
-            detail = ptGeo;
-        }
-        else {
-            // merge unpacked and packed point data
-
-            nonConstDetail.mergePoints(*ptGeo);
-            detail = &nonConstDetail;
-        }
-
```
Idclip, line 863: Altered and moved this logic to *before* the calculation of the voxel size

openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
jmlait: The &=, -=, etc overrides make a lot of sense as you can only apply that with matching group types.  So you can't have 
pointgroup -= primgroup;
As far as I understand, the overrides then enforce that restriction.
The issue is that |= does implicit conversion of the group type.  The ElementGroup version used to be a private method until GA_ElementGroup was merged into GA_ATIGroupBool. The latter was always meant to be a private interface, but got exposed as ElementGroup must be public.  (This was 4y9m ago, so likely compatible with all current OpenVDB targets)

Long story short, the |= working with different group types wasn't meant to be exposed :>  I think the best solution is to have the GA_PointGroup overrides accept an ElementGroup as the second input.  I'll add an RFE to this effect and see if there are any issues.

jmlait:
@danrbailey : Isn't this such a filter?

    parms.add(hutil::ParmFactory(PRM_STRING, "group", "Group")
        .setChoiceList(&hutil::PrimGroupMenu)
        .setTooltip("Specify a subset of the input point data grids to convert.")

Or do you mean to also filter the primitives to unpack?

openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
jmlait: Why not use combine() rather than this awkward |=?

Idclip: I wasn't entirely sure if combine did the same thing but will definitely switch if so! Is this the main way to "cast" one group type to another?

danrbailey:
> @danrbailey : Isn't this such a filter?
> 
> ```
> parms.add(hutil::ParmFactory(PRM_STRING, "group", "Group")
>     .setChoiceList(&hutil::PrimGroupMenu)
>     .setTooltip("Specify a subset of the input point data grids to convert.")
> ```
> 
> Or do you mean to also filter the primitives to unpack?

Well, this parameter is the primitive group filter for conversion in the other direction (VDB -> Houdini). It's hidden for this direction (Houdini -> VDB) because it doesn't make a lot of sense to have a primitive group filter for points. However, it would at least make sense to include a points group filter, but whether or not that's a good idea or not, I'm not sure. I notice that some of the other VDB sops that handle points (such as VDB from particles) also don't expose a points group filter, so maybe it's just not convention to do this? I can't remember if we talked about this at some point and then decided against it.

jmlait:
I think I decided against the same named parameter being used as different filter types.  Having a separate ptgroup filter does make sense here though.  It could also subsume some of the implicit logic that is going on, like rejecting points that are part of packed primitives.

Idclip:
> It could also subsume some of the implicit logic that is going on, like rejecting points that are part of packed primitives.

Are you suggesting we shouldn't always be ignoring points representing packed geo? I was going to suggest we extend this behaviour for points representing other geo too (i.e. volumes). Would converting these points ever be useful/wanted?

openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
jmlait: Yep, combine will chain to |= if the src is an element group. It then handles in addition weird stuff like edge groups.

The |= in ElementGroup will be marked deprecated in 18.5.

jmlait:
I think we should always ignore points of packed geo.  But it definitely is the case one could imagine other sets of points to be removed as well.  Having a group makes this easier to do without an extra blast operation.  (Which also may avoid a defragment, so be a performance win)

Idclip:
Will merge this soon

openvdb_houdini/SOP_OpenVDB_Points_Convert.cc
Idclip: Switch to combine!

PR #777, opened 2020-08-11
TSC Meeting notes

danrbailey:


tsc/meetings/2020-08-11.md
```diff
@@ -0,0 +1,69 @@
...
+two PRs are 746 (fast sweeping) and 775 (openexr 2.5 fix). Nick to merge 775,
+Ken / Dan to look at DCO issue and then once merged, 7.1 can be released.
+
+Decision to hold off on merging other PRs until after 7.1. Next time, a separate
+release branch would be preferable here.
+
+6) NanoVDB
+
+Question about how to deploy. Jeff and Dan favour doing a lightweight
+integration into OpenVDB. All agreed that this is the best choice in light of
```
kmuseth, line 63: can you elaborate on what "lightweight integration means". My recollection is that we where debating two different approaches 1) adding NanoVDB into the existing opevenvdb repository as a "feature branch" or 2) Initially releasing NanoVDB as a separate repository and then integrate it into openvdb later. Jeff and Dan were in favor of 1), whereas Nick and Ken preferred 2). We agreed to continue this discussion at the next meeting.

danrbailey: I can clarify that - what I meant by lightweight integration was the "feature branch" option. 

Initially, yes I believe we were divided on how to proceed here. However, it was my understanding that the deciding factor here was the realization that NanoVDB would have to be hosted by the ASWF. For the separate repo option (2), we would thus need to request a new NanoVDB ASWF repo which would also require setting up CLAs, license agreements, etc. Finally, we coalesced around the feature branch option (1). It sounds like that wasn't your take away from this discussion, so I'm happy to leave this as open discussion and put this on the next agenda to discuss further?

kmuseth: Actually, we did have a vote in the last few seconds didn't we - so we decided on 1 - ALL of us. Let's write that instead.

Regarding 2 and the additional repo - I talked to the TAC about that and it's not a problem since openexr has already do this.

danrbailey: I updated the wording here.

PR #782, opened 2020-08-13
Update next release to 7.1.1

danrbailey:


danrbailey:
> it's probably optimistic to think that our next release is a patch release, but it's a good startingpoint

Agreed. I hesitated on jumping straight to 7.2.0

PR #785, opened 2020-08-14
Faster CSG Operations

danrbailey:
Update tools::csgUnion(), tools::csgIntersection() and tools::csgDifference() with an efficient breadth-first, threaded implementation that steals data and is significantly faster.

Current implementation:

CSG Union (Overlap) ... completed in 9 seconds and 726.057 milliseconds (9726.057ms)
CSG Union (Disjoint) ... completed in 0.087 milliseconds

New implementation:

CSG Union (Overlap) ... completed in 771.590 milliseconds
CSG Union (Disjoint) ... completed in 0.044 milliseconds

There are a lot of changes included in this PR:

* LeafManager and NodeManager construction is now threaded
* NodeManager construction now uses the tree from a LeafManager not the nodes during construction
* LeafManager::getNodes() is now marked deprecated
* New DynamicNodeManager class that offers foreachTopDown() and reduceTopDown() modes that work efficiently with branch-filtered, topology-changing operations
* New TreeToMerge class that can be used with a const or non-const Tree to steal or deep-copy nodes
* New tools::CsgUnionOp, tools::CsgIntersectionOp and tools::CsgDifferenceOp operators that can be used with a DynamicNodeManager and offers the ability to merge an array of grids
* Adapt tools::csgUnion(), tools::csgIntersection() and tools::csgDifference() to use these new operators
* Remove all CSG visitor classes and mark all Tree visitor methods as deprecated


danrbailey:
There's a sporadic unit test failure on Mac and Windows when merging partially-constructed leaf nodes. This is a rare edge case, which I'll fix soon. Please don't hold off reviewing this PR in the interim. :)

openvdb/openvdb/tree/LeafManager.h
```diff
@@ -231,17 +227,10 @@ class LeafManager
...
     }
 
-    virtual ~LeafManager()
-    {
-        if (mIsMaster) {
-            delete [] mLeafs;
-            delete [] mAuxBuffers;
-        }
-    }
+    virtual ~LeafManager() { }
```
Idclip, line 233: Unrelated to your PR, but why is this even virtual

openvdb/openvdb/tree/LeafManager.h
```diff
@@ -635,34 +615,99 @@ class LeafManager
...
-    };
-
-    void initLeafArray()
+    void initLeafArray(bool serial = false)
     {
-        const size_t leafCount = mTree->leafCount();
+        // Build an array of all nodes that have leaf nodes as their immediate children
+
+        using NodeChainT = typename NodeChain<RootNodeType, RootNodeType::LEVEL>::Type;
+        using NonConstLeafParentT = typename NodeChainT::template Get</*Level=*/1>;
```
Idclip, line 623: Guess this isn't guaranteed to be the leaf node parent! Should probably be:
`
using NonConstLeafParentT = typename NodeChainT::PopBack::Back;
`

openvdb/openvdb/tree/LeafManager.h
```diff
@@ -635,34 +615,99 @@ class LeafManager
...
+        std::vector<Index32> leafCounts;
+        if (serial) {
+            leafCounts.reserve(leafParents.size());
+            for (LeafParentT* leafParent : leafParents) {
+                leafCounts.push_back(leafParent->childCount());
+            }
+        } else {
+            leafCounts.resize(leafParents.size());
+            tbb::parallel_for(
+                tbb::blocked_range<size_t>(0, leafParents.size(), /*grainsize=*/64),
```
Idclip, line 640: remove arbitrary grain size maybe? Or expose as option

openvdb/openvdb/tree/LeafManager.h
```diff
@@ -763,8 +808,8 @@ class LeafManager
             for (typename LeafRange::Iterator it = range.begin(); it; ++it) op(*it, it.pos());
         }
         void join(const LeafReducer& other) { mLeafOp->join(*(other.mLeafOp)); }
-        LeafOp *mLeafOp;
-        const bool mOwnsOp;
+        std::unique_ptr<LeafOp> mLeafOpPtr;
+        LeafOp *mLeafOp = nullptr;
```
Idclip, line 812: Is extra complexity worth the `LeafReducer` simply copying the op in the main constructor?

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -31,32 +30,152 @@ template<typename TreeOrLeafManagerT, Index LEVELS = TreeOrLeafManagerT::RootNod
...
+        if (serial) {
+            nodeCounts.reserve(parents.nodeCount());
+            for (size_t i = 0; i < parents.nodeCount(); i++) {
+                if (!nodeFilter.valid(i))   nodeCounts.push_back(0);
+                else                        nodeCounts.push_back(parents(i).childCount());
+            }
+        } else {
+            nodeCounts.resize(parents.nodeCount());
+            tbb::parallel_for(
+                tbb::blocked_range<Index64>(0, parents.nodeCount(), /*grainsize=*/64),
```
Idclip, line 115: hardcoded grainsize

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -530,13 +681,204 @@ class NodeManager
...
+        if (!mList.initNodeChildren(parent, nodeFilterOp, !threaded))   return;
+        mList.reduceWithIndex(nodeFilterOp.op(), threaded, grainSize);
+    }
+
+protected:
+    NodeList<NodeT> mList;
+};// DynamicNodeManagerLink class
+
+
+template<typename TreeOrLeafManagerT, Index _LEVELS>
```
Idclip, line 846: You mentioned you removed the ability to construct NodeManagers with LeafManagers - perhaps this should just be `TreeT`? Guess it will work either way, but you could then further guarantee the type with something like
`static_assert(std::is_base_of<TreeBase, TreeOrLeafManagerT>::value, "Type to DynamicNodeManager is not a valid tree type");`

openvdb/openvdb/tree/LeafManager.h
danrbailey: Good question, I'm not sure. Perhaps I can even remove it here.

openvdb/openvdb/tree/LeafManager.h
danrbailey: I tend to follow these two bits of advice from TBB:

* A rule of thumb is that grainsize iterations of operator() should take at least 100,000 clock cycles to execute
* You do not have to set the grainsize too precisely

It's not arbitrary - my general approach is to pick between grain sizes of 1, 64 and 1024 and err on the low side. In this case, I opted for 64 as for a typical node size, this will translate to 64 popcount() instructions which will likely not be sufficient to maximise throughput. You'll notice that the subsequent parallel_for has no grain size because there is plenty of work being done by the operation, hence why exposing it doesn't really make sense.

openvdb/openvdb/tree/NodeManager.h
danrbailey: I removed the ability to construct NodeManagers from LeafManager **leaf arrays** - it is still possible to construct a NodeManager from a LeafManager as the call to `tree.root()` will map to `LeafManager::root()`. It's just that the root node will be used for the construction rather than the internal leaf array.

openvdb/openvdb/tree/LeafManager.h
danrbailey: It entirely depends on how expensive a deep-copy of the op is and how many threads you're running the reduction on? There's no stipulation that an op should _not_ store large amounts of data inside it. I mean we would probably advise against it, but this pattern at least mitigates against the worst outcome.

openvdb/openvdb/tree/LeafManager.h
Idclip: I would then add a comment as to why 64 is a better value here from your reasons given, otherwise it does look arbitrary and could very well have come from a value you chose based on your particular hardware.

danrbailey: Sure, I can add a quick comment about this here.

danrbailey:
I believe I've now addressed all the feedback given in the review and on the PR. In particular, I've modified the constructors for TreeToMerge and CSG Operators to include a tag dispatch class. I've also added more documentation to the DynamicNodeManager class.

openvdb/openvdb/tree/LeafManager.h
Idclip: Thanks - I think as a rule, if you select a particular grainsize other than 1, then a brief comment should accompany it

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -3,13 +3,12 @@
 
 /// @file tree/NodeManager.h
 ///
-/// @author Ken Museth
+/// @authors Ken Museth, Dan Bailey
```
Idclip, line 6: Do we have any guidelines on this? I've seen some project use `Modified by:`. Fine by me, I just haven't been changing authors of existing headers in PRs (filter/morphology etc)

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+    // returns true if mask has been initialized
+    bool hasMask() const;
+
+private:
+    struct MaskUnionOp;
+
+    typename TreeType::Ptr mTreePtr;
+    TreeType* mTree = nullptr;
+    const TreeType* mConstTree = nullptr;
+    typename MaskTreeType::Ptr mMaskTree;
```
Idclip, line 122: Personally I'm not a fan of a single class representing two states of constness. I think a better design would be for a specialisation of this convenience class which is templated on a const type. `stealOrDeepCopyNode` could then be called something like `getNode` and be defined with respect to the specialisation - this also better matches the naming of `addTile` with respect to its behaviour. It also removes this tagging constructor stuff (though you could still keep it to ensure the correct specialisation has been selected) and simplifies the members (no mask on non-const). That would be my preference as this class is pretty much boilerplate except for those methods. This wouldn't be as much of an issue if this wasn't part of the public API and only used internally. Also, do you need to store a shared pointer to the tree here? Could these be references instead and remove the ability to pass nullptrs? Could the mask be a `unique_ptr` or have I misunderstood how that's used?

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+
+    // Processes the leaf nodes. Required by the NodeManager
+    bool operator()(LeafT& leaf, size_t idx) const;
+
+private:
+    // on processing the root node, the background value is stored, retrieve it
+    // and check that the root node has already been processed
+    const ValueT& background() const;
+
+    mutable std::vector<TreeToMerge<TreeT>> mTreesToMerge;
```
Idclip, line 217: minor: it's weird that this has to be mutable when the container doesn't change. Maybe this should be a vector of pointers.

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+    if (!mOtherBackground)  mOtherBackground = &mTree.rootPtr()->background();
+
+    // find all tile values in this root and track inside/outside and active state
+    // note that level sets should never contain active tiles, but we handle them anyway
+
+    constexpr uint8_t ACTIVE_TILE = 0x1;
+    constexpr uint8_t INSIDE_TILE = 0x2;
+    constexpr uint8_t CHILD = 0x4;
+
+    auto getTileFlag = [&](auto valueIter) -> uint8_t
```
Idclip, line 746: minor: pass iterator by ref

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+
+private:
+    // on processing the root node, the background values are stored, retrieve them
+    // and check that the root nodes have already been processed
+    const ValueT& background() const;
+    const ValueT& otherBackground() const;
+
+    // note that this vector is copied in NodeTransformer every time a foreach call is made,
+    // however in typical use cases this cost will be dwarfed by the actual merge algorithm
+    mutable TreeToMerge<TreeT> mTree;
```
Idclip, line 274: No vector here - will this only ever work on 2 trees?

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+
+template <typename TreeT>
+const typename CsgDifferenceOp<TreeT>::ValueT&
+CsgDifferenceOp<TreeT>::otherBackground() const
+{
+    if (!mOtherBackground) {
+        OPENVDB_THROW(RuntimeError, "Other background value not set. "
+            "This operator is only intended to be used with foreachTopDown().");
+    }
+    return *mOtherBackground;
```
Idclip, line 923: Can we remove these runtime errors in favour of asserts? The same at line 721. nounwind/noexcept  opts are always nice (though I realise some other internal methods might throw, we can still avoid unnecessary ones - these are private after all).

Idclip (approved):
I'm approving - left a few more comments on Merge.h but in general it looks great

openvdb/openvdb/tree/NodeManager.h
jmlait: Author tags are a bit of a weird thing anyways, I mean, that is what version control is for :>  It's fine by me to leave it unspecified.

openvdb/openvdb/tools/Merge.h
jmlait: So there'd be two TreeToMerge specializations:

TreeToMergeDeepCopy
TreeToMergeSteal

The Deep copy would take a const pointer, and the Steal would take a non-const?

Won't this run into the same issue where we'd still need something to represent *either* type of tree?  For when we have a std::vector<TreeToMerge> and want it to have a mix of deepcopy and steal trees?  We're back to a virtual base class; or building another opaque wrapper around TreeToMergeDeepCopy and TreeToMergeSteal to alias them...  Since this is rather special case and lightweight, I don't see anything wrong with just collapsing it into a single structure.

Here's a similar class we ended up with in the HDK, FWIW:
https://www.sidefx.com/docs/hdk16.0/class_g_o_p___group_parse_1_1_group_creator.html

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+    /// lightweight mask tree with the same topology is created that can be
+    /// pruned to use as a reference. Initialization of this mask tree can
+    /// optionally be disabled for delayed construction.
+    TreeToMerge(TreeType* tree, DeepCopy tag, bool initialize = true)
+        : TreeToMerge(static_cast<const TreeType*>(tree), tag, initialize) { }
+
+    /// @brief Reset the non-const tree shared pointer. This is primarily
+    /// used to preserve the order of trees to merge in a container but have
+    /// the data in the tree be lazily loaded or resampled.
+    void reset(typename TreeType::Ptr treePtr)
```
jmlait, line 77: This bypasses the Steal flag?   Should it take the Steal Flag?

Does it make sense to reset() a tree to a DeepCopy mode for the same delay-load argument?

Or is reset() entirely internal to the operators working on TreeToMerge?

openvdb/openvdb/tools/Merge.h
jmlait: Thinking on it, it isn't the case that TreeToMerge is "Const" or "Non-Const".  It is either "Stealable" or "Deep-Copy"

I'd be rather tempted to just const_cast away the const inside of TreeToMerge so it only needs a single tree pointer type, but instead uses a state flag if it is in DeepCopy or Steal mode.

openvdb/openvdb/tools/Merge.h
```diff
@@ -0,0 +1,931 @@
...
+    {
+        mTreePtr = treePtr;
+        mTree = mTreePtr.get();
+        mConstTree = nullptr;
+    }
+
+    /// @brief Return a pointer to the non-const tree.
+    TreeType* tree() { return mTree; }
+    /// @brief Return a pointer to the const tree.
+    const TreeType* constTree() { return mConstTree; }
```
jmlait, line 87: These are a bit of a misnomer.
stealTree()
deepCopyTree()
would be more accurate as they reflect if it is in that mode or not.

Otherwise, it seems odd that constTree() doesn't return mTree if mConstTree is nullptr; as that is also a valid const tree!

Those sound like verbs though.  treeToStealFrom().  treeToDeepCopyFrom()?

openvdb/openvdb/tools/Merge.h
jmlait: I think this is why it isn't included in the Union/Intersect operator.  I kind of expected difference to be there, and had a comment to that effect before I noticed an intersection class.  I believe the logic is that difference isn't as well defined on a list as the first member is treated differently.

Given (T0, T1, T2, ...)
it becomes
T0 & ~( T1 | T2 | ... )
ie, union of the tail, then intersection of the inversion of the tail with the head.

openvdb/openvdb/tools/Merge.h
jmlait: I agree, however, I note that we should have non-crashing recovery code for mOtherBackground that makes sense.

(I'm also reminded I want us to use a VDB_ASSERT for better control as well...)

jmlait (approved):

openvdb/openvdb/tools/Merge.h
Idclip: > The Deep copy would take a const pointer, and the Steal would take a non-const?
> Won't this run into the same issue where we'd still need something to represent either type of tree?  For when we have a std::vector and want it to have a mix of deepcopy and steal trees?

I think we'd just have two vectors, one for each specialization - no need for a shared base or any more infrastructure

openvdb/openvdb/tree/NodeManager.h
danrbailey: Perhaps a topic for a TSC meeting? There is a small benefit to me at least in knowing who to direct a question to without going through version control history. Up until ASWF adoption, Git commits were typically from the OpenVDB DWA GitHub account, whereas these author comments are often more informative. I sort of go by the assumption that everyone that's made a substantial contribution should be listed as primary authors for quick reference, but we can discuss changing this...

openvdb/openvdb/tools/Merge.h
danrbailey: > I think we'd just have two vectors, one for each specialization - no need for a shared base or any more infrastructure

I think this a point I should make clearer in the documentation. It is _really_ important to preserve the order of grids regardless of whether or not they are deep-copy or steal. The reason is that at least doing certain operations over floating-point trees, you'll get different results due to floating-point associativity. I believe strongly that there should never be a change in the resulting grid by changing how you're pulling data from it (ie stealing instead of deep-copying), so we need to mix the two types while preserving the order. I tried a few different implementations of this such as using two vectors as you're suggesting @Idclip and simply storing an integer parameter for the order. I found that it ended up with a more complicated public API and was more complicated to manage internally too, so I settled on this wrapped tree approach where the deep-copy/steal state was an abstraction.

> I'd be rather tempted to just const_cast away the const inside of TreeToMerge so it only needs a single tree pointer type, but instead uses a state flag if it is in DeepCopy or Steal mode.

I think that's a good suggestion. As long as there's no way to pass in a const tree to be stolen, there should be no opportunity for abuse. I'll try that out.

> Also, do you need to store a shared pointer to the tree here?

I found this useful when I was resampling from an existing const tree so as to match the target transform (a very common operation!). In that case, you want to be able to steal from this new temporary tree, but you also need to hold on to a shared pointer somewhere to maintain ownership. It didn't seem like a major extension to this class to support this and cleaned up the client code quite a bit not having to worry about storing a buffer of these temporary trees.

> Could these be references instead and remove the ability to pass nullptrs?

It's quite convenient to be able to store a TreeToMerge class with a nullptr because you can use it as a placeholder for where a tree will be inserted. This means you can do things asynchoronously - loading a tree or resampling another tree as discussed above.

> Could the mask be a unique_ptr or have I misunderstood how that's used?

Yes, the mask can be a unique pointer, I can change that. It would be nice to have a unique pointer typedef or using alias in the Tree class.

openvdb/openvdb/tools/Merge.h
danrbailey: Yeah, I don't love this aspect of this implementation. There's a lot I successfully managed to change about the NodeManager implementation, but the requirement for having a const operator when doing a foreach operation was a bridge too far. Right now, it's just a detail of the class, so it's something we could theoretically come back and change at a later date.

openvdb/openvdb/tools/Merge.h
danrbailey: Yes, as @jmlait says, it's not totally clear what difference should do if there are multiple grids. You can achieve `T0 & ~( T1 | T2 | ...)` by doing `Difference(T0, Merge(T1, T2, ...))` so I felt it was clearer to enforce that especially as there's no performance penalty in doing so. There's currently no difference operation in my implementation of the VDB Merge SOP for this exact reason. That suggests that we would always want to keep the VDB Combine SOP (or equivalent) around in some form or another for binary operations.

openvdb/openvdb/tools/Merge.h
danrbailey: I'll take a look at this, thanks.

openvdb/openvdb/tools/Merge.h
danrbailey: No, reset() is not entirely internal. I think for consistency it makes sense to add it, especially if we represent the mode by a state flag instead of whether the tree is const/non-const. I don't really see a good use case for this, but perhaps there's one I'm missing. I'll change this.

openvdb/openvdb/tools/Merge.h
danrbailey: Yes, I agree with you, let's remove the const/non-const differentiation entirely and use steal/deep-copy. `stealTree()` is clearly confusing, perhaps we just drop the From - `TreeToMerge::treeToSteal()`?

openvdb/openvdb/tools/Merge.h
Idclip: >  I tried a few different implementations of this such as using two vectors as you're suggesting @Idclip and simply storing an integer parameter for the order. I found that it ended up with a more complicated public API and was more complicated to manage internally too, so I settled on this wrapped tree approach where the deep-copy/steal state was an abstraction.

Fair enough - I would say however that I'm not sure how this would create a more complicated **public** API, unless you do mean the need for this index tracking. You reference fp arithmetic which is absolutely a valid case for retaining order, but ops like CsgUnion are a great example where this does not need to be respected. You're providing a container which can be utilised within a custom visitor, but it's down to you to implement this visitor correctly - and, if by correctly we mean must adhere to deterministic ordering, then that's something you as a user must do. Having a single container just so it can be stored in a single vector to help with this feels a bit arbitrary, especially along with the other issues identified. I'll abstain from any further comments on this though.

> In that case, you want to be able to steal from this new temporary tree, but you also need to hold on to a shared pointer somewhere to maintain ownership.
>  It didn't seem like a major extension to this class to support this and cleaned up the client code quite a bit not having to worry about storing a buffer of these temporary trees.

Right so you're using this container for new storage too? When you say buffer of tmp trees, this could be achieved with another vector of tree ptrs? Removing the ability for this class to hold a null tree imo would be a worthwhile thing to consider (you'd remove a bunch of if(tree) checks too). Along with "Pay for what you use" etcetc

danrbailey: > Fair enough - I would say however that I'm not sure how this would create a more complicated public API, unless you do mean the need for this index tracking. 

Essentially, yes you need to track the index on the client end to these types of operations. Where this complicates a public API is for asynchronous workflows such as the resampling example and where you care about order. In this case, you want to swap the contents of one of these TreeToMerge objects with a new tree. To do so, you have to maintain a reference to this TreeToMerge object or if you use two tree vectors, an index and a flag to select the right element from the right vector. 

> You reference fp arithmetic which is absolutely a valid case for retaining order, but ops like CsgUnion are a great example where this does not need to be respected. You're providing a container which can be utilised within a custom visitor, but it's down to you to implement this visitor correctly - and, if by correctly we mean must adhere to deterministic ordering, then that's something you as a user must do.

It is a valid point and I did consider this. On balance, I think maintaining a consistency of APIs to all multi-tree operators is nice. That means not having a CsgUnionOp with one set of constructors and a FogVolumeAddOp (for example) with a different set of constructors because it needs to be deterministic. Internally, we could extract the data from one of these TreeToMerge arrays on construction if we know determinism isn't required, but the benefits in doing so are (relatively) limited and this change won't affect the public API.

> Right so you're using this container for new storage too? When you say buffer of tmp trees, this could be achieved with another vector of tree ptrs? 

Yes, this could be achieved with another vector of tree ptrs. That was where I started with this. However, as you can construct a TreeToMerge array _before_ constructing the operator, you'd end up having to store this vector of tree pointers in the client code. It is a convenience not to have to do this. If you'll still thinking about this along the lines of doing away with the TreeToMerge array for the CsgUnionOp, then consider it in the context of a deterministic operator such as FogVolumeAddOp.

> Removing the ability for this class to hold a null tree imo would be a worthwhile thing to consider (you'd remove a bunch of if(tree) checks too). Along with "Pay for what you use" etcetc

This is probably a tiny optimization. Nonetheless, I will try and get rid of the ability to store a nullptr - with the slight change to the internal storage, that might be more feasible. A related question is not being able to optimize away the check if a tree can be stolen from or not in the inner loop, despite being able to trivially check if _any_ tree in the array can be stolen from. I think the simple solution to this is to tweak the implementation of the three methods on the critical path to add a new template parameter:

```
    template<typename NodeT, bool DeepCopyOnly = false>
    const NodeT* probeConstNode(const Coord& ijk) const;
```

That way, you could modify the internals of the operator to take advantage of this. I'm not going to do that here, just highlighting that these kind of optimizations could theoretically be introduced without breaking the public API. Let's not lose sight of the huge performance boost in using this technique in the first place. :)

Idclip: > This is probably a tiny optimization.
> Let's not lose sight of the huge performance boost in using this technique in the first place. :)

Absolutely - this feedback isn't related to performance optimizations (i.e. removing the nullptr check is imo a better overall design). Feel free to use or discard these points as you see fit!

openvdb/openvdb/tools/Merge.h
jmlait: Yep, dropping the From sounds wise.

openvdb/openvdb/tools/Merge.h
jmlait: > Yes, the mask can be a unique pointer, I can change that. It would be nice to have a unique pointer typedef or using alias in the Tree class.

We ran into this ourselves, as we often have a Ptr typedef.  In our case Ptr is supposed to be a raw pointer, so we use SPtr and UPtr as the defaults for shared and unique pointers.  Personally I don't like Ptr being an implicit shared pointer, as there are performance and ownership implications that should be be made explicit.

danrbailey:
I believe I've addressed all the remaining feedback now. Thanks for all the suggestions!

openvdb/openvdb/tools/Merge.h
danrbailey: Yes, it's not ideal having Ptr being a shared pointer. I'm not sure how we'd go about changing that though, it would break so much code! One tricky aspect of using a unique pointer as a member of a class is not being able to implicitly define the copy constructor. In this case, I wrapped in it a little struct that only contains the unique pointer and does a deep-copy on copy construction so that it's not necessary to add logic for copying all the other member variables.

PR #786, opened 2020-08-14
Synchrnoize with 18.5.  

jmlait:
Main changes
- More meta attribute support
- use overrides
- Fix bug with empty grids early-terminating the enlargeBoundingBox.

Note this is just a reference implementation.


Idclip:
might be worth adding release notes but I'll leave it down to you!

PR #788, opened 2020-08-17
Active tile count

Megidd:
### Goal

There are situations in which there is curiosity to know how many active tiles are at each tree level. Hence, a new method is developed to return a vector containing active tiles at each level. 

### Note

This new method is developed by using bitmask counting which is faster than old technique of tree-iterator.

### Unit and benchmark test

Unit test is developed by comparing this new method with the older tree-iterator method. Also, benchmark test can be done by using commented-out `timer` statements in unit test.

### Notice

This PR is developed by closely following PR #536 steps.

### DIFF
--- a/openvdb/tree/RootNode.h
+++ b/openvdb/tree/RootNode.h
@@ -1642,6 +1642,20 @@ RootNode<ChildT>::nodeCount(std::vector<Index32> &vec) const
     vec[ChildNodeType::LEVEL] = sum;
 }
 
+template<typename ChildT>
+inline void
+RootNode<ChildT>::activeTileCountByLevel(std::vector<Index32>& vec) const
+{
+    assert(vec.size() > LEVEL);
+    Index32 sum = 0;
+    for (MapCIter i = mTable.begin(), e = mTable.end(); i != e; ++i) {
+        if (isChild(i)) {
+            getChild(i).activeTileCountByLevel(vec);
+        } else if (isTileOn(i)) ++sum;
+    }
+    vec[LEVEL] = 0;// zero tiles at root level
+    vec[ChildNodeType::LEVEL] = sum;
+}
 ////////////////////////////////////////
 
 
--- a/openvdb/tree/RootNode.h
+++ b/openvdb/tree/RootNode.h
@@ -486,7 +486,7 @@ public:
     Index64 offLeafVoxelCount() const;
     Index64 onTileCount() const;
     void nodeCount(std::vector<Index32> &vec) const;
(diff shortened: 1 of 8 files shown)


openvdb/tree/Tree.h
```diff
@@ -113,6 +113,10 @@ class OPENVDB_API TreeBase
     /// is given as element NodeType::LEVEL in the return vector. Thus, the size
     /// of this vector corresponds to the height (or depth) of this tree.
     virtual std::vector<Index32> nodeCount() const = 0;
+    /// Return a vector with active tile counts. The number of active tiles at level NodeType
+    /// is given as element NodeType::LEVEL in the return vector. Thus, the size
+    /// of this vector corresponds to the height (or depth) of this tree.
```
jmlait, line 118: This should include a discussion of why this is different than the nodeCount() above.

"active tile counts" could be read ambiguously, sometimes people use that to refer to active tiles without children (ie, active constant regions) but not active tiles with children (ie, active nodes).

Suggested text.
Return a vector with active tile counts.   While nodeCounts returns the number of allocated nodes at each level, regions of the tree may have data but not have nodes because they have activated constant regions.  This returns both the count of both of these types of tiles.

Megidd:
> ... are you contributing this with a particular use case in mind?

Well, not specifically =) At times we are curious to know how many active tiles are at each level just to have more insight.

openvdb/tree/RootNode.h
```diff
@@ -1641,6 +1642,20 @@ RootNode<ChildT>::nodeCount(std::vector<Index32> &vec) const
...
+RootNode<ChildT>::activeTileCount(std::vector<Index32>& vec) const
+{
+    assert(vec.size() > LEVEL);
+    Index32 sum = 0;
+    for (MapCIter i = mTable.begin(), e = mTable.end(); i != e; ++i) {
+        if (isChild(i)) {
+            getChild(i).activeTileCount(vec);
+        } else if (isTileOn(i)) ++sum;
+    }
+    vec[LEVEL] = 0;// zero tiles at root level
```
kmuseth, line 1656: what's the point of having an entry for the Root::LEVEL if it is always zero?

kmuseth (approved):
approved but please see question - seems to me the vector can be truncated by the root level

openvdb/tree/RootNode.h
Megidd: @kmuseth Just to make sure the size of the vector corresponds to the height (or depth) of this tree. To be consistent, I guess. Not sure 🙂

PR #791, opened 2020-08-19
TSC Meeting Notes

jmlait:
Meeting notes 2020-08-18

### DIFF
--- a/tsc/meetings/2020-08-18.md
+++ b/tsc/meetings/2020-08-18.md
@@ -1,0 +1,66 @@
+Minutes from 61st OpenVDB TSC meeting, Aug 4th, 2020, (EDT)
+
+Attendees: *Nick* A., *Jeff* L., *Ken* M.
+
+Additional Attendees: Johannes Meng (Intel), JT Nelson (Blender),
+Bruce Chernia (Intel), Peter Cheng (DW), Andre Pradhana (DW)
+
+Regrets: *Peter* C, *Dan* B..
+
+Agenda:
+
+1) Confirm quorum
+2) Secretary
+3) Forum
+4) NanoVDB
+5) Repository Structure
+6) Makefile
+7) Next Meeting
+
+1) Quorum was confirmed.
+
+2) Secretary was Jeff Lait
+
+3) Forum
+
+Questions about computing averages of leaf nodes.  Unclear if the questioner wants it baked in the tree or just computed.  "On load" implies it is to be computed on construction.
+
+Another post is on large VDBs and Out of Core (OOC).  Wants to be able to stream it and unload it.  Ken has in the past written a minimal version that does delayed loading and unloading.  Loads in only the leaf data into a preset pool.
+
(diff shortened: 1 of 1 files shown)


tsc/meetings/2020-08-18.md
```diff
@@ -0,0 +1,66 @@
...
+
+2) Secretary was Jeff Lait
+
+3) Forum
+
+Questions about computing averages of leaf nodes.  Unclear if the questioner wants it baked in the tree or just computed.  "On load" implies it is to be computed on construction.
+
+Another post is on large VDBs and Out of Core (OOC).  Wants to be able to stream it and unload it.  Ken has in the past written a minimal version that does delayed loading and unloading.  Loads in only the leaf data into a preset pool.
+
+A mistake we've tried to do is make VDB try to do everything.  So we kept putting everything in one structure, so it is not ideal for any application.  Out of core is substandard, and random access is harmed by the out of core.  We're trying to simplify the tree, and reduce dependency.  Ken would like to contribute his streamable read only tree at some point.  Ken will reply to the thread.
```
danrbailey, line 30: Sounds good. I think we should try and measure the difference in performance when using out-of-core, I expect you're right that it is not negligible, but would be good to understand precisely what the cost is. I'll try and prioritise adding a CMake flag that allows us to disable it soon.

tsc/meetings/2020-08-18.md
```diff
@@ -0,0 +1,66 @@
...
+
+Ken can keep the current private repo private.  He can manually send synchronization pull requests by copying over from the private repo to the public.  It is expected development will occur on the private repo and push to the public for the next few months until it transfers fully.  Any PRs to the public repo can be hand-merged by Ken into the private repo to apply.
+
+5) Repository structure.
+
+Proposal to change the structure to allow for different versions to mix together for different submodules.  Requires headers not be off the root directory, so you can -I include submodules from different locations.  Each module can get a subdirectory of its same name to store the C and header files.  This will affect existing tooling.   Doing this before AX and Nano will be good.  We then also end up with a clean source folder.  Build system can ensure that the correct componets are built, so you never include anything from the local repo that is supposed to go from the installation on disk.
+
+This will break existing PRs so we should make sure they are relatively up to date.  The actual change is easy, the implications are harder.
```
danrbailey, line 46: Have we verified this breakage? I'm sure I've done this in the past and Git has been smart enough to know a file has been moved and to follow the new location when doing a merge. We should test if not.

tsc/meetings/2020-08-18.md
```diff
@@ -0,0 +1,66 @@
...
+
+All present vote in favour.
```
danrbailey, line 48: Not totally clear to me what was voted on here? Is this with or without an additional 'src' directory as discussed in the recent email thread? Or is this a vote to approve in theory but details still to be figured out?

tsc/meetings/2020-08-18.md
kmuseth: Just a note of caution here. It's surprisingly difficult to evaluation this perform overhead because it depends highly the CPUs ability to do branch-prediction. Good branch-prediction results in a relatively low overhead of having out-of-core (since in most cases data are in fact in-core).

tsc/meetings/2020-08-18.md
jmlait: No, I believe this was a theoretical concern.

tsc/meetings/2020-08-18.md
jmlait: This was without a src directory, just a double-nesting, IIRC.

PR #800, opened 2020-08-25
Restructure repo modules to resolve ambiguous include paths

Idclip:
This is the *bare* minimum of restructuring required to achieve the desired result of direct include paths to each vdb component not including others. There's other potential improvements that can be done as a result of this, but I've left this PR to consist of the minimum change set. 

```
openvdb
    ├── openvdb
    │   └── openvdb
    ├── openvdb_ax
    │   └── openvdb_ax
    └── openvdb_houdini
        └── openvdb_houdini
```

Another structure that's been mentioned is something more substantial as described below (from the exr repo @cary-ilm). Note that even with this approach, each lib would still have to be double nested to support building against existing installations - we can always build upon this initial change set in the future.

```
openvdb
└── src
    ├── bin
    │   ├── vdb_lod
    │   ├── vdb_print
    │   ├── vdb_render
    │   └── ...
    ├── contrib
    ├── examples
    ├── lib
    │   ├── openvdb
    │   │    └── openvdb
    │   ├── openvdb_ax
    │   │    └── openvdb_ax
    │   ├── openvdb_houdini
    │   │    └── openvdb_houdini
    │   └── ...
    └── test
        ├── vdb_test
        ├── vdb_ax_test
        └── ...
```


Idclip:
> Looks good to me, thanks! I'd fully support moving the COPYRIGHT, LICENSE, README files up into the parent directory as part of this change too.

Think I'll do that separately as their might be more discussion needed i.e. CMakeLists, bin directory, the viewer directory, e.t.c - this PR means that it's feasible for nano/ax merges!

Idclip:
@danrbailey @kmuseth @jmlait  It seems that merging this PR has caused conflicts on other PRs which have edited files. I could revert this until an appropriate time? Or just push on ahead and revisit older PRs as and when necessary

danrbailey:
@Idclip - no that's fine, let's keep this change. You have to merge master into each PR to resolve the conflicts, but in my test they got auto-resolved so it's a trivial step. I just updated #805 to resolve the conflicts to be sure and it didn't involve any manual conflict resolution.

PR #801, opened 2020-08-26
Bump to CMake 3.12

Idclip:
Turns out the the vfx-18 docker image ("testabi5") uses CMake 3.9.4. There's nothing in the VFX platform to define what CMake version we should be using and I'm aware of other tools in the aswf that have bumped to 3.12. I think we have 3 options

1) Install a custom version of CMake for that test
2) See if we can bump the vfx18 image to use 3.12
3) Don't bump CMake until VDB 8 - but it would be nice if we could do it now.

Opening this PR for discussion.


danrbailey:
Ironically, I believe it was me that proposed having the CMake version in vfx18 be 3.9.4 in the first place with the logic being that we should use the most recent patch version of the latest major.minor version of CMake that was available at the time the VFX Reference Platform year was ratified. I think that logic still makes sense and ideally the docker images should be immutable once published based on the ratified vfx reference platform so I'd rather we don't modify the cmake version in the the docker image itself.

It's annoying we can't get rid of ABI=5 yet as it's pretty old now. That being said, I don't see a reason not to push OpenVDB forwards with adopting 3.12 though and I agree it would be nice to align with other ASWF libraries in using a later version. I would say that solution (1) is the most pragmatic if it's not too complicated.

danrbailey:
@Idclip - It looks the Houdini schedule build also needs updating:

https://github.com/AcademySoftwareFoundation/openvdb/runs/1047544947?check_suite_focus=true

PR #805, opened 2020-08-28
Remove OpenVDB Static Specialization Macro

danrbailey:
This macro is a bit confusing. As far as I can tell, none of the places where it is currently being used are templated function specializations and I can't find any reference to needing to make function templates static inline for ODR on _any_ compiler.

Perhaps this macro was being used in a different way at one stage? Regardless, this pattern is used a lot _without_ this static keyword and it is only defined as static for non-GCC compilers and GCC compilers < 4.4 so I think it is pretty safe to remove it now. The CI passes across all platforms and compilers that we test against.


Idclip (approved):
release notes, but looks good

PR #806, opened 2020-08-28
Delete Deprecated Functionality and Use C++14 Attribute

danrbailey:


Idclip:
I'm wondering if it's still worth keeping the macro around and instead aliasing it to `[deprecated](msg)`? Happy either way

openvdb_houdini/SOP_NodeVDB.cc
```diff
@@ -353,107 +352,6 @@ SOP_NodeVDB::getNodeSpecificInfoText(OP_Context &context, OP_NodeInfoParms &parm
...
-SOP_NodeVDB::duplicateSourceStealable(const unsigned index, OP_Context& context)
-{
-    OPENVDB_NO_DEPRECATION_WARNING_BEGIN
-    auto error = this->duplicateSourceStealable(index, context, &gdp, myGdpHandle, true);
-    OPENVDB_NO_DEPRECATION_WARNING_END
-    return error;
-}
-
-
-////////////////////////////////////////
```
Idclip, line 454: So we still use these heavily. I would propose that these aren't removed until we have full DS support. If we really want to get rid of them I can re-introduce them on my end so it's not a huge problem, but it's worth pointing out that they're still in use somewhere

danrbailey:
> I'm wondering if it's still worth keeping the macro around and instead aliasing it to `[deprecated](msg)`? Happy either way

I think because this is part of the C++14 standard, we should just use the feature directly. It's simpler and I don't see us ever supporting less than a C++14 compiler again.

openvdb_houdini/SOP_NodeVDB.cc
danrbailey: Is there any specific features of DS that you're after that's holding you up in migrating? As much as possible, I think we should avoid maintaining code that's only there for one studio to use. As you say it's easy enough for that studio to maintain it in their own fork of the code so that should be preferred in these types of cases. Happy to discuss in our next TSC meeting though.

Idclip: Two parts to this. The first is that, when I last profiled, this method was still slightly faster than the compiled blocks, primarily for small-ish VDB geo streams, due to what I can only assume is the way some parameters were being evaluated. The second and more important issue is that we derive from this SOP type extensively throughout a good number of nodes which rely on this method. Along with the standard VDB Houdini nodes, these have subsequently been used in many digital assets (a good 50+ at least) and in general networks too. They rely on this unload behaviour for performance as there's no way we've been able to go through every asset and upgrade to compiled blocks. 

I don't believe this to be studio specific; anyone that previously relied on this behaviour will suddenly notice a change in performance. Thinking about it, this deprecation warning is actually a bit pointless. Really we could have done with something like a SOP warning when `duplicateSourceSteable` was being used in an unload context. Lets definitely discuss in TSC!

Idclip: I ran a very quick test with some of our networks by switching to the C++ verb support (our software already supports it, it's just turned off). As expected, a bunch of HDA's started evaluating every timestep when they didn't need to - however more concerning was that fact that some of the SOPs in question, primarily smaller operators, increased in cook time by a factor of x2/3 according to the performance monitor. Our internal operator timers come out with the same time execution so perhaps this is expected? i.e. maybe the performance monitor isn't actually working correctly with duplicate source stealable? @jmlait any thoughts?

jmlait: Marshalling through the UT_Options the way the NodeParmsOptions does would be slower than directly accessing evalString().  But I'm a bit surprised at 2/3x!

On the other hand, our perf monitor is known to have strange things, so maybe it is miscounting?  Maybe wall clock the time to cook the chain?

danrbailey:
Closing this PR. We ultimately decided to reverse this decision to go with the standard C++14 deprecated functionality so this is mostly no longer relevant. I'll create a new PR to handle the removal of long deprecated methods.

PR #812, opened 2020-09-01
Overlap compilation of vdb library and vdb tests

danrbailey:
With this change, the source code for the vdb library and for the unit tests get built in parallel. @Idclip - putting this in as a draft PR as I'm not sure it's ready to be merged, but just looking for feedback at this stage?


danrbailey:
Closing in favor of #1416.

PR #813, opened 2020-09-01
Explicit Template Instantiation

danrbailey:
I had a go at introducing explicit template instantiation to the library as an investigation into locking down the tree hierarchy. It has some tradeoffs which we can discuss. The unit tests build around 30% faster with this change.


danrbailey:
Closing this PR. This investigation was useful in understanding the different possible directions to head in with this feature, but ultimately is quite different to the final version so doesn't seem to present any value in keeping this alive.

PR #817, opened 2020-09-04
Make ZLib an Optional Dependency

danrbailey:


Idclip:
This looks great, thanks Dan. One thing worth noting; I believe blosc relies on or at least builds and links to zlib anyway. I'm not exactly clear on why it does this, I haven't looked into it further than the readme:

> Currently C-Blosc comes with support of BloscLZ, a compressor heavily based on FastLZ (http://fastlz.org/), LZ4 and LZ4HC (https://github.com/Cyan4973/lz4), Snappy (https://github.com/google/snappy), Zlib (http://www.zlib.net/) and Zstd (http://www.zstd.net).

and

> C-Blosc comes with full sources for LZ4, LZ4HC, Snappy, Zlib and Zstd 

and in blosc.h

```
/* Names for the different compressors shipped with Blosc */
#define BLOSC_BLOSCLZ_COMPNAME   "blosclz"
#define BLOSC_LZ4_COMPNAME       "lz4"
#define BLOSC_LZ4HC_COMPNAME     "lz4hc"
#define BLOSC_SNAPPY_COMPNAME    "snappy"
#define BLOSC_ZLIB_COMPNAME      "zlib"
```

In any case, this manifested in the static build on Windows through vcpkg with the whole `BLOSC_USE_EXTERNAL_SOURCES` addition to `FindBlosc`, which isn't an ideal solution but works. The question(s) though becomes, what exactly is blosc building into its lib from zlib, can this conflict with how we search for zlib currently and do we even need zlib if blosc is on?

danrbailey:
> This looks great, thanks Dan. One thing worth noting; I believe blosc relies on or at least builds and links to zlib anyway. I'm not exactly clear on why it does this, I haven't looked into it further than the readme:
> 
> > Currently C-Blosc comes with support of BloscLZ, a compressor heavily based on FastLZ (http://fastlz.org/), LZ4 and LZ4HC (https://github.com/Cyan4973/lz4), Snappy (https://github.com/google/snappy), Zlib (http://www.zlib.net/) and Zstd (http://www.zstd.net).
> 
> and
> 
> > C-Blosc comes with full sources for LZ4, LZ4HC, Snappy, Zlib and Zstd
> 
> and in blosc.h
> 
> ```
> /* Names for the different compressors shipped with Blosc */
> #define BLOSC_BLOSCLZ_COMPNAME   "blosclz"
> #define BLOSC_LZ4_COMPNAME       "lz4"
> #define BLOSC_LZ4HC_COMPNAME     "lz4hc"
> #define BLOSC_SNAPPY_COMPNAME    "snappy"
> #define BLOSC_ZLIB_COMPNAME      "zlib"
> ```
> 
> In any case, this manifested in the static build on Windows through vcpkg with the whole `BLOSC_USE_EXTERNAL_SOURCES` addition to `FindBlosc`, which isn't an ideal solution but works. The question(s) though becomes, what exactly is blosc building into its lib from zlib, can this conflict with how we search for zlib currently and do we even need zlib if blosc is on?

Good questions, I hadn't considered this. Blosc can be built without ZLib (using -DDEACTIVATE_ZLIB=ON) but otherwise it will always be built with it as even if it's not found externally, it will use it's own sources. We do need ZLib even if Blosc is on, because we include the zlib.h header which isn't distributed with Blosc (ie we use ZLib compression directly not through Blosc). 

For dynamic library compatibility, Blosc links against libz.so.1. If the zlib.h header we include has a different API than the one used in Blosc when that was being compiled and the SONAME doesn't change, there's potential for a conflict. The best way to resolve this is probably to just build Blosc with ZLib disabled (considering we don't even use it). I don't think there's any way to figure out from an installed version of Blosc which ZLib was used to compile it so I'm not even sure how we'd go about resolving this.

This PR isn't really intended to address the binary compatibility issue although it's worth asking this question. I'll change this PR so that you can only disable ZLib if you also disable Blosc. Even though it's technically possible, it seems reasonable to assume that if you have Blosc, you also have ZLib.

Idclip:
> We do need ZLib even if Blosc is on, because we include the zlib.h header which isn't distributed with Blosc (ie we use ZLib compression directly not through Blosc).

Yeah so I guess this is what I was getting at with the last point. Could we instead use `blosc_set_compressor` and use zlib through blosc? We'd obviously then need compile guards which switch to the zlib API or blosc API  i.e. you can choose either blosc or zlib, but you no longer need both. Not required for this PR though.

> I don't think there's any way to figure out from an installed version of Blosc which ZLib was used to compile it so I'm not even sure how we'd go about resolving this.

Yes, it's hard enough trying to detect if it was compiled with zlib let alone the version

danrbailey:
> > We do need ZLib even if Blosc is on, because we include the zlib.h header which isn't distributed with Blosc (ie we use ZLib compression directly not through Blosc).
> 
> Yeah so I guess this is what I was getting at with the last point. Could we instead use `blosc_set_compressor` and use zlib through blosc? We'd obviously then need compile guards which switch to the zlib API or blosc API i.e. you can choose either blosc or zlib, but you no longer need both. Not required for this PR though.

Maybe it's possible, I'm not sure. I agree that it would be nice.

> > I don't think there's any way to figure out from an installed version of Blosc which ZLib was used to compile it so I'm not even sure how we'd go about resolving this.
> 
> Yes, it's hard enough trying to detect if it was compiled with zlib let alone the version

Agreed.

I updated this PR to disallow Blosc and !ZLib.

Idclip (approved):
> Maybe it's possible, I'm not sure. I agree that it would be nice.

Lets create a task so we don't forget about this. lgtm!

PR #818, opened 2020-09-04
Introduce OpenVDB AX core library

Idclip:
This PR introduces the core OpenVDB AX Library, binaries, documentation and required CMake changes. The Houdini AX SOP can come after this has been reviewed. A few "high level" thoughts:

- Language docs. Still need work, but the core specification w.r.t types and operators is available https://idclip.github.io/openvdbax-doxygen/ax.html I think this may require its own meeting to fully review.
- `vdb_ax` thoughts on the command line binaries interface and commands would be good.
- Directory structure. With the new nested infrastructure, it's entirely possible to instead of `openvdb_ax` have `openvdb/ax` as an include path.
- Boost::random. As far as the dependency goes, I'm planning to remove it and switch to std::mt19937/mt19937_64. I can update this PR once this is available - however I'd like to find some time in a TSC meeting to discuss the actual implementation of methods such as `rand()`.
- AX CI - Currently AX is tested for vfx platforms 19 and 20, and llvm versions 6.0 through to 10.0. It builds AX separately and depends on an existing installation of VDB. I like this as it splits out the CI somewhat, but means that building VDB and AX together isn't tested. 
- Lib version. It currently has it's own `version.h` and `.so` version. Does anyone feel strongly on this being kept separate? 

There is one final change that I'm hopeful will make it in to AX before 7.2 is released in regards to how exceptions are handled throughout the lib - however this doesn't stall reviewing/merging this work and actually may function as a good first test PR once this is all in!


danrbailey:
Rather than doing one mega review, I think I'll just do a few rounds of reviewing as and when I put aside some time. I'll start with CI because that's easy to review. 

It would be nice to align more closely with OpenVDB. We intentionally didn't go with the matrix-style approach to prevent the explosion of CI jobs. Instead we test along two primary axes - (1) latest ABI and all configuration options and platforms and (2) all other ABIs and DCCs. I'd like us to take a similar approach here so we don't explode the CI jobs from ~21 => ~55. As most of these new jobs are in excess of an hour (perhaps that's excluding caching?), we're really going to slow down our turnaround time for CI completing given the limited resources provided for open-source projects. Also, I think the limits for Mac OS jobs are lower than Linux jobs, so we should factor that in. If possible, I'd also like to fold AX builds into the Build.yml and we shouldn't definitely test building AX and VDB together. 

We have to be judicious with our use of the GitHub Actions cache. It's a fairly dumb mechanism because it just evicts the oldest cache. If the Houdini install is evicted from the cache, all the Houdini jobs will fail, so it's reliant on keeping all required Houdini install caches below 5GB (right now Houdini is ~220MB as I've stripped out everything that's not required for our builds). I would prefer a cron mechanism similar to the Houdini one that updates the cache once a week, then at least multiple concurrent small platform changes wouldn't trigger new caches each time that could quickly blow through the 5GB of available space. It's for this reason, I've been hesitant to introduce ccache to the CI. As an aside, GitHub Actions caching is fairly new and it's improving rapidly so some of these limitations will likely disappear over time. Also we should check if a particular version of LLVM is in the docker containers, I can't remember?

Idclip:
Thanks for the feedback Dan. I appreciate what you're saying and I think I agree with the main two takeaways (correct me if I'm wrong):
 - 1) Do less, more specific builds because of resources
 - 2) Test AX building with VDB

But I have some other comments.

> We intentionally didn't go with the matrix-style approach to prevent the explosion of CI jobs

Matrix jobs do not equal an "explosion" of CI jobs. Matrix configurations provide more homogenised settings which are less error prone and easier to edit/read. They can of course be used to create hundreds of different build settings but it's not a requirement. I would in fact advocate that we use matrix builds in VDB.

>  I'd like us to take a similar approach here so we don't explode the CI jobs from ~21 => ~55

This is not the case. As you mentioned, this feature branch had none of the LLVM caches. Builds will not be this long. Your new Half PR took 1h to build:
https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/240437646
A standard build of all the CI jobs for AX took 1h 18min
https://github.com/dneg/openvdb_ax/actions/runs/233486759
The builds are going to get slightly longer - there's not much we can do about that. Additionally, as **soon** as we surpass the github actions max concurrent limit, we sort of gain free resources. i.e. if the max job limit is 20 and we have 21 jobs, then we may as well have 40. This is a generalised view as it of course depends on the min/max job times, but in practise I haven't seen much difference.

> If possible, I'd also like to fold AX builds into the Build.yml 

I'll have a go at restructuring and see what it looks like.
 
>  Finally, we should check if a particular version of LLVM is in the docker containers, I can't remember?

They don't - or at least I had problems using what was installed/pre-built versions of LLVM. Essentially I found that I had to build it from source and cache it.

Idclip:
My current WIP for this can be found here https://github.com/Idclip/openvdb/tree/test_ci

Idclip:
I've reduced the number of AX jobs from 34 to 23. We can hopefully further reduce them by deciding on what minimum version of LLVM we'd like to support (it was 5, I've bumped it to 6). Each LLVM install takes around ~100mb thats x5 (for each version) x2 (for gcc/clang) which takes the cache size to ~1gb - way under the max. I've kept the AX jobs separate as, at the moment, we have file "global" workflow settings like `on` and `paths-ignore` which means that I can keep the core builds off if something only changes in `openvdb_ax`. Additionally I've transitioned the VDB core builds over to a matrix strategy to demonstrate how it can be used effectively.

danrbailey:
Thanks for looking at the CI. I don't mind the matrix syntax itself and it is (slightly) less code, though I'm not sure it buys us a lot. I don't see a clean way to include all the platform logic without a bunch of confusing conditional branching which makes it harder to read and harder to test the flow of logic when using just one configuration. It seems most useful when you wish to test across a mostly or completely dense matrix.

When I said 21 => 55 I was referring to the number of jobs not time taken to run them. I'd prefer us to use fewer jobs to increase the turnaround time. Your example here in reference to the max job limit is assuming there's just one workflow running, when you have multiple workflows running (which I find is common), the CI limit is applied across _all_ workflows so there's still a desire to keep the number of jobs to a minimum. For example, I think we should cut down the number of Mac builds - perhaps to even just a single build, then use Linux to build all the variations (LLVM versions, debug, etc).

I'm a little nervous about mixing the implicit caching with the explicit caching models. I can see a pattern where we quickly end up creating way more caches than we expect and the Houdini caches become the least recently used and get evicted. There are a bunch of additional cache rules such as creating unique caches for each feature branch that might catch us out. We need to be cautious about how we create these caches and to make sure that we re-use keys wherever possible to use the fewest caches (for example using the same cache key for g++/clang++ builds). I don't mind trying it if we're mindful of the need to reuse caches wherever possible.

Idclip:
> Your example here in reference to the max job limit is assuming there's just one workflow running

I wasn't working under that assumption - I realise the max tasks are limited across all workflows and agree that less more specific jobs would be ideal. I'll see about cutting the mac builds down but I'd really like to test as many versions of LLVM as possible across the different platforms. I realise this might not be feasible.

> I'm a little nervous about mixing the implicit caching with the explicit caching models. I can see a pattern where we quickly end up creating way more caches than we expect and the Houdini caches become the least recently used and get evicted.

That's fair - I'm happy to try and switch to an explicit caching workflow, but it's quite hard to get that working without these changes going in. If you're open to it, I propose we go with this method for the first merge and I'm happy to immediately switch to a more explicit method. Would the explicit way still mean CI forks would work as expected?

danrbailey:
> > I'm a little nervous about mixing the implicit caching with the explicit caching models. I can see a pattern where we quickly end up creating way more caches than we expect and the Houdini caches become the least recently used and get evicted.
> 
> That's fair - I'm happy to try and switch to an explicit caching workflow, but it's quite hard to get that working without these changes going in. If you're open to it, I propose we go with this method for the first merge and I'm happy to immediately switch to a more explicit method. Would the explicit way still mean CI forks would work as expected?

To be honest, we're trying to tackle this in the wrong way. The reason Houdini isn't in an ASWF Docker container is because it's a licensed piece of software that requires a signed EULA to use it. That's not the case for LLVM, so really we should be packaging this inside a purpose-built Docker container. That would free up potential cache space to accelerate our builds with ccache.

danrbailey:
@Idclip - did you have a chance to look at the way that OSL handles LLVM dependencies? Their build script for LLVM is here:

https://github.com/imageworks/OpenShadingLanguage/blob/master/src/build-scripts/build_llvm.bash

It might be worth trying the pre-compiled route again to see if we can easily mitigate any issues with using the actions cache?

Idclip:
> @Idclip - did you have a chance to look at the way that OSL handles LLVM dependencies? 

Yes I did try this, it didn't work - I suspect ABI issues with libstdc++ relating to `GLIBCXX_USE_CXX11_ABI`. They build with a CMake command here:

https://github.com/imageworks/OpenShadingLanguage/blob/master/src/build-scripts/ci-build-and-test.bash#L29

But this option is off here by default:

https://github.com/imageworks/OpenShadingLanguage/blob/9fdaf2cb1fbefe8983388d85dca3f5112d9913b4/src/cmake/compiler.cmake#L241

danrbailey:
> > @Idclip - did you have a chance to look at the way that OSL handles LLVM dependencies?
> 
> Yes I did try this, it didn't work - I suspect ABI issues with libstdc++ relating to `GLIBCXX_USE_CXX11_ABI`. They build with a CMake command here:
> 
> https://github.com/imageworks/OpenShadingLanguage/blob/master/src/build-scripts/ci-build-and-test.bash#L29
> 
> But this option is off here by default:
> 
> https://github.com/imageworks/OpenShadingLanguage/blob/9fdaf2cb1fbefe8983388d85dca3f5112d9913b4/src/cmake/compiler.cmake#L241
> 
> edit: perhaps not as g++ should be already configured in the vfx docker container with `GLIBCXX_USE_CXX11_ABI=0`? I can try again

Yeah, think it's worth trying that again and also checking the version shipped with the docker containers.

Idclip:
> Yeah, think it's worth trying that again and also checking the version shipped with the docker containers.

Tried this again, it's pretty painful! Currently running into these undefined symbols:
https://github.com/Idclip/openvdb/actions/runs/270373249

```
../../libopenvdb_ax.so.0.3.0: undefined reference to `llvm::TargetRegistry::lookupTarget(std::string const&, std::string&)'
../../libopenvdb_ax.so.0.3.0: undefined reference to `llvm::Triple::normalize(llvm::StringRef)'
../../libopenvdb_ax.so.0.3.0: undefined reference to `llvm::SubtargetFeatures::getString() const'
../../libopenvdb_ax.so.0.3.0: undefined reference to `llvm::sys::getDefaultTargetTriple()'
```

I'm fairly confident that this is ABI incompatibility between compilers. I've tried linking explicitly against every LLVM lib provided as part of the download (they're all static) and these still show up. I can replicate this locally and building with `-D_GLIBCXX_USE_CXX11_ABI=1` fixes these issues - but I'm still yet to be successful getting this to work in the current docker container.

Idclip:
I can confirm that using the version of LLVM that's provided by the container (7.0.0) and building with **g++** works:
https://github.com/Idclip/openvdb/runs/1160966918?check_suite_focus=true

Building with clang++ does not:
https://github.com/Idclip/openvdb/runs/1160870690?check_suite_focus=true

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
```diff
@@ -0,0 +1,1982 @@
...
+        llvm::Value* result = binaryOperator(one, x, ast::tokens::MINUS, B);
+        result = binaryOperator(result, a, ast::tokens::MULTIPLY, B);
+        llvm::Value* right = binaryOperator(x, b, ast::tokens::MULTIPLY, B);
+        result = binaryOperator(result, right, ast::tokens::PLUS, B);
+        return result;
+    };
+
+    static auto lerp = [](auto a, auto b, auto x) -> auto {
+        using ValueT = decltype(a);
+        return (ValueT(1.0) - x) * a + x * b;
```
jmlait, line 506: ```a + (b -a) * x```
is a better formulation, or you have
```
lerp(x, x, t) != x
```
which is surprising.

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
```diff
@@ -0,0 +1,1982 @@
...
+    };
+
+    static auto min = [](auto a, auto b) -> auto {
+        return std::min(a, b);
+    };
+
+    return FunctionBuilder("min")
+        .addSignature<double(double,double)>(generate, (double(*)(double,double))(min))
+        .addSignature<float(float,float)>(generate, (float(*)(float,float))(min))
+        .addSignature<int32_t(int32_t,int32_t)>(generate, (int32_t(*)(int32_t,int32_t))(min))
```
jmlait, line 544: Does this mean no support for long?

What about max(vec4, vec4, vec4) to do component wise?

or float max(vec4)?

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
```diff
@@ -0,0 +1,1982 @@
...
+    {
+        //         (outMax - outMin)(x - inMin)
+        // f(x) = ----------------------------  + outMin
+        //                inMax - inMin
+        // if inMax == inMin, f(x) = (outMax + outMin) / 2.0
+
+        // NOTE: this also performs a clamp on the ordered input range
+        // @TODO revisit. If this is the best thing to do, should add conditional
+        // branching so that the clamping math is never executed when the value
+        // is inside
```
jmlait, line 633: Input range should not be forced to be ordered.
Clamping the input range makes sense; VEX uses efit() for the case you don't want clamping.

Here's the VEX definition:
``` 
    float d = omax - omin;
    float tmp = select(d == 0, (nmin + nmax) * 0.5,
            nmin + (nmax - nmin) * (val - omin) / d);
    return select(nmin < nmax,
            clamp(tmp, nmin, nmax),
            clamp(tmp, nmax, nmin));
```

n is New, o is Old, so will look a bit backwards.

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
```diff
@@ -0,0 +1,1982 @@
...
+            "rand() with a given seed. If rand is not called with a seed, the generator "
+            "advances continuously across different elements which can produce non-"
+            "deterministic results. It is important that rand is always called with a "
+            "seed at least once for deterministic results.")
+        .get();
+}
+
+inline FunctionGroup::Ptr axsignbit(const FunctionOptions& op)
+{
+    return FunctionBuilder("signbit")
```
jmlait, line 793: I'd go with
```sign```
which would include integers and return 1/0/-1.

File: openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc [hunk 5]
```diff
@@ -0,0 +1,1982 @@
...
+        .setDocumentation("Multiplies two matrices together and returns the result")
+        .get();
+}
+
+inline FunctionGroup::Ptr axpolardecompose(const FunctionOptions& op)
+{
+    static auto polardecompose = [](auto in, auto orth, auto symm) -> bool {
+        bool success = false;
+        try {
+            success = openvdb::math::polarDecomposition(*in, *orth, *symm);
```
jmlait, line 1140: Does this do negative determinant checking?

openvdb_ax/openvdb_ax/version.h
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+/// @file openvdb_ax/version.h
+///
+/// @brief Library and file format version numbers
+///
+/// @details  Based off of the versioning system within openvdb. There is
+///   currently no AX namespace versioning which instead leverages openvdb's
+///   version namespace.
+///
```
jmlait, line 11: I feel that since this controls the API/ABI, there isn't a reason for a secondary version structure here.  Either OpenVDB AX should use its own namespace based on its own version; or use the core name space and version.  Using the core namespace means for all practical purposes it is locked to the core version numbers!

I'd also be against using separate namespace as that encourages the idea these can desynchronize.  And suggests we need to say for Version X of AX, versions Y-Z of core are compatible. Which sounds a crazy explosion of issues.   If possible, I'd want to instead say that each version of AX is only compatible directly with the corresponding Core version, and any other cross over would require ABI guarantees and void *.

openvdb_ax/openvdb_ax/ax.cc
```diff
@@ -0,0 +1,94 @@
...
+
+namespace openvdb {
+OPENVDB_USE_VERSION_NAMESPACE
+namespace OPENVDB_VERSION_NAME {
+namespace ax {
+
+/// @note Implementation for initialize, isInitialized and unitialized
+///       reamins in compiler/Compiler.cc
+
+void run(const char* ax, openvdb::GridBase& grid)
```
danrbailey, line 18: I find the wording of the API a little confusing in places - the mix of executable, execute, executor, run, etc.

I also find repeating words harder to mentally parse:

```
VolumeExecutable executable;
executable.execute();
```

Wondering it might read better to "execute" a "binary" rather than executing an executable (ie VolumeBinary instead of VolumeExecutable). 

I find most programming languages go with eval or evaluate instead of run for evaluating an expression too.

openvdb_ax/openvdb_ax/ax.cc
```diff
@@ -0,0 +1,94 @@
...
+        // Execute on the provided numerical grid
+        // @note  Throws on invalid grid inputs such as mismatching types
+        exe->execute(grid);
+    }
+}
+
+void run(const char* ax, openvdb::GridPtrVec& grids)
+{
+    if (grids.empty()) return;
+    // Check the type of all grids. If they are all points, run for point data.
```
danrbailey, line 52: I know this is not an easy thing to solve, but this logic seems weird to me. You have an array of point grids and accidentally add a density grid on the end and it totally changes the way the other grids are processed. Also, what happens if you want to add a volumesample() operation? Wouldn't you have to adjust this logic?

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
```diff
@@ -0,0 +1,148 @@
...
+    ///   effect of disabling multi-threading.
+    /// @param grain The grain size
+    void setGrainSize(const size_t grain);
+    /// @return  The current grain size
+    size_t getGrainSize() const;
+
+    ////////////////////////////////////////////////////////
+
+    // @brief deprecated methods
+    OPENVDB_DEPRECATED void
```
danrbailey, line 86: These deprecated methods should be removed before merging, no point adding methods that we know we plan to remove.

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
```diff
@@ -0,0 +1,148 @@
...
+
+private:
+    // The Context and ExecutionEngine must exist _only_ for object lifetime
+    // management. The ExecutionEngine must be destroyed before the Context
+    const std::shared_ptr<const llvm::LLVMContext> mContext;
+    const std::shared_ptr<const llvm::ExecutionEngine> mExecutionEngine;
+    const AttributeRegistry::ConstPtr mAttributeRegistry;
+    const CustomData::ConstPtr mCustomData;
+    const std::unordered_map<std::string, uint64_t> mFunctionAddresses;
+    std::unique_ptr<Settings> mSettings;
```
danrbailey, line 140: What's the reason for using this pimpl pattern here for settings - is it an attempt to try not to break ABI? More generally, do we need to worry about ABI for these types of objects. If we are going to attempt to do this, why not put all of this data into the implementation as none of it is used in the public interface?

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
```diff
@@ -0,0 +1,148 @@
...
+    void setCreateMissing(const bool flag);
+    /// @return  Whether this executable will generate new point attributes.
+    bool getCreateMissing() const;
+
+    /// @brief  Set the threading grain size. Default is 1. A value of 0 has the
+    ///   effect of disabling multi-threading.
+    /// @param grain The grain size
+    void setGrainSize(const size_t grain);
+    /// @return  The current grain size
+    size_t getGrainSize() const;
```
danrbailey, line 81: I don't like the grain size being a setting of the executable. I think it should be an argument to the execute() method with a default value. I'd also like to see a threaded=true argument to align with other similar methods in VDB such as LeafManager::foreach().

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
```diff
@@ -0,0 +1,148 @@
...
+namespace ax {
+
+class Compiler;
+
+/// @brief Object that encapsulates compiled AX code which can be executed on a target point grid
+class PointExecutable
+{
+public:
+    using Ptr = std::shared_ptr<PointExecutable>;
+    ~PointExecutable();
```
danrbailey, line 42: Empty deconstructors should just be `= default` or made implicit.

openvdb_ax/openvdb_ax/compiler/CompilerOptions.h
```diff
@@ -0,0 +1,74 @@
...
+    ///         to determine whether they are allowed to be called during code generation
+    ///         to evaluate call sites with purely constant arguments and replace the call
+    ///         with the result.
+    /// @note   This does not impact IR functions which we leave to LLVM's CF during
+    ///         IR optimization.
+    /// @note   We used to bind IR methods to corresponding C bindings, however it can be
+    ///         very easy to implement incorrectly, leading to discrepancies in the CF
+    ///         results. Fundamentally, LLVM's support for CF IR is far superior and our
+    ///         framework only supports some types of folding (see codegen/ConstantFolding.h)
+    bool mConstantFoldCBindings = true;
```
danrbailey, line 35: I tend to think of m prefixed variables as being private or protected members. For these type of structs, I would suggest just including the name itself so you don't need the user to include the m prefix or otherwise provide getters and setters, but I don't think that's necessary here.

openvdb_ax/openvdb_ax/compiler/Compiler.h
```diff
@@ -0,0 +1,116 @@
...
+    static UniquePtr create(const CompilerOptions& options = CompilerOptions());
+
+    /// @brief Compile/build a given AST into an executable object of the given type.
+    /// @param syntaxTree An abstract syntax tree to compile
+    /// @param data External/custom data which is to be referenced by the executable object. It
+    ///        allows one to reference data held elsewhere, such as inside of a DCC, inside of the
+    ///        executable
+    /// @param compilerErrors A vector of strings where errors are inserted into
+    template <typename ExecutableT>
+    typename ExecutableT::Ptr
```
danrbailey, line 75: Why return a shared pointer here and not a unique pointer?

openvdb_ax/openvdb_ax/compiler/Compiler.h
```diff
@@ -0,0 +1,116 @@
...
+    void setFunctionRegistry(std::unique_ptr<codegen::FunctionRegistry>&& functionRegistry);
+
+private:
+    std::shared_ptr<llvm::LLVMContext> mContext;
+    const CompilerOptions mCompilerOptions;
+    std::shared_ptr<codegen::FunctionRegistry> mFunctionRegistry;
+};
+
+
+}
```
danrbailey, line 111: Minor style thing - VDB almost always includes the namespace comments to these. For consistency, it would be nice to add it:

```
} // namespace tree
} // namespace OPENVDB_VERSION_NAME
} // namespace openvdb
```

openvdb_ax/openvdb_ax/compiler/Compiler.h
```diff
@@ -0,0 +1,116 @@
...
+///   string). The class object exists to cache various structures,
+///   primarily LLVM constructs, which benefit from existing across
+///   additional compilation runs.
+///
+
+#ifndef OPENVDB_AX_COMPILER_HAS_BEEN_INCLUDED
+#define OPENVDB_AX_COMPILER_HAS_BEEN_INCLUDED
+
+#include "../ast/AST.h"
+#include "../compiler/CompilerOptions.h"
```
danrbailey, line 19: Probably left over from an old refactor - but this relative include isn't needed - should simply be "CompilerOptions.h"

openvdb_ax/openvdb_ax/compiler/LeafLocalData.h
```diff
@@ -0,0 +1,229 @@
...
+///         adding new string attributes or updating positions.
+///
+/// @note  Due to the way string handles work, string write attribute handles cannot
+///        be constructed in parallel, nor can read handles retrieve values in parallel
+///        if there is a chance the shared metadata is being written to (with set()).
+///        As the compiler allows for any arbitrary string setting/getting, leaf local
+///        maps are used for temporary storage per point. The maps use the string array
+///        pointers as a key for later synchronization.
+///
+struct LeafLocalData
```
danrbailey, line 40: If this is just point-related, might not be a bad idea to include Point in the name of the class and the header?

openvdb_ax/openvdb_ax/compiler/LeafLocalData.h
```diff
@@ -0,0 +1,229 @@
...
+    ///
+    inline GroupHandleT* getOrInsert(const std::string& name)
+    {
+        GroupHandleT* ptr = get(name);
+        if (ptr) return ptr;
+
+        static const size_t maxGroupsInArray =
+#if (OPENVDB_LIBRARY_MAJOR_VERSION_NUMBER > 7 ||  \
+    (OPENVDB_LIBRARY_MAJOR_VERSION_NUMBER >= 7 && \
+     OPENVDB_LIBRARY_MINOR_VERSION_NUMBER >= 1))
```
danrbailey, line 83: A good reason to synchronize the versions IMO. How far back should we support this branching logic and when do we remove it?

openvdb_ax/openvdb_ax/codegen/ComputeGenerator.h
```diff
@@ -0,0 +1,217 @@
...
+    bool traverse(const ast::DeclareLocal* decl)
+    {
+        if (!decl) return true;
+        if (!this->visit(decl)) return false;
+        return true;
+    }
+
+    virtual bool visit(const ast::CommaOperator*);
+    virtual bool visit(const ast::AssignExpression*);
+    virtual bool visit(const ast::Crement*);
```
danrbailey, line 141: I think this was meant to be a temporary name! ;)

danrbailey:
Another piece of feedback I forgot to mention. This is related to Jeff's comment about aiming for parity with C++ in terms of performance. There's a performance cost in using "Unknown" codec for accessing point attributes, so we should consider a mechanism for allowing you to use an explicit codec when referencing point attributes.

danrbailey:
Another couple of comments related to building without using CMake:

* I had issues attempting to use AX due to the way axparser.h is included. The challenge is the custom logic in the CMake means it's not possible to simply copy the source tree and compile the code as the generated grammar files are not found. Would it be possible to adjust the way this header is included to something like this:

```
#ifdef OPENVDB_AX_GRAMMAR
#include "axparser.h"
#else
#include "grammar/generated/axparser.h"
#endif
```

or otherwise to copy to a path under the CMake build folder that has "grammar/generated" in it and keep the longer include. The ideal behavior I think is that it just works out of the box with the pre-generated grammar files, rather than having to mess with include paths or modify the code.

* I ran into problems with matrices. I was initially getting compiler errors such as Abs(Mat3) is not found, etc. Would it be possible to change it so that you include a compiler define to _enable_ matrices rather than one to _disable_ them (ie OPENVDB_AX_MATRIX). That way it aligns with the default behavior in OpenVDB where matrix grids are not supported. There's also a parameter in the CMake - MATRIX_SUPPORT which doesn't appear to be documented.

danrbailey:
Minor request - in the unit tests, could the relative paths change to be absolute paths (`#include "../ast/AST.h"` => `#include <openvdb_ax/ast/AST.h>`). This is more in line with how the OpenVDB unit tests work and makes it easier to build if you wish to split the unit tests out from the source tree.

openvdb_ax/openvdb_ax/ast/AST.h
```diff
@@ -0,0 +1,2414 @@
...
+        if (name) *name = token.substr(at + 1, token.size());
+        return true;
+    }
+private:
+    const tokens::CoreType mType;
+    const bool mTypeInferred;
+};
+
+/// @brief  ExternalVariable represent any access to external (custom) data,
+///         typically associated with the '$' symbol syntax. Note that the AST
```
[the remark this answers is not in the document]
Idclip: As discussed, the sop supports ch/chs/chramp! However I checked and currently does not support backticks. This should be relatively easy to add by looking for \`exp\` expressions, extracting, evaluating and replacing

openvdb_ax/openvdb_ax/ast/Scanners.h
```diff
@@ -0,0 +1,203 @@
...
+///         write only attributes in the 'writeOnly' container @code @a=1; @endcode
+///         and readWrite attributes in the 'readWrite' container @code @a+=1; @endcode
+/// @note   Note that the code generator is able to do this far more efficiently, however
+///         this provides simple front-end support for detecting these types of operations
+///
+/// @param node       The AST to analyze
+/// @param readOnly   The unique list of attributes which are only read from
+/// @param writeOnly  The unique list of attributes which are only written too
+/// @param readWrite  The unique list of attributes which both read from and written too
+///
```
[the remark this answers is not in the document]
Idclip: Yeah this will currently mark a as writable. No plan to change this behaviour, at least not from the default - and we're doing code evaluation then this may be better suited to the IR level with llvms value visitors than at the AST

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: The problem with your version is that `lerp(a,b,1.0f) != b`. However our version is also only monotonic when `a < 0 && b < 0` or `a > 0 and b > 0`... I took a further look at this and turns out the standard introduced lerp in c++20... here's how clang implements it:

``` 
_Fp __lerp(_Fp __a, _Fp __b, _Fp __t) noexcept {
    if ((__a <= 0 && __b >= 0) || (__a >= 0 && __b <= 0))
        return __t * __b + (1 - __t) * __a;

    if (__t == 1) return __b;
    const _Fp __x = __a + __t * (__b - __a);
    if (__t > 1 == __b > __a)
    	return __b < __x ? __x : __b;
    else
    	return __x < __b ? __x : __b;
}
```

https://github.com/llvm/llvm-project/blob/d1dbda10cefeaa124e28eb289cdc92c049c3d973/libcxx/include/cmath#L620

I mean... feels a bit overkill, but this does resolve all cases. And could always then add a `fastlerp` or somesuch.

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: I believe our version is the same as this, but it clamps `val` to `omin` and `omax` first rather than the result to `nmin` and `nmax`. Something like

```
val =  select(omin< omax,  clamp(val , omin, omax),  clamp(val , omax, omin));
float d = omax - omin;
float tmp = select(d == 0, (nmin + nmax) * 0.5,
            nmin + (nmax - nmin) * (val - omin) / d);
```

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
Idclip: Will do a final pass of all deprecated methods once there is no more comments

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
Idclip: This can't be defaulted here because there definition of `Settings` lives in the .cc. If we change this class then perhaps this can be done

openvdb_ax/openvdb_ax/compiler/LeafLocalData.h
Idclip: With the port into VDB, like with the deprecated code branching, I'll be removing all of this

openvdb_ax/openvdb_ax/ast/Tokens.h
```diff
@@ -0,0 +1,364 @@
...
+    return UNKNOWN_OPERATOR;
+}
+
+inline OperatorToken operatorTokenFromName(const std::string& name)
+{
+    if (name == "+")    return PLUS;
+    if (name == "-")    return MINUS;
+    if (name == "*")    return MULTIPLY;
+    if (name == "/")    return DIVIDE;
+    if (name == "%")    return MODULO;
```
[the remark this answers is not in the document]
Idclip: I've gone down a bit of a rabbit hole with this. It seems that most languages I've looked into (bar c) implement a "floored modulo" operation, where the divisor in `a % b` is assumed to be positive - essentially `((a % b) + b) % b;` (assuming % has been implemented as a "truncated modulo"). As I understand it, this still isn't a true euclidean mod, which should always produce positive results. i.e.
```
x = a%b;
if (x < 0) {
   if (b > 0) x = x + b;
   else         x = x - b;
}
```
This is essentially the most performant implementation (using truncated mod) of a pure euclidean mode where as the floored impl can be improved to:

```
   (a < 0 ? ((a % b) + b) % b : a % b);
```

This version preserves the sign of the divisor, however it's debatable how important or how "correct" propagating this information is. In terms of its implementation and regularity, f-mod seems to be the best balance between the three, although e-mod covers all cases in the most uniform way. As well as speed and ... "precision?" we also care about compatibility. The fact that AX/VEX is so similar to C concerns me a bit, though within applications like Houdini users are perhaps expected to be more familiar with python. This is something I've never considered before... Part of me is now wondering if it should even be supported in favour of explicit functions :cold_sweat: 

http://norayr.am/papers/divmodnote.pdf
https://core.ac.uk/download/pdf/187613369.pdf

openvdb_ax/openvdb_ax/ast/Tokens.h
```diff
@@ -0,0 +1,364 @@
...
+
+template<typename T> CoreType tokenFromType()                    { return UNKNOWN; }
+template<> inline CoreType tokenFromType<bool>()                 { return BOOL; }
+template<> inline CoreType tokenFromType<char>()                 { return CHAR; }
+template<> inline CoreType tokenFromType<int16_t>()              { return SHORT; }
+template<> inline CoreType tokenFromType<int32_t>()              { return INT; }
+template<> inline CoreType tokenFromType<int64_t>()              { return LONG; }
+template<> inline CoreType tokenFromType<float>()                { return FLOAT; }
+template<> inline CoreType tokenFromType<double>()               { return DOUBLE; }
+template<> inline CoreType tokenFromType<math::Vec2<int32_t>>()  { return VEC2I; }
```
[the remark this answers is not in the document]
Idclip: Off the back of this comment we decided to remove `short`/`int16` local variable support and allow `int` to be aliased to 32 or 64bit. There remains the question of what to do with integer literals as it would be nice to remove the `l` suffix for longer types. There's two options in my mind:
1) all literals are 64 bit regardless of your `int` type selection. This is slightly strange as literal arithmetic may be done at a higher precision. Additionally, because other container types like vec3i are 32bit, you get implicit warnings with things like `vec3i a = {1,2,3};` or `int a = 1;` unless we explicitly ignore assignments from int literals. 
2) literals are automatically stored as 64 bit if the value is larger than `limit<int32_t>::max()`. This is nicer, but has the strange side effect of producing different results if, say, an evaluation of two literals produces a 64bit value. i.e. `1 + 2147483648` vs `1 + 2147483649`. The latter will return a int64 value, the prior will wrap and return a 32bit value.
Otherwise, if literals are tied to the type alias of `int` (e.g. 32bit) then we'd have to keep a suffix to distinguish 64bit values

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: Added missing in64_t signatures. Have put vector min/max component methods on the todo list

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: Added sign which implements signum. Left signbit purely for the +-0.0 case

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: Added more documentation

openvdb_ax/openvdb_ax/version.h
Idclip: Removed this file completely. AX is now always coupled to/part of the vdb version namespace

openvdb_ax/openvdb_ax/ax.cc
```diff
@@ -0,0 +1,94 @@
...
+#include "compiler/Compiler.h"
+#include "compiler/PointExecutable.h"
+#include "compiler/VolumeExecutable.h"
+
+namespace openvdb {
+OPENVDB_USE_VERSION_NAMESPACE
+namespace OPENVDB_VERSION_NAME {
+namespace ax {
+
+/// @note Implementation for initialize, isInitialized and unitialized
```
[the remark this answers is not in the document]
Idclip: Fixed

openvdb_ax/openvdb_ax/compiler/Compiler.h
Idclip: Fixed

openvdb_ax/openvdb_ax/compiler/Compiler.h
Idclip: Fixed

openvdb_ax/openvdb_ax/ax.h
```diff
@@ -0,0 +1,95 @@
...
+///           only refer to the provided single grid. On success, the grid will
+///           have its voxels or point data modified as dictated by the provided
+///           AX code.
+/// @note  Various defaults are applied to this pipeline to provide a simple
+///        run signature. For OpenVDB Numerical grids, only active voxels are
+///        processed. For OpenVDB Points grids, all points are processed. Any
+///        warnings generated by the parser, compiler or executable will be
+///        ignored.
+/// @note  Various runtime errors may be thrown from the different AX pipeline
+///        stages. See Exceptions.h for the possible different errors.
```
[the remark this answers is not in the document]
Idclip: I've incorporated the logging changes (see `compiler/Logger.h`). These run functions haven't been updated but will add methods in the future which take a `Logger` here too. On the compiler, you can now call `exe = C.compiler<T>("@a = 1;", &logger);`. This will return a nullptr on failure and update the logger accordingly. Exceptions have been limited to those which cause unexpected behaviour (incorrect dev api usage) and can probably be further reduced to asserts in the future.

Idclip:
@danrbailey 

> I had issues attempting to use AX due to the way axparser.h is included. The challenge is the custom logic in the CMake means it's not possible to simply copy the source tree and compile the code as the generated grammar files are not found. Would it be possible to adjust the way this header is included to something like this:

I've reworked this as you requested

> I ran into problems with matrices. I was initially getting compiler errors such as Abs(Mat3) is not found, etc. Would it be possible to change it so that you include a compiler define to enable matrices rather than one to disable them (ie OPENVDB_AX_MATRIX). That way it aligns with the default behavior in OpenVDB where matrix grids are not supported. 

Also flipped this logic

> There's also a parameter in the CMake - MATRIX_SUPPORT which doesn't appear to be documented.

This CMake parameter, along with the version branching in AX, can be complete removed on merge as it's coupled to VDB. - so when #723 is merged, this code can be removed

> Minor request - in the unit tests, could the relative paths change to be absolute paths (#include "../ast/AST.h" => #include <openvdb_ax/ast/AST.h>). 

Done

openvdb_ax/openvdb_ax/ast/Tokens.h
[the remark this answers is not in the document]
Idclip: I'm most likely not going to add this integer precision switch for this PR. I have removed `int16` types but haven't been able to come up with an implementation that I'm happy with yet w.r.t aliasing int to either 32 or 64 bit. This would be much simpler if a) AX wasn't bound to closely to VDB types and b) floats were treated in a similar way. We can add this later at no change to the frontend and I'll be able to address all three of these issues together which will make for a much better implementation

openvdb_ax/openvdb_ax/ast/Tokens.h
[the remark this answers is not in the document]
Idclip: One last concern with this is that python, which implements floored mod, also implements floored division. This means that in both C and python, the following expressions are true:

```
  (-7/3)*3 + (-7%3) == -7
  (-7/-3)*-3 + (-7%-3) == -7
  (7/-3)*-3 + (7%-3) == 7
```

These are **not** true in VEX. Are we sure this is the right decision? I would think that keeping these properties and relationship is more important than the behaviour of mod. @jmlait?

openvdb_ax/openvdb_ax/ax.cc
Idclip: I've switched this to throw if the vector contains both points and volumes - as you say, when this functionality becomes available it'll change the result

openvdb_ax/openvdb_ax/compiler/LeafLocalData.h
Idclip: Done and moved to codegen/

Idclip:
A summary of changes made to this PR:

 - Refactored CI with AX workflow and aswf clang containers
 - Removed frontend short and long types - replaced with int16, int64. Added int32 type for consistency. Removed the ability to create int16 locals and literals. I have **not** added int64 bit mode. See https://github.com/AcademySoftwareFoundation/openvdb/pull/818#discussion_r515919648
 - I have **not** implemented floored modulus. See https://github.com/AcademySoftwareFoundation/openvdb/pull/818#discussion_r511061527
 - Added a compiler `Logger` which removes most exceptions
 - Updated lerp to be C++20 standards compliant 
 - Added `sign()`
 - Removed AX versioning - lib and namespace uses VDB version
 - Fixed relation include paths in binary projects

There are a number of outstanding API requests which I haven't actioned yet (@danrbailey) however I believe there's nothing that can't now be changed in the future either transiently or with deprecation support fairly easily. I am also open to changing more until 7.2 gets release and we can discuss should anyway feel strongly about any of the outstanding points. However it would be great to tie this particular PR off!

openvdb_ax/openvdb_ax/ast/Tokens.h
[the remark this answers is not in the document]
jmlait: That is a good point about the non-floored division not matching the floored mod.  Ideally, those would match.

But I disagree that maintaining that invariant is the most important thing.  The most important thing is for people writing selection functions with mod that they get an expected range.

int choice = something % nchoice;

is a very common idiom that fails utterly when something is signed.  I've seen people forget about sign in that case too many times to be complacent about this.

The other use of mod is for wrapping/tiling.  In this case, the lack of floored division means that to get the tile number you can't just use integer division.   But I'll note that maintaining that invariant won't help someone trying to convert a coord into a tile/tileoffset:

int tile = x / resx;
int tileoffset = x % resx;

The invariant that this will rebuild to the original x with
x = tile * resx + tileoffset;
is rather moot - you have a negative tile offset in the C'style, and a repeated tile index around 0, so you already have serious bugs with the behaviour at zero.

The break in the division invariant is annoying, but is also not without precedent.  Ada also breaks mod from div, likely for this very reason - the variant of mod that preserves the second-term's sign is the safer choice to avoid bugs.  They interestingly separate the concept of mod from the concept of remainder - remainder should satisfy that property, mod does not have to.

Support % vs explicit functions is a bit moot, you still have to argue what the `mod` function does :>  We do end up with SYSsafemod() (C-style) and SYSwrapmod() (Python style) in Houdini.

openvdb_ax/openvdb_ax/ast/Tokens.h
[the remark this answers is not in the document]
jmlait: Sounds good!

openvdb_ax/openvdb_ax/ast/Tokens.h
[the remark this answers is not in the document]
Idclip: Ad discussed, I've made this change! AX now performs floored mod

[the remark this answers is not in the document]
Idclip: As discussed I've made this change! AX now performs floored mod

Idclip:
Looking to merge this @danrbailey @jmlait

.github/workflows/houdini.yml
```diff
@@ -2,73 +2,62 @@
...
+      - '**/*.md'
+  pull_request:
+    branches:
+      - '**'
+    paths-ignore:
+      - 'CHANGES'
+      - 'openvdb/doc/**'
+      - 'openvdb_maya/**'
+      - 'openvdb_ax/**'
+      - '**/*.md'
```
danrbailey, line 22: Can we move this back into the build.yml job? I would prefer all the build jobs to remain under the same heading. I think the original reason for wanting to split this was due to having different triggers, but it looks like they're the same now?

ci/install_llvm.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# Download, build and install LLVM from source
```
danrbailey, line 5: This file isn't used anymore? Should we leave it out until there's a need for it?

.github/workflows/houdini.yml
Idclip: This also allows for individual status badges (see here https://github.com/dneg/openvdb). tbh I prefer it this way, it's much clearer and the jobs are completely separate. The Houdini one may also change with the introduction of the AX SOP. Any of this worth it to you?

ci/install_llvm.sh
Idclip: Removed

.github/workflows/houdini.yml
danrbailey: IMO, those status badges are not particularly useful, maybe as a small motivation to us to fix CI issues in a timely fashion? As a user, I think it's fine to know if "something" is broken.

I'd like to optimise this for the developer experience and there are two issues that this implementation exposes in the GitHub Actions UI:

* if you group under multiple workflows, when you view one "run", you cannot easily jump to another that was triggered from the same commit, you have to go back to "actions" and then find the next from the list of all actions. Here's an example:

https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/361703656

Now try and find the "houdini" run that was triggered from the same 184151f commit?

It's fine if you go via the checks page from the PR such as here:

https://github.com/AcademySoftwareFoundation/openvdb/pull/818/checks

But that means you have to navigate to the PR first and then to this page rather than go directly via actions.

* your automated matrix naming truncates the name of the build. From the same link as above - can you find which one is ABI=5 without having to hover over the name? I don't mind adopting this matrix approach, but I would at least like to fix the naming.

Idclip: I see how the first issue could catch somebody out, though I think it's a bit pedantic... I'll move it back for the sake of not stalling this PR

> your automated matrix naming truncates the name of the build. 

This has nothing to do with moving it, but I'll fix the naming

danrbailey: Thanks. Agreed, it definitely reads as pedantic when viewed in light of this massive PR. :) I think in future we should probably make this kind of change in a separate PR. Many of these CI changes are unrelated to AX and could be reviewed in isolation.

Yes, the naming could have been a separate comment.

Idclip: I've fixed both issues, though when I took a look at the houdini naming it seems the old naming was even worse:

https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/362887233
vs
https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/361703653

"testhou175" vs " linux-vfx-hou:17_5-vfx:2018-cxx:cl..." (where the ... is the truncation)

So I'm not sure what you were referring too.. I'm removed the " linux-vfx-" prefix anyway

Idclip: Just realised I think you were referring to the linux build names? I've thus moved the abi number to the front

danrbailey: Great, thank you!

danrbailey (approved):

openvdb_ax/openvdb_ax/ax.h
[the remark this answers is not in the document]
jmlait: Looking forward to the logger based run() method in future PRs.

jmlait (approved):

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+@par
+Where @b `grids` can comprise of any number of OpenVDB volumes. In this program,
+only a floating point grid with the name @b `myattribute` will be updated. Each
+voxel in this grid will have it's value compared in the same way as the points
+example given above.
+
+
+@subsection axexecutionxcontext OpenVDB Execution Context
+@par
+AX programs are designed to run over different types of geometry, specifically
```
kmuseth, line 165: "types of geometry" -> "types of grids". While points could be referred to as geometry I don't think it's a good description for voxels

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+OpenVDB points and OpenVDB volumes, the latter of which is the set of all
+numerical OpenVDB grids. Programs must be compiled separately for points and
```
kmuseth, line 167: what does "numerical" refer to here? Integer?

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+    specifically for point attributes @b or voxel values, not both.
+    - Whilst the AX grammar does not change syntactically, @ref axattribaccess
+    "attribute accesses" may be affected. This is due to the fundamental
+    differences of executing over different types of VDBs/geometry.
+@par Execution over Points
+AX kernels compiled for OpenVDB Points run @b individually on <b>each point</b>
+in every <b>active voxel</b> in a points VDB. Attributes that have been
+accessed in the program are provided such that the program has access to all
+data available on the currently processing point. Multiple OpenVDB Point grids
+can be processed by the same AX program, but only a single OpenVDB Points VDB
```
kmuseth, line 183: "OpenVDB Points VDB" -> "OpenVDB Point Grid" ?!

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+can be processed at a time. The default behaviour for AX point programs is to
+process every point which exists in active OpenVDB voxels.
+@par Execution over Volumes
+AX kernels for OpenVDB volumes run individually on every <b>active voxel</b> of
+VDBs which are @b written to (whilst retaining access to all available VDBs).
+Volumes can be thought of as only holding a single "attribute". Whilst points
+hold relevant attributes on the same element in the geometry (assuming no
+neighbour accesses), multiple volume accesses require accessing multiple volume
+primitives. The location of this access is the <b>world space position</b> of
```
kmuseth, line 192: I think the terms "geometry" and "primitives" in this context could use a more explicit definition

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+
+
+<br/><hr>
+
+
+@section axdatatypes Data Types
+@par
+AX supports a rich variety of native types to be able to maximise the efficiency
+of arithmetic operations and facilitate accessing underlying geometry. As AX was
+originaly designed for OpenVDB, native AX types aim to provide direct access to
```
kmuseth, line 206: the use of "originally" seems to suggest that it's not longer the case (that since it's initial development it has been extended to other grid types than OpenVDB). Is that the case?

kmuseth (approved):
Minor comments in the documentation.

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+- @ref axfunctions
+    - @ref axuserfunctions
+- @ref axvexsupport
+- @ref axexamples
+
+
+@section axintro Introduction
+@par
+AX is a fast and portable JIT compiled expression language, originally designed
+for writing volume and point kernels specifically for OpenVDB. It was originally
```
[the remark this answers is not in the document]
Idclip: Done

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+    - @ref axloops
+- @ref axfunctions
+    - @ref axuserfunctions
+- @ref axvexsupport
+- @ref axexamples
+
+
+@section axintro Introduction
+@par
+AX is a fast and portable JIT compiled expression language, originally designed
```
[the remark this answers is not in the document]
Idclip: Done

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+whilst retaining the performance achievable by custom C++ operators.
+@par
+The language is predominately inspired by <A HREF="https://www.sidefx.com/docs/houdini/vex/index.html">SideFX's VEX</A>
+language which itself takes elements from C, C++ and the Renderman Shading
+Language. The design of AX uses concepts from these languages and refrains
+from deviating from their syntax too significantly. Specifically, it does not
+aim to change what are already well established and heavily used language
+syntaxes within the Visual Effects industry; on the contrary, AX aims to provide
+an interface to which a user with previous or little programming experience can
+easily transition to. However it does introduce new concepts where the the
```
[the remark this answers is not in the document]
Idclip: Done

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+Note that native AX only supports the modification of OpenVDB data. Currently,
+that means that native AX programs can be written and built for execution over
+OpenVDB Points or OpenVDB Volumes. For any given AX program, the execution context
+(what it's processing) may change what functionality is available and, more
+importantly, the behaviour of reading/writing from attributes. See the @ref axexecutionxcontext
+for details.
+@par
+For details of extending AX for custom geometry, see the @subpage openvdbaxapi
+developer documentation. Note that custom geometry support requires C++ extension.
+
```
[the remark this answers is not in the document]
Idclip: I've added a new paragraph which talks about this in this section!

doc/ax/ax.txt
```diff
@@ -0,0 +1,2095 @@
...
+float temp = float@myattribute;
+// if the value is less than 0, clamp it to 0
+if (temp < 0.0f) float@myattribute = 0.0f;
+@endcode
+Note that there is no other required logic here to make this a compatible
+program i.e. feeding this code to the @ref vdbaxbinary "vdb_ax command line binary"
+will compile and execute over provided OpenVDB files.
+@par
+The following explains the above example in relation to OpenVDB data; OpenVDB
+points and OpenVDB volumes. The example demonstrates reading from a value on
```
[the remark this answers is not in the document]
Idclip: I've left this as it is, specifically because the idea of "reading" in this instance pertains to the language as well as the vdb type and I want to make the important distinction between reading and writing (where as "access" is slightly more ambiguous)

doc/ax/ax.txt
Idclip: Done

doc/ax/ax.txt
Idclip: Done

doc/ax/ax.txt
Idclip: I've reworded this section to better describe VDB behaviour

doc/ax/ax.txt
Idclip: Removed "originally"

doc/ax/ax.txt
Idclip: It was an attempt to distinguish between point grids and all other natively supported vdb types but I've reworded this

PR #819, opened 2020-09-05
Make Half an Optional Dependency

danrbailey:
This is a follow-up PR to #651, thanks @BenFrantzDale. I've incorporated your changes and made a few additional edits to get this ready for merging into OpenVDB.

It is now possible to use the CMake option `-DUSE_HALF=OFF` to disable the Half library. In this case, it is no longer possible to read or write Half grids and any attempt to call these methods without OpenVDB being built with Half support will throw an exception.

I also made a small change to the way that EXR is supported in OpenVDB - the RayTracer now always has a `saveEXR()` method which throws an exception if OpenVDB is built without EXR support. I also renamed the macro to `OPENVDB_USE_EXR` to fall in line with other similar macros.


openvdb/openvdb/Types.h
```diff
@@ -256,6 +264,25 @@ struct ValueTraits<T, false>
...
+template <> struct IsFloatingPoint<half> { enum { value = true }; };
+#endif
+
+
+// @brief Extend std::is_class<T>::value to work with half
+template <typename T>
+struct IsClass { enum { value = std::is_class<T>::value }; };
+#if OPENVDB_USE_HALF
+template <> struct IsClass<half> { enum { value = false }; };
+#endif
```
Idclip, line 280: Not sure I agree with this - half is technically a class right?

openvdb/openvdb/tools/RayTracer.h
```diff
@@ -358,6 +358,11 @@ class Film
         file.setFrameBuffer(framebuffer);
         file.writePixels(mHeight);
     }
+#else
+    void saveEXR(const std::string& /*fileName*/, size_t /*compression*/ = 2, size_t /*threads*/ = 8)
+    {
+        OPENVDB_THROW(IoError, "OpenVDB has been compiled without EXR support");
+    }
```
Idclip, line 365: This class isn't templated - can we just move the implementation and all EXR includes to a source file? It would mean there's one less define people have to worry about in their own projects

openvdb/openvdb/Types.h
danrbailey: Half is a class containing one member - an unsigned short. What we're really after knowing here is if the type is a component type (ie vector, matrix, quaternion) or a simple POD-type (half, short, float, etc).

openvdb/openvdb/tools/RayTracer.h
danrbailey: The define is handled for you by CMake which is our only supported build system at this stage and it already existed but under a different name, I'm just aligning it with the other defines. I imagine this method is rarely used, so let's not over-complicate this.

BenFrantzDale:
I still liked my version that wouldn't compile requests to read/write `half` files rather than leaving it to a runtime error, but ultimately all I care is that the dependency chain is broken so openvdb doesn't require OpenEXR.

danrbailey:
> I still liked my version that wouldn't compile requests to read/write `half` files rather than leaving it to a runtime error, but ultimately all I care is that the dependency chain is broken so openvdb doesn't require IlmBase.

For writing and for reading caches you've written yourself, I would agree that a compile time error is preferable. The problem is that when you attempt to read a .vdb file with a half grid inside it which was written by someone else (who did have half enabled), ultimately you'll still end up with a runtime error. It's just that it would be something much more obtuse, something like `cannot read grid - unregistered grid type "tree_half_5_4_3"`. I think it's better to have a clearer runtime error that's more meaningful in this scenario.

BenFrantzDale:
@danrbailey I agree that a runtime error is required for reading unregistered types. Still, I think requesting to write `half` could be prevented at compile time. But again, I don't want to get in the way of this PR.

openvdb/openvdb/Types.h
Idclip: > Half is a class containing one member

That's it then! Doesn't matter if it's a single POD member or just an interface, you'd expect the implementation of `std::is_class` to return true if the type is a class. Additionally this new struct you've introduced is called `IsClass` which you'd expect to also return true.

Taking a step back and looking at its use here:

```
template<typename ValueType_, typename Codec_>
bool
TypedAttributeArray<ValueType_, Codec_>::valueTypeIsClass() const
{
    // half is not defined as a non-class type as expected, so explicitly exclude it
    return std::is_class<ValueType>::value && !std::is_same<half, ValueType>::value;
}
```

It feels like this method is poorly named for the intended result. As you say:

> What we're really after knowing here is if the type is a component type

So perhaps it should be renamed as such. Either way, I definitely don't think we should be introducing an `IsClass` struct in this way for `half`

BenFrantzDale: Would an `IsScalar` be more appropriate?

openvdb/openvdb/tools/RayTracer.h
Idclip: >  I imagine this method is rarely used, so let's not over-complicate this.

Sorry I don't agree with this! I believe the change I'm proposing would in fact simplify this implementation. Although CMake handles it for you doesn't mean we shouldn't continue to simplify the build system for downstream projects - in essence that's the reason we're making these changes! 

I think the changes are trivial - If you like I'm happy to make them and submit a PR. 1) move the implementation to a new .cc, 2) add the new .cc to CMake and 3) remove the need for downstream projects to worry about this define.

openvdb/openvdb/Types.h
Idclip: I think the root problem here is that I disagree with the behaviour of `TypedAttributeArray::valueTypeIsClass`. As this logic is only used there (I think) there's two solutions. The first doesn't address the issue but, for this PR, we simply remove `IsClass` and introduce the compiler branching logic in `valueTypeIsClass()` to call `!std::is_same<half, ValueType>::value` if `half` if enabled. The second and better solution would be to deprecate `valueTypeIsClass()` and introduce more appropriate methods such as `valueTypeIsPOD()` (`std::is_pod<>`) or better `valueTypeIsTrivial() valueTypeStandardLayout()` (`std::is_trivial<> std::is_standard_layout()`)

danrbailey: None of these alternatives are really appropriate - std::is_pod is being deprecated in C++20, std::is_trivial and std::is_standard_layout would all return true for simple component structs (Vec, Quat, Matrix). Again, what we really want to know is if the type contains more than one (nested) value. std::is_scalar or std::is_arithmetic might work. We could change valueTypeIsClass to valueTypeIsScalar, I think I originally considered this but can't remember why I went with class in the end.

Anyway, I don't want to derail this PR as it's a bit away from the scope of this change, so I'll put this into the method itself and we can discuss changing it in a subsequent effort, which is a bit more involved as it requires phasing out through ABI.

openvdb/openvdb/tools/RayTracer.h
danrbailey: Actually I think your suggestion is more limiting. Your solution would require the user to go back and recompile the core OpenVDB library with OpenEXR. In addition, it wouldn't be a clear error, it would be a linking error complaining about this method not having been defined. With my solution, you _could_ simply link OpenEXR into the final application and set this define yourself without having to touch the core library.

openvdb/openvdb/Types.h
BenFrantzDale: Yeah, it seems like the standard doesn't want people using its `type_traits` metafunctions to talk about scalar-like user-defined types: Both `std::is_floating_point` and `std::is_arithmetic` say it's UB to specialize. :-/

`openvdb/Types.h` does have `ValueTraits<T>` including `ValueTraits<T>::IsScalar`, FWIW.

danrbailey:
I've updated this PR to remove `GridBase::setSaveFloatAsHalf` when Half is not enabled and made the associated fixes in the Python module and Houdini plugin.

openvdb/openvdb/Types.h
danrbailey: That's a very good point, in fact this may have been what motivated me to add ValueTraits. Thank you for reminding me about this one. :+1: 

I've updated this PR to convert all these methods over to use ValueTraits now.

jmlait:
I don't think I agree that GridBase::setSaveFloatAsHalf should be erased in this build.

This is just metadata on the grid.  A plugin that is built with no half support should still support preserving this preference and setting it.

A compile time error doesn't really work because you never know where your grids came from - they can come from any ABI compliant build of OpenVDB, not necessarily another optionalhalf build!  So I think the right thing is the runtime error if you try and save a SaveFloatAsHalf grid inside of an optionalhalf build.  But optionalhalf builds should have full and truthful access to the metadata value.

kmuseth:
Would like to discuss this PR in our next TSC meeting. While I agree that it would be great to remove dependency on ILMBase I don't agree that this means stripping out support for fp26 from OpenVDB. In fact I've received requests from Autodesk to expand our support of FP16 to in-core as well as out-of-core. So, instead I suggest we add our own implementation of Half.h which of course is compatible with the one in ilmbase. We might even have one here at nvidia that we can share. FP16 will also be useful for NanoVDB.

danrbailey:
Closing this PR as a result of the recent decision to embed the half library into OpenVDB (#927).

PR #822, opened 2020-09-11
TypeList::Unique compiler fix

Idclip:
Fixed an issue where `TypeList::Unique` would be instantiated continuously on every `TypeList`.

It seems that as soon as a `TypeList` of any kind is created, `TypeList::Unique` would also be instantiated. As this also creates a `TypeList` this causes the compiler to explode for larger lists. I tried changing the implementation of `TypeList::Unique` to use a different container but this doesn't help. Really the `Unique` typedef should be a separate template struct. However this fix stops automatic instantiation so it can continue to live on the `TypeList`.

See an example of the issue here:

https://github.com/Idclip/openvdb/runs/1101152129?check_suite_focus=true

### DIFF
--- a/openvdb/openvdb/unittest/TestTypes.cc
+++ b/openvdb/openvdb/unittest/TestTypes.cc
@@ -536,13 +536,13 @@ TestTypes::testTypeList()
     //
 
     // Test some methods on lists with duplicate types
-    using DulplicateIntTypes = TypeList<Int32, Int16, Int64, Int16>;
-    using DulplicateRealTypes = TypeList<float, float, float, float>;
-    static_assert(DulplicateIntTypes::Size == 4, "");
-    static_assert(DulplicateRealTypes::Size == 4, "");
-    static_assert(DulplicateIntTypes::Index<Int16> == 1, "");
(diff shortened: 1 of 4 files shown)


openvdb/openvdb/unittest/TestTypes.cc
```diff
@@ -541,8 +541,8 @@ TestTypes::testTypeList()
     static_assert(DulplicateIntTypes::Size == 4, "");
     static_assert(DulplicateRealTypes::Size == 4, "");
     static_assert(DulplicateIntTypes::Index<Int16> == 1, "");
-    static_assert(std::is_same<DulplicateIntTypes::Unique, TypeList<Int32, Int16, Int64>>::value, "");
-    static_assert(std::is_same<DulplicateRealTypes::Unique, TypeList<float>>::value, "");
+    static_assert(std::is_same<DulplicateIntTypes::Unique<>, TypeList<Int32, Int16, Int64>>::value, "");
```
danrbailey, line 544: While you're touching this file, maybe fix these typos too -> Dulplicate -> Duplicate ?

danrbailey (approved):

PR #823, opened 2020-09-12
Make Delayed Loading and Boost IOStreams Optional 

danrbailey:
Allow disabling delayed-loading when building the core library. This turns boost::iostreams into an optional dependency and is one step closer to making Boost optional!

I opted to add more macro defines than strictly necessary to eliminate the iostreams dependency because I've wrapped them around synchronisation primitives such as mutexes and atomics. Hopefully this gives us some insight into any performance cost that might be associated with using delayed loading.


jmlait (commented):
It isn't clear from the documentation of the flag that this results in a very different ABI?

Someone might enable this just to make interprocess/etc optional, and not realize it isn't compatible?

Either we should keep the binary layout so it remains compatible, or make it more clear what is happening when this flag go es on.

danrbailey:
> It isn't clear from the documentation of the flag that this results in a very different ABI?
> 
> Someone might enable this just to make interprocess/etc optional, and not realize it isn't compatible?
> 
> Either we should keep the binary layout so it remains compatible, or make it more clear what is happening when this flag go es on.

This PR should _not_ be changing Grid ABI. Can you point out where the ABI is being changed?

Idclip:
@danrbailey If you have the time it would be awesome to revist this

richhones:
Is there any way to improve the error if trying to compile without delayed loading if trying to use the signature that doesnt exist?

jmlait (approved):

danrbailey:
> Is there any way to improve the error if trying to compile without delayed loading if trying to use the signature that doesnt exist?

It shouldn't fail for the first time when you attempt to call `File.open()` because the class required for the second argument is also gated by this macro, so you shouldn't even be able to create an object. I hope that's obvious enough, though honestly I think the chance that anyone is using this functionality is practically zero.

PR #826, opened 2020-09-14
Added the OPENVDB_STATICLIB define to all static builds

Idclip:


openvdb/openvdb/CMakeLists.txt
```diff
@@ -438,7 +438,7 @@ if(OPENVDB_CORE_STATIC)
   )
 
   target_compile_definitions(openvdb_static
-    PUBLIC ${OPENVDB_CORE_PUBLIC_DEFINES}
+    PUBLIC ${OPENVDB_CORE_PUBLIC_DEFINES} -DOPENVDB_STATICLIB
```
diiigle, line 441: This is applying to all platforms.
In FindOpenVDB above it is only `WIN32`

Idclip: Correct - this matching the behaviour of `OPENVDB_DLL`. Perhaps these should also be in `WIN32` branches though there is obviously no difference on other platforms.

danrbailey (approved):

PR #837, opened 2020-10-02
Fixing mapping between index and world referentials.

### LINKED ISSUE #834: NanoVDB Grid worldToIndexDirF / indexToWorldDirF problem ?
In the Grid class the functions to transform a direction are not working correctly. 
While in transforming a point with worldToIndexF the matF in struct Map is used, the transforming of the direction uses the inverse mInvMatF. When a ray is transformed with ray.worldToIndexF(grid); this results in a wrong transformation. I see that the working code is commented out. Is there any reason why it's changed? but in my code it results in a wrong ray transformation. 
I can workaround this easily for now but it should be checked

webanck:
Fixes #834

### DIFF
--- a/nanovdb/nanovdb/NanoVDB.h
+++ b/nanovdb/nanovdb/NanoVDB.h
@@ -1381,12 +1381,12 @@ public:
     /// @brief transformation from index space direction to world space direction
     /// @warning assumes dir to be normalized
     template<typename Vec3T>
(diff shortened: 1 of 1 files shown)


kmuseth (commented):
Thanks, I've added this fix to a larger commit - so I'll be closing this PR

kmuseth:
This was add in a separate PR together with several other changes. So I'll be closing this PR

PR #839, opened 2020-10-05
Improve find active values

kmuseth:
New features:
    - Added tools::anyActiveVoxels and tools::anyActiveTiles that check
      if respectively active voxels or tiles in a tree intersect a given
      CoordBBox.
    - Added tools::activeTiles that returns a vector containing all the active
      tiles in a tree that intersects a given CoordBBox.


kmuseth:
can someone please review this - it's a relatively small PR and people are requesting it

danrbailey:
I only noticed one error per-se and it's not even part of this PR, but I noticed it while reviewing this. The definition of the standalone countActiveValues() method has the wrong signature:

```
inline bool
countActiveValues(const TreeT& tree, const CoordBBox &bbox)
{
}
```

The `bool` return type should be `Index64`. I don't think this method is being unit tested either which is probably why it wasn't caught.

Other than that, I don't think there are any technical issues here. My main feedback is that I think you're leaving a lot of money on the table in terms of performance. Here's where I see room for improvement:

* None of the standalone anyActiveValues(), anyActiveVoxels(), anyActiveTiles(), noActiveValues() methods are threaded. The most effective parallelization here would be filtered, breadth-first parallelization as provided by the DynamicNodeManager.
* The standalone countActiveValues() and activeTiles() methods are threaded using depth-first parallelization, it would be faster to do filtered, breadth-first parallelization again through DynamicNodeManager.
* I question the value of the FindActiveValues class itself. While it might be _marginally_ faster - if the computation of the root node child arrays were parallelized (ie `FindActiveValues::init()`), would the difference even be that noticeable?
* The ValueAccessor methods don't seem to be providing much benefit here - it's only used as an early-out for a single call, which would mean the cache of the ValueAccessor is being populated top-down anyway, so it'll actually be slightly slower than just doing a straight `tree().isValueOn()`? The only benefit I can see would be if you call it repeatedly on the same tree with very similar bounding boxes? Is that a common use-case here?

I wouldn't hold this PR up over any of these suggestions. Ultimately, having the functionality is more beneficial than not having it at all, but still voicing some areas where I think there are potential improvements to be had.

kmuseth:
I fixed the signature bug and will look at the suggestions for additional multi-threading at some other time (it has already proven plenty fast to be useful)

danrbailey (approved):

PR #843, opened 2020-10-09
Updated the VDB Filter class/SOP to support tiles

Idclip:


danrbailey:
Hi @Idclip, I tried to evaluate your unit tests here to better understand your changes, but they're all still based on CppUnit - would you mind updating this PR so that I can try this out?

openvdb/openvdb/tools/Filter.h
```diff
@@ -5,8 +5,7 @@
 ///
 /// @file tools/Filter.h
 ///
-/// @brief Filtering of VDB volumes. Note that only the values in the
-/// grid are changed, not its topology! All operations can optionally
+/// @brief Filtering of VDB volumes. All operations can optionally
 /// be masked with another grid that acts as an alpha-mask.
```
jmlait, line 9: This should comment in the new version that topology *can* be modified if setProcessTiles is set (and comment that it is not set by default, if I read it right?)

Likewise, mention the dilation is "chunky" (if it is...)

jmlait (approved):

Idclip:
> Hi @Idclip, I tried to evaluate your unit tests here to better understand your changes, but they're all still based on CppUnit - would you mind updating this PR so that I can try this out?

done!

openvdb/openvdb/tools/Filter.h
Idclip: > Likewise, mention the dilation is "chunky" (if it is...)

Updated the comment, though I'm not sure what you're referring to with this!

danrbailey:
This change looks great. One question I have is whether you're redundantly storing mask tree leaf nodes when you don't care about the voxel values in the bitmask? I ask because I ran into this recently in the new merge functionality. With very large trees the mask leaf nodes were starting to take up performance and increase memory, so I opted to create a tree that has active tiles instead of leaf nodes:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/Merge.h#L446

Is that something that applies here too?

Idclip:
> This change looks great. One question I have is whether you're redundantly storing mask tree leaf nodes when you don't care about the voxel values in the bitmask? I ask because I ran into this recently in the new merge functionality. With very large trees the mask leaf nodes were starting to take up performance and increase memory, so I opted to create a tree that has active tiles instead of leaf nodes:
> 
> https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/Merge.h#L446
> 
> Is that something that applies here too?

Yes, see this comment: https://github.com/AcademySoftwareFoundation/openvdb/blob/7c6981dc5fa6874a090dfb162e0ddc39483e3f1e/openvdb/openvdb/tools/Filter.h#L251-L252

Because I'm using `topologyUnion` to activate the leaf buffers in the dest tree, the mask tree needs to be densified anyway. Ideally what I need is to write a foreach across a sparse/tiled mask tree that touches leaf nodes in the dest tree, but topUnion is faster than the naïve serial approach and a mt version doesn't exist yet i.e:
```
src.addTile(...);
dest.topologyUnion(src, /*preserveTilesInDest*/false, /*densifyTilesFromSrc*/true)
```

danrbailey:
> > This change looks great. One question I have is whether you're redundantly storing mask tree leaf nodes when you don't care about the voxel values in the bitmask? I ask because I ran into this recently in the new merge functionality. With very large trees the mask leaf nodes were starting to take up performance and increase memory, so I opted to create a tree that has active tiles instead of leaf nodes:
> > https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/Merge.h#L446
> > Is that something that applies here too?
> 
> Yes, see this comment:
> 
> https://github.com/AcademySoftwareFoundation/openvdb/blob/7c6981dc5fa6874a090dfb162e0ddc39483e3f1e/openvdb/openvdb/tools/Filter.h#L251-L252
> 
> Because I'm using `topologyUnion` to activate the leaf buffers in the dest tree, the mask tree needs to be densified anyway. Ideally what I need is to write a foreach across a sparse/tiled mask tree that touches leaf nodes in the dest tree, but topUnion is faster than the naïve serial approach and a mt version doesn't exist yet i.e:
> 
> ```
> src.addTile(...);
> dest.topologyUnion(src, /*preserveTilesInDest*/false, /*densifyTilesFromSrc*/true)
> ```

In this case, you would expect densifyTilesFromSrc to only apply to tiles that get densified into leaf nodes right (ie not RootNode tiles for example)?

Idclip:
> > > This change looks great. One question I have is whether you're redundantly storing mask tree leaf nodes when you don't care about the voxel values in the bitmask? I ask because I ran into this recently in the new merge functionality. With very large trees the mask leaf nodes were starting to take up performance and increase memory, so I opted to create a tree that has active tiles instead of leaf nodes:
> > > https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/Merge.h#L446
> > > Is that something that applies here too?
> > 
> > 
> > Yes, see this comment:
> > https://github.com/AcademySoftwareFoundation/openvdb/blob/7c6981dc5fa6874a090dfb162e0ddc39483e3f1e/openvdb/openvdb/tools/Filter.h#L251-L252
> > 
> > Because I'm using `topologyUnion` to activate the leaf buffers in the dest tree, the mask tree needs to be densified anyway. Ideally what I need is to write a foreach across a sparse/tiled mask tree that touches leaf nodes in the dest tree, but topUnion is faster than the naïve serial approach and a mt version doesn't exist yet i.e:
> > ```
> > src.addTile(...);
> > dest.topologyUnion(src, /*preserveTilesInDest*/false, /*densifyTilesFromSrc*/true)
> > ```
> 
> In this case, you would expect densifyTilesFromSrc to only apply to tiles that get densified into leaf nodes right (ie not RootNode tiles for example)?

In this particular case I can guarantee that src only contains leaf level tiles (as that's how I'm constructing it) - in the more general case, should this theoretical function exist, I guess I would expect any active tiles at any level of the tree create the corresponding dense topology

PR #844, opened 2020-10-13
Fix order of operations in GridTransfomer constructor

tom-cnops:
The local_util::decompose function returns components intended for rotate-scale-translate reconstruction, but were used as scale-rotate-translate

See also the reconstruction of the decomposed parts in:
https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/GridTransformer.h#L315
https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/unittest/TestGridTransformer.cc#L294


tom-cnops:
@danrbailey my apologies if it is inappropriate to @ a developer (I'm pretty new here). 
I was just wondering if I had missed a step to get a review (should I open a corresponding issue or modify a test to illustrate the changed behavior)?

jmlait:
Thank you for pinging us.   Sorry about letting this PR languish.   It looks like you have done all the right things for review.

Looking at the spots you identified,I can see why rst might be the right choice over srt.  But I've got matrices backwards too many times in my life to necessarily trust my judgement.

Do you have an example of a context in which this occurred that can illustrate why the old code was buggy?  Likewise, adding a test modification that causes this to pass (where it failed before) would help us understand the context of the change.

I always fear these changes because if our unit tests were ideal; something should have broken from the change :>  So when I reverse the order of something and nothing happens, I get worried :>

Thank you again for finding this issue, going through the CLA hoops to get this far, and not losing interest when you were neglected.

tom-cnops:
Hi, thanks for your reply!

The existing test for the decomposition actually has a few cases that trigger the bug, but it didn't verify that GridTransformer would be constructed correctly. 
I expanded the test to verify that the transform of a GridTransformer matches its constructor argument. This fails with the `srt` order.

tom-cnops:
Not really sure what caused [681c923](https://github.com/AcademySoftwareFoundation/openvdb/pull/844/commits/681c92375f9b86f61828b08318fdb6680942deb0 ) to not build since it worked with my install, but it seems to build now.

danrbailey:
> Not really sure what caused [681c923](https://github.com/AcademySoftwareFoundation/openvdb/pull/844/commits/681c92375f9b86f61828b08318fdb6680942deb0) to not build since it worked with my install, so this is the best I could think of.

Thank you for adding a unit test, that's very helpful.

Since you submitted this PR we have migrated from CppUnit to GoogleTest for our core library unit tests, hence CPPUNIT_ASSERT() should now be EXPECT_TRUE(). To test this yourself locally, I would fetch the latest changes and merge in the master branch to your branch. You'll need to install google test in order to be able to run the unit tests now.

tom-cnops:
That explains a lot! I changed it to gtest now and at least locally the test works as intended.

danrbailey (approved):
This looks good to me. I tested your unit test change (which I would have expected to work) and it does indeed fail without your fix. Thanks for submitting this.

@jmlait - can you verify this change and I'll merge this in for the 8.0 release?

jmlait:
I see, we had the test to verify it but did the round trip by explictily rebuilding.  That is a valid thing to do to avoid the case where both decompose & construction were lying, but meant we didn't actually test the rebuild.

jmlait (approved):

danrbailey:
Merged, thank you! This will go into the upcoming 8.0 release.

If you're looking for other ways to contribute or want to ask questions about a potential contribution, please don't hesitate to get in touch with us - openvdb-dev@lists.aswf.io. :)

PR #849, opened 2020-10-19
Fix build of feature/nanovdb: explicit return type instead of auto

theHamsta:
auto is not a valid return type in C++11. We either have to use decltype(auto)
or the explicit return type.

> 'auto' return without trailing return type; deduced return types are a C++14 extension

Signed-off-by: Stephan Seitz <stephan.seitz@fau.de>

### DIFF
--- a/nanovdb/nanovdb/NanoVDB.h
+++ b/nanovdb/nanovdb/NanoVDB.h
@@ -3303,19 +3303,19 @@ template<typename ValueT>
 using NanoGrid = Grid<NanoTree<ValueT>>;
 
 template <int CacheLevels = 3, typename ValueT = float>
-auto createAccessor(const NanoGrid<ValueT> &grid)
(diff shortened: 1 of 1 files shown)


theHamsta:
On clang12:
```
../nanovdb/./nanovdb/NanoVDB.h(3306): error: expected a type specifier

../nanovdb/./nanovdb/NanoVDB.h(3312): error: expected a type specifier

../nanovdb/./nanovdb/NanoVDB.h(3318): error: expected a type specifier
```

theHamsta:
Could also be fixed by changing this line.
`set(CMAKE_CXX_STANDARD 11)`

 Then, nanovdb would use the same standard as OpenVDB.

kmuseth:
gcc 8 doesn't seem to have an issue with this but I'm happy to merge your changes if it's an issue in clang 12. Incidentally this function (createAccessor) will change (and maybe go away) in my next commit ... hopefully later this week. I'm allowing for the ReadAccessor to be completely customized with respect to the nodes that it caches (more on this later).

PR #865, opened 2020-11-08
Improved the layout and interface of some trivial VDB types

Idclip:
- `Vec`, `Quat`, `Mat` now conform to `is_trivial` (https://en.cppreference.com/w/cpp/named_req/TrivialType)
 - Removed unnecessary NodeUnion specialization


openvdb/openvdb/tree/NodeUnion.h
```diff
@@ -44,42 +42,28 @@ class NodeUnion
...
-    ValueT& getValue() { return mValue; }
-    void setValue(const ValueT& val) { mValue = val; }
+    // Small check to ensure this class isn't
+    // selected for some expected types
+    static_assert(!ValueTraits<ValueT>::IsVec &&
+        !ValueTraits<ValueT>::IsMat &&
+        !ValueTraits<ValueT>::IsQuat &&
+        !std::is_same<ValueT, math::Coord>::value &&
+        !std::is_arithmetic<ValueT>::value,
+        "Unexpected instantiation of NodeUnion");
```
Idclip, line 53: I'll probably remove this `static_assert` as it isn't really doing the right job - I wanted a check to ensure the right specialization was being selected, not the wrong one

Idclip:
As discussed, will remove all defaulted methods and add comments instead

danrbailey:
> As discussed, will remove all defaulted methods and add comments instead

I like leaving just the constructor defaulted. The rule of five doesn't apply to this one and it's expected that deconstructors, copy-constructors, etc will live nearby if they're defined so helps with navigating the class. Plus if anyone adds a new non-empty constructor, it won't delete the implicit one.

Idclip:
> > As discussed, will remove all defaulted methods and add comments instead
> 
> I like leaving just the constructor defaulted. The rule of five doesn't apply to this one and it's expected that deconstructors, copy-constructors, etc will live nearby if they're defined so helps with navigating the class. Plus if anyone adds a new non-empty constructor, it won't delete the implicit one.

Done

danrbailey (approved):

PR #866, opened 2020-11-08
Split out TypeList into its own header and moved macros to Platform

Idclip:


jmlait (approved):

openvdb/openvdb/TypeList.h
```diff
@@ -0,0 +1,531 @@
...
+
+template<typename... Ts> struct TypeList; // forward declaration
+
+namespace internal {
+
+// Implementation details of @c TypeList
+
+/// @brief Dummy struct, used as the return type from invalid or out-of-range
+///        @c TypeList queries.
+struct NullType {};
```
danrbailey, line 34: While we're moving this about - could we rename the namespace from internal to something that's less likely to clash (typelist_internal?). 

Using `internal::NullType` is rather generic.

openvdb/openvdb/Platform.h
```diff
@@ -229,4 +229,54 @@
     #define OPENVDB_HOUDINI_API OPENVDB_IMPORT
 #endif
 
+#if defined(__ICC)
```
danrbailey, line 232: Tempted to say we strip all of this out - we don't build or test for ICC so it's hard to know if any of this is still relevant. They're just suppressing warnings after all.

danrbailey (approved):

openvdb/openvdb/Platform.h
Idclip: Some of these defines are still in use so no harm leaving them in for now

openvdb/openvdb/TypeList.h
Idclip: I've included `TypeList.h` back into `Types.h` so as not to to break anyone's code so doesn't really matter that it's being moved - but I'm happy to make this change

Issue #869, opened 2020-11-09
initialize() should use double-checked locking


BenFrantzDale:
`openvdb::initialize()` in `openvdb.cc` is

```
namespace {
// Declare this at file scope to ensure thread-safe initialization.
Mutex sInitMutex;
bool sIsInitialized = false;
}

void
initialize()
{
    Lock lock(sInitMutex);
    if (sIsInitialized) return;

    ...
```
which means each call requires a mutex lock/unlock. There is an `uninitialize()` which also locks, although calling `uninitialize()` seems pretty dangerous in that anyone using anything that gets removed by `uninitialize()` will likely trigger UB. (For safe uninitialization to work, we'd probably be better off not to have a singleton!) If `uninitialize()` weren't available, we could just use `std::once_flag`.

All that said, the already-initialized ("hot") path for `initialize()` shouldn't require locking but should use double-checked locking <https://en.wikipedia.org/wiki/Double-checked_locking>. That is
```
namespace {
// Declare this at file scope to ensure thread-safe initialization.
Mutex sInitMutex;
std::atomic<bool> sIsInitialized{false};
}

void
initialize()
{
    if (sIsInitialized.load(std::memory_order_acquire)) return; // Hot path.
    Lock lock(sInitMutex);
    if (sIsInitialized.load(std::memory_order_relaxed)) return; // Someone else beat us to it while we were locking.

    ...
```

Idclip:
This sounds reasonable to me! Would you mind making a PR?

BenFrantzDale:
> This sounds reasonable to me! Would you mind making a PR?

Give me a few days. 

Thoughts about the safety of `uninitialize()`? It seems like it allows the caller to rip things out from under OpenVDB invoking UB. Is it actually used? I’m not volunteering to rearchitect that, but a non-singleton instance seems like the principled approach.

Idclip:
[...]

PR #873, opened 2020-11-11
Make sIsInitialized (used by openvdb::initialize() and openvdb::unini…

BenFrantzDale:
…tialize()) use double-checked locking so the hot path is just one atomic load

Closes https://github.com/AcademySoftwareFoundation/openvdb/issues/869

### DIFF
--- a/openvdb/openvdb/openvdb.cc
+++ b/openvdb/openvdb/openvdb.cc
@@ -8,6 +8,7 @@
 //#endif
 #include "tools/PointIndexGrid.h"
 #include "util/logging.h"
+#include <tbb/atomic.h>
 #include <tbb/mutex.h>
 #ifdef OPENVDB_USE_BLOSC
 #include <blosc.h>
@@ -36,14 +37,15 @@ typedef Mutex::scoped_lock Lock;
 namespace {
 // Declare this at file scope to ensure thread-safe initialization.
 Mutex sInitMutex;
-bool sIsInitialized = false;
+tbb::atomic<bool> sIsInitialized{false};
 }
 
 void
 initialize()
 {
+    if (sIsInitialized.load<tbb::memory_semantics::acquire>()) return;
     Lock lock(sInitMutex);
-    if (sIsInitialized) return;
+    if (sIsInitialized.load<tbb::memory_semantics::acquire>()) return; // Double-checked lock
 
     logging::initialize();
 
@@ -118,7 +120,7 @@ initialize()
 __pragma(warning(disable:1711))
 #endif
 
-    sIsInitialized = true;
+    sIsInitialized.store<tbb::memory_semantics::release>(true);
 
 #ifdef __ICC
 __pragma(warning(default:1711))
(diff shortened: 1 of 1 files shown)


Idclip:
I would use `std::atomic` as you first suggested - tbb is actually deprecating a bunch of their containers in favour of the standard:
https://software.intel.com/content/www/us/en/develop/articles/tbb-revamp.html

Also don't forget to sign off your commits with -s `git commit -s -m "blah"` - this is required for the DCO check you see failing (you can ignore the other build failures, they are unrelated)

Idclip:
> I would use `std::atomic` as you first suggested - tbb is actually deprecating a bunch of their containers in favour of the standard:

Though I've just realised the mutex in use is a tbb::mutex not an std::mutex - nevermind, sticking with tbb is probably better for consistency

Idclip (approved):

danrbailey:
Apologies for not replying to the original question. Just to clarify, there's nothing wrong with the thread-safety of the original implementation, it's just slower than it could be. While this is a valid optimization and we have used this kind of pattern elsewhere, I'm not convinced it's totally necessary to add the extra complexity to these methods. If the user is calling `openvdb::initialize()` / `openvdb::uninitialize()` frequently enough to benefit from this change - they're probably not using the library quite as we intended and perhaps an update to the docs would be better? Or perhaps I'm just not aware of a valid use case in which this becomes useful?

Idclip:
>  Just to clarify, there's nothing wrong with the thread-safety of the original implementation, it's just slower than it could be. 

Correct

> I'm not convinced it's totally necessary to add the extra complexity to these methods.

I considered this but quickly decided against it. The PR is 6 new lines of code for what is an obvious performance boost. This method can be called hundreds if not thousands of times and it's naïve to assume we know all use cases. Sure, in an ideal scenario the initialisation and execution of vdb library methods would be separate but that is not always the case. Lets take Houdini as an example. Each time a SOP is created initialize has to be invoked; a production scene could hold thousands of nodes inside digital assets. 

>  they're probably not using the library quite as we intended and perhaps an update to the docs would be better?

The alternative would be to force users to implement their own locks/atomics (i.e. `std::call_once` or whatever) in cases where the application they're leveraging doesn't offer the appropriate callbacks? Seems a bit silly when we can simply use the above proposal.

BenFrantzDale:
For reference, I was motivated not by measured performance, but by reading the source and noticing it could be very cheap to ensure that it’s initialized. Better would be to create a handle but that’s a bigger proposal that I personally don’t feel is justified.

jmlait (approved):

Idclip:
@BenFrantzDale Can you sign off your commit please? This will fix the DCO status check (https://github.com/AcademySoftwareFoundation/openvdb/pull/873/checks?check_run_id=1385694575) and then we can look to merge this

BenFrantzDale:
@Idclip sure. While I'm at it should I switch to std::atomic?

Idclip:
> @Idclip sure. While I'm at it should I switch to std::atomic?

I wouldn't for now as it'll require re-approval! We can merge as is

BenFrantzDale:
OK, I think I did the right incantation.

danrbailey (approved):

BenFrantzDale:
Is this all set to merge? It's blocking https://github.com/AcademySoftwareFoundation/openvdb/pull/969/files which needs to be rebased on this.

PR #874, opened 2020-11-12
Add Houdini 18.5 to CI

danrbailey:


jmlait (approved):
I presume our installer changes weren't a problem as you can just un-tar to get enough files working for CI?

danrbailey:
> I presume our installer changes weren't a problem as you can just un-tar to get enough files working for CI?

Yep, exactly. The only difference between this and previous Houdini installs was adding the lzma library which is now needed by the upgrade to Boost 1.72.

PR #875, opened 2020-11-13
Migrate from CppUnit to GoogleTest

danrbailey:
This PR migrates all our existing unit tests over to use GTest.

After our discussion this week, I was looking at GTest docs and realized that pretty much all the commands we use in CppUnit have a 1-to-1 with an equivalent in GTest. I did a quick test and found that we could get ~90% of the way through a migration by simply doing a bunch of blanket find-replaces. Here are some examples:

`CPPUNIT_ASSERT` => `EXPECT_TRUE`
`CPPUNIT_ASSERT_EQUAL` => `EXPECT_EQ`
`CPPUNIT_ASSERT_THROW` => `EXPECT_THROW`
`CPPUNIT_ASSERT_DOUBLES_EQUAL` => `EXPECT_NEAR`
`CPPUNIT_FAIL` => `FAIL`

Here's an example of a class before:

```
class TopologyToLevelSet: public CppUnit::TestCase
{
public:
    CPPUNIT_TEST_SUITE(TopologyToLevelSet);
    CPPUNIT_TEST(testConversion);
    CPPUNIT_TEST_SUITE_END();

    void testConversion();
};

CPPUNIT_TEST_SUITE_REGISTRATION(TopologyToLevelSet);

void TopologyToLevelSet::testConversion()
{
    ...
}
```

and after migration:

```
class TopologyToLevelSet: public ::testing::Test
{
};

TEST_F(TopologyToLevelSet, testConversion)
{
    ....
}
```

There was a little more work with the classes that access private internals using a friend declaration, then I ended up leaving a method on the original class and just calling that method, like this:

```
class TopologyToLevelSet: public ::testing::Test
{
public:
    void testConversion();
};

void TopologyToLevelSet::testConversion()
{
    ....
}
TEST_F(TopologyToLevelSet, testConversion) { testConversion(); }
```

Finally, I've updated the CI and the documentation. 

There are many ways in which we could use GTest better, but this is trying to do the least work to bring the existing test functionality across. I've also deleted all the custom functionality in main.cc, so there is no longer any support for the old unit test flags (-t, -v, etc). I opted for the lowest supported version being GTest 1.8.0 which works and was released in 2016. We should look at getting GTest into our OpenVDB docker containers too, but for now this just builds and installs it from source (it's pretty lightweight).


Idclip:
AX will need updating too :\

danrbailey:
> AX will need updating too :\

Hopefully not to painful, I skimmed your unit tests and it looks fairly conventional which should make it pretty straight-forward to convert.

jmlait (approved):
A surprisingly mechanical conversion!  Should wait for AX to merge in so we don't break that PR with dependency removals, though.

I noticed a few descriptions of tests were in the registration rather than the implementation, and were lost in the conversion.  Not sure if that is worth rescuing though.  It is sort of unfortunate to lose the central control board of tests; but it also is rather fortunate to have less boiler plate to create new tests.

The main problem is we have no way to test our tests to see if we are still testing everything :>

danrbailey:
I've merged in the latest master including AX changes and everything builds correctly - core library uses gtest, ax uses cppunit. I recommend we merge this PR as-is and then update AX to use gtest in a separate pass, as it already has it's own CMakeLists for the unittests. @Idclip - thoughts?

Idclip:
Hey @danrbailey thanks - I would prefer this doesn't get rushed for 7.2. This would give me time to see if I can migrate AX over to gtest and would avoid having this dual dependency for the unit tests. It would also avoid having to update the documentation twice. If I remember correctly, the main instigation for this was nanovdb so we have some time?

danrbailey:
> Hey @danrbailey thanks - I would prefer this doesn't get rushed for 7.2. This would give me time to see if I can migrate AX over to gtest and would avoid having this dual dependency for the unit tests. It would also avoid having to update the documentation twice. If I remember correctly, the main instigation for this was nanovdb so we have some time?

My main concern is just not letting this PR stale and then have to spend time fixing conflicts and getting this PR working again. I'm fine with this not going into 7.2 though. Can we create a 7.2 release branch so that I can merge this into master?

Idclip:
> My main concern is just not letting this PR stale and then have to spend time fixing conflicts and getting this PR working again. I'm fine with this not going into 7.2 though. Can we create a 7.2 release branch so that I can merge this into master?

I understand, lets discuss the 7.2 release next TSC - I'll propose we hold off with other merges which would cause this to fall behind. I'll mention however that this did come out of nowhere. Had we planned for this we could have stalled other merges/AX. The work is obviously appreciated but I'm not sure that the right approach would be to adapt entirely to a new PR because we're afraid of it becoming stale.

w.r.t the 7.2 release branch, I'm not sure it's worth it as we're so close to the deadline? I doubt much more will go in.

danrbailey:
> > My main concern is just not letting this PR stale and then have to spend time fixing conflicts and getting this PR working again. I'm fine with this not going into 7.2 though. Can we create a 7.2 release branch so that I can merge this into master?
> 
> I understand, lets discuss the 7.2 release next TSC - I'll propose we hold off with other merges which would cause this to fall behind. I'll mention however that this did come out of nowhere. Had we planned for this we could have stalled other merges/AX. The work is obviously appreciated but I'm not sure that the right approach would be to adapt entirely to a new PR because we're afraid of it becoming stale (edit: at least without discussion).
> 
> w.r.t the 7.2 release branch, I'm not sure it's worth it as we're so close to the deadline? I doubt much more will go in.

I think we just need to adapt our process to using release branches. The master should be an "in development" branch where these types of PR can be safely merged. In this case perhaps it's not worth it, but then again creating a release branch is trivial. Is it documentation or release note conflicts you're concerned about?

PR #888, opened 2020-12-01
Add new GridBase::isTreeUnique() virtual function (ABI=8)

danrbailey:


Idclip (approved):

openvdb/openvdb/Grid.h
```diff
@@ -898,8 +903,12 @@ class Grid: public GridBase
     TreeBase::ConstPtr constBaseTreePtr() const override { return mTree; }
     //@}
     /// @brief Return true if tree is not shared with another grid.
-    /// @todo Make this into a virtual function with ABI=8
+    /// @note This is a virtual function with ABI=8
+#if OPENVDB_ABI_VERSION_NUMBER >= 8
+    bool isTreeUnique() const override;
+#else
```
jmlait, line 909: Do we care that isTreeUnique() now appears to be virtual, so will be slower to invoke?  Often we end up with fully-specialized trees so would be able to directly call the method without virtual dispatch?

jmlait (approved):

openvdb/openvdb/Grid.h
danrbailey: Good point. I've marked this method as final to allow devirtualization. Calling this method on a fully-specialized tree should not require virtual dispatch now.

PR #889, opened 2020-12-01
Remove RootNode::getChildCount() (ABI=8)

danrbailey:
My intuition would be that removing a private non-virtual class method could not cause ABI breakage, but according to the [C++ ABI Itanium](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts) rules:

> You can:
> * remove private non-virtual functions if they are not called by any inline functions (and have never been)

That's not the case here (it's called by an inline function), so simply drop it in ABI=8 instead. Doesn't need a deprecation because it's private...


jmlait (approved):
Hmm.. .Might explain why removing default constructors broke ABI in Nick's change?

danrbailey:
> Hmm.. .Might explain why removing default constructors broke ABI in Nick's change?

Perhaps. The standard says:

> An implicitly-declared default constructor is an inline public member of its class

So if the explicitly defined constructor was not marked inline (as I think was the case), this could well explain what went wrong here.

PR #890, opened 2020-12-01
Add final keyword to map classes and methods (ABI=8)

danrbailey:
Not conclusive whether or not adding a final keyword can affect ABI, so to be safe, make these ABI-critical classes final using ABI=8 only.


jmlait (approved):
Approve of use of macro to centralize the ABI 8 decisions.

PR #891, opened 2020-12-01
Remove virtual LeafManager destructor

danrbailey:
I don't see a good reason why the LeafManager ever had a virtual destructor, so remove it. I don't believe this needs to be done with an ABI flag.


Idclip:
could possibly do with a changelog update should anyone be deriving (it's not marked as final)

Idclip (approved):

danrbailey:
> could possibly do with a changelog update should anyone be deriving (it's not marked as final)

Yes will add this when I merge it.

PR #895, opened 2020-12-03
Revert Node Union changes (ABI breaking)

Idclip:
Revert "d6bd5dd2e0d46288dea4d863507170a590509623" and "b97c2738da1efa889d949e9f36eaff078efeba1c"
Reverts: https://github.com/AcademySoftwareFoundation/openvdb/pull/865

Breaks ABI :( I will re-apply for VDB 8 with ABI guards


Idclip:
Again, Windows/Mac CI failures are unrelated...

jmlait (approved):
Gah.  Frustrating that this broke ABI.   Any idea how?  Only thing I can think of is if a UT_Array is built on the Houdini side it might be acting differently as we try to treat POD differently...  But that shouldn't affect already compiled Houdini code...

Idclip:
@jmlait What caught us was a crash using MMB calling through to  `SOP_Node::fillInfoTreeNodeSpecific`. It manifested in a super weird stack trace inside sesi namespace vdb using a `LeafIteratorbase`. We do incremental releases and these were the things that had changed:

 - Trivial types: https://github.com/AcademySoftwareFoundation/openvdb/pull/865
 - Matrix grid support https://github.com/AcademySoftwareFoundation/openvdb/pull/723
 - @danrbailey NodeManager stuff https://github.com/AcademySoftwareFoundation/openvdb/pull/785

As this is a bit time sensitive I didn't investigate further (and I'm now away!), but everything is now stable after reverting this. I'll send you through the full stack trace when I can and will definitely take a closer look after 7.2.

danrbailey:
Very glad we caught this before pushing out a broken build. :/

Disappointed that the ABI checks I introduced didn't catch this though, need to keep expanding these checks so that we have more confidence in making these types of changes. At least this serves as a good test case for making sure the ABI checks fail.

danrbailey (approved):

Idclip:
> Very glad we caught this before pushing out a broken build. :/

We should discuss a proposal for locking down a release branch a certain time in advance of a release. This was a similar issue to 6.2 (or was it 6.1?) but just so happened that I was able to get some users to test it before committing to 7.2.

> Disappointed that the ABI checks I introduced didn't catch this though, need to keep expanding these checks so that we have more confidence in making these types of changes. At least this serves as a good test case for making sure the ABI checks fail.

Yes, when I come back to this PR I'll see if I can introduce a test which breaks with these changes.

PR #898, opened 2020-12-11
Start Development on 8.0.0

danrbailey:
Bump the version to 8.0.0, remove support for VFX Reference Platform 2018 as dictated by our deprecation policy and introduce new deprecation warnings and update the CI.


danrbailey:
@kmuseth @jmlait @Idclip - would one of you mind approving please? We have a short window for the 8.0 release.

CHANGES
```diff
@@ -1,6 +1,23 @@
 OpenVDB Version History
 =======================
 
+
+Version 8.0.0 - In development
+
+      Some changes in this release (see "ABI changes" below) alter
```
kmuseth, line 7: Rather verbose and probably unnecessary (the definition of a new major release is an ABI change). How about instead:

This version introduces ABI changes relative to older major releases, so to preserve ABI compatibility it might be necessary to define the macro OPENVDB_ABI_VERSION_NUMBER=N, where, for example, N is 6 for Houdini 18.0 and 7 for Houdini 18.5.

doc/changes.txt
```diff
@@ -2,6 +2,28 @@
 
 @page changes Release Notes
 
+@htmlonly <a name="v8_0_0_changes"></a>@endhtmlonly
+@par
+<B>Version 8.0.0</B> - <I>In development</I>
+
+@par
+<BLOCKQUOTE>
+Some changes in this release (see @ref v8_0_0_ABI_changes "ABI changes" below)
```
kmuseth, line 11: please see the suggested change above

kmuseth (changes requested):
Except for the two minor changes in my comment it looks good

jmlait (approved):

PR #901, opened 2020-12-14
VDB To Spheres SOP had bad logic in the resolution of obsolete parameters

jmlait:
It was using the mere existence of a parm to trigger an update if the
world size was set.  Obsolete parms exist even if the .hip file lacked
a parm for them; the only way to tell if it was an old .hip file is
by checking factory defaults.


danrbailey (approved):
@jmlait - I skimmed the codebase and only spotted one other place where I believe this also needs fixing:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Write.cc#L199

Is that right?

This is such a common pattern, it suggests that there might be room for a new method something like PRM_ParmList::getNonFactoryDefaultParmPtr() ?

jmlait:
Thank you for scanning!

I'm not even sure what the correct pattern should be :<  Obsolete Parms are a bit of a mess for this reason - ideally they'd only exist if they were in the .hip, the problem is the .hip loader will drop them if there isn't an obsolete to acquire it.

I'll fix the Write case when I update this for the new changes :>

Issue #902, opened 2020-12-15
7.2.0 fails to compile


svenstaro:
This is on Arch Linux, gcc 10.2.0, boost 1.75.0. I'm trying to compile 7.2.0 (and also tried master). I get: 
[build.txt](https://github.com/AcademySoftwareFoundation/openvdb/files/5693255/build.txt)

If I compile with clang 11.0.0 instead, I get tons of these:
```
../openvdb/openvdb/./TypeList.h:318:36: error: unknown type name 'size_t'; did you mean 'std::size_t'?
```

danrbailey:
Hi @svenstaro - thanks for reporting. We weren't aware that there were issues compiling OpenVDB on GCC 10. 

From the release notes (https://gcc.gnu.org/gcc-10/changes.html), it looks like this change is responsible:

> Reduced header dependencies, leading to faster compilation for some code.

I've fixed the issue by adding the missing header include and will submit a PR shortly to address this.

danrbailey:
This PR resolves this issue - #905.

svenstaro:
I confirm that this works fine. A patch release for this would be appreciated.

danrbailey:
> I confirm that this works fine. A patch release for this would be appreciated.

I've just published a 7.2.1 patch release which includes this header fix.

PR #905, opened 2020-12-17
Add new GCC10 and Clang10 CI builds and fix size_t issue

danrbailey:
This addresses #902 and uses the apt package manager with ubuntu 20.04 to verify GCC10 and Clang10 both of which are now part of the GitHub Actions environment.


kmuseth (approved):
looks good

PR #908, opened 2020-12-17
fixed copy-right

kmuseth:


danrbailey:
Hi @kmuseth - would you mind submitting a PR to OpenVDB to make this fix too? Also, I think the footer and the unit test header and footer need fixing too.

PR #912, opened 2020-12-17
Active Tile Count

danrbailey:
This is a new PR based on #788 (@Megidd), we're in the window to make changes to ABI so I'm bringing your contribution in.

I've made a few small changes here - I renamed the method Tree::activeTileNodeCount() instead of Tree::activeTileCount() mainly because I didn't want there to be any confusion with the variation that returns a value for the tile count in the whole tree. I admit the naming isn't perfect, but I thought it might be best to align it close to the existing nodeCount() method that does something very similar.


CHANGES
```diff
@@ -27,7 +27,8 @@ Version 8.0.0 - In development
       CoordBBox.
     - Added tools::activeTiles that returns a vector containing all the active
       tiles in a tree that intersects a given CoordBBox.
-
+    - Added Tree::activeTileNodeCount, which counts the number and type of
```
kmuseth, line 30: I really don't like this name change! This is nothing called a "TileNode", only a Tile and a Node and to make matters worse they are mutually exclusive so this new makes me nauseous :) If this method still returns the total number of active tiles is could remain "activeTileCount".

openvdb/openvdb/tree/InternalNode.h
```diff
@@ -1029,6 +1030,18 @@ InternalNode<ChildT, Log2Dim>::nodeCount(std::vector<Index32> &vec) const
     vec[ChildNodeType::LEVEL] += count;
 }
 
+template<typename ChildT, Index Log2Dim>
+inline void
+InternalNode<ChildT, Log2Dim>::activeTileNodeCount(std::vector<Index32>& vec) const
+{
+    assert(vec.size() > ChildNodeType::LEVEL);
+    const auto countTile = mValueMask.countOn();
+    const auto countChild = mChildMask.countOn();
```
kmuseth, line 1039: why compute countCount? If there are no child nodes "for (auto iter = this->cbeginChildOn(); iter; ++iter)" will terminate as fast as a call to  mChildMask.countOn() (because the iterator works on 64 words) and if there are any child nodes you just wasted time computing  countChild. If you really insist on checking the child bit mask (whish I would advise against since the iterator is super fast) then instead call mChildMask.isOff() which might perform an early-out

openvdb/openvdb/tree/LeafNode.h
```diff
@@ -131,6 +131,8 @@ class LeafNode
     static Index32 leafCount() { return 1; }
     /// no-op
     void nodeCount(std::vector<Index32> &) const {}
+    /// no-op
+    void activeTileNodeCount(std::vector<Index32>&) const {}
```
kmuseth, line 135: change to activeTileCount

openvdb/openvdb/tree/LeafNodeBool.h
```diff
@@ -111,6 +111,8 @@ class LeafNode<bool, Log2Dim>
     static Index32 leafCount() { return 1; }
     /// no-op
     void nodeCount(std::vector<Index32> &) const {}
+    /// no-op
+    void activeTileNodeCount(std::vector<Index32>&) const {}
```
kmuseth, line 115: change to activeTileCount

openvdb/openvdb/tree/LeafNodeMask.h
```diff
@@ -114,6 +114,8 @@ class LeafNode<ValueMask, Log2Dim>
     static Index32 leafCount() { return 1; }
     /// no-op
     void nodeCount(std::vector<Index32> &) const {}
+    /// no-op
+    void activeTileNodeCount(std::vector<Index32>&) const {}
```
kmuseth, line 118: change to activeTileCount

openvdb/openvdb/tree/RootNode.h
```diff
@@ -486,6 +486,7 @@ class RootNode
     Index64 offLeafVoxelCount() const;
     Index64 onTileCount() const;
     void nodeCount(std::vector<Index32> &vec) const;
+    void activeTileNodeCount(std::vector<Index32>& vec) const;
```
kmuseth, line 489: change to activeTileCount

openvdb/openvdb/tree/RootNode.h
```diff
@@ -1645,6 +1646,20 @@ RootNode<ChildT>::nodeCount(std::vector<Index32> &vec) const
     vec[ChildNodeType::LEVEL] = sum;
 }
 
+template<typename ChildT>
+inline void
+RootNode<ChildT>::activeTileNodeCount(std::vector<Index32>& vec) const
```
kmuseth, line 1651: change to activeTileCount

openvdb/openvdb/tree/Tree.h
```diff
@@ -113,6 +113,15 @@ class OPENVDB_API TreeBase
...
     virtual std::vector<Index32> nodeCount() const = 0;
+#endif
+#if OPENVDB_ABI_VERSION_NUMBER >= 8
+    /// Return in @a vec the active tile count for each level. The number of active tiles at level NodeType
+    /// is given as element NodeType::LEVEL in the vector. Thus, the size
+    /// of this vector corresponds to the height (or depth) of this tree.
+    /// @note While nodeCount() returns the number of allocated nodes at each level,
+    /// regions of the tree may have data but not have nodes because
+    /// they have activated constant regions.
+    virtual std::vector<Index32> activeTileNodeCount() const = 0;
```
kmuseth, line 124: change to activeTileCount

openvdb/openvdb/unittest/TestGrid.cc
```diff
@@ -87,6 +87,9 @@ class ProxyTree: public openvdb::TreeBase
 #if OPENVDB_ABI_VERSION_NUMBER >= 7
     std::vector<openvdb::Index32> nodeCount() const override
         { return std::vector<openvdb::Index32>(DEPTH, 0); }
+#endif
+#if OPENVDB_ABI_VERSION_NUMBER >= 8
+    std::vector<openvdb::Index32> activeTileNodeCount() const override { return std::vector<openvdb::Index32>(); }
```
kmuseth, line 92: change to activeTileCount

kmuseth (changes requested):
Please see my comments (rename method and get rid of coundNode). On a related note I don't understand why we're adding this method?! I've adding something very similar in FindActiveValues.h and and I was under the impression that we had agreed to remove methods on the grid nodes and replace them with (more performant multi-threaded) standalone tools. If so, than I would argue that this PR goes contrary to that decision/trend.

CHANGES
danrbailey: This method cannot remain `activeTileCount()` because there's already a method called `Tree::activeTileCount()` which returns the total number of active tiles in the whole tree. This method differs in that it returns the number of active tiles per-level of the tree as a vector (similar to `Tree::nodeCount`). You obviously cannot have two methods with the same name returning something different and we cannot change the return type of the existing method without breaking API with no warning.

danrbailey (commented):
> Please see my comments (rename method and get rid of coundNode). On a related note I don't understand why we're adding this method?! I've adding something very similar in FindActiveValues.h and I was under the impression that we had agreed to remove methods on the grid nodes and replace them with (more performant multi-threaded) standalone tools. If so, than I would argue that this PR goes contrary to that decision/trend.

Yes, that is the plan. Until we've migrated other counting-related functionality across like `Tree::nodeCount()`, I think it would feel confusing to introduce this one method as a new tool. There's probably minimal work in porting one more method that works in a very similar way to the others. Ultimately, I think if it provides value (which I believe this does), then we might as well add it to the tree now.

I see your FindActiveValues tool being different to this method because it (1) doesn't consider the nodes/tiles at different levels of the tree when counting, (2) requires you to supply a bounding box argument.

All that being said, I agree that this method doesn't fit neatly and I'm not convinced we'll find a suitable way to introduce this in time to release 8.0 so I think we'll have to drop this from making it into the next release.

PR #913, opened 2020-12-17
Drop minimum IlmBase/OpenEXR version back to 2.2

danrbailey:
There are minimal differences between IlmBase/OpenEXR 2.2 and 2.3 and as Houdini 18.0 adheres to VFX Reference Platform 2019 *except* for shipping with IlmBase 2.2 instead of 2.3, it's a common error when you try and build against Houdini 18. I don't particularly want to encourage people to disable all dependency version checks for this case so I think it's worth more liberally applying our deprecation policy in this case.


Idclip:
Maybe a small comment in the CMake to explain why these versions haven't changed? Or in the dependency docs (those versions also need updating)

Idclip (approved):

danrbailey:
> Maybe a small comment in the CMake to explain why these versions haven't changed? Or in the dependency docs (those versions also need updating)

Thanks, I put a quick comment in the dependency docs and updated them.

PR #914, opened 2020-12-18
Enable visit2 on const tree

tom-cnops:
There already was a const version of `Tree::visit2`, but this could not be called as it would end up at [RootNode.h#L3439-L3440](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/RootNode.h#L3439-L3440) with `RootNodeT` const qualified.

The modified test tries to call `visit2` on a const tree and it fails to compile.

The changes allow the  const version of `Tree::visit2` to be used, and also allow mixing a const and a non-const tree (the corresponding iterators passed to visitor will be const iterators or not).

### DIFF
--- a/openvdb/openvdb/tree/InternalNode.h
+++ b/openvdb/openvdb/tree/InternalNode.h
@@ -787,8 +787,7 @@ protected:
     template<typename NodeT, typename VisitorOp, typename ChildAllIterT>
     static inline void doVisit(NodeT&, VisitorOp&);
 
-    template<typename NodeT, typename OtherNodeT, typename VisitorOp,
-        typename ChildAllIterT, typename OtherChildAllIterT>
+    template<typename NodeT, typename OtherNodeT, typename VisitorOp>
     static inline void doVisit2Node(NodeT&, OtherNodeT&, VisitorOp&);
 
     template<typename NodeT, typename VisitorOp,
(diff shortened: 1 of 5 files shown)


danrbailey:
Hi Tom, thanks for submitting this. We're actually in the process of removing all the visitor methods on the Tree classes. I've relatively recently added a new DynamicNodeManager class and marked all the visitor methods as deprecated (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/Tree.h#L951). They're likely be removed in an 8.1.0 release sometime early next year. 

The reason for wanting to migrate away from these methods is that the visitor pattern is inherently single-threaded, it visits each node of the tree in turn. The existing NodeManager provides methods to operate on a const tree at each level of the tree in parallel (such as https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/NodeManager.h#L610). The one remaining area in which we were previously still relying on the visitor pattern was doing CSG operations that modified the tree during traversal. We've now introduced the DynamicNodeManager that allows for using a const or non-const tree. Like the NodeManager::foreachTopDown() it still navigates over the tree in parallel at each level of the tree. However unlike the NodeManager, it offers the ability to edit the tree in levels below the current level during traversal and provides filtering mechanisms to prune portions of the tree from the traversal algorithm. As a result, it's now the fastest way to traverse over a tree in parallel while modifying it and all our CSG operations have been migrated towards using this mechanism.

tom-cnops:
Makes sense, thanks for the explanation!

PR #924, opened 2021-01-05
Add new VDB Visualize As Points tool to Houdini

danrbailey:
Add a new VDB Visualize As Points tool that places a VDB Visualize Tree SOP with various parameters changed to visualize voxels as points with values.

This is done through HOM so that it shows up as a tab menu option.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Visualize.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Visualize.cc
@@ -26,6 +26,7 @@
 #include <GU/GU_PolyReduce.h>
 #include <GU/GU_Surfacer.h>
 #include <PRM/PRM_Parm.h>
+#include <PY/PY_CompiledCode.h>
 #include <UT/UT_Interrupt.h>
 #include <UT/UT_VectorTypes.h> // for UT_Vector3i
 #include <UT/UT_UniquePtr.h>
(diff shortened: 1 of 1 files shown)


openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Visualize.cc
```diff
@@ -380,6 +384,46 @@ of VDB volumes as well as to examine their extents and the values of individual
...
+node.parm("drawleafnodes").set(False);
+node.parm("drawtiles").set(False);
+node.parm("drawvoxels").set(True);
+node.parm("addvalue").set(True);
+node.parm("usegridname").set(True);
+'''
+
+if not hou.shelves.tool('vdbvisualizeaspoints'):
+ tool = hou.shelves.newTool(name='vdbvisualizeaspoints')
+ tool.setLabel('VDB Visualize As Points')
```
jmlait, line 410: VDB Visualize as Points
Title Case has the preposition as be lower case.

jmlait (approved):

danrbailey:
I'm going to close this PR. @jmlait - I never quite got this to work reliably, so we switched to use local shelf tools at ILM (there's now one for VDB Visualize as Points and VDB Visualize as Slice). It would be great if there was some mechanism in Houdini to be able to add new tab menu options like this from C++ without having to go through shelves and HOM to try and achieve this.

PR #926, opened 2021-01-07
Use a templated function to avoid invoking csg operations on bool

jmlait:
Use a templated function to avoid invoking csg operations on bool
grids.  These aren't supported on MSVC as the unary negation operation
is invalid for bool.


danrbailey (approved):

jmlait:
Trying out the pendingchanges directory, asking for a re-review based on that :>

danrbailey (approved):

PR #927, opened 2021-01-08
Inline Half

jmlait:
Copy the half implementation from openexr2.4. 

Inline the lookup tables rather than using separate files.
Converts OpenVDB-style export tags.
Namespace into math library.


danrbailey:
Looks good. I think I'd prefer putting the implementation into openvdb/3rdparty/half/half.h just so there's a clear distinction between our code and 3rd-party code for licensing and maintenance, then we could just have the alias in openvdb/Types.h (ie openvdb::math::half = openvdb::3rdparty::half).

Also would be nice to split the "Inline the OpenEXR 2.4 Half implementation." commit into two commits if possible - one that does the raw ingest and one that updates the namespaces so that we can more trivially apply the namespace patch/commit to a later version of half.h in the future without having to resolve conflicts.

Looking at the CI failures, I think we'll also need to wrap the #include with something that suppresses some of the additional warnings that was previously being provided by using -isystem.

jmlait:
> Looks good. I think I'd prefer putting the implementation into openvdb/3rdparty/half/half.h just so there's a clear distinction between our code and 3rd-party code for licensing and maintenance, then we could just have the alias in openvdb/Types.h (ie openvdb::math::half = openvdb::3rdparty::half).

Hmm... I really didn't want to get that deep into changing the CMake rules and versioning rules.  I actually also spent a while wondering if this should be naturalized as Half.h and Half to match the rest of our convention before backing off.  Fortunately, the math:: prefix makes it easy to disambiguate.

Putting the needs of licensing ahead of proper semantic organization seems backwards.  If half.h didn't exist, and we wrote it, we would have put it in openvdb::math.  So why send people through hoops just because it is ingested from an external source?  If it were a truly raw import, I'd be more happy with a 3rdparty directory.  Something like blosc or something, if we wanted to in-line a native version, I'd see doing that with.  But precisely because this has to be transformed I think it makes most sense as a native file.

I would, however, like to send this PR past ASWF before we approve it in any case to make sure we are dealing with licensing correctly.   But the whole reason files have their SPDX prefix in front of them is because it is already assumed each file is licensed individually.

> Also would be nice to split the "Inline the OpenEXR 2.4 Half implementation." commit into two commits if possible - one that does the raw ingest and one that updates the namespaces so that we can more trivially apply the namespace patch/commit to a later version of half.h in the future without having to resolve conflicts.

I'm not very convinced that is useful as this library evolves so slowly a manual patch is more than reasonable, and an automated patch will already have conflicts as I had to inline the lookup tables.

> Looking at the CI failures, I think we'll also need to wrap the #include with something that suppresses some of the additional warnings that was previously being provided by using -isystem.

Yeah, I need to look into the CI issues still.  I did manage to build on internal platforms and run on my machine :>

danrbailey:
I had a quick look at these CI failures and I think it might be related to not having updated the installed cmake module to remove references to half:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/cmake/FindOpenVDB.cmake#L451
https://github.com/AcademySoftwareFoundation/openvdb/blob/master/cmake/FindOpenVDB.cmake#L630

jmlait:
Ah, that would make sense!   I'll try and excise those.

danrbailey:
Hmm, doesn't look like that was very successful. By extracting Half from VDB, we're now in the situation where libopenvdb.so doesn't require Half, but libopenvdb_sesi.so still does. In that case, perhaps the solution is to now treat half like hboost (ie a Houdini-only dependency) and explicitly specify it in the cmake wrapper for the official Houdini cmake:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/cmake/OpenVDBHoudiniSetup.cmake#L201

I guess the question remains what will happen to libopenvdb_sesi.so in the future? Will it still link against Half or embed it like this open-source version?

I'm still speculating a bit here though, perhaps @Idclip has a better idea of how to resolve this?

Idclip:
We shouldn't be finding or linking in half in `OpenVDBHoudiniSetup.cmake` if we don't need it. The way the Houdini CMake works is that it builds a single "Houdini" target with `_houdini_create_libraries`. So, in a single run of CMake, you can't have 2x"Houdini" deps (one with half, one without). 

One solution would be to run the ABI tests in a separate run of CMake and expose the `_HOUDINI_EXTRA_LIBRARY_NAMES` as an option (i.e. `-D_HOUDINI_EXTRA_LIBRARY_NAMES=half.lib`). Another, and probably better, way would be to find it in `openvdb_houdini/abitest/CMakeLists.txt` like so:

```
# Find libopenvdb_sesi library
find_library(OPENVDB_SESI_LIB openvdb_sesi
  ${_FIND_HOUDINI_ADDITIONAL_OPTIONS}
  PATHS ${HOUDINI_LIB_DIR}
)

# Find sesi half library
find_library(SESI_HALF Half
  ${_FIND_HOUDINI_ADDITIONAL_OPTIONS}
  PATHS ${HOUDINI_LIB_DIR}
)

...

target_link_libraries(openvdb_houdini_abi_test PUBLIC
  ${OPENVDB_LIB}
  ${OPENVDB_SESI_LIB}
  ${SESI_HALF}
)
```

jmlait:
> I guess the question remains what will happen to libopenvdb_sesi.so in the future? Will it still link against Half or embed it like this open-source version?

Once libopenvdb.so stops linking to half, so will libopenvdb_sesi.so.

I think I now understand this failure though.  I think Nick's solution makes sense, and appears to at least link locally.  When Houdini catches up to this change and stops having Half, we can then remove this from the abitest.

jmlait:
Okay, feel like I'm getting there.  However, openvdb_render is still an issue it seems.  I thought USE_EXR was false though, judging from the code paths in the CMakeList that I used to disable the Half include.

openvdb_render.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) int __cdecl Imf_2_5::globalThreadCount(void)" (__imp_?globalThreadCount@Imf_2_5@@YAHXZ) referenced in function "void __cdecl `anonymous namespace'::saveEXR(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class openvdb::v8_0::tools::Film const &,struct A0x995b8c99::RenderOpts const &)" (?saveEXR@?A0x995b8c99@@YAXAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBVFilm@tools@v8_0@openvdb@@AEBURenderOpts@1@@Z) [D:\a\openvdb\openvdb\build\openvdb\openvdb\cmd\vdb_render.vcxproj]
266
openvdb_render.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) void __cdecl Imf_2_5::setGlobalThreadCount(int)" (__imp_?setGlobalThreadCount@Imf_2_5@@YAXH@Z) referenced in function "void __cdecl `anonymous namespace'::saveEXR(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &,class openvdb::v8_0::tools::Film const &,struct A0x995b8c99::RenderOpts const &)" (?saveEXR@?A0x995b8c99@@YAXAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBVFilm@tools@v8_0@openvdb@@AEBURenderOpts@1@@Z)

Idclip:
@jmlait This the code causing the issue (in PlatformConfig.h):

```
    // By default, assume that we're dynamically linking OpenEXR, unless
    // OPENVDB_OPENEXR_STATICLIB is defined.
    #if !defined(OPENVDB_OPENEXR_STATICLIB) && !defined(OPENEXR_DLL)
        #define OPENEXR_DLL
    #endif
```

Most likely the MD build on windows is linking dynamically against OpenEXR. In `openvdb/CMakeLists.txt` you have this code:

```
if(WIN32)
  # @note OPENVDB_OPENEXR_STATICLIB is old functionality from the makefiles
  #       used in PlatformConfig.h to configure EXR exports. Once this file
  #       is completely removed, this define can be too
  get_target_property(ILMBASE_LIB_TYPE IlmBase::Half TYPE)
  if(OPENEXR_USE_STATIC_LIBS OR (${ILMBASE_LIB_TYPE} STREQUAL STATIC_LIBRARY))
    list(APPEND OPENVDB_CORE_PUBLIC_DEFINES -DOPENVDB_OPENEXR_STATICLIB)
  endif()
```

The variable `OPENVDB_CORE_PUBLIC_DEFINES` propagates it's values to dependant targets - so if `OPENEXR_USE_STATIC_LIBS ` is `OFF` and `ILMBASE_LIB_TYPE` is not `STATIC`, the define `OPENVDB_OPENEXR_STATICLIB` will propagate to the build of `vdb_render`. Your change to wrap this in `if(USE_EXR)` means that this is now never propagated. 

We could previously do this here because we **always** needed Half. Now, because OpenEXR is purely optional for the core lib, it's possible this isn't hit but `vdb_render` still wants to link dynamically. So we need to duplicate this CMake in `openvdb/cmd/CMakeLists.txtx`. Around line 142, we should add:

```
if(WIN32)
  get_target_property(ILMBASE_LIB_TYPE IlmBase::Half TYPE)
  if(OPENEXR_USE_STATIC_LIBS OR (${ILMBASE_LIB_TYPE} STREQUAL STATIC_LIBRARY))
    target_compile_definitions(vdb_render  PUBLIC -DOPENVDB_OPENEXR_STATICLIB)
  endif()
endif()
```

We should probably also specifically add `IlmBase::Half` to the above lib list (line 135, same file)

jmlait:
Thank you, that certainly makes sense.... *crosses fingers*

jmlait:
Now it seems to be failing in an AX mac build?  But if I read the records properly, that passed before?  And the failure looks quite odd, as it is not finding tbb...  I'd not expect the CMakelist of cmd to affect that...  Is this a separate configuration issue?

Idclip:
Yer no idea why that failed, most likely something changes on the docker image or github actions env in between commits. Seems to be passing now!

jmlait:
Thank you all for your help resolving these build issues!

We still have the issues Dan raised to wrangle out, possibly at the next meeting.  Primarily I think is the location, where I'm reluctant to move this out of math to 3rdparty/Half.  The pain of getting this to work at all through our maze of CI hasn't really increased my desire to try adding a brand new directory :>

The two phase commit I think is a good idea, and wish I had done it that way in the first place, but not sure I want to now go back and re-do it all just for that effect :>

danrbailey:
Thanks Jeff! I don't think it's a big deal for us to re-apply your changes whenever we next update the half headers, it should be pretty obvious what to do here. 

As for moving this to a different location, let's just leave the namespace as openvdb::math and just physically move to a different directory and update the [header include](https://github.com/AcademySoftwareFoundation/openvdb/pull/927/files#diff-71749886d193d442668fac684bb66491997400172050c806c7435d60b01e407eR10) and [CMake path](https://github.com/AcademySoftwareFoundation/openvdb/pull/927/files#diff-9a25698ed115f3e617878780568fd7e09699c8e5b76daf0454ee190f72c995b3R255)?

The Mac failures can be ignored I think (the master branch CI failed recently too). They seem non-deterministic and I know there was some re-provisioning of the Mac hardware over the holidays so hopefully it just resolves itself. Retry until it succeeds!

Idclip:
I'm going to attempt to summarise the discussion we had in the TSC (though @danrbailey will have notes he can link). Currently, this PR provides a modified version of Half which has been altered in 3 main ways: namespaced, casting and lookup tables. These modifications by themselves are probably enough of a reason to warrant them not needing a 3rdparty folder. Additionally, because of the way this has been set up, you can **no longer use OpenEXR Half** with OpenVDB library methods; that is, you can still build against OpenEXR, but all OpenVDB methods will use this internal representation of Half. This side effect also justifies it not to appear in a 3rdparty directory.

However... location in the repository aside, is this what we want? Though we are fairly confident that no significant changes (API/ABI/behaviour) will be made to the half implementation prior to an introduction to the standard, we can not be certain. This change set means that from this point on, OpenVDB would not be able to use said changes until it itself was updated. Additionally, **any subsequent release of the library from this PR would be unable to leverage any future changes made to OpenEXR Half**. This does not pose too much of an issue should these changes be minor, but anything more substantial could cause problems. 

Note that these issues are specific to Half because the changes it has seen have historically been very small. I think that, were this PR focused on migrating blosc, the decision would be a lot clearer i.e. 3rdparty folder, compile time switch to use either use blosc source from VDB or from a previous installation. With this in mind, does it make sense for Half to behave the same way.

jmlait:
> Additionally, because of the way this has been set up, you can **no longer use OpenEXR Half** with OpenVDB library methods;

Thinking on it some more, I think this is very much the crux of the issue.  The other way of looking at it is that if you want to use OpenEXR Half you don't have to worry about conflicting with OpenVDB anymore.   If we make it an internal 3rd party file we *still*  have a dependency on a particular version of OpenEXR, we've just shipped OpenVDB with that dependency pre-installed.  To remove the dependency requires something like this that fully namespaces it away.

Idclip:
>  If we make it an internal 3rd party file we still have a dependency on a particular version of OpenEXR, we've just shipped OpenVDB with that dependency pre-installed.

Agreed. I think any introduction of Half here should be properly namespaced (not just a carbon copy) so it can be used alongside OpenEXR. Whether it lives in math/ or not I'm not particularly bothered about anymore - but should we allow users to optionally select the Half implementation from their existing OpenEXR installation. We would want to avoid another compile time define for downstream projects (i.e. I don't want users to have to provide -DOPENVDB_USE_INTERNAL_HALF or something). We could achieve this by updating all references to half coming from a single VDB header, say math/half_impl.h, and generate the contents of this header depending on the original CMake configuration. For example. the default could be to continue to use OpenEXR, where by this header is something like:

```
namespace openvdb { namespace math { 
#include <OpenEXR/half.h> 
} }
```
or otherwise, if using this internal representation, we would replace it with your header. I need to think about if there are better ways to structure this, but I get the impression that this support would reduce everyone's concerns?

jmlait:
I think we are already there with this setup?  The new Half is namespaced in this PR.  If someone wants to inject their own variant, they are free to do so by either replacing half.h (so as to gain the same namespace effect) or just updating Types.h to something like that line.  We only include half.h in one location.  But note there are linking changes, one change I made was for

class OPENVDB_API half

as we put the symbol visiblity on the class as a whole, rather than using HALF_EXPORT on each symbol.

Also note there is the halfLimits.h.   This has to, by design, inject into the std namespace so can't just be wrapped with a namespace + include.   And I think this line

class numeric_limits <openvdb::OPENVDB_VERSION_NAME::math::half>

requires knowledge of the full path.  (And I'd like validation I did that right...)

Built-in half support is going to be more necessary as I think we'll want a native half grid sooner than later.  At some point we should just inline leftpad to stop the tower of dependencies.

When/if someone has a use case where they need to use an external half implementation, they can send a PR with an approach to that effect.  But I really think it is more likely someone will want to replace it with a native half before they'll want to do that.

cary-ilm:
Dan and I talked offline. Dan explained that your ultimate goal is to reduce dependencies so that OpenVDB is less cumbersome to build. Here are some further considerations regarding the half type, we’re happy to discuss further.

The OpenEXR project is working towards a 3.0 release in which Imath is split into its own independent project:

https://github.com/AcademySoftwareFoundation/Imath

Along with ImathVec.h, ImathMatrix.h, etc, this also includes the half type. There are two goals behind this:

1. Make these utilities available  to projects that want to avoid a dependency on OpenEXR.  
2. Make OpenEXR less cumbersome for novices to build.

OpenEXR 3.0 will have an external dependency on Imath, which it will reference via a CMake ExternalProject, so building OpenEXR automatically downloads and builds Imath if it's not already installed, which ideally means it “just works.”  The Imath repo also includes the python bindings but they’re off by default, which further simplifies the build process for most users.

This isn’t released yet but the Imath master branch works, and the OpenEXR master branch CMake files build against it, so you can give it a spin and see if it works for you.

In the process, we’ve also made some modifications to Imath/half:

- Constexpr everywhere possible
- Noexcept everywhere possible
- CUDA support (via __host__ and __device__ qualifiers)
- Remove dependency on Iex

Technically, we’re not yet done with adding constexpr to the half type, but that will be complete shortly. We’re also planning to add intrinsic half-to-float conversion instructions that avoid the lookup table altogether, although we’ll likely need to retain the tables for the benefit of platforms that don’t support those instructions. The OpenEXR TSC has discussed the coming addition of 16-bit floating types to the C++ standard but we don’t yet have a firm strategy around it, except to note that OpenEXR is primarily a data representation, and the coming standard will presumably have compatible bit layout even if it differs in rounding behavior, etc.

So, after sitting dormant for more than a decade, this code is about to go through a flurry of minor but helpful improvements, after which it will likely remain stable for the foreseeable future.

Also note that OpenEXR 2.5 retired the compile-time lookup table header generation. The toFloat.h and eLut.h headers are regular include files now, which removes the error-prone generation step that existed in OpenEXR 2.4 and before. toFloat.h is 16,398 lines, big but not big enough to warrant jumping through hoops to generate it on the fly, as it was in the early days.

We’d like to do whatever we can to encourage adoption of Imath and half, especially by ASWF projects, so if there are other changes to the code or build process that would help your use case, we’d be happy to discuss it. Dan explained that you hope to simplify the build process by eliminating a dependency altogether. If you’re going to duplicate and copy the half code into OpenVDB, you should at least wait for the improvements we have in progress, but if you’re willing to settle on a simpler and less cumbersome external dependency, hopefully we can make it suit your needs.

kmuseth:
Thanks Cary, we’ll discuss this further during our next TSC meeting. In the future we also need to support half in NanoVDB which implies hardware support for fp16  (e.g offered though cuda header files). This is just to say that it’s a complex issue :)

Cheers,
Ken

jmlait:
Okay, this sucks.   The problem is every CMakeList has to include the build directory to see the BuildConfig.h artifact.   Which is very annoying as it treats this header different from the other headers.

But any actual external user will see them as the same thing; as the install rule will have put it in the right place.   I think it is a sign we are not properly building the parallel directories.   The openvdb_ax directory should be able to build against any openvdb, not just the one in the same library.  Thus it would make sense if the build rule for openvdb_core always "installed" itself into build, and then the other parallel libraries like openvdb_ax (or openvdb/cmd!) would use the installed openvdb's includes, *not* the ones in the source directory...

But that seems like a lot as well...  I'll see if I can just add the relevant include paths for now.

jmlait:
Good news is that the add_target_include does a better job of propagating the built header's directory location, so my long rant earlier may be misdirected.  Latest commit rips out the hacking includes and hopefully will go through our CI.

Idclip:
This looks great, thanks for taking the time to look into the build config stuff. I'll try and test this week

Idclip:
@jmlait, based on the discussion yesterday, are you happy to remove the EXR support from the ray tracer in this PR? I have a few CMake tweaks to make here, but will do that after if

jmlait:
Yes, I am willing to remove that support.

Idclip:
Thanks @jmlait - I'd like to contribute some CMake improvements, any objections if I commit to this branch?

jmlait:
Please do commit to this branch!

The change Dan wanted was the CMakeList variable
OPENVDB_BUILDCONFIG_USE_EXR
which we use to trigger Half in the BuildConfig and in the Types.h to not be USE_EXR, but be something like IMATH_HALF.  So maybe
OPENVDB_BUILDCONFIG_USE_IMATH_HALF
But I'll let you fold that into your cmake changes rather than delaying anything.

Idclip:
> Please do commit to this branch!

Hmm I can't seem to push to this, most likely it's due to permission settings on the sidefx fork. @jmlait how would you like to proceed? Could you change these settings? I can send you the modifications. Or we can merge this (perhaps into a temporary branch) and I'll push up a new PR. My changes are minor but solve a few remaining issues.

jmlait:
Oh... that is interesting....  It seems org-owned forks can't enable maintainer edits???   The button I'm supposed to press isn't there, and searching leads to stuff as recent as 2020 complaining the correct workflow is to create a user-fork?

So I can't change the settings.  I guess you can send me a .patch like bad old days?  or do you create a fork of the PR somehow?

danrbailey:
> Oh... that is interesting.... It seems org-owned forks can't enable maintainer edits??? The button I'm supposed to press isn't there, and searching leads to stuff as recent as 2020 complaining the correct workflow is to create a user-fork?
> 
> So I can't change the settings. I guess you can send me a .patch like bad old days? or do you create a fork of the PR somehow?

I'd suggest creating a feature branch in the repo for this and then modifying the target of this PR to point at the temporary branch then just merge it and @idclip or you can make further PRs into this branch until it's ready to merge to master (ie same as the feature/nanovdb or feature/ax workflows).

jmlait:
Okay, it is done...  We now have a feature/inlinehalf branch which we can all target.   Kind of annoying that it means this PR ends up "closed" and further edits are separate PRs.  Feels like we are being forced up a layer of granularity for no good reason.

PR #934, opened 2021-01-18
Initial attempt at fixing the root-level CSG being wrong.  

jmlait:
Correctly stops the importing of nodes missing on the primary target, but doesn't
seem to properly erase root tiles when the secondary merges are missing
roots.

### DIFF
--- a/openvdb/openvdb/tools/Merge.h
+++ b/openvdb/openvdb/tools/Merge.h
@@ -533,7 +533,7 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
     };
 
     std::unordered_map<Coord, /*flags*/uint8_t> tiles;
-    std::unordered_map<Coord, /*tilecount*/int> tilecount;
+    std::unordered_map<Coord, /*tilecount*/size_t> tilecount;
 
     if (root.getTableSize() > 0) {
         for (auto valueIter = root.cbeginValueAll(); valueIter; ++valueIter) {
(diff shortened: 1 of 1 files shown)


openvdb/openvdb/tools/Merge.h
```diff
@@ -603,6 +645,9 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
             auto it = tiles.find(key);
             if (it != tiles.end() && it->second == INSIDE_STATE)     continue;
 
+            // Missing tiles are outside, so if we are intersection we skip
+            if (Intersect)  continue;
+
```
jmlait, line 650: This line successfully causes one direction of the intersection to work.  When the dest has fewer roots than a source, they aren't erroneously added.  But if you reverse the order the problem returns.

jmlait:
[strange_isect.zip](https://github.com/AcademySoftwareFoundation/openvdb/files/5832444/strange_isect.zip)

Here's a pair of VDBs that fail to do intersection as expected.  Reversing the order can switch if the missing root tiles are in the A or B path.

danrbailey:
Thanks for tracking this down Jeff and for the test case. This seems like the correct thing to do to me - I'll try and turn this into a unit test and follow the logic through in a bit more detail to make sure I've fully understand it. This kind of points out the flaw in my thinking that I could combine Union and Intersection into a single CSG operator. Maybe it now makes sense to split them back out...

jmlait:
Yeah, to clarify, only half of this works.  And I don't know why the other half doesn't.  I thought from the surrounding code that calling addTile() on the root node would erase tiles in the A section, but like the proverbial cat (https://www.nfb.ca/film/the-cat-came-back/), they seem to come back the next day.  I'm curious if maybe the dynamic node manager has already pre-cached them?

jmlait:
Closed because PR#936 does this right.

PR #935, opened 2021-01-19
Drop Boost requirements back to 1.61 due to H18

Idclip:

My main concern here is our CI no longer tests against Boost 1.61 - even against H18 we use the VFX19 platform which uses Boost 1.66.


.github/workflows/build.yml
```diff
@@ -66,10 +66,10 @@ jobs:
         config:
           # disable dependency version checks for H18.0 as it ships with OpenEXR 2.2,
           # not OpenEXR 2.3 as stated in the VFX Reference Platform 2019
```
Idclip, line 68: need to remove this comment

jmlait:
I'm not sure about this.  The version of boost Houdini uses internally should be irrelevant - this is why Houdini renames all the symbols as hboost.  So 18.0 *is* VFX 2019.

Houdini . is fully compliant with the VFX Reference Platform CY2019 specification (http://www.vfxplatform.com).

Under boost, there is the note:

Yes
The Boost libraries are renamed with symbols namespaced

The fact the "hboost" named symbols may be from 1.61 should no affect on what version of Boost one builds OpenVDB with.  Any more than if Houdini didn't contain any boost at all...

The same is true for OpenEXR:

Yes
The OpenEXR libraries are renamed with symbols namespaced

Idclip:
@jmlait historically we have used the deps bundled with Houdini, is that not true anymore? Do we not still use the version of OpenEXR that comes with Houdini or at least want to have that as an option? 8.0.0 won't build using the versions Houdini comes with without disabling this check. If we should no longer be using these dependencies then yes, we can bump these to the correct VFX2019 versions and remove the documentation which tells users that they can use the 3rdparty deps that come with Houdini (and instead have to install them themselves).

jmlait:
But haven't the dependencies bundled with houdini been renamed to hboost?  There is no libboost to link with for Houdini, only libhboost which will have different symbols....

er... But I see...  The openvdb_houdini source all has been set to use hboost, as they've been synced with what we use locally.  In particular, when we split from main-line boost to hboost; we also swept in the openvdb_houdini source to use hboost in the newer build targets of Houdini.

When we build libopenvdb_sesi we do a pre-process over all of the OpenVDB core source to replace boost with hboost, so generate a openvdb core that is built specifically against Houdini's version of boost.

This does mean, though, that if boost & hboost are not ABI compliant, we're going to run into trouble.  Removing hboost from openvdb_houdini won't solve the issue either, I think, because native nodes will be constructing VDBs with hboost and trying to pass them off with void * casts to boost.  In practice you may be safe if the data layout is the same, but  yuck.

There is a separate question though, of why we have to increase the min spec for boost & openexr in openvdb for 8.0?  VFX platform requires we support the VFX boost & openexr, but does not say we have to not support earlier versions if people build against them.

danrbailey (approved):

Idclip:
As discussed in TSC (02/02/2021), best thing to do here is to drop the min requirement to 1.61 until support for H18 is removed. It was noted that the CI currently doesn't build against the version of boost shipped with Houdini, but this PR can be merged for the next patch release and we should aim to update the CI soon

PR #936, opened 2021-01-23
Fix CSG intersection

danrbailey:
This isn't for final review as it needs a bit more work, but it resolves the issue and demonstrates that it doesn't require changes to the DynamicNodeManager.

As you identified @jmlait, the issue is that for the root node in CSG intersection mode, if any trees being merged don't have a particular key, that element needs to be deleted from the resulting tree, meaning unlike CSG union and CSG difference, _absence_ of a root node child takes precedence over outside/inside tiles or child nodes. This doesn't apply to InternalNodes in CSG intersection mode as every element must be either a tile or a child.

I took a slightly different approach to yours here and first detect any keys that need deleting from the target tree, then delete them. Adding tiles or children in CSG intersection mode is then only permissible if they already exist in the root node. The root node has a pretty strict API so it's a bit of convoluted logic to even delete root node children - I found the easiest solution was to convert any elements I wanted to delete into background tiles and then call root.eraseBackgroundTiles(). There's definitely an argument for extending the RootNode API to make this kind of operation easier, but I don't think that should hold up this fix necessarily.

My next step is to modify the unit tests to run through the original CSG visitors to make sure I haven't baked in any other incorrect assumptions (as a test only - Merge.h should remain upstream of Composite.h). I also need a few additional unit tests to test the dropping of tiles and children in CSG Intersection mode.

### DIFF
--- a/openvdb/openvdb/tools/Merge.h
+++ b/openvdb/openvdb/tools/Merge.h
@@ -525,6 +525,9 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
         return true;
     };
 
+    // delete any background tiles
+    root.eraseBackgroundTiles();
+
     // for intersection, delete any root node keys that are not present in all trees
     if (Intersect) {
         // find all tile coordinates to delete
@@ -538,25 +541,10 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
             const Coord& key = childIter.getCoord();
             if (!keyExistsInAllTrees(key))   toDelete.push_back(key);
         }
-        // mark any background tiles as active to prevent them from being deleted
-        std::vector<Coord> toPreserve;
-        for (auto valueIter = root.beginValueAll(); valueIter; ++valueIter) {
-            const Coord& key = valueIter.getCoord();
-            // background tiles are always inactive
-            if (valueIter.isValueOff() &&
-                math::isApproxEqual(*valueIter, root.background())) {
-                toPreserve.push_back(key);
-                valueIter.setValueOn();
-            }
-        }
         // only mechanism to delete elements in root node is to delete background tiles,
         // so insert background tiles (which will replace any child nodes) and then delete
         for (Coord& key : toDelete)     root.addTile(key, *mBackground, false);
         root.eraseBackgroundTiles();
-        // now re-insert the background tiles
-        for (Coord& key : toPreserve) {
-            root.addTile(key, *mBackground, false);
-        }
     }
 
     // find all tile values in this root and track inside/outside and active state
@@ -663,6 +651,7 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
             if (it != tiles.end() && it->second == INSIDE_STATE)     continue;
 
             auto childPtr = mergeTree.template stealOrDeepCopyNode<typename RootT::ChildNodeType>(key);
+            childPtr->resetBackground(mergeRoot->background(), root.background());
             if (childPtr)   root.addChild(childPtr.release());
 
             children.insert(key);
@@ -684,6 +673,9 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(RootT& root, size_t) con
             }
         }
     }
+
+    // finish by removing any background tiles
+    root.eraseBackgroundTiles();
 
     return continueRecurse;
 }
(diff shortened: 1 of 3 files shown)


jmlait:
This works both ways for my test case.

In line 630/732 (sadly I can't comment on lines not touched in github!) we have

 if (!mergeNode)     continue;

This read to me suspicously like a similar potential problem.  Where does this flag trigger, and why doesn't it have the same implications?

danrbailey:
> This works both ways for my test case.
> 
> In line 630/732 (sadly I can't comment on lines not touched in github!) we have
> 
> if (!mergeNode) continue;
> 
> This read to me suspicously like a similar potential problem. Where does this flag trigger, and why doesn't it have the same implications?

I understand your concern. 

Firstly, I'll just highlight that the DynamicNodeManager only caches at the level it is about to execute which you can see through the interleaving of the "init" methods and the "foreach" methods here:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/NodeManager.h#L1592

so any manipulation of a target node directly impacts how the subsequent source children are merged into the target children. For example, replacing a target node with a target tile eliminates any more recursion down this branch _before_ caching happens. 

Having made the change in this PR, if you are evaluating a node _below_ a root node (ie InternalNode1, InternalNode2, LeafNode) then all of your source merge trees _must_ contain an element in the root node hash table that has the same key as the one that this node is rooted under. That means that if a source merge tree has no node at this level of the tree, it is because one of it's ancestors is a tile. With CSG intersection, if that case arose, the tile could _only_ be an inside tile because an outside tile would have taken precedence and forced termination before recursion reached this level of the tree. A child node takes precedence over an inside tile so it can be safely skipped.

danrbailey:
Just as a note to this ongoing work - a "valid" levelset is expected to have _only_ inactive tiles and for those tiles to have either positive or negative background values. I've found that the visitor-based CSG intersection method doesn't extrapolate in a predictable way to "invalid" levelsets. This means that if you provide any "invalid" levelsets as the input, the resulting levelset is not always a "valid" levelset itself and that the CSG intersection method isn't _always_ commutative. I'm treating this as "undefined behavior" and only strictly matching the visitor-based implementation for valid inputs, so really I'm just highlighting that there may be differences in the result of these new methods compared to the visitor-based CSG methods if the inputs are not valid levelsets.

jmlait (approved):

danrbailey:
I've extended and improved the unit tests and validated against the original composite methods now. This highlighted a couple of additional differences between the implementations that I've now addressed.

There is one change in behavior that I've intentionally made, which is to make the root node CSG intersection and CSG union commutative in one specific case where it was not previously as I believe this is more correct. This is demonstrated by the following:

```cpp
FloatGrid::Ptr grid1 = createLevelSet<FloatGrid>();
grid1->tree().root().addTile(Coord(0, 0, 0), 10.0f, false);
FloatGrid::Ptr grid2 = createLevelSet<FloatGrid>();
grid2->tree().touchLeaf(Coord(0, 0, 0));
tools::csgIntersection(grid1->tree(), grid2->tree());
```

In this case, the resulting grid has one inactive tile with value 10.0. However, if you reverse the order of the inputs, the tile value is mapped to the background value on merge and then gets removed from the grid (as do all background tiles at the root node level). With the change in behavior, the resulting grid from the above routine now also re-maps the tile value to the background value and removes it from the grid so the result is identical irrelevant of the order of inputs. This remapping of non-background values is only performed at the root node so there is no change in behavior in the rest of the tree.

openvdb/openvdb/tools/Merge.h
```diff
@@ -501,11 +501,52 @@ struct UnallocatedBuffer<BufferT, bool>
...
+        }
+        // find all child coordinates to delete
+        for (auto childIter = root.cbeginChildOn(); childIter; ++childIter) {
+            const Coord& key = childIter.getCoord();
+            if (!keyExistsInAllTrees(key))   toDelete.push_back(key);
+        }
+        // only mechanism to delete elements in root node is to delete background tiles,
+        // so insert background tiles (which will replace any child nodes) and then delete
+        for (Coord& key : toDelete)     root.addTile(key, *mBackground, false);
+        root.eraseBackgroundTiles();
```
Idclip, line 547: This is also called at the end of this function - do you need to call it twice?

Idclip:
Approving as this now passes @jmlait's tests

Idclip (approved):

openvdb/openvdb/tools/Merge.h
danrbailey: Yes, later portions of this routine check if the key exists in the target tree and then skips any merges if not. However, where the key is valid, it is still possible to merge in a background tile (as this method isn't first called on the source trees), so it needs to be called at the end again to ensure that they all finally get removed.

Typically these root node hash tables are very small, so this method is not expensive.

PR #944, opened 2021-01-31
Build Epydoc documentation through CI

danrbailey:
* Add Epydoc to CI documentation (and python module)
* Build Doxygen from source to ensure the version matches 1.8.11 used for releases
* Fix a couple of doxygen issues

The next step following this PR is to append a GitHub action to upload the documentation so as to simplify the release process.


.github/workflows/docs.yml
```diff
@@ -20,9 +20,26 @@ on:
...
+      run: pip install epydoc
+    - name: install_latex
+      run: yum -y install texlive-latex-bin texlive-dvips
     - name: build
-      run: ./ci/build.sh Release None OFF None "doc" -DCMAKE_INSTALL_PREFIX=$HOME
+      run: ./ci/build.sh Release None OFF None "core,python,doc" -DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON -DDISABLE_DEPENDENCY_VERSION_CHECKS=ON
+    - name: mkdir
+      run: mkdir -p $HOME/docs
+    - name: install
+      run: cp -r /usr/local/doc/html/* $HOME/docs
```
Idclip, line 40: is this dir guaranteed to always only contain the vdb installed docs?

doc/CMakeLists.txt
```diff
@@ -109,7 +109,7 @@ set(DOXYGEN_ALIASES
   )
 
 set(DOXYGEN_PREDEFINED
-  "OPENVDB_VERSION_NAME=v8_2"
+  "OPENVDB_VERSION_NAME=v8_0"
```
Idclip, line 112: oops, what happened here

Idclip (approved):

.github/workflows/docs.yml
danrbailey: Yes - for now this just gets thrown away when the job finishes, but the next step is to try and publish this somewhere so we don't have to ever generate documentation offline.

doc/CMakeLists.txt
danrbailey: Looks like I bumped only bumped the major version:

https://github.com/AcademySoftwareFoundation/openvdb/pull/898/files#diff-652f011e4f45934b3684d4596670e3847750ec3289b5092d4f8847b1ca6a75edR112

There are warnings that Doxygen spits out about the links being broken, but unfortunately they don't get turned into errors so they were silently ignored. I looked this up and there's a FAIL_ON_WARNINGS flag we can add to Doxygen, but it's only recently been introduced and I think is released or is going to be released as 1.9.0 (https://github.com/doxygen/doxygen/issues/8023). We could attempt to upgrade or maybe error through parsing the log, but I didn't go further than this - I think getting the release process automated is probably more important.

.github/workflows/docs.yml
Idclip: sure, I was just concerned that another dep or pkg might install stuff here that we'd end up copying

danrbailey: It doesn't seem to be the case right now, but I can rename this to make this location more likely to be unique if you're concerned.

Idclip: I think it's actually `/usr/share/doc` I'm thinking of (maybe we should be installing there!) but a unique location would be good

danrbailey: Ah, I thought you were talking about the target ($HOME/docs). I see what you mean, other packages could feasibly have installed something inside an html directory here. I'll look to change the install prefix.

Idclip: For future consideration:  https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html has been updated to include `CMAKE_INSTALL_DOCDIR` which evals to `share/doc/OpenVDB` on my system

danrbailey: ok, thanks - we could use this in the future, for now I just changed the install location

danrbailey:
This now builds the docs into the share/doc/OpenVDB location. I've also added a deployment step which only gets run if you launch this workflow manually and will copy the documentation into a new gh-pages branch that deploys it to here:

https://academysoftwarefoundation.github.io/openvdb/

This page is empty currently as it needs to be submitted to master for the workflow dispatch option to appear, but I tested on my fork (https://danrbailey.github.io/openvdb/). Note that you can choose which branch to run the action on and it will build and deploy the documentation from that branch.

Idclip:
> I've also added a deployment step which only gets run if you launch this workflow manually and will copy the documentation into a new gh-pages 

This is great, but how does this tie in with our current system where we don't use github pages?

danrbailey:
> > I've also added a deployment step which only gets run if you launch this workflow manually and will copy the documentation into a new gh-pages
> 
> This is great, but how does this tie in with our current system where we don't use github pages?

I propose a two-step migration - (1) we publish the docs to gh-pages, (2) we update the openvdb-website repo to change the URL of the online documentation to point to this gh-pages location.

There's a lot more that we can do here to improve things in the future, but for now this greatly simplifies one of the most time-consuming steps in doing a release.

PR #950, opened 2021-02-02
Updated CODEOWNERS

Idclip:


danrbailey:
Seeing as you're changing this - can we also update this to remove Peter and add Andre?

Idclip:
> Seeing as you're changing this - can we also update this to remove Peter and add Andre?

I believe this is better done in a single PR with everyone's approval

danrbailey (approved):

PR #952, opened 2021-02-03
VDB AX SOP layout + CI

Idclip:
Summary of changes:
  - Moved the AX SOP and header util file to live alongside all other SOPs
  - Exposed all Houdini installation paths as cache variables so the full installation can be configured via CMake arguments.
  - Added per SOP options to disable/enable SOPs in the format `OPENVDB_BUILD_<SOP_NAME>`. If a variable of this name is provided, it can be set to ON/OFF to manually toggle SOPs. By default all SOPs are ON, so it only makes sense to set this to OFF for most SOPs (except for the AX SOP)
  - The AX SOP now builds by default if both the AX lib and the Houdini plugin are ON. 
    - If the AX lib is off the SOP is off by default.
    - If **only** the Houdini plugin is being built, you can define `OPENVDB_BUILD_SOP_OpenVDB_AX` to ON to find the openvdb_ax lib and build the SOP


.github/workflows/build.yml
```diff
@@ -54,43 +58,6 @@ jobs:
...
-    - name: validate_houdini
-      if: github.repository_owner == 'AcademySoftwareFoundation'
-      run: test -f "hou/hou.tar.gz"
-    - name: houdini
-      run: ./ci/install_houdini.sh
-    - name: build
-      run: ./ci/build_houdini.sh ${{ matrix.config.build }} ${{ matrix.config.extras }} -DDISABLE_DEPENDENCY_VERSION_CHECKS=${{ matrix.config.disable_checks }}
-    - name: test
-      run: ./ci/test.sh
-
```
Idclip, line 93: This has moved, again, due to the need to change the path-ignore rules

jmlait (approved):
I like the idea of allowing the openvdb_houdini to contain both AX and Core SOPs.

Idclip:
@danrbailey I don't think what we discussed yesterday is possible, or at least I don't think it's worth it - to get it to work I'll need to add the build triggers to every config in the `build.yml` file. I'm still of the belief that the separation is actually better, but maybe you could take a look and see if there's a better way to structure it?

.github/workflows/build.yml
```diff
@@ -7,19 +7,23 @@ on:
...
     paths-ignore:
       - 'CHANGES'
-      - 'openvdb/doc/**'
+      - 'doc/**'
       - 'openvdb_maya/**'
+      - 'openvdb_houdini/**'
       - 'openvdb_ax/**'
-      - '**/*.md'
+      - '**.md'
+      - '**.txt'
```
danrbailey, line 15: Won't this miss changes to CMakeLists.txt?

danrbailey (changes requested):
Could you extract all your CI changes to build.yml, docs.yml and houdini.yml to a new PR so we can discuss separately please? Unless I'm mistaken, I don't believe there are any changes related to the AX content of this PR in the changes for these files and I don't want our discussion on this to hold up being able to merge this PR. Plus it's always nice to make these PRs small and targeted in case there's any need for reverting down the line and to avoid potential conflicts.

Idclip:
> Could you extract all your CI changes to build.yml, docs.yml and houdini.yml to a new PR so we can discuss separately please? Unless I'm mistaken, I don't believe there are any changes related to the AX content of this PR in the changes for these files and I don't want our discussion on this to hold up being able to merge this PR. Plus it's always nice to make these PRs small and targeted in case there's any need for reverting down the line and to avoid potential conflicts.

This is disappointing, I structured this PR very specifically to include the CI and CMake changes and see no reason to split this out further. It makes sense to include these together. I already split out the UI changes so have been submitting smaller PRs. This is "small and targeted" .

The reason `houdini.yml` exists is due to the 'openvdb_houdini/**'` build trigger not requiring the core lib builds to run with changes there. With the introduction of the AX SOP to this folder, changes there would previous invoke all builds in build.yml. This is obviously redundant....  I think this improvement is far more significant to the fact that you might need to click twice in github.

.github/workflows/build.yml
Idclip: Yes, will fix this

Idclip:
The only other thing I can suggest, should this really be a problem, is to attempt to share the build trigger settings with [anchors and aliases](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd) i.e.
```
&core_build_triggers
  on:
    push:
      branches:
        - '**'
      paths-ignore:
        - 'CHANGES'
        - ....


jobs:
  build1:
    - on: *core_build_triggers
```

I did this successfully in travis a while ago but am no sure if this will work and don't really have a desire to do this here.

danrbailey:
> > Could you extract all your CI changes to build.yml, docs.yml and houdini.yml to a new PR so we can discuss separately please? Unless I'm mistaken, I don't believe there are any changes related to the AX content of this PR in the changes for these files and I don't want our discussion on this to hold up being able to merge this PR. Plus it's always nice to make these PRs small and targeted in case there's any need for reverting down the line and to avoid potential conflicts.
> 
> This is disappointing, I structured this PR very specifically to include the CI and CMake changes and see no reason to split this out further. It makes sense to include these together. I already split out the UI changes so have been submitting smaller PRs. This is "small and targeted" .

I don't agree with this. The only change you _need_ to make here is removing "openvdb_ax" from the exclusion list in build.yml, everything else is a separate optimization that could have been discussed in a subsequent pass. There's no overlap between this optimization and any other changes to the CMake as far as I can tell.

> The reason `houdini.yml` exists is due to the 'openvdb_houdini/**'` build trigger not requiring the core lib builds to run with changes there. With the introduction of the AX SOP to this folder, changes there would previous invoke all builds in build.yml. This is obviously redundant.... I think this improvement is far more significant to the fact that you might need to click twice in github.

My rationale is not just about navigating between workflow pages on GitHub Actions which is annoying. Your change splits the main build logic between two files, creates more actions, more scheduled actions and an additional workflow complicating the CI further and making navigation around the actions tab slower. 

Having said all that, you're right that there are different triggers required here and your change will reduce the overall computation. There are other ways to reduce computation such as re-introducing ccache, but they're admittedly more time-intensive to implement and GitHub Actions doesn't support anchors as far as I'm aware. If you're happy to support reducing the number of build workflows again should GitHub Actions introduce functionality to do so, I'm happy to support this (with a minor change or two)?

danrbailey:
> I like the idea of allowing the openvdb_houdini to contain both AX and Core SOPs.

Yes, I like this too.

.github/workflows/houdini.yml
```diff
@@ -0,0 +1,63 @@
+
+name: Houdini
```
danrbailey, line 2: This has the same name as the workflow to download and cache Houdini I believe. Can we change the name of workflow in houdini_cache_update.yml so they're at least distinct in the UI?

Idclip:
> I don't agree with this. The only change you need to make here is removing "openvdb_ax" from the exclusion list in build.yml, everything else is a separate optimization that could have been discussed in a subsequent pass. There's no overlap between this optimization and any other changes to the CMake as far as I can tell.

This is not an "optimization". The CI is broken as it is. Changes to the AX SOP should not rebuild the entire stack of tests in build.yml. It is wrong and should not be part of anything that is merged with this PR or any CI which is introduced for the SOP component. I am open to other ways to structure the CI but I do not want this merged with that incorrect behaviour.

> Your change splits the main build logic between two files, creates more actions, more scheduled actions and an additional workflow complicating the CI further and making navigation around the actions tab slower.

It does not create more actions - the action count remains exactly the same. It absolutely does not complicate the CI, in fact is simplifies it greatly. The build in houdini.yml are *completely* independent and have nothing to do with the builds in build.yml. The only thing I agree here is that navigation across the tabs is every so slightly slower. But even this is such a huge stretch. I'm beginning to think I've missed something significant with the CI, as I'm not sure how I could solve any of these concerns.

> There are other ways to reduce computation such as re-introducing ccache, but they're admittedly more time-intensive 

Perhaps this is where our mindsets differ. This is *not* an optimisation. Changes to the SOP should not trigger the aforementioned builds. This is wrong and is different to introducing something like ccache.

danrbailey:
> > I don't agree with this. The only change you need to make here is removing "openvdb_ax" from the exclusion list in build.yml, everything else is a separate optimization that could have been discussed in a subsequent pass. There's no overlap between this optimization and any other changes to the CMake as far as I can tell.
> 
> This is not an "optimization". The CI is broken as it is. Changes to the AX SOP should not rebuild the entire stack of tests in build.yml. It is wrong and should not be part of anything that is merged with this PR or any CI which is introduced for the SOP component. I am open to other ways to structure the CI but I do not want this merged with that incorrect behaviour.
> 
> > Your change splits the main build logic between two files, creates more actions, more scheduled actions and an additional workflow complicating the CI further and making navigation around the actions tab slower.
> 
> It does not create more actions - the action count remains exactly the same. It absolutely does not complicate the CI, in fact is simplifies it greatly. The build in houdini.yml are _completely_ independent and have nothing to do with the builds in build.yml. The only thing I agree here is that navigation across the tabs is every so slightly slower. But even this is such a huge stretch. I'm beginning to think I've missed something significant with the CI, as I'm not sure how I could solve any of these concerns.
> 
> > There are other ways to reduce computation such as re-introducing ccache, but they're admittedly more time-intensive
> 
> Perhaps this is where our mindsets differ. This is _not_ an optimisation. Changes to the SOP should not trigger the aforementioned builds. This is wrong and is different to introducing something like ccache.

I think we're firmly in bikeshedding territory here, so if it wasn't clear from my earlier comment - I'm backing down, let's do it your way, I accept the compromise. Fix the houdini workflow name and let's move forward. :)

Without trying to inflame the discussion further, I do think you're presenting this in a more black-and-white light when the reality is more nuanced. There's nothing "wrong" with overbuilding - we do this _all_ the time. If you make a change to the VDB Advect SOP, the build should _not_ build the VDB Activate SOP, but it does. We're frequently rebuilding _lots_ of stuff we don't need to, so then it merely comes down to a matter of granularity with your argument being that if the granularity exists in the build system, we should take advantage of it to reduce redundant compilation. I accept that, but I don't think that the alternative solution equates to the CI being incorrect or broken or that it's fundamentally any different to reducing redundant compilation through ccache. Broadly speaking, if an error in a PR results in the PR checks failing and fixing that error results in them passing, then I would say that the CI is correct and doing it's job regardless of how it's implemented or presented to the user.

Idclip:
>  There's nothing "wrong" with overbuilding - we do this all the time. If you make a change to the VDB Advect SOP, the build should not build the VDB Activate SOP, but it does.

I appreciate you accepting this change, but we will have to agree to disagree with this. What's frustrating is that during the integration of AX I had to strip out a bunch of valid CI tests due to concerns about CI turnaround times, CI resources and job slots (see all comments on this PR https://github.com/AcademySoftwareFoundation/openvdb/pull/818#issuecomment-687399535) - yet you're now saying that overbuilding is fundamentally the same as adding ccache. I tried to address this problem then without success, so it's important to have this discussion.

danrbailey:
> > There's nothing "wrong" with overbuilding - we do this all the time. If you make a change to the VDB Advect SOP, the build should not build the VDB Activate SOP, but it does.
> 
> I appreciate you accepting this change, but we will have to agree to disagree with this. What's frustrating is that during the integration of AX I had to strip out a bunch of valid CI tests due to concerns about CI turnaround times, CI resources and job slots (see all comments on this PR [#818 (comment)](https://github.com/AcademySoftwareFoundation/openvdb/pull/818#issuecomment-687399535)) - yet you're now saying that overbuilding is fundamentally the same as adding ccache. I tried to address this problem then without success, so it's important to have this discussion.

I think there was a short discussion to be had about the trade offs in optimizing paths-ignore vs the number of separate workflows in which we would likely have ended up in the same place as this. What I think we disagreed on is my proposal of merging this AX PR and then doing overbuilding while we evaluate the trade-offs. That's ok, I don't think we need to agree on that point.

Anyway, yes I do agree that the larger discussion here is important. When it starts to impact the project such that we feel we're putting up flimsier guard rails, we do need to start considering all mechanisms to reduce build times, which includes paths-ignore and ccache, but also more radical solutions like explicit template instantiation (#813).

Idclip:
I've renamed the houdini cache workflow and removed the **.txt build rules

danrbailey (approved):
That's great, thanks Nick!

PR #953, opened 2021-02-03
Run CI against VFX Ref 2018

danrbailey:
Although we don't officially support VFX Reference 2018, our codebase seems to still build against it, so I suggest we keep this test for the time being. When we deprecate support for Houdini 18.0 later this year, we can jump to VFX Reference 2020 as our minimum supported version - which will make Boost 1.70 our minimum version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,6 +64,8 @@ jobs:
     strategy:
       matrix:
         config:
+          # build against VFX Ref 2018 even though it's unsupported, just to make sure that we test Boost 1.61
+          - { cxx: clang++, image: '2018', hou: '18_0', build: 'Release', extras: 'ON' }
           - { cxx: clang++, image: '2019', hou: '18_0', build: 'Release', extras: 'ON' }
           - { cxx: clang++, image: '2020', hou: '18_5', build: 'Release', extras: 'ON' }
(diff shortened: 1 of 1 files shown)


jmlait (approved):

Idclip (changes requested):
The whole point of the boost change was so you can have `disable_checks` OFF for H18... your previous builds were failing due to the older CMake version so you should be able to revert that change?

danrbailey:
> The whole point of the boost change was so you can have `disable_checks` OFF for H18... your previous builds were failing due to the older CMake version so you should be able to revert that change?

My untested assumption was that there were other old deps in VFX 2018 that would require disabling checks, such as TBB 2017. I'll revert this change and see if it still passes.

danrbailey:
> > The whole point of the boost change was so you can have `disable_checks` OFF for H18... your previous builds were failing due to the older CMake version so you should be able to revert that change?
> 
> My untested assumption was that there were other old deps in VFX 2018 that would require disabling checks, such as TBB 2017. I'll revert this change and see if it still passes.

Nope, failed on attempting to pull in TBB 2017. Un-re-reverted.

Idclip:
> Nope, failed on attempting to pull in TBB 2017. Un-re-reverted.

Oh I misunderstood what you are doing - I thought it would be better to explicitly tell our CMake to build against Houdini 18 boost on the VFX19 platform. But this is okay too.

Idclip (approved):

PR #963, opened 2021-02-10
Merge feature/inlinehalf

Idclip:


Idclip:
@danrbailey githib won't let me interactively resolve conflicts because "feature/inlinehalf branch is a protected branch". I also can't merge master->feature/inlinehalf for the same reason. So I've had to create *another* branch which merges master->feature/inlinehalf and then PR that... can we change this setting?

Idclip:
I've resolve this for this case by doing the above (opening a new PR with master->feature/inline and merging that)

danrbailey:
> I've resolved this for this case by doing the above (opening a new PR with master->feature/inline and merging that)

I've enabled "Allow force pushes" on feature branches, but that may not fix it because I suspect this might be caused by the DCO check. We'll have to test it.

danrbailey:
Just added a CI check into this feature branch for the USE_IMATH_HALF variable:

https://github.com/AcademySoftwareFoundation/openvdb/pull/966

jmlait (approved):

danrbailey (approved):

PR #972, opened 2021-02-18
feature/ax->master

Idclip:
Final changes from the feature/ax branch, introducing the SOP into master


Idclip:
Had a weird CI failure here on testabi8clang10 (https://github.com/AcademySoftwareFoundation/openvdb/pull/972/checks?check_run_id=1926815847)

```
1: [ RUN      ] TestFile.testAsync
1: /home/runner/work/openvdb/openvdb/openvdb/openvdb/unittest/TestFile.cc:2284: Failure
1: Expected equality of these values:
1:   0
1:   ::stat(filenames[id].c_str(), &buf)
1:     Which is: -1
1: /home/runner/work/openvdb/openvdb/openvdb/openvdb/unittest/TestFile.cc:2285: Failure
1: Expected equality of these values:
1:   Index64(refFileSize)
1:     Which is: 476396
1:   Index64(buf.st_size)
1:     Which is: 0
1: [  FAILED  ] TestFile.testAsync (4041 ms)
```

danrbailey (approved):
Approved, but I notice the release notes are being added to 8.0.1 instead of 8.0.2, so would be good to fix that before merging.

Idclip:
> Approved, but I notice the release notes are being added to 8.0.1 instead of 8.0.2, so would be good to fix that before merging.

Good catch, fixed

Idclip:
@jmlait, same here, could you re-approving - this is the final merge of the feature AX branch with the SOP into master

PR #975, opened 2021-02-22
Fix Stream Compression Unit Test

danrbailey:
This unit test is designed to pass a sequence of 32 integers to Blosc that cannot be compressed to test the failure case in our library, however the random shuffling wasn't being seeded correctly so it was generating a different sequence every time and sporadically failing.

I've replaced the random shuffle with a hard-coded deterministic sequence that always fails to compress instead.


jmlait (approved):
Ha!  Our CI test was validating our RNG was "sufficiently random" to be uncompressible :>

Idclip (approved):

danrbailey:
> Ha! Our CI test was validating our RNG was "sufficiently random" to be uncompressible :>

Indeed it was. :)

PR #982, opened 2021-03-02
Minor fixes to the OpenVDB AX SOP and FindOpenVDB AX CMake

Idclip:

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_AX.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_AX.cc
@@ -129,6 +129,7 @@ void checkAttributesAgainstList(const std::string& list,
         throw std::runtime_error(msg.c_str());
     }
 }
+
 ////////////////////////////////////////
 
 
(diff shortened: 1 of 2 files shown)


jmlait (approved):

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_AX.cc
```diff
@@ -853,35 +870,7 @@ void SOP_OpenVDB_AX::syncNodeVersion(const char* old_version,
...
-        "8.0.0", {
-            // ax, ax sop and vdb versions re-synced at this version
-            { "attributestocreate",
-                [](const SOP_OpenVDB_AX& node) -> std::string {
-                    const bool createMissing = static_cast<bool>(node.evalInt("createattributes", 0, 0));
-                    if (createMissing == 1) return "*";
-                    else return "";
-                }
-            },
+        "8.0.2", {
```
danrbailey, line 873: 8.0.2 is now 8.1.0

danrbailey (approved):

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_AX.cc
Idclip: Still needs to be 8.0.2 as this is when this change was introduced (i.e. `OPENVDB_LIBRARY_PATCH_VERSION_NUMBER` was 2 at this point in time). This system will apply this change to anything syncing before this version to this version (inclusive) and later

danrbailey: Ok, that makes sense.

PR #983, opened 2021-03-02
Fix Performance Regression in DynamicNodeManager

danrbailey:
In #921, I reverted the change to the NodeManager foreach methods to use copy-by-value instead of copy-by-reference. As it turns out this has a significant performance cost for the DynamicNodeManager that would take a lot of extra complexity in the API to resolve. This now forks the implementation so that NodeManager foreach uses copy-by-value and DynamicNodeManager foreach uses copy-by-reference.

I had tried to keep these two structures as aligned as possible, but they have already diverged because the DynamicNodeManager passes an index to the operator() and returns a bool whereas the NodeManager does not. This is not ideal, but I think is probably the most pragmatic way of resolving this.

### DIFF
--- a/openvdb/openvdb/tree/NodeManager.h
+++ b/openvdb/openvdb/tree/NodeManager.h
@@ -278,7 +278,8 @@ public:
         transform.run(this->nodeRange(grainSize), threaded);
     }
 
-    // identical to foreach except the operator() method has a node index
+    // identical to foreach except the operator() method has a node index and
+    // the operator is referenced instead of copied in NodeTransformer
     template<typename NodeOp>
     void foreachWithIndex(const NodeOp& op, bool threaded = true, size_t grainSize=1)
     {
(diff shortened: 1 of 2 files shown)


danrbailey:
@kmuseth @jmlait @Idclip @apradhana - would someone mind approving this PR so I can do the 7.2.3 patch release?

Idclip (approved):

Idclip:
Might be worth commenting that `foreachWithIndex` uses the reference mode! otherwise it's a bit confusing

danrbailey:
> Might be worth commenting that `foreachWithIndex` uses the reference mode! otherwise it's a bit confusing. The function description also says its identical when it fact its not anymore.

Thanks, I changed the function description of foreachWithIndex to mention this. There's already a note in the function description of the `DynamicNodeManager::foreachTopDown()` that highlights the difference, which I expect to be the main entry point.

PR #986, opened 2021-03-03
Use process_time in test

jspricke:
And comment it at the time as it is not used.


chrisflesher (approved):
Looks good!

Idclip:
This change has been incorporated in #1248, thanks!

PR #995, opened 2021-03-09
Fixed an issue which stopped TestPointCount test running concurrently

Idclip:
This could cause the test to fail if executing multiple copies of `vdb_test` on the same machine.

### DIFF
--- a/openvdb/openvdb/unittest/TestPointCount.cc
+++ b/openvdb/openvdb/unittest/TestPointCount.cc
@@ -5,6 +5,7 @@
 
 #include <openvdb/points/PointDataGrid.h>
 #include <openvdb/openvdb.h>
+#include <openvdb/io/TempFile.h>
 
 #include <openvdb/points/PointGroup.h>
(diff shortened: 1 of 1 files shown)


danrbailey (approved):
Thanks Nick!

Idclip:
This is failing due to the MT builds building the dynamic lib against the static CRT on windows. Now I know this is bad, but one of the things I wanted to make sure we **didn't** do was restrict this behaviour (i.e. do not explicitly disallow linking against the static CRT for .dlls and vice versa, though we should probably warn somewhere). I will fix this by using `openvdb_static` for the unit tests on the MT windows build... but I'd really like to understand why I'm suddenly seeing this failure. `TempFile` inherits from `std::ostream` and, as it's currently building a dll, is exporting the entire class interface. If we're using the static CRT, then I think this is why we're getting multiple symbols being defined but want to investigate further. Looking at `TempFile`, it would probably be better to use composition rather than inheritance anyway.

PR #998, opened 2021-03-11
Fix Header Includes

danrbailey:
This adds in any missing headers to make them isolated - ie putting `#include "header.h"` into an empty main.cpp should compile without error.

This also addresses use of the `openvdb_ax/test/util.h` header. This header is included fairly inconsistently in the unit tests ("openvdb_ax/test/util.h", "../util.h", "../test/util.h", "util.h"). I believe the correct pattern to use is "../util.h" because that makes the openvdb_ax/test directory relocatable and _only_ requires the openvdb_ax headers being on the include path.


Idclip (approved):
Thanks for this @danrbailey!

PR #1003, opened 2021-03-16
CMake multi-configuration generators

Idclip:
Added support for multi-config generators such as MSVS and improved the logic related to the MSVC runtime. In summary, our find modules now search for debug and release library variants and configure them properly for generators which support multi-configs.

 - Blosc, Log4cplus, IlmBase, OpenEXR and TBB find modules now configure release and debug library variants
 - Fixed an issue where FindBlosc wouldn't add zstd as a dep in non static builds BLOSC_USE_EXTERNAL_SOURCES
 - Fixed Log4cplus interface language and missing interface libraries on Win32
 - Removed Jemalloc and Tbbmalloc logic from the core library builds
 - Fixed a CMake bug when using static builds of ilmbase Half
 - The variable CMAKE_MSVC_RUNTIME_LIBRARY is no longer set by the OpenVDB CMake
 - CMake promotes a warning if CMAKE_MSVC_RUNTIME_LIBRARY is set to an unexpected variant relative to the build aritfacts
 - CMake promotes a warning if VCPKG is being used and the triplet is set to an unexpected variant relative to the build aritfacts
 - If CMAKE_MSVC_RUNTIME_LIBRARY is not set, OpenVDB configures the static library with /MT(d) and the dynamic library with /MD(d)

I've also added a variable `MSVC_MP_THREAD_COUNT ` to allow for configuring parallel builds on Windows. Finally, I consolidated the windows builds and added a weekly cron job for the debug builds.


openvdb/openvdb/CMakeLists.txt
```diff
@@ -154,25 +210,12 @@ if(USE_BLOSC OR USE_ZLIB)
...
-if(TARGET Jemalloc::jemalloc)
-  list(APPEND OPENVDB_CORE_DEPENDENT_LIBS Jemalloc::jemalloc)
-endif()
-
-if(TARGET TBB::tbbmalloc AND TARGET TBB::tbbmalloc_proxy)
-  list(APPEND OPENVDB_CORE_DEPENDENT_LIBS
-    TBB::tbbmalloc
-    TBB::tbbmalloc_proxy
-  )
-endif()
```
Idclip, line 174: Removed this, we agreed not to do this correct?

e4lam: You're probably not asking me, but yes, I fully agree. :)

e4lam:
Thanks for this Nick as I just had ran into this last night myself! For some reason, ever since we switched to cmake for openvdb, we haven't noticed that Windows debug builds were wrong for OpenVDB until I upgraded our entire toolchain to Visual Studio 2019 version 16.9.1. Now it just crashes on start up when when the OpenVDB nodes try to register themselves, due to a seemingly ABI issue with std::string.

I haven't tried this change but is it possible to make it so that we can customize the debug postfix for dependencies? At least for Blosc. As you commented that that vcpkg builds debug Blosc libraries strangely, we build Blosc ourselves with `CMAKE_DEBUG_POSTFIX="_d"`. So we need to get the debug OpenVDB to build against `blosc_d.lib` / `libblosc_d.lib`.

Idclip:
@e4lam yikes, I hadn't realized quite how prevalent this issue was! I was just discussing with @jmlait the issue of debug suffix naming - I think it's really only blosc that doesn't enforce its own convention, but yeah I should probably provide an option for this per module e.g. `BLOSC_DEBUG_SUFFIX` which defaults to a suitable value or the expected case if there is one. Will add this!

.github/workflows/build.yml
```diff
@@ -25,6 +25,8 @@ on:
   schedule:
     # run this workflow every day at 7am UTC
     - cron:  '0 7 * * *'
+    # run this workflow every midnight Tuesday
```
danrbailey, line 28: Minor comment - I would suggest making this run Monday morning UTC so there's a little more time to investigate any issues before our TSC meeting in case there's anything to discuss. Also would prefer to avoid any times that aren't immediately obvious (ie does 'midnight Tuesday' mean Monday/Tuesday or Tuesday/Wednesday?).

.github/workflows/build.yml
```diff
@@ -37,12 +39,12 @@ jobs:
     strategy:
       matrix:
         config:
-          - { cxx: clang++, image: '2019', abi: '6', build: 'Release',     test: 'ON'  }
-          - { cxx: clang++, image: '2020', abi: '7', build: 'Release',     test: 'ON'  }
-          - { cxx: clang++, image: '2021', abi: '8', build: 'Release',     test: 'ON'  }
-          - { cxx: clang++, image: '2021', abi: '8', build: 'DebugNoInfo', test: 'OFF' } # tests take too long..
```
danrbailey, line 43: The "DebugNoInfo" mode was because the job would occasionally run out of disk space when building the Houdini plugins in debug mode with all symbols:

https://github.com/AcademySoftwareFoundation/openvdb/blame/master/tsc/meetings/2020-04-02.md#L56

The limit is still 14GB of disk space and I believe we are over that if we enable symbols again, so to change this would just rely on not having the limit strictly enforced?

danrbailey (approved):
A couple of minor comments regarding the CI changes. I don't have enough knowledge about the Windows side to review, but approving anyway so you're not blocked in getting this merged in.

.github/workflows/build.yml
Idclip: This workflow doesn't build the houdini plugins - and interestingly, the Houdini workflow in houdini.yml uses 'Debug'! This might have been a mistake when moving this around, but they're not failing. I think it's okay to change this one toDebug but, as the Houdini stuff takes so long, it might be ever so slightly faster to put that back to `DebugNoInfo`

Idclip:
@e4lam I've added the requested control variables with the following defaults: `Blosc = _d`, `tbb = _debug`, `ilmbase/openexr = _d`, `log4cplus = D` ... I've taken these default values from vcpkg and, depending on how the port was made, I'm not exactly sure if they match the conventions of the source pkgs. If you have any reservations here I'd be happy to change them and instead, in our root CMake, I'd check for `VCPKG_TOOLCHAIN` and set accordingly.

e4lam:
> I'm not exactly sure if they match the conventions of the source pkgs

@Idclip Those defaults are perfect! Other than using _d for blosc, we use the source defaults for the result of those packages (tbb and openexr, we don't build with log4cplus) so I believe that with these new changes, we don't even need to set anything when building openvdb.

jmlait (approved):

e4lam:
@Idclip My apologies that I have been commenting without actually trying these changes. I finally got a chance to do this and noticed that we're still linking against the release zlib. I took a look and it looks that they currently build with a `CMAKE_DEBUG_SUFFIX=d` at https://github.com/madler/zlib but Houdini's zlib 1.2.7 dates from before they used CMAKE_DEBUG_SUFFIX and is currently building it with `CMAKE_DEBUG_SUFFIX=_d`. I believe that for openvdb, it makes sense for it to default to `ZLIB_DEBUG_SUFFIX=d`.

e4lam:
PS. Having said that, it looks like fixing ZLib might require a non-trivial change so maybe it's best to have it as a separate PR.

e4lam:
A separate problem here I just noticed is that there's no support for linking in the static zlib instead of the shared zlib?

Idclip:
> A separate problem here I just noticed is that there's no support for linking in the static zlib instead of the shared zlib?

Yeah that's true, I believe I can handle this by setting `CMAKE_FIND_LIBRARY_SUFFIXES` before calling `find_package(ZLIB)` so that's an easy change.

> I took a look and it looks that they currently build with a `CMAKE_DEBUG_SUFFIX=d` at https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/CMakeLists.txt#L63 but Houdini's zlib 1.2.7 dates from before they used CMAKE_DEBUG_SUFFIX and is currently building it with `CMAKE_DEBUG_SUFFIX=_d`. I believe that for openvdb, it makes sense for it to default to `ZLIB_DEBUG_SUFFIX=d` (actually, we might switch ourselves now that I know this).

This is slightly trickier - I'm not sure what's best CMake practise in terms of _who_ should be providing these find modules. I think, ideally, these would already be available as config files either from the source packages themselves or from applications which distribute them. For zlib we're using CMakes find module (https://github.com/Kitware/CMake/blob/master/Modules/FindZLIB.cmake). You can see where they're setting the zlib debug names here: (https://github.com/Kitware/CMake/blob/master/Modules/FindZLIB.cmake#L76). We can't override this, so I think the only way we can support a custom zlib suffix is if we introduce a duplicate module. This won't be difficult, we can just copy it over and add this custom support, but I've been wondering exactly where the line is with what find functionality lives in VDB. Though I ultimately don't see any other way to solve this.

e4lam:
> You can see where they're setting the zlib debug names here: (https://github.com/Kitware/CMake/blob/master/Modules/FindZLIB.cmake#L76). We can't override this, so I think the only way we can support a custom zlib suffix is if we introduce a duplicate module. This won't be difficult, we can just copy it over and add this custom support, but I've been wondering exactly where the line is with what find functionality lives in VDB. Though I ultimately don't see any other way to solve this.

I'm fine with switching over to something that works with CMake's default FindZLIB.cmake. The more important issue is to have the debug OpenVDB link against the debug zlib.

Idclip:
> > You can see where they're setting the zlib debug names here: (https://github.com/Kitware/CMake/blob/master/Modules/FindZLIB.cmake#L76). We can't override this, so I think the only way we can support a custom zlib suffix is if we introduce a duplicate module. This won't be difficult, we can just copy it over and add this custom support, but I've been wondering exactly where the line is with what find functionality lives in VDB. Though I ultimately don't see any other way to solve this.
> 
> I'm fine with switching over to something that works with CMake's default FindZLIB.cmake. The more important issue is to have the debug OpenVDB link against the debug zlib.
> 

Hmm sorry a bit confused, are you saying that's not the case currently even with the 'zlibd' naming convention? Looking at that find module I think that should be working correctly but that's on the latest branch, I'll investigate a bit

e4lam:
> Hmm sorry a bit confused, are you saying that's not the case currently even with the 'zlibd' naming convention? Looking at that find module I think that should be working correctly but that's on the latest branch, I'll investigate a bit

Ah sorry, you're right. I had wrongly assumed that it wasn't working because we didn't handle it. I made a copy of zlib_d.lib to zlibd.lib and it does indeed link against the right library if I have that.

Idclip:
@e4lam awesome, in which case I'll look to merge this PR as is and fix the static ZLIB linking asap. And, just to confirm, is the lack of a debug suffix control for zlib going to be a problem? Do we need a solution in VDB?

openvdb/openvdb/CMakeLists.txt
```diff
@@ -489,6 +538,16 @@ if(OPENVDB_CORE_SHARED)
...
+
+  if(WIN32)
+    if(NOT CMAKE_MSVC_RUNTIME_LIBRARY)
+      # NOTE: MSVC_RUNTIME_LIBRARY does not propagate to targets, so
+      # also add it explicitly as a compile option
+      set_target_properties(openvdb_shared PROPERTIES
+        MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
+      target_compile_options(openvdb_shared PUBLIC "/MD$<$<CONFIG:Debug>:d>")
+    endif()
+  endif()
```
Idclip, line 550: @e4lam sorry to drag you into the joy that is CMake and the MSVC runtime, but what do you think of this change? In summary, if `CMAKE_MSVC_RUNTIME_LIBRARY` hasn't been explicitly provided, force the static build of VDB to build with `/MT` and the shared build to build with `/MD` (and their debug variants). Without this, I believe both would build with `/MD` or `/MDd` which they take from the default value of `CMAKE_MSVC_RUNTIME_LIBRARY`. So it's slightly different behaviour for the static build.

e4lam:
> awesome, in which case I'll look to merge this PR as is and fix the static ZLIB linking asap. And, just to confirm, is the lack of a debug suffix control for zlib going to be a problem? Do we need a solution in VDB?

No, we actually were already invoking cmake with CMAKE_MODULE_PATH set to a custom directory. So I just forked FindZLIB.cmake ourselves in this directory and adjusted it so that it looked for the right debug names, as well as picking up the shared import library before the static library.

openvdb/openvdb/CMakeLists.txt
e4lam: @Idclip You're right, I did not notice this when I rebuilt OpenVDB ourselves with the quick fix of removing the lines that set CMAKE_MSVC_RUNTIME_LIBRARY to the empty string. Strangely though, it all seems to link fine in Houdini land (eg. the static libUT.lib) and it doesn't seem to cause a dependency in our statically linked executables to the shared CRT. Regardless, I'm all for this change so that the resulting static lib is built with `/MT` and `/MTd`.

The other consideration here is that some people _want_ to compile the static library with `/MD`. ie. they want to statically link in OpenVDB but still link in the shared CRT ([discussion thread](https://www.openvdb.org/forum/?place=msg%2Fopenvdb-forum%2FI5Q_Jn4IGGM%2F8LB0tuuUXdQJ)).  ). But I think that's fine, as long as they can override CMAKE_MSVC_RUNTIME_LIBRARY with the appropriate generator expression.

e4lam: Ah, I think I know why compiling the static openvdb with /MD is not a problem for Houdini. None of our static binaries need to call anything into openvdb, so I suspect maybe the linker is smart enough to not link in openvdb in even though it's specified on the linker line.

Idclip: > The other consideration here is that some people _want_ to compile the static library with `/MD`. ie. they want to statically link in OpenVDB but still link in the shared CRT ([discussion thread](https://www.openvdb.org/forum/?place=msg%2Fopenvdb-forum%2FI5Q_Jn4IGGM%2F8LB0tuuUXdQJ)). But I think that's fine, as long as they can override CMAKE_MSVC_RUNTIME_LIBRARY with the appropriate generator expression.

Absolutely, you can provide `-DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"` as an argument which will set /MD and /MDd as expected to all artefacts.

.github/workflows/build.yml
Idclip: Done!

e4lam:
Thanks again for all the Windows cmake work!

PR #1008, opened 2021-03-17
OpenVDB Points -> Volume transfer framework & kernels

Idclip:
This is a draft pull request of the point transfer/rasterization framework and spherical, ZB and trilinear kernels. The final commit adds a dummy Houdini SOP which can be used to test the surfacing methods in Houdini, but I anticipate this to be left out of a final PR. Note that this PR also includes the PointStatistics #997 work.

Below are some preliminary timings. More data to be added (timings, images of results, surface comparisons etc).
**Input**: https://artifacts.aswf.io/io/aswf/openvdb/models/waterfall_points.vdb/1.0.0/waterfall_points.vdb-1.0.0.zip

----------------------------------------------------------------------------------------------------------------------------------------------

`tools::rasterizeSpheres()` vs `tools::ParticlesToLevelSet` (From Particle SOP)
No attribute transfers, no pscale on points (reads from parameter)

Target Voxel Size 0.1, Target  Half Band 3 
Input pscale | Old Time (ms) | New Time (ms) | Speedup
-- | -- | -- | -- 
0.1 | 2588.47 | 1764.82 | x1.467 
0.2 | 4435.08 | 2524.45 | x1.757 
0.5 | 17639.2 | 6946.8 | x2.539 
1.0 | 57238.1 | 16061.5 | x3.567

Input  pscale 0.1, Target  Half Band 3 
Target Voxel Size | Old Time (ms) | New Time (ms) | Speedup
-- | -- | -- | -- 
0.01 | 53369.9 | 25959.1 | x2.056 
0.1  | 2588.47 | 1764.82 | x1.467  
1.0  | 4260.55 | 2801.18 | x1.521 
5.0  | 8869.62 | 4194.94 | x2.114

Target Voxel Size 0.1, Input pscale 0.1,
Target Half Band | Old Time (ms) | New Time (ms) | Speedup
-- | -- | -- | -- 
1 | 739.048 | 479.244 | x1.542 
3 | 2588.47 | 1764.82 |  x1.467 
10 | 37725.9 | 19165.4 | x1.968 
15 | 118521 | 51222.2 |  x2.314

----------------------------------------------------------------------------------------------------------------------------------------------

`tools::rasterizeSpheres()` vs `tools::ParticlesToLevelSet` (VDB From Particle SOP)
6 attributes to transfer `v@v` (varying), `v@Cd` (varying), `int@id` (varying), `int64@a` (uniform), `float@b` (uniform). `float@c` (varying)
no pscale on points (reads from parameter)

Target Voxel Size 0.1, Target Half Band 3

Input pscale | Old Time (ms) | New Time (ms) | Speedup
-- | -- | -- | -- 
0.1 | 2815.53 | 1659.43 | x1.697 
0.2 | 4818.51 | 2828.60 | x1.703 
0.5 | 16674.90 | 6948.99 |  x2.4 
1   | 66442.40 | 16395.70 | x4.052 

----------------------------------------------------------------------------------------------------------------------------------------------

`tools::rasterizeSpheres()` vs `tools::ParticlesToLevelSet` (VDB From Particle SOP)
pscale on points (uniform and varying)

// more to document / upload

----------------------------------------------------------------------------------------------------------------------------------------------

`tools::rasterizeSpheres()` vs `tools::rasterizeSmoothSpheres`

This section compares the two new methods against each other and additionally compares `tools::rasterizeSmoothSpheres` against the VDB From Particle Fluid SOP. Note that the latter comparison is not strictly fair; the VDB From Particle Fluid SOP does not perform the exact same functionality as `tools::rasterizeSmoothSpheres`, but in all tests the `Limit Refinement` parameter is on and set to `0`, and the `Resample Input` parameter is `OFF`.

// more to document / upload

Target Voxel Size 0.075, Input Pscale 0.15, Influence Scale (for Smooth Spheres) 3
rasterizeSpheres  |  rasterizeSmoothSpheres
-- | --  
 | 
2643.64ms | 4361.73ms (x0.6)

Target Voxel Size 0.075, Input Pscale 0.15, Influence Scale (for Smooth Spheres) 3
VDB From Particle Fluid  |  rasterizeSmoothSpheres
-- | --  
 | 
5466.03ms | 4361.73ms (x1.253)

### DIFF
--- a/openvdb/openvdb/points/PointRasterizeSDF.h
+++ b/openvdb/openvdb/points/PointRasterizeSDF.h
@@ -1,0 +1,1701 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+//
+/// @author Nick Avramoussis
+///
+/// @file PointRasterizeSDF.h
+///
+/// @brief Transfer schemes for rasterizing point positional and radius data to
+///  signed distance fields with optional closest point attribute transfers.
+///  All methods support arbitrary target linear transformations, fixed or
+///  varying point radius, filtering of point data and arbitrary types for
+///  attribute transferring.
+///
+/// @details There are two main transfer implementations; rasterizeSpheres and
+///  rasterizeSmoothSpheres. The prior performs trivial narrow band stamping
+///  of spheres for each point, where as the latter calculates an averaged
+///  position of influence per voxel as described in:
+///    [Animating Sand as a Fluid - Zhu Bridson 05].
+///
+///  rasterizeSpheres() is an extremely fast and efficient way to produce both a
+///  valid symmetrical narrow band level set and transfer attributes using
+///  closest point lookups.
+///
+///  rasterizeSmoothSpheres() produces smoother, more blended connections
+///  between points which is ideal for generating a more artistically pleasant
+///  surface directly from point distributions. It aims to avoid typical post
+///  filtering operations used to smooth surface volumes. Note however that
+///  rasterizeSmoothSpheres may not necessarily produce a *symmetrical* narrow
+///  band level set; the exterior band may be smaller than desired depending on
+///  the search radius. The surface can be rebuilt or resized if necessary.
+///  The same closet point algorithm is used to transfer attributes.
+///
+///  In general, it is recommended to consider post rebuilding/renormalizing the
+///  generated surface using either tools::levelSetRebuild() or
+///  tools::LevelSetTracker::normalize() tools::LevelSetTracker::resize().
+///
+/// @note These methods use the framework provided in PointTransfer.h
+///
+
+#ifndef OPENVEB_POINTS_RASTERIZE_SDF_HAS_BEEN_INCLUDED
+#define OPENVEB_POINTS_RASTERIZE_SDF_HAS_BEEN_INCLUDED
+
+#include "PointDataGrid.h"
+#include "PointTransfer.h"
+#include "PointStatistics.h"
+
+#include <openvdb/openvdb.h>
+#include <openvdb/Types.h>
+#include <openvdb/tools/Prune.h>
+#include <openvdb/tools/ValueTransformer.h>
+#include <openvdb/thread/Threading.h>
+#include <openvdb/util/NullInterrupter.h>
+
+#include <unordered_map>
+
+#include <tbb/task_group.h>
+#include <tbb/parallel_reduce.h>
+
+namespace openvdb {
+OPENVDB_USE_VERSION_NAMESPACE
+namespace OPENVDB_VERSION_NAME {
+namespace points {
+
+/// @brief Narrow band sphere stamping with a uniform radius.
+/// @details Rasterizes points into a level set using basic sphere stamping with
+///   a uniform radius. The radius parameter is given in world space units and
+///   is applied to every point to generate a fixed surface mask and consequent
+///   distance values.
+/// @param points       the point data grid to rasterize
+/// @param radius       the world space radius of every point
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return The signed distance field.
+template <typename PointDataGridT,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+typename SdfT::Ptr
+rasterizeSpheres(const PointDataGridT& points,
+             const Real radius,
+             const Real halfband = LEVEL_SET_HALF_WIDTH,
+             math::Transform::Ptr transform = nullptr,
+             const FilterT& filter = NullFilter(),
+             InterrupterT* interrupter = nullptr);
+
+/// @brief Narrow band sphere stamping with a varying radius.
+/// @details Rasterizes points into a level set using basic sphere stamping with
+///   a variable radius. The radius string parameter expects a point attribute
+///   of type RadiusT to exist.
+/// @param points       the point data grid to rasterize
+/// @param radius       the name of the radius attribute
+/// @param scale        an optional scale to apply to each per point radius
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return The signed distance field.
+template <typename PointDataGridT,
+    typename RadiusT = float,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+typename SdfT::Ptr
+rasterizeSpheres(const PointDataGridT& points,
+             const std::string& radius,
+             const Real scale = 1.0,
+             const Real halfband = LEVEL_SET_HALF_WIDTH,
+             math::Transform::Ptr transform = nullptr,
+             const FilterT& filter = NullFilter(),
+             InterrupterT* interrupter = nullptr);
+
+/// @brief Narrow band sphere stamping with a uniform radius and closest point
+///   attribute transfer.
+/// @details Rasterizes points into a level set using basic sphere stamping with
+///   a uniform radius. The radius parameter is given in world space units and
+///   is applied to every point to generate a fixed surface mask and consequent
+///   distance values. Every voxel's closest point is used to transfer each
+///   attribute in the attributes parameter to a new grid of matching topology.
+///   The destination types of these grids is equal to the ValueConverter result
+///   of the attribute type applied to the PointDataGridT.
+/// @note The AttributeTypes template parameter should be a TypeList of the
+///   required or possible attributes types. i.e. TypeList<int, float, double>.
+///   A runtime error will be thrown if no equivalent type for a given attribute
+////  is found in the AttributeTypes TypeList.
+/// @param points       the point data grid to rasterize
+/// @param radius       the world space radius of every point
+/// @param attributes   list of attributes to transfer
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return A vector of grids. The signed distance field is guaranteed to be
+///   first and at the type specified by SdfT. Successive grids are the closest
+///   point attribute grids. These grids are guaranteed to have a topology
+///   and transform equal to the surface.
+template <typename PointDataGridT,
+    typename AttributeTypes,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+GridPtrVec
+rasterizeSpheres(const PointDataGridT& points,
+             const Real radius,
+             const std::vector<std::string>& attributes,
+             const Real halfband = LEVEL_SET_HALF_WIDTH,
+             math::Transform::Ptr transform = nullptr,
+             const FilterT& filter = NullFilter(),
+             InterrupterT* interrupter = nullptr);
+
+/// @brief Narrow band sphere stamping with a varying radius and closest point
+///   attribute transfer.
+/// @details Rasterizes points into a level set using basic sphere stamping with
+///   a variable radius. The radius string parameter expects a point attribute
+///   of type RadiusT to exist. Every voxel's closest point is used to transfer
+///   each attribute in the attributes parameter to a new grid of matching
+///   topology. The destination types of these grids is equal to the
+///   ValueConverter result of the attribute type applied to the PointDataGridT.
+/// @note The AttributeTypes template parameter should be a TypeList of the
+///   required or possible attributes types. i.e. TypeList<int, float, double>.
+///   A runtime error will be thrown if no equivalent type for a given attribute
+////  is found in the AttributeTypes TypeList.
+/// @param points       the point data grid to rasterize
+/// @param radius       the name of the radius attribute
+/// @param attributes   list of attributes to transfer
+/// @param scale        scale to apply to each per point radius
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return A vector of grids. The signed distance field is guaranteed to be
+///   first and at the type specified by SdfT. Successive grids are the closest
+///   point attribute grids. These grids are guaranteed to have a topology
+///   and transform equal to the surface.
+template <typename PointDataGridT,
+    typename AttributeTypes,
+    typename RadiusT = float,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+GridPtrVec
+rasterizeSpheres(const PointDataGridT& points,
+             const std::string& radius,
+             const std::vector<std::string>& attributes,
+             const Real scale = 1.0,
+             const Real halfband = LEVEL_SET_HALF_WIDTH,
+             math::Transform::Ptr transform = nullptr,
+             const FilterT& filter = NullFilter(),
+             InterrupterT* interrupter = nullptr);
+
+/// @brief Smoothed point distribution based sphere stamping with a uniform radius.
+/// @details Rasterizes points into a level set using [Zhu Bridson 05] sphere
+///   stamping with a uniform radius. The radius and search radius parameters
+///   are given in world space units and are applied to every point to generate
+///   a fixed surface mask and consequent distance values. The search radius is
+///   each points points maximum contribution to the target level set. The search
+///   radius should always have a value equal to or larger than the point radius.
+/// @warning The width of the exterior half band *may* be smaller than the
+///   specified half band if the search radius is less than the equivalent
+///   world space halfband distance.
+/// @param points       the point data grid to rasterize
+/// @param radius       the world space radius of every point
+/// @param searchRadius the maximum search distance of every point
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return The signed distance field.
+template <typename PointDataGridT,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+typename SdfT::Ptr
+rasterizeSmoothSpheres(const PointDataGridT& points,
+             const Real radius,
+             const Real searchRadius,
+             const Real halfband = LEVEL_SET_HALF_WIDTH,
+             math::Transform::Ptr transform = nullptr,
+             const FilterT& filter = NullFilter(),
+             InterrupterT* interrupter = nullptr);
+
+/// @brief Smoothed point distribution based sphere stamping with a varying radius.
+/// @details Rasterizes points into a level set using [Zhu Bridson 05] sphere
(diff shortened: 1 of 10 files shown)


Idclip:
@danrbailey, one thing of note might be the commit which modifies some of the point count/mask tools (https://github.com/AcademySoftwareFoundation/openvdb/pull/1008/commits/e65338d0e2d5c996bfa30e76cf42b3bd02ed7514). It's an isolated/fairly small commit which provides tree only function signatures and faster mask construction. Wanted to provide it in context but happy to split out.

danrbailey:
> @danrbailey, one thing of note might be the commit which modifies some of the point count/mask tools ([e65338d](https://github.com/AcademySoftwareFoundation/openvdb/commit/e65338d0e2d5c996bfa30e76cf42b3bd02ed7514)). It's an isolated/fairly small commit which provides tree only function signatures and faster mask construction. Wanted to provide it in context but happy to split out.

Thanks, yeah I skimmed this bit and it looks good, feel free to split this out and I think I can review this more quickly...

danrbailey:
Thanks for sharing @Idclip, this is really great to see it all in context. I'm going to have to put aside a fair bit of time to review this. In terms of the core library API at least, one of my main considerations is to try and avoid ending up with _a lot_ of different rasterization methods. It might be useful to me if you can share a more comprehensive list of features and what this rasterization does and does not support. Perhaps based on the all the existing rasterization tools such as VDB Rasterize Points SOP and the Houdini ones such as Volume Rasterize Particles SOP, etc. I'm thinking of features like - varying particle radii, ability to rasterize into frustum grids, velocity blur, arbitrary attribute rasterization support, etc, etc.

danrbailey:
Seeing as you're working on updating this PR, I wanted to throw out some additional thoughts / considerations on my end. I mentioned some of this in my previous comment so this is not particularly new. 

Rasterization is a massive topic and there are lots of features that we _can_ support. I believe this implementation is not exhaustive in the sense that it does not rasterize to frustum grids, does not support motion blur (see Houdini's Volume Rasterize Attributes SOP), does not support multi-grid rasterization (such as rasterizing thousands of input point grids into a single density grid without having to do them one-by-one), does not support streaming (removes attributes or grids from memory after rasterization completes). To be clear, I'm not suggesting that we should only ever accept an über algorithm that does everything, it seems to me that certain styles of algorithms will be more suitable in some situations than to others. For example, I would think that a scatter-style algorithm would work more efficiently with frustum grids than this type of gather-style algorithm. I just wonder how to organize a world where we might have a bunch of different tools doing some but not all of the features that we associate with rasterization. I'm in one mind to say that all rasterization-related methods should just live in a single header `points/PointRasterize.h` even if there's some overlap and it ups the compile time (but we alleviate that somewhat with explicit instantiation).

One thing that does feel reasonable to me to try and cut down the massive problem space is to refer to points => surface as "surfacing" and points => density/arbitrary grids as "rasterization". So `points/PointRasterizeSDF.h` should be renamed `points/PointSurfacer.h` in my opinion.

Idclip:
Thanks for your comments @danrbailey - I pretty much agree with what you've said!

> One thing that does feel reasonable to me to try and cut down the massive problem space is to refer to points => surface as "surfacing" and points => density/arbitrary grids as "rasterization". So points/PointRasterizeSDF.h should be renamed points/PointSurfacer.h in my opinion.

The reason I went with `PointRasterizeSDF` was because I also have a `PointRasterizeTrillinear` header ready to PR, so I foresaw a bunch of `PointRasterize<Details>` headers

danrbailey:
There's a lot to review here, so just for the sake of making some progress with this one, I'm going to add comments as and when I get time and hopefully keep returning to this to get it over the line. 

I'll start with the easy stuff - all the changes to PointMask and PointCount look good, the only thing I would request is that we put a call to the new tree-based methods in the unit tests somewhere just to prevent any subsequent refactoring from unintentionally changing the public API.

openvdb/openvdb/unittest/TestPointRasterizeTrilinear.cc
```diff
@@ -0,0 +1,492 @@
+///////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2017-2020 Double Negative Visual Effects
+//
```
apradhana, line 4: Needs the correct copyright.

BSalem:
It seems that something was updated which causing rasterizeSpheres() to not be multi-threaded! By applying on the same water_fall.vdb shared here, it's working on a single thread and takes around 17mins! (radius 0.1), I tested on a patched older vdb build (8.0) and it looks like the examples above (it's multi-threaded and taking similar time to process as were shared above).

Any idea about what would be the reason why it's single threaded now?

Idclip:
> Any idea about what would be the reason why it's single threaded now?

@BSalem thanks for reporting, I remember running into this myself when rebasing these changes. I can't remember exactly what caused it, think it was to do with some threading changes between VDB versions. Currently a bit blocked on proceeding but will investigate this asap.

Idclip:
@BSalem I beieve I've resolved this - the method now always runs multithreaded and I'll look to expose a control for running it single-threaded in the future. Also started a discussion regarding this issue here: https://lists.aswf.io/g/openvdb-dev/message/1714

BSalem:
Thank you very much for the update @Idclip ... I will give it a try

BSalem:
@Idclip it works now as expected... thanks

Idclip:
I'm going to remove the SOP from this branch, at which point it should be ready to merge @danrbailey

Idclip:
@danrbailey this is good to go. I have more work to do on top of this (such as testing moving the impl out) and we can of course rename and move these files as we progress. Let me know if this looks good to approve.

apradhana (approved):
Looks good to me.

openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -0,0 +1,1701 @@
...
+///
+///  In general, it is recommended to consider post rebuilding/renormalizing the
+///  generated surface using either tools::levelSetRebuild() or
+///  tools::LevelSetTracker::normalize() tools::LevelSetTracker::resize().
+///
+/// @note These methods use the framework provided in PointTransfer.h
+///
+
+#ifndef OPENVEB_POINTS_RASTERIZE_SDF_HAS_BEEN_INCLUDED
+#define OPENVEB_POINTS_RASTERIZE_SDF_HAS_BEEN_INCLUDED
```
danrbailey, line 41: Typo: OPENVEB

openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -0,0 +1,1701 @@
...
+/// @param radius       the world space radius of every point
+/// @param halfband     the half band width
+/// @param transform    the target transform for the surface
+/// @param filter       a filter to apply to points
+/// @param interrupter  optional interrupter
+/// @return The signed distance field.
+template <typename PointDataGridT,
+    typename SdfT = typename PointDataGridT::template ValueConverter<float>::Type,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
```
danrbailey, line 79: The interrupter should not be provided as a template argument. This prevents us doing explicit template instantiation. There are lots of old methods that still do this that we haven't changed, but for new methods, it should really just be part of the function signature (ie `util::NullInterrupter* interrupter = nullptr`).

openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -0,0 +1,1701 @@
...
+    }
+
+private:
+    const RadiusTreeT& mMin;
+    const RadiusTreeT& mMax;
+    const Real mScale, mHalfband;
+};
+
+template <typename SdfT, typename InterrupterT, typename PointDataGridT>
+inline typename SdfT::Ptr
```
danrbailey, line 1113: Template functions should not be inline, this does nothing except prevent explicit instantiation from being used.

openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -0,0 +1,1701 @@
...
+        surface->tree().topologyUnion(*mask);
+        mask.reset();
+        // set maskoff values to -background
+        tree::ValueAccessor<const MaskTreeT> acc(*maskoff);
+        auto setOffOp = [acc](auto& iter) {
+            if (acc.isValueOn(iter.getCoord())) {
+                iter.modifyValue([](auto& v) { v = -v; });
+            }
+        };
+        tools::foreach(surface->beginValueOff(), setOffOp,
```
danrbailey, line 1151: Unless I'm mis-reading this, isn't this using the slow tree value iterator?

openvdb/openvdb/points/PointRasterizeTrilinear.h
```diff
@@ -0,0 +1,444 @@
...
+OPENVDB_USE_VERSION_NAMESPACE
+namespace OPENVDB_VERSION_NAME {
+namespace points {
+
+///
+template <typename ValueT, bool Staggered = true>
+struct TrilinearTraits
+{
+    using ResultT = typename std::conditional<
+        VecTraits<ValueT>::IsVec, ValueT, math::Vec3<ValueT>>::type;
```
danrbailey, line 38: Isn't this the same as `VecTraits<ValueT>::ElementType` ?

openvdb/openvdb/points/PointTransfer.h
```diff
@@ -0,0 +1,579 @@
...
+/// @param points       the point data grid to rasterize
+/// @param transfer     the transfer scheme
+/// @param filter       optional point filter
+/// @param interrupter  optional interrupter
+template <typename PointDataTreeOrGridT,
+    typename TransferT,
+    typename FilterT = NullFilter,
+    typename InterrupterT = util::NullInterrupter>
+inline void
+rasterize(const PointDataTreeOrGridT& points,
```
danrbailey, line 168: Same comment about removing inline and the InterrupterT template arg.

openvdb/openvdb/points/PointRasterizeSDF.h
Idclip: Whilst I agree, we can't do explicit template instantiation for this anyway as it takes a `FilterT` :(

openvdb/openvdb/points/PointRasterizeSDF.h
Idclip: Yes this should really use the dynamic node manager but it didn't exist when I wrote this

RodenLuo:
Hi, I would like to implement a SOP in Houdini for the `rasterizeSmoothSpheres` mentioned in this PR (and adapt it, mainly I would like to have different particle separations for different points when building the SDF). I am freshly new to OpenVDB. I know how to compile an HDK project. But I'm confused about how to compile OpenVDB and use the self-compiled OpenVDB nodes in Houdini.  Could you please kindly guide me or point me to the relevant manual? Thanks!

Idclip:
Hi @RodenLuo,

The OpenVDB Houdini Plugin can be a bit tricky to compile, but we have some instructions for it here:

https://www.openvdb.org/documentation/doxygen/build.html#buildBuildHou

Also worth mentioning we have the following PR which implements a SOP for this purpose (but is not yet finished, you can use it as a reference if you wish): https://github.com/AcademySoftwareFoundation/openvdb/pull/1390

Nick

RodenLuo:
Hi Nick, Thanks very much! I was just reading the HDK Doc (and noticed your name in the author filed [here](https://www.sidefx.com/docs/hdk/_point_rasterize_s_d_f_8h.html#a9623e4343ff8ed49dbe3372c9b538517)). It was a headache to wire all the variables correctly. #1390 came right in time! 

Thanks,
Roden

RodenLuo:
Hi Nick @Idclip,

I compiled the `SOP_OpenVDB_Points_Surfacer` on Win10, Intel i7-9700F, RTX 2060 in Release x64 (see #1591). However, using the default parameter of that SOP, it takes  4min43sec to cook (Fig 1). I then apply `Input Pscale 0.15` by converting the demo input to points, add pscale and then convert back to vdb points. And then using `Voxel Size 0.075, Input Pscale 0.15, Influence Scale (for Smooth Spheres) 3`, the reconstruction is not even completed (Fig 2). If I check `Use World Space Influence Radius` and the values automatically goes to 0.1, it takes 4107ms and the result looks decent but not as good as you showed (Fig 3). The VDB From Particle Fluid SOP with default parameters takes 4563ms and gives decent result (Fig 4). Could you please comment the above procedures and guide on how to reproduce your visual at your convenience? The HIP file is attached in [VDBPointsSurfacer.zip](https://github.com/AcademySoftwareFoundation/openvdb/files/10877944/VDBPointsSurfacer.zip). 

As you mentioned "the VDB From Particle Fluid SOP does not perform the exact same functionality as `tools::rasterizeSmoothSpheres`", I wonder is VDB From Particle Fluid SOP's implementation available somewhere, either as source code or as algorithm description? 

Thanks,
Roden

image
image
image
image

Idclip:
Hi Roden,

In terms of producing the exact same result as in the description of this PR, this was achieved with the following values:
- voxel size: 0.075, pscale: 0.15, use world space radius: off, influence radius: 3
image

I will say that part of the reason this SOP is not complete is that using the average position controls as they are can be very fiddly, especially with varying radii (though that's not the case here). You need to be careful with some of option bounds as there aren't any heuristic checks with respect to radius lookup or voxel size. Depending on your input, if you decrease or increase (respectively) these values too far, you'll quickly end up with a slow/memory consuming process. This might be necessary depending on your input, but the best way to iterate on a point cloud is to clip a portion of the point cloud down to a small subset and run that through the surfacer multiple times, tweaking values as you see fit, until you see a result you like.

If you're using a uniform pscale, you don't need to apply pscale directly to the points and can instead use the "Particle Radius Scale" value directly (as in the above image). The influence radius scale is a *either* a multiplier on the particle radius (when "Use World Space Influence Radius" is *OFF*) _or_ a specific world space search radius (when "Use World Space Influence Radius" is *ON*). So for a pscale of 0.15 and a influence radius of 3, your point kernels will have a maximum world space lookup of 0.45 units. If however, "Use World Space Influence Radius" is *ON*, then your search radius would be exactly 3 world space units. In this example your input point cache voxel size is 0.05. A world space search of 0.45 would mean that every point receives contributions from roughly `(0.45/0.05)=` 9 voxels around it or, in other words, a maximum of `pi*(4/3.0)*(9^3)=`  ~3053 voxels. A world space search of 3 would result in ~904778 voxel lookups for each point!

In regards to the "VDB From Particle Fluid SOP", this source code is not public. My understanding however is that it implements a somewhat similar algorithm, with the main difference being the additional controls which perform some refinement operations on the incoming points _before_ they are fed to the surfacing operation. These controls are the "Limit Refinement" and "Resample Input" options. I honestly can't remember exactly what these operations do, @jmlait may be able to articulate this further.

RodenLuo:
Hi Nick, Thanks for the detailed info. I'm able to reproduce the same result. Apart from what you mentioned, one of the key differences between them is shown in the below GIF and is causing some problems for me (screenshot below). It seems there is a continuous filling of the field in the built-in node once the distant fields meet each other. I am going to dive into the code and see if I can generate a similar result as the built-in one. Could you please kindly give some hints on which part of the code I should look into? Thanks. 

Top: VDB From Particle Fluid SOP; Bottom: SOP_OpenVDB_Points_Surfacer
ParticleFluidSurface_vs_RasterrizeSmoothSurface

Top: Particle Fluid Surface SOP (which uses VDB From Particle Fluid SOP); Bottom: SOP_OpenVDB_Points_Surfacer
image

Idclip:
@RodenLuo very interesting, thanks for the breakdown - there may very well be a bug or mistake I've made somewhere, I'd have to take a closer look.

> Could you please kindly give some hints on which part of the code I should look into?

The generic framework which handles the parallel rasterization/transfer schemes is in `PointTransfer.h`. You write a transfer scheme which implements this required interface. So both the spherical/avg modes have their own schemes which implement per point raster kernels. These kernels are in `PointRasterizeSDFImpl.h`. The avg position scheme is called `AveragePositionTransfer` ~line 353. The important methods these schemes implement are `initialize` ~line 402, which is run for every node surface node, `rasterizaPoint` ~line 425 which is run for every point which _may_ contribute to the current surface node being processed and `finalize` ~line 506, run once a surface node has finished being processed. 

The framework implements a hybrid gather/scatter approach - initially the surface topology is created by activating regions of interest using `initFixedSdf` in  `PointRasterizeSDFImpl.h` (think of it as scattering active topology based on point radii). Once the topology is initiased, the transfer schemes perform a gather opertion for each surface node (so each node is only processed once), whereby points are collected and stamped into the surface using the above methods.

The Avg transfer schemes does something like this: 

1. Init the active surface node values to 0. (initialize)
2. For each surface voxel, collect and average out point positions and weighted radii which contribute (rasterizePoint)
3. Assign the final weighted/accumulated surface values to the voxel (finalize)

Hopefully this gives you a good start, do let me know what you find!

Idclip:
I would add however that the centre artefacts that you're seeing from the vdb points surfacer are an expected unfortunate byproduct of the algorithm. There has been work in this space to fix these which I have tried successfully (primarily [Solenthaler 07 - see Surface Reconstruction](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwi20avHwsL9AhXSQEEAHZUVAJUQFnoECAkQAQ&url=https%3A%2F%2Fpeople.inf.ethz.ch%2F~sobarbar%2Fpapers%2FSol07b%2FSol07b.pdf&usg=AOvVaw1RYFouapQKIT7i0mz-Oz4z)) but aren't part of the current implementation (there's significant overhead to doing this)

RodenLuo:
Thanks very much! While I'm going through the code to learn its implementation, I wonder do you have your fix (even though it's computationally heavy) publically available somewhere? Also, while looking into the paper you sent and checking a few following works, I'm curious whether the following methods are in OpenVDB or not: 

1) Quispe, Filomen Incahuanaco, and Afonso Paiva. "[Counting Particles: a simple and fast surface reconstruction method for particle-based fluids.](https://ieeexplore.ieee.org/document/9991770)" 2022 35th SIBGRAPI Conference on Graphics, Patterns and Images (SIBGRAPI). Vol. 1. IEEE, 2022.

2) Yu, Jihun, and Greg Turk. "[Reconstructing surfaces of particle-based fluids using anisotropic kernels.](https://dl.acm.org/doi/abs/10.1145/2421636.2421641)" ACM Transactions on Graphics (TOG) 32.1 (2013): 1-12. 

[fluid-engine-dev](https://github.com/doyubkim/fluid-engine-dev) implements the second one and also [Zhu & Bridson](https://dl.acm.org/doi/abs/10.1145/1073204.1073298)'s (which SOP_OpenVDB_Points_Surfacer is using if I'm not mistaken).

Idclip:
> While I'm going through the code to learn its implementation, I wonder do you have your fix (even though it's computationally heavy) publically available somewhere?

Not publically available at the moment. I wasn't planning to ever contribute it as it's a significant amount of extra code for something I don't think will be used much but I'll see if I can clean it up.

> Quispe, Filomen Incahuanaco, and Afonso Paiva. "[Counting Particles: a simple and fast surface reconstruction method for particle-based fluids.](https://ieeexplore.ieee.org/document/9991770)" 2022 35th SIBGRAPI Conference on Graphics, Patterns and Images (SIBGRAPI). Vol. 1. IEEE, 2022.

Interesting, I have not read this, thanks!

> Yu, Jihun, and Greg Turk. "[Reconstructing surfaces of particle-based fluids using anisotropic kernels.](https://dl.acm.org/doi/abs/10.1145/2421636.2421641)" ACM Transactions on Graphics (TOG) 32.1 (2013): 1-12.

Ah yes, this always comes up. This isn't available in VDB at the moment but, similiar to the Solenthaler work, we have an implemenation which we're hoping to share in the future. I can't provide any ETA's on either unfortunately.

> also [Zhu & Bridson](https://dl.acm.org/doi/abs/10.1145/1073204.1073298)'s (which SOP_OpenVDB_Points_Surfacer is using if I'm not mistaken).

Yes that's correct

PR #1010, opened 2021-03-18
Introduce multi-threaded methods to count active voxels and used memory

danrbailey:
Thread the activeVoxelCount() method using the DynamicNodeManager. Performance on a 1 billion voxel VDB (32 threads):

Original: 203 milliseconds
New: 26 milliseconds

There's still a bool to be able to optionally run single-threaded if desired.

~~I didn't add the deprecations to the original method, but my proposal here is to move all the counting methods over to this new header and to deprecate all these recursive tree methods.~~ Edit: after some discussion we decided to call these tools directly from the Tree methods - for example `Tree::activeVoxelCount()` now calls `tools::countActiveVoxels()`.


openvdb/openvdb/tools/Count.h
```diff
@@ -0,0 +1,83 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+//
+/// @file   Count.h
```
Idclip, line 4: Maybe this should live in FindActiveValues? Or the two can be combined?

Idclip:
> Performance on a 1 billion voxel VDB (32 threads):
> 
> Original: 203 milliseconds
> New: 26 milliseconds

Do you have timings for smaller/other size data sets? Is it always faster?

danrbailey:
> > Performance on a 1 billion voxel VDB (32 threads):
> > Original: 203 milliseconds
> > New: 26 milliseconds
> 
> Do you have timings for smaller/other size data sets? Is it always faster?

The break-even point is around 3 million voxels when total compute is < 0.2ms. Here's an idea of how these algorithms scale:

```
      750,000 voxels -   0.057ms vs  0.15ms
    3,000,000 voxels -   0.19ms  vs  0.18ms
    7,000,000 voxels -   0.40ms  vs  0.27ms
   37,000,000 voxels -   5.8ms   vs  0.92ms
  109,000,000 voxels -  18ms     vs  3.2ms
  302,000,000 voxels -  56ms     vs  8.4ms
  820,000,000 voxels - 170ms     vs 22ms
1,000,000,000 voxels - 200ms     vs 26ms
```

Note that I experimented with changing the grain size, but there's currently no way to use one grain size for the internal nodes and a different one of the leaf nodes, so in the end leaving it as 1 seemed optimum.

openvdb/openvdb/tools/Count.h
danrbailey: This is a good question, I was thinking about combining them. Tagging @kmuseth for visibility.

I've just pushed up a change to rename `tools::activeVoxelCount` to `tools::countActiveVoxels` and migrated just one of the methods from FindActiveValues - the `tools::countActiveValues` method. I renamed that to `tools::countActiveVoxels` and rewrote it to use the DynamicNodeManager as well, so these are the two signatures:

```
template <typename TreeT>
Index64 countActiveVoxels(const TreeT& tree, bool threaded = true);
template <typename TreeT>
Index64 countActiveVoxels(const TreeT& tree, const CoordBBox& bbox, bool threaded = true);
```

I tested the performance of this new method against the FindActiveValues one for three different cases:

```
Entire BBox -  FindActiveValues 290ms  vs Count 28ms
Tiny BBox -    FindActiveValues 0.02ms vs Count 0.43ms
Partial BBox - FindActiveValues 59ms   vs Count 9.3ms
```

It's not entirely a fair test because I noticed a significant optimization for the test case I was using (a sphere level set) which is to handle the common case where the leaf node bounding box is entirely contained by the testing bounding box. If I apply that optimization to FindActiveValues, they're much closer:

```
Entire BBox - FindActiveValues 39ms vs Count 28ms
```

It doesn't do everything that the FindActiveValues header does, but it's a lot less code - I'm happy to see the DynamicNodeManager working well for this use case too.

Finally, I added `tools::memUsage` to this Count header and achieved a similar speed up there too:

`Tree::memUsage 184ms vs tools::memUsage 22ms`

I _think_ it makes sense to put this method here as it is counting the bytes of memory it occupies.

jmlait (approved):

danrbailey:
Updated this PR with the changes we discussed last week. `Tree::activeVoxelCount()` and `Tree::memUsage()` now use the multi-threaded tools directly and I've marked the `RootNode` and `InternalNode` methods as deprecated. This also removes some of the bbox counting implementation in `FindActiveValues.h` and marks `tools::countActiveValues` as deprecated (in favour of `tools::countActiveVoxels`).

danrbailey:
I believe this is ready to go in. Jeff has already approved, but I'll give this a week until our next TSC meeting to allow anyone else to comment before I finally merge it.

danrbailey:
I have modified this PR to address the feedback from our discussion this week - reverting changes to remove and deprecate methods in FindActiveValues.h and reverting deprecations from `onVoxelCount()` methods in RootNode and InternalNode. I have left `tools::countActiveValues()` calling `tools::countActiveVoxels()` and applied the optimization I discovered to `FindActiveValues::count`.

openvdb/openvdb/tools/Count.h
```diff
@@ -0,0 +1,235 @@
...
+Index64 memUsage(const TreeT& tree, bool threaded = true);
+
+
+////////////////////////////////////////
+
+
+namespace count_internal {
+
+template<typename TreeType>
+struct ActiveVoxelCountOp
```
kmuseth, line 45: please add documentation for all the member methods of this struct

openvdb/openvdb/tools/Count.h
```diff
@@ -0,0 +1,235 @@
...
+    void join(const ActiveVoxelCountOp& other)
+    {
+        count += other.count;
+    }
+
+    openvdb::Index64 count{0};
+};
+
+template<typename TreeType>
+struct ActiveVoxelCountBBoxOp
```
kmuseth, line 76: this struct is also missing documentaiton

openvdb/openvdb/tools/Count.h
```diff
@@ -0,0 +1,235 @@
...
+        count += other.count;
+    }
+
+    openvdb::Index64 count{0};
+private:
+    CoordBBox mBBox;
+};
+
+template<typename TreeType>
+struct MemUsageOp
```
kmuseth, line 155: same here

kmuseth (approved):
please add missing documentation

danrbailey:
> please add missing documentation

I've added some more documentation to these internally namespaced classes now.

PR #1013, opened 2021-03-23
CI Sanitizers

Idclip:
Adds support for address, memory, thread and UB sanitizers and reworks the coverage option

An example run: https://github.com/Idclip/openvdb/actions/runs/679869827

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,43 @@
+
+# This workflow is for running various sanitizers on OpenVDB. We currently
+# only use clang over gcc (for no particular reason). Note that some of the
+# errors produces by these sanitizers may be fairly obscure as the
+# llvm-symbolizer is not installed on the openvdb image. The address
+# sanitizer should never fail.
+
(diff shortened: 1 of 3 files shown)


.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,43 @@
...
+          #- { build: 'tsan', cmake: '' } # requires full stack rebuild for valid results (including libc++)
+          #- { build: 'msan', cmake: '' } # requires full stack rebuild for valid results (including libc++)
+          #- { build: 'coverage', cmake: '' } # todo
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@v1
+    - name: install_gtest
+      run: ./ci/install_gtest.sh 1.8.0
+    - name: build
+      run: ./ci/build.sh ${{ matrix.config.build }} 8 ON None "core,test" \
```
Idclip, line 38: AX and python components to add (some issues in AX which I will fix before adding here)

danrbailey (approved):

PR #1021, opened 2021-03-30
Updated AX example docs

Idclip:
- Improved AX examples in doxygen documentation
 - Updated `ci/extract_test_examples.sh` to also compile AX code in documentation (would previously just parse)

@richhones


danrbailey (approved):
Very neat, would be good to have something like this for all our code samples too. :)

PR #1024, opened 2021-04-08
DynamicNodeManager Grain Size

danrbailey:
After some consideration, I decided to make the current grain size apply only to threading across leaf nodes and to add an additional grain size to thread across non-leaf nodes.

For most cases, adjusting the grain size for the leaf node threading is what you want, but at least with this implementation there's the option to provide an additional grain size for internal node threading rather than having this hard-coded to any particular value. By default, non-leaf node threading is set to 1.

### DIFF
--- a/openvdb/openvdb/tree/NodeManager.h
+++ b/openvdb/openvdb/tree/NodeManager.h
@@ -808,41 +808,45 @@ public:
     DynamicNodeManagerLink() = default;
 
     template<typename NodeOpT, typename RootT>
-    void foreachTopDown(const NodeOpT& op, RootT& root, bool threaded, size_t grainSize)
+    void foreachTopDown(const NodeOpT& op, RootT& root, bool threaded,
+        size_t leafGrainSize, size_t nonLeafGrainSize)
     {
         if (!op(root, /*index=*/0))         return;
         if (!mList.initRootChildren(root))  return;
         ForeachFilterOp<NodeOpT> filterOp(op, mList.nodeCount());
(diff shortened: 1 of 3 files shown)


jmlait (commented):
Does this fix the problem of having 8 root nodes with a trivial tree?  Don't we still then always spawn 8 threads even though the total leaf count may be small?

(I guess we can also argue we should fix the 8-root node probelm anyways, as that probably is a big speed win...)

danrbailey:
> Does this fix the problem of having 8 root nodes with a trivial tree? Don't we still then always spawn 8 threads even though the total leaf count may be small?
> 
> (I guess we can also argue we should fix the 8-root node probelm anyways, as that probably is a big speed win...)

If you were to choose a grain size of 8 for non-leaf nodes and 1 for leaf nodes, then yes. It would dispatch just 1 thread for the 8 root node children but still parallelize across the leaf nodes with 1 thread per leaf. Is that what you meant?

jmlait (approved):
I think my objections still stand, but the real problem is that we have 8 root nodes when we shoudn't.  Hopefully we fix that at some point and my issues are moot, so I think I shouldn't be holding this PR back.

danrbailey:
> I think my objections still stand, but the real problem is that we have 8 root nodes when we shoudn't. Hopefully we fix that at some point and my issues are moot, so I think I shouldn't be holding this PR back.

Thanks @jmlait, yes agreed fixing the 8 root node children would reduce the threads being spawned for this use case.

PR #1025, opened 2021-04-08
Add optional CMake variable to build a subset of the unit tests

danrbailey:
I very frequently do not wish to build all the unit tests and only wish to iterate on one or even a few. Right now, this requires modifying the CMakeLists.txt for the unit tests and carefully managing this edit, so as not to commit this change.

This PR adds a new optional CMake variable to define the unit tests you wish to build such as:

cmake -DOPENVDB_TESTS=NodeManager
cmake -DOPENVDB_TESTS="Activate;NodeManager"

If this variable isn't provided by the user, all the unit tests will be built. You'll note that this list assumes the format "Test<Name>.cc" which is our convention.

### DIFF
--- a/openvdb/openvdb/unittest/CMakeLists.txt
+++ b/openvdb/openvdb/unittest/CMakeLists.txt
@@ -12,6 +12,12 @@ project(OpenVDBUnitTests LANGUAGES CXX)
 
 include(GNUInstallDirs)
 
+set(OPENVDB_TESTS "" CACHE STRING [=[
+  An optional list of unit tests to build. If not provided, all unit tests will be built.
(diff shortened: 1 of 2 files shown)


Idclip (approved):

jmlait (approved):
... i guess since CMake doesn't glob easily, this can't easily take glob patterns?

danrbailey:
> ... i guess since CMake doesn't glob easily, this can't easily take glob patterns?

Good thinking. Actually, CMake does have file glob functionality - I made a small change to this to expand the file names using glob. In addition to the above instantiations, you can now also do:

`cmake -DOPENVDB_TESTS="Leaf*"`

which will build all the tests that match TestLeaf*.cc

PR #1030, opened 2021-04-14
Fix PointData and PointIndex aliases

danrbailey:
This fixes all cases where PointIndexTree, PointDataGrid and PointDataTree were used incorrectly instead of the templated T type and also renames any templated typenames that would override these aliases.

Finally, I have added a new unit test that builds a custom PointDataGrid with a different value type and leaf node size and added a macro to disable the PointDataGrid and PointDataTree aliases to ensure that the functionality in the headers only ever uses templated types.

### DIFF
--- a/openvdb/openvdb/points/PointAttribute.h
+++ b/openvdb/openvdb/points/PointAttribute.h
@@ -284,7 +284,7 @@ inline void appendAttribute(PointDataTreeT& tree,
 
     tree::LeafManager<PointDataTreeT> leafManager(tree);
     leafManager.foreach(
-        [&](typename PointDataTree::LeafNodeType& leaf, size_t /*idx*/) {
+        [&](typename PointDataTreeT::LeafNodeType& leaf, size_t /*idx*/) {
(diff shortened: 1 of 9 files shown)


openvdb/openvdb/points/PointAttribute.h
```diff
@@ -67,7 +67,7 @@ inline void appendAttribute(PointDataTreeT& tree,
 /// @param transient          mark attribute as transient
 template <typename ValueType,
           typename CodecType = NullCodec,
-          typename PointDataTreeT = PointDataTree>
+          typename PointDataTreeT>
```
Idclip, line 70: I guess we're not calling this with <=1 template args, but I'd leave this defaulted

danrbailey: It is a bit surprising, I had thought that all template parameters had to be default from the first one you make default onwards. It seems that's not the case for function template where the template type can be deduced.

Idclip (approved):

PR #1031, opened 2021-04-14
VDB Merge

danrbailey:
This PR extends `tools/merge.h` to add a new `SumMergeOp` for parallel breadth-first summing of multiple aligned VDBs using the DynamicNodeManager. The implementation carefully preserves the order of operations to ensure the same result in spite of the non-associativity of floating-point addition. As an example, collapsing a constant value leaf node into a tile of one of the input grids will not change the resulting merged grid.

This also adds a new VDB Merge SOP which partially implements the prototype from #623. At present, the only two options available on this SOP are for merging SDF VDBs and fog volume VDBs. This implementation works well in parallel when merging a few very large VDBs and a lot of very small VDBs or some combination of both and can deep-copy and/or resample concurrently with the merging of other grids.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
@@ -70,7 +70,7 @@ newSopOperator(OP_OperatorTable* table)
         .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
             "name",             "Name",
             "primitive_number", "Primitive Number",
-            "none",             "None"
+            "all",              "All"
         })
         .setTooltip(
             "Criteria under which groups of grids are merged.")
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
@@ -66,26 +66,33 @@ newSopOperator(OP_OperatorTable* table)
             " (see [specifying volumes|/model/volumes#group])"));
 
     parms.add(hutil::ParmFactory(PRM_STRING, "collation", "Collation")
-        .setDefault("nameclassandtype")
+        .setDefault("name")
         .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
-            "nameclassandtype", "Name, VDB Class and Value Type",
-            "classandtype",     "VDB Class and Value Type"
+            "name",             "Name",
+            "primitive_number", "Primitive Number",
+            "none",             "None"
         })
         .setTooltip(
             "Criteria under which groups of grids are merged.")
         .setDocumentation(
-            "The criteria under which groups of grids are merged.\n\n"
-            "Grids with different names will be merged if name is not part"
-            " of the collation criteria"));
+            "The criteria under which groups of grids are merged. In addition to these collation"
+            " options, only grids with the same class (fog volume, level set, etc) and"
+            " value type (float, int, etc) are merged.\n\n"
+            "Name:\n"
+            "   Collate VDBs with the same grid name.\n\n"
+            "Primitive Number:\n"
+            "   Collate first primitives from each input, then second primitives from each input, etc.\n\n"
+            "None:\n"
+            "   Collate all VDBs (provided they have the same class and value type).\n\n"));
 
-    // Menu of resampling options
-    parms.add(hutil::ParmFactory(PRM_STRING, "resample", "Resample VDBs")
-        .setDefault("first")
-        .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
-            "first",        "To Match First VDB"
-        })
-        .setTypeExtended(PRM_TYPE_JOIN_PAIR)
-        .setDocumentation("Specify which grid to use as reference when resampling."));
+    // // Menu of resampling options
+    // parms.add(hutil::ParmFactory(PRM_STRING, "resample", "Resample VDBs")
+    //     .setDefault("first")
+    //     .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
+    //         "first",        "To Match First VDB"
+    //     })
+    //     .setTypeExtended(PRM_TYPE_JOIN_PAIR)
+    //     .setDocumentation("Specify which grid to use as reference when resampling."));
 
     // Menu of resampling interpolation order options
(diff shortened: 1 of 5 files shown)


openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,682 @@
...
+        .setChoiceList(&hutil::PrimGroupMenuInput1)
+        .setDocumentation(
+            "A subset of the input VDBs to be modified"
+            " (see [specifying volumes|/model/volumes#group])"));
+
+    parms.add(hutil::ParmFactory(PRM_STRING, "collation", "Collation")
+        .setDefault("nameclassandtype")
+        .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
+            "nameclassandtype", "Name, VDB Class and Value Type",
+            "classandtype",     "VDB Class and Value Type"
```
jmlait, line 72: Since we are doing germancase here I'm not sure why we have the embedded and
I'd prefer
nameclasstype
classtype
Or maybe since these are tokens anyways,
name_class_type
class_type

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,682 @@
...
+
+    // Menu of resampling options
+    parms.add(hutil::ParmFactory(PRM_STRING, "resample", "Resample VDBs")
+        .setDefault("first")
+        .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
+            "first",        "To Match First VDB"
+        })
+        .setTypeExtended(PRM_TYPE_JOIN_PAIR)
+        .setDocumentation("Specify which grid to use as reference when resampling."));
+
```
jmlait, line 89: Do we need to add this now if it doesn't have any options yet?  Isn't it safe just to not have it; and add it with First as the default when we have more than one choice?

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,682 @@
...
+        .setDefault("add")
+        .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
+            "none",             "None            ",
+            "add",              "Add"
+        })
+        .setTooltip("Merge operation for Fog VDBs.")
+        .setDocumentation(
+            "Merge operation for Fog VDBs.\n\n"
+            "None:\n"
+            "   Add grids to the output of the SOP without merging them.\n\n"
```
jmlait, line 122: Weird to have two meanings of Add here.
"Passes the grids to the output without affecting them"
perhaps?
"Leaves input fog grids unchanged"

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,682 @@
...
+            "Add:\n"
+            "    Generate the sum of all fog VDBs with same collation.\n\n"));
```
jmlait, line 124: Perhaps... "...fog VDBS within the same collation."

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,682 @@
...
+\n\
+\"\"\"Merge VDB SDF grids.\"\"\"\n\
+\n\
+@overview\n\
+\n\
+Merges VDB SDF grids.\n\
+\n\
+Unlike the VDB Combine SOP, it provides a multi-input so can merge more than two grids without needing an \
+additional merge SOP. \n\
+Float and double VDB SDF grids can be merged, all other VDB grids are added to the output without merging. \n\
```
jmlait, line 165: Again, i don't see it as "adding to the output" so much as the node leaving them unchanged.  (Technically, adding to  the output is what we are doing, but people have many mental models of what actually happens :>)

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
jmlait: One trivial collation type is "Primitive Number".   Ie, collate the first primitive of all inputs, then the second, etc.  Useful perhaps when people are lazy about naming.

danrbailey: Yes, good idea. I took a quick look at doing this and thought it was going to be much more complicated to implement without some significant refactoring then realized I could simply store the primitive number in a hash map keyed by the GEO_PrimVDB::getUniqueId().

After some hesitation around this parameter, I decided it might be better to drop the "VDB Class and Value Type" part for each option and simply make that something that applies to each collation, so it now reads:
* Name
* Primitive Number
* None

If and when I implement something that can merge grids of different value types (ie merging a float grid with a double grid, etc), then this parameter will probably need to change, but this is not a trivial thing to implement anyway so best to keep it simple at this stage I think.

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
danrbailey: Fair point, I've made this an obsolete parameter for the time being.

danrbailey:
@jmlait - thanks for the feedback, lots of good improvements. I've addressed all your comments now, would be good to know what you think about these edits.

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Merge.cc
```diff
@@ -0,0 +1,712 @@
...
+        .setDocumentation(
+            "A subset of the input VDBs to be modified"
+            " (see [specifying volumes|/model/volumes#group])"));
+
+    parms.add(hutil::ParmFactory(PRM_STRING, "collation", "Collation")
+        .setDefault("name")
+        .setChoiceListItems(PRM_CHOICELIST_SINGLE, {
+            "name",             "Name",
+            "primitive_number", "Primitive Number",
+            "none",             "None"
```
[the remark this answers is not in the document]
danrbailey: Yes, I think you're right, "All" is more intuitive for this use case. I can change that.

PR #1032, opened 2021-04-14
Fix Rebuild LevelSet Metadata

danrbailey:
This fixes a bug where metadata was not being copied from the input grid when doing tools::levelSetRebuild().

This showed up in Houdini where using the VDB Rebuild SDF SOP was resulting in a VDB primitive that was returning a value of "" for the vdb_value_type intrinsic instead of "float".

I also added a TestLevelSetRebuild unit test because this method was only being tangentially tested through TestFastSweeping. For now it only calls one of these methods and tests that metadata persists but it may be worth expanding this to validate the API properly.

### DIFF
--- a/openvdb/openvdb/tools/LevelSetRebuild.h
+++ b/openvdb/openvdb/tools/LevelSetRebuild.h
@@ -245,13 +245,18 @@ doLevelSetRebuild(const GridType& grid, typename GridType::ValueType iso,
 
     QuadAndTriangleDataAdapter<Vec3s, Vec4I> mesh(points, primitives);
 
+    typename GridType::Ptr newGrid;
+
     if (interrupter) {
-        return meshToVolume<GridType>(*interrupter, mesh, *transform, exBandWidth, inBandWidth,
+        newGrid = meshToVolume<GridType>(*interrupter, mesh, *transform, exBandWidth, inBandWidth,
+            DISABLE_RENORMALIZATION, nullptr);
+    } else {
+        newGrid = meshToVolume<GridType>(mesh, *transform, exBandWidth, inBandWidth,
             DISABLE_RENORMALIZATION, nullptr);
     }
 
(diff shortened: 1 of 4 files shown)


Idclip:
We can't use this solution. There are a number of algorithms across VDB where metadata copying is ignored - this behaviour is probably relied upon somewhere. Additionally, I've called this method internally plenty of times without caring about the state of the grid metadata - I don't want to pay this penalty. 

You can either add a new argument which defaults to the old behaviour, or add a new signature which takes a tree and a transform which avoids this problem entirely.

I suggest adding a new argument which defaults to ignoring

danrbailey:
> We can't use this solution. There are a number of algorithms across VDB where metadata copying is ignored - this behaviour is probably relied upon somewhere. Additionally, I've called this method internally plenty of times without caring about the state of the grid metadata - I don't want to pay this penalty.
> 
> You can either add a new argument which defaults to the old behaviour, or add a new signature which takes a tree and a transform which avoids this problem entirely.

I agree with you. Looking at the documentation around the core library API, it is clearly stating that this is a new grid:
```
/// @brief Return a new grid of type @c GridType that contains a narrow-band level set
/// representation of an isosurface of a given grid.
```
so it is expected that this new grid would not have the metadata carried across from the older one.

However, I think in this case, the expectations are different in the client code (Houdini), but I'll open a new PR to address the issue there.

Incidentally, just to highlight that you can use the new Grid methods to very cheaply drop metadata:
```
/// @brief Return a new grid of the same type as this grid whose tree and transform
/// is shared with this grid and whose metadata is provided as an argument.
GridBase::ConstPtr Grid::copyGridReplacingMetadata(const MetaMap& meta) const;
```

There are a lot of algorithms that do a deepCopy under the hood that will carry the metadata with them, in case you didn't want to pay that penalty there.

PR #1033, opened 2021-04-14
Fix VDB Rebuild SDF SOP Metadata

danrbailey:
This showed up in Houdini where using the VDB Rebuild SDF SOP was resulting in a VDB primitive that was returning a value of "" for the vdb_value_type intrinsic instead of "float".

I believe the expectation here is that the VDB Rebuild SDF SOP should behave like the VDB Renormalize SDF SOP and carry across metadata from the input grid. The language here "Repair level sets represented by VDB volumes" leads me to believe that this SOP is modifying the input grid rather than creating a new grid. 

The alternative is perhaps to follow the example of SOPs like the VDB Resample SOP which copies across attributes and group membership by using `hvdb::replaceVdbPrimitive()` and then reconstructs the metadata.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Rebuild_Level_Set.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Rebuild_Level_Set.cc
@@ -292,8 +292,11 @@ SOP_OpenVDB_Rebuild_Level_Set::Cache::cookVDBSop(
 
                 openvdb::FloatGrid& grid = UTvdbGridCast<openvdb::FloatGrid>(vdbPrim->getGrid());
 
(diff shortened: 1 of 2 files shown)


Idclip:
> The language here "Repair level sets represented by VDB volumes" leads me to believe that this SOP is modifying the input grid rather than creating a new grid.

Agreed. With this mindset, `rebuildLevelSet` should really take a reference to a grid rather than return a new one. Most methods that do this preserve the grid data and only modify the tree.

Idclip (approved):

Idclip:
Approved, though I'd suggest waiting for approval from @jmlait too

apradhana (approved):

jmlait (approved):

PR #1037, opened 2021-04-22
Re-introduced the OPENVDB_DEPRECATED macros

Idclip:

This macro controls the C++ standard `[[deprecated]]` attribute's behaviour.

### DIFF
--- a/openvdb/openvdb/Platform.h
+++ b/openvdb/openvdb/Platform.h
@@ -93,13 +93,22 @@
     #define OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
 #endif
 
+/// Deprecation macros. Define OPENVDB_NO_DEPRECATION_WARNINGS to disable all
+/// deprecation warnings in OpenVDB.
+#ifndef OPENVDB_NO_DEPRECATION_WARNINGS
+#define OPENVDB_DEPRECATED [[deprecated]]
+#define OPENVDB_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
+#else
+#define OPENVDB_DEPRECATED
+#define OPENVDB_DEPRECATED_MESSAGE(msg)
+#endif
 
 /// @brief Bracket code with OPENVDB_NO_DEPRECATION_WARNING_BEGIN/_END,
 /// to inhibit warnings about deprecated code.
 /// @note Use this sparingly.  Remove references to deprecated code if at all possible.
 /// @details Example:
 /// @code
-/// [[deprecated]] void myDeprecatedFunction() {}
+/// OPENVDB_DEPRECATED void myDeprecatedFunction() {}
 ///
 /// {
 ///     OPENVDB_NO_DEPRECATION_WARNING_BEGIN
--- a/openvdb/openvdb/points/AttributeArray.h
+++ b/openvdb/openvdb/points/AttributeArray.h
@@ -146,7 +146,7 @@ public:
 
     /// Return a copy of this attribute.
 #ifndef _MSC_VER
-    [[deprecated("In-memory compression no longer supported, use AttributeArray::copy() instead")]]
+    OPENVDB_DEPRECATED_MESSAGE("In-memory compression no longer supported, use AttributeArray::copy() instead")
 #endif
     virtual AttributeArray::Ptr copyUncompressed() const = 0;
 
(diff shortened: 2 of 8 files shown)


openvdb/openvdb/Platform.h
```diff
@@ -93,13 +93,22 @@
...
 
+/// Deprecation macros. Define OPENVDB_NO_DEPRECATION_WARNINGS to disable all
+/// deprecation warnings in OpenVDB.
+#ifndef OPENVDB_NO_DEPRECATION_WARNINGS
+#define OPENVDB_DEPRECATED [[deprecated]]
+#define OPENVDB_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
+#else
+#define OPENVDB_DEPRECATED
+#define OPENVDB_DEPRECATED_MESSAGE(msg)
+#endif
```
Idclip, line 104: This does **not** change the behaviour of the `OPENVDB_NO_DEPRECATION_WARNING_BEGIN` and `OPENVDB_NO_DEPRECATION_WARNING_END` macros. That is, should you `#define OPENVDB_NO_DEPRECATION_WARNINGS` somewhere, the `BEGIN`/`END` macros will still produce pragma messages. These pragmas are not tied to the `-Wdeprecated-declarations` compiler options.

Perhaps it should also suppress those too? Then we have a macro to suppress a macro which suppress deprecation warnings 💯 I personally hate #pragma messages, IMO we should remove the BEGIN/END defines too and simply allow 
`OPENVDB_NO_DEPRECATION_WARNINGS` to supress VDB deprecations. The only benefit of the BEGIN/END macros is that you could potentially supress upstream deprecations... though I don't think that's actually a benefit :\

jmlait: Isn't BEGIN/END about being able to locally accept specific deprecated calls? If one just used this for that, you'd not detect any new deprecations in a file when you upgrade.

I personally don't like the #pragma output either, I think if you have an explicit BEGIN/END block you've already paid your price.  Having the compiler *also* spam output is excessive.

jmlait (approved):

danrbailey (approved):
Thanks for doing this Nick, this has been on my list for a while. 

Does this not need to be exposed in the CMake as well?

openvdb/openvdb/Platform.h
Idclip: > Isn't BEGIN/END about being able to locally accept specific deprecated calls?

You mean the case where a VDB function wants to use deprecated VDB code? e.g:

```
[[deprecated]] void openvdb::foo();
void openvdb::bar() { 
OPENVDB_NO_DEPRECATION_WARNING_BEGIN 
openvdb::foo(); 
OPENVDB_NO_DEPRECATION_WARNING_END
}
```

In theory the above *should* never happen, what with all our discussions regarding only marking things as deprecated when there are appropriate replacements. Indeed, we currently only use these macros in 3 places, twice in the unit tests and once in SOP_NodeVDB for `duplicateSourceStealable` (though I was also against this). The unit tests should probably just be built with `-Wno-error=deprecated-declarations` as there are occasions where we still want to test deprecated methods. I may have misunderstood what you meant here though.

Either way, I can leave these in for now.

Idclip:
> Does this not need to be exposed in the CMake as well?

I don't think this is necessarily required. This option is not specifically designed for use when *building* VDB (although it can be leveraged there as well), it's designed to provide downstream projects more control with VDB's deprecations. Putting it in our CMake for the VDB build is slightly awkward as, IMO, the appropriate way to solve that is to build with `-Wno-error=deprecated-declarations`.

openvdb/openvdb/Platform.h
jmlait: I agree VDB functions shouldn't use deprecated VDB.  But I use that OPENVDB_NO_DEPRECATION_WARNING_BEGIN when the external, non-VDB, function wants to use deprecated VDB.  Without it I'd have to fully turn off deprecations just because there are 1 or 2 things that can't be swept yet.

Idclip: So the scenario you're describing is that you can't update your code immediately but don't want to define `OPENVDB_NO_DEPRECATION_WARNINGS` because you don't want new warnings to be silenced should you update again with your code still not "fixed"? That makes sense, but what I'm proposing is that these macros should be part of the downstream implementation and not VDBs responsibility. As above, there should in theory be zero use for them in VDB itself.

danrbailey:
> > Does this not need to be exposed in the CMake as well?
> 
> I don't think this is necessarily required. This option is not specifically designed for use when _building_ VDB (although it can be leveraged there as well), it's designed to provide downstream projects more control with VDB's deprecations. Putting it in our CMake for the VDB build is slightly awkward as, IMO, the appropriate way to solve that is to build with `-Wno-error=deprecated-declarations`.

Right, that makes sense.

PR #1038, opened 2021-04-23
Fix VDB Resample SOP - Voxel Size + Voxel Scale

danrbailey:
Both of these modes have text such as this in the help:

> Using Voxel Scale Only
> 
> Keep the transform of the input VDB but scale the voxel size, increasing or decreasing the resolution.

However, the VDB Resample SOP was always creating a new linear transform and applying the voxel size/scale. This fix takes the transform from the input VDB as I believe was intended.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Resample.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Resample.cc
@@ -48,6 +48,7 @@ public:
     class Cache: public SOP_VDBCacheOptions { OP_ERROR cookVDBSop(OP_Context&) override; };
 
 protected:
+    void syncNodeVersion(const char*, const char*, bool*) override;
     void resolveObsoleteParms(PRM_ParmList*) override;
     bool updateParmsFlags() override;
 };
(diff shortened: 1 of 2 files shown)


danrbailey:
I've just pushed a change to this PR to add back in the axis-aligned linear transform option for voxel size and voxel scale based on our discussion in the TSC this week. This also adds syncNodeVersion() behavior to enable this new toggle so as to preserve the current behavior for nodes created prior to 8.1.0.

When putting down a new node, the behavior is to apply the voxel size changes to the input VDB transform.

danrbailey:
@jmlait - we didn't finally agree on the new for the new UI parameter, so I went with "linearxform" in the end which is disabled unless the mode is voxel size or voxel scale, let me know if you think that is clear or not:
axisalignedlineartransform

jmlait:
> @jmlait - we didn't finally agree on the new for the new UI parameter, so I went with "linearxform" in the end which is disabled unless the mode is voxel size or voxel scale, let me know if you think that is clear or not:
> axisalignedlineartransform

Internal name linearxform is fine.  Label doesn't sound like it is doing anything.
Maybe:
"Force Axis Aligned Transform"

Or maybe drop the transform as I'm not thinking of VDBs having transforms; it is that they are transformed.
"Force Axis Aligned"
"Make Axis Aligned"

I think the linear transform is there to suggest that it destroys taper.  But I'd contend trying to be too precise adds to confusion here as people aren't sure what linear transform means.   Axis Aligned implies no taper (as you can't have taper and be axis aligned) so I think it is sufficient to have that, and maybe have the help explain about taper, etc.

jmlait (approved):

PR #1042, opened 2021-04-27
AX feature PR

Idclip:
New Features:
 - Significant performance improvements to AX volume kernels, primarily due to
 improved vectorisation. Performance scales depending on the available host CPU
 instruction set and scalability of the AX kernel. Trivial assignments profile
 upwards of 3x faster on AVX.
 - Introduced Active Tile Streaming for volumes kernels. AX can now dynamically
 expand and collapse active nodes at all levels of a VDB tree on demand, switching
 on or off by checking the spatial access pattern of AX programs.

Improvements:
  - Refactored the backend AX representation of strings with SSO support and heap allocation.
  - Added new methods on the VolumeExecutable to control active tile streaming and node execution levels
  - The deletepoint() AX function is now natively supported by the PointExecutable

Houdini:
  - The AX SOP utilises the new Active Tile Streaming feature for Volumes. Grids
  will only be densified in areas where unique values are created.

Bug Fix:
  - Fixed a bug which could cause string allocations in loops to overflow the stack
  - Fixed a bug where the ValueOff iterator for the VolumeExecutable could remove
  child branches of a VDB tree.

### DIFF
--- a/openvdb_ax/openvdb_ax/CMakeLists.txt
+++ b/openvdb_ax/openvdb_ax/CMakeLists.txt
@@ -111,6 +111,7 @@ set(OPENVDB_AX_LIBRARY_SOURCE_FILES
   codegen/PointComputeGenerator.cc
   codegen/PointFunctions.cc
   codegen/StandardFunctions.cc
+  codegen/StringFunctions.cc
   codegen/Types.cc
   codegen/VolumeComputeGenerator.cc
   codegen/VolumeFunctions.cc
@@ -148,6 +149,7 @@ set(OPENVDB_AX_CODEGEN_INCLUDE_FILES
   codegen/FunctionTypes.h
   codegen/PointComputeGenerator.h
   codegen/PointLeafLocalData.h
+  codegen/String.h
   codegen/SymbolTable.h
   codegen/Types.h
   codegen/Utils.h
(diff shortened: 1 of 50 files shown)


danrbailey (approved):
Very cool.

Idclip:
Thanks @danrbailey! Waiting another week or so with these so they can be completed tested. If you find some spare time I'd love to hear any feedback you have on the performance/tile streaming feature (if you can run the same tests you were running).

Idclip:
The vectorisation changes are specific to the volume kernels that are JIT'ed. You can test this performance in isolation of the active tile changes by running AX **only** across leaf nodes in a volume i.e. run some AX before these changes on a VDB which only contains leaf nodes, then run the same AX with these changes on the same input. Without going into too much detail, the invocations of the kernels themselves have been refactored so that LLVM's optimiser is better able to apply vectorisation across multiple instructions. 

> (I'm testing using SSE 4.2)

To make sure we're on the same page here; AX always targets the host architecture irrespective of what OpenVDB was compiled with. So if you're running AX kernels on a host with AVX, it'll utilise those instructions where possible even if OpenVDB was built with SSE42. Conversely if the host doesn't have any vector extension support, it won't apply any vectorisation (we'll expose this setting in the future, but control for this is only useful for testing or JIT cross compilation).

PR #1045, opened 2021-04-28
Support for worldspace dilation specification in the VDB Activate SOP.

jmlait:
VDB Activate is also switched to working with tiles, which changes behaviour
but the old behaviour is likely a bug.  Houdini's version will also make
this change.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
@@ -211,17 +211,27 @@ an inclusive range, so includes the maximum voxel.)"));
 
     parms.addFolder("Expand");
 /*
-    Expand the active area by the specified number of voxels.  Does not support
+    Expand the active area by at least the specified number of voxels.  Does not
     operation or setting of values.
 */
     parms.add(hutil::ParmFactory(PRM_INT, "expand", "Voxels to Expand")
                 .setDefault(PRMoneDefaults)
                 .setRange(PRM_RANGE_FREE, -5, PRM_RANGE_FREE, 5)
-                .setTooltip("Expand the active area by the specified number of voxels.")
+	      .setTooltip("Expand the active area by at least the specified number of voxels.")
                 .setDocumentation(
-R"(Expand the active area by the specified number of voxels.  Does not support
+R"(Expand the active area by at least the specified number of voxels.  Does not support
 operation or setting of values.)"));
 
+/*
+    Expand the active area by at least the specified distance. Does not support
+    operation or setting of values.
+*/
+    parms.add(hutil::ParmFactory(PRM_FLT, "expanddist", "Expand Distance")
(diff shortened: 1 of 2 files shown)


danrbailey:
Looks like the CI is failing due to some casting warnings as errors.

If I'm reading the code correctly, if "expanddist" and "expand" have different signs, will it not do a dilation followed by an erosion?

Idclip:
> Looks like the CI is failing due to some casting warnings as errors.

Also some whitespace errors

jmlait:
> Looks like the CI is failing due to some casting warnings as errors.

Yep, hopefully fixed along with the white space.

I really think this warning should not exist, it is a very dangerous warning.  The "fix" of adding static_cast<int> doesn't actually affect code safety and instead hard-codes the down conversion.  If some future version of OpenVDB were to change the dilate functions to take a 64bit int; this code would continue artificially constrict the type to 32bits.

> If I'm reading the code correctly, if "expanddist" and "expand" have different signs, will it not do a dilation followed by an erosion?

That is correct.  We spent some time wondering what the correct solution is in this case.   Importantly, we wanted 0 to be a no-op in all cases allowing one to set both and have it do the maximum, as that is a common requirement.  This seemed the best choice for when someone has a confused input.

Idclip:
> That is correct. We spent some time wondering what the correct solution is in this case. Importantly, we wanted 0 to be a no-op in all cases allowing one to set both and have it do the maximum, as that is a common requirement. This seemed the best choice for when someone has a confused input.

Some of the other SOPs have a worldspace/indexspace toggle which switches the available param. It would avoid this problem; is that something we want to consider here?

jmlait:
A suggestion at the meeting is that we should generate a warning when this happens.

jmlait:
I've added a warning if the signs mis-match.

Idclip:
> A suggestion at the meeting is that we should generate a warning when this happens.

Was my suggestion discussed? Happy either way, but am curious if you had specific advice for ws/is options like this for the future.

jmlait:
> > A suggestion at the meeting is that we should generate a warning when this happens.
> 
> Was my suggestion discussed? Happy either way, but am curious if you had specific advice for ws/is options like this for the future.

Yes, it was.  Forgot you weren't there so was confused by this message :>

The logic of having both active here is it isn't unusual to have two constraints on a field: to have a minimum voxel pad and a minimum distance pad, and want to enforce both.  Trying to do this with a max operation is back to the sort of expressions that led to wanting this parameter in the first place :> 

This is stuff like maybe a displacement noise, where you want to ensure there is always X fresh voxels, but also have a displacement bound to enforce...

Idclip:
> The logic of having both active here is it isn't unusual to have two constraints on a field: to have a minimum voxel pad and a minimum distance pad, and want to enforce both. Trying to do this with a max operation is back to the sort of expressions that led to wanting this parameter in the first place :>

understood, makes sense!

Idclip (approved):

PR #1046, opened 2021-04-29
Minor vdb_view improvements

Idclip:
- Fix a bug with vdb_view which wouldn't reset the camera speed on focus
- Removed unnecessary CMake for vdb_view 
- Improved the drawing of PointDataGrids
- Improved the exception handling of BufferObjects and added support for drawing an array without index values

Before/After with VDB Points:
image


danrbailey (approved):
Looks good to me, just some fixes needed for Mac.

PR #929, opened 2021-01-12
CMake RPATH and 3.12 updates

Idclip:

PR #1048, opened 2021-04-30
Nacho/fix_rpaths

nachovizzo:
# Descriptions

I'm working under Ubuntu 20.04 systems, more specifically using docker containers. The problem I had(only when using docker, not when trying in my local system) was that the binaries and the python library path were being set to just `lib`. Which of course is a directory that doesn't exist. Doing some criminology I found that this change was introduced in PR #929. Before this change, I can build and install with no problem. I always build the library with `cmake -DOPENVDB_BUILD_PYTHON_MODULE=ON`

### Before

-- Installing: /usr/local/lib/python3.8/dist-packages/pyopenvdb.so
-- Set runtime path of "/usr/local/lib/python3.8/dist-packages/pyopenvdb.so" to **"/usr/lib/x86_64-linux-gnu:lib"**
-- Installing: /usr/local/include/openvdb/python/pyopenvdb.h
-- Installing: /usr/local/bin/vdb_print
-- Set runtime path of "/usr/local/bin/vdb_print" to **"/usr/lib/x86_64-linux-gnu:lib"**

### After

-- Installing: /usr/local/lib/python3.8/dist-packages/pyopenvdb.so
-- Set runtime path of "/usr/local/lib/python3.8/dist-packages/pyopenvdb.so" to **"/usr/lib/x86_64-linux-gnu:/usr/local/lib"**
-- Installing: /usr/local/include/openvdb/python/pyopenvdb.h
-- Installing: /usr/local/bin/vdb_print
-- Set runtime path of "/usr/local/bin/vdb_print" to **"/usr/lib/x86_64-linux-gnu:/usr/local/lib"**

## Minimal CMake example

This is to show that this "problem" is not related to  `OpenVDB` but to CMake

```cmake
cmake_minimum_required(VERSION 3.16)
project(test_install)
include(GNUInstallDirs)
message(STATUS "CMAKE_LIBRARY_ARCHITECTURE=${CMAKE_LIBRARY_ARCHITECTURE}")
message(STATUS "CMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}")
```

## Background information

The [official documentation says](https://cmake.org/cmake/help/v3.16/module/GNUInstallDirs.html#result-variables):

`CMAKE_INSTALL_<dir>`

Destination for files of a given type. This value may be passed to the DESTINATION options of install() commands for the corresponding file type.

`CMAKE_INSTALL_FULL_<dir>`

The absolute path generated from the corresponding `CMAKE_INSTALL_<dir>` value. If the value is not already an absolute path, an absolute path is constructed typically by prepending the value of the `CMAKE_INSTALL_PREFIX` variable.

So basically the `CMAKE_INSTALL_FULL_<dir>` variable it's not needed for the `install()` commands, but it's needed for other purposes, like changing the `RPATH`

## Related issues

 - https://gitlab.kitware.com/cmake/cmake/-/issues/20565
 - https://gitlab.kitware.com/cmake/cmake/-/issues/20250


Idclip (approved):
Good catch @nachovizzo, thanks very much for the contribution. looks good to me

openvdb_houdini/openvdb_houdini/CMakeLists.txt
```diff
@@ -207,7 +207,7 @@ if(OPENVDB_ENABLE_RPATH)
     ${Tbb_LIBRARY_DIRS}
   )
   if(OPENVDB_BUILD_CORE)
-    list(APPEND "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+    list(APPEND RPATHS ${CMAKE_INSTALL_FULL_LIBDIR})
   else()
     list(APPEND ${OpenVDB_LIBRARY_DIRS})
```
nachovizzo, line 212: @Idclip I guess that the `RPATHS` is missing here, am I right?

Idclip: Yeah another good spot - our CI builds with RPATHS enabled so I'm surprised neither of these commands cause a CMake error :|

nachovizzo: Fixed on 1f506b8709!

danrbailey (approved):

PR #1050, opened 2021-05-04
Blind Data (ABI=9)

danrbailey:
Adding a transitive offset to the RootNode, InternalNode and LeafNodes. This is an ABI=9 feature and doesn't change the size of the nodes in memory, just adds a 32-bit integer to each.

I hesitated on the naming and decided to call this a "transitive offset" - we use offset quite frequently in various contexts and making the transitive nature of this explicit seemed like a good idea.

The other aspect I hesitated on is the size of the integer type. In this implementation, the underlying size is a 32-bit integer (so as not the need to expand the node), but input and output is a 64-bit integer. As mentioned previously, we can steal a few unused bits from the origin to stretch the size of the offset we can store - by making the interface 64-bit, this saves needing to make an ABI change at a later date.

### DIFF
--- a/openvdb/openvdb/tree/InternalNode.h
+++ b/openvdb/openvdb/tree/InternalNode.h
@@ -268,6 +268,18 @@ public:
     /// Set the grid index coordinates of this node's local origin.
     void setOrigin(const Coord& origin) { mOrigin = origin; }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transitive offset value.
+    Index64 transitiveOffset() const { return Index64(mTransitiveOffset); }
+    /// Set the transitive offset value.
+    void setTransitiveOffset(Index64 transitiveOffset)
+    {
+        // assert that offset does not exceed 32-bit limit
+        assert(transitiveOffset <= Index64(std::numeric_limits<Index32>::max()));
+        mTransitiveOffset = static_cast<Index32>(transitiveOffset);
+    }
+#endif
+
     Index32 leafCount() const;
     void nodeCount(std::vector<Index32> &vec) const;
     Index32 nonLeafCount() const;
@@ -818,6 +830,10 @@ protected:
     NodeMaskType mChildMask, mValueMask;
     /// Global grid index coordinates (x,y,z) of the local origin of this node
     Coord mOrigin;
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Transitive offset
+    Index32 mTransitiveOffset = 0;
+#endif
 }; // class InternalNode
 
 
@@ -897,10 +913,13 @@ struct InternalNode<ChildT, Log2Dim>::DeepCopy
 
 template<typename ChildT, Index Log2Dim>
 inline
-InternalNode<ChildT, Log2Dim>::InternalNode(const InternalNode& other):
-    mChildMask(other.mChildMask),
-    mValueMask(other.mValueMask),
-    mOrigin(other.mOrigin)
+InternalNode<ChildT, Log2Dim>::InternalNode(const InternalNode& other)
+    : mChildMask(other.mChildMask)
+    , mValueMask(other.mValueMask)
+    , mOrigin(other.mOrigin)
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    , mTransitiveOffset(other.mTransitiveOffset)
+#endif
 {
     DeepCopy<InternalNode<ChildT, Log2Dim> > tmp(&other, this);
 }
@@ -914,6 +933,9 @@ InternalNode<ChildT, Log2Dim>::InternalNode(const InternalNode<OtherChildNodeTyp
     : mChildMask(other.mChildMask)
     , mValueMask(other.mValueMask)
     , mOrigin(other.mOrigin)
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    , mTransitiveOffset(other.mTransitiveOffset)
+#endif
 {
     DeepCopy<InternalNode<OtherChildNodeType, Log2Dim> > tmp(&other, this);
 }
@@ -946,10 +968,13 @@ template<typename ChildT, Index Log2Dim>
 template<typename OtherChildNodeType>
 inline
 InternalNode<ChildT, Log2Dim>::InternalNode(const InternalNode<OtherChildNodeType, Log2Dim>& other,
-                                            const ValueType& background, TopologyCopy):
-    mChildMask(other.mChildMask),
-    mValueMask(other.mValueMask),
-    mOrigin(other.mOrigin)
+                                            const ValueType& background, TopologyCopy)
+    : mChildMask(other.mChildMask)
+    , mValueMask(other.mValueMask)
+    , mOrigin(other.mOrigin)
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    , mTransitiveOffset(other.mTransitiveOffset)
+#endif
 {
     TopologyCopy1<InternalNode<OtherChildNodeType, Log2Dim> > tmp(&other, this, background);
 }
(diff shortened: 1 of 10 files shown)


openvdb/openvdb/tree/RootNode.h
```diff
@@ -966,6 +978,10 @@ class RootNode
 
     MapType mTable;
     ValueType mBackground;
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Transient Data (not serialized)
+    Index32 mTransientData = 0;
+#endif
```
Idclip, line 984: @danrbailey as discussed, could you clarify why this is needed on the RootNode

openvdb/openvdb/tree/InternalNode.h
```diff
@@ -268,6 +268,18 @@ class InternalNode
     /// Set the grid index coordinates of this node's local origin.
     void setOrigin(const Coord& origin) { mOrigin = origin; }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transient data value.
+    Index64 transientData() const { return Index64(mTransientData); }
```
kmuseth, line 273: This API seems dangerous. Why expose it as uint64_t when in fact it's internally only uint32_t? I think it should be uint32_t all the way. Obviously uint64_t would be better but uint32_t incurs no extra memory so that's the right choice.

openvdb/openvdb/tree/LeafNode.h
```diff
@@ -182,6 +182,18 @@ class LeafNode
     /// Return the global coordinates for a linear table offset.
     Coord offsetToGlobalCoord(Index n) const;
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transient data value.
+    Index64 transientData() const { return Index64(mTransientData); }
```
kmuseth, line 187: Same comment here - use Index32 instead of Index64

openvdb/openvdb/tree/LeafNodeBool.h
```diff
@@ -164,6 +164,18 @@ class LeafNode<bool, Log2Dim>
     /// Return the global coordinates for a linear table offset.
     Coord offsetToGlobalCoord(Index n) const;
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transient data value.
+    Index64 transientData() const { return Index64(mTransientData); }
```
kmuseth, line 169: An same again, use Index32 instead of Index64

openvdb/openvdb/tree/LeafNodeMask.h
```diff
@@ -168,6 +168,18 @@ class LeafNode<ValueMask, Log2Dim>
     /// Return the global coordinates for a linear table offset.
     Coord offsetToGlobalCoord(Index n) const;
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transient data value.
+    Index64 transientData() const { return Index64(mTransientData); }
```
kmuseth, line 173: Same comment here - use Index32 instead of Index64

openvdb/openvdb/tree/RootNode.h
```diff
@@ -405,6 +405,18 @@ class RootNode
     /// Return the bounding box of this RootNode, i.e., an infinite bounding box.
     static CoordBBox getNodeBoundingBox() { return CoordBBox::inf(); }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the transient data value.
+    Index64 transientData() const { return Index64(mTransientData); }
```
kmuseth, line 410: Same comment here - use Index32 instead of Index64

kmuseth (changes requested):
Please see my inline comments

openvdb/openvdb/tree/RootNode.h
danrbailey: The RootNode _is_ a node. This change adds a 32-bit transient integer to every node for use with tools like `NodeManager` and `tools::visitNodesDepthFirst` that visit every node. Why would we exclude the RootNode?

openvdb/openvdb/tree/InternalNode.h
danrbailey: The main reason is because the lower three bits of the origin Coord can be "stolen" to increase the range of this transient data beyond 32-bits. By stealing these three bits, we can extend the range from 4.3 billion => 34.4 billion. By making this API return 64-bit integers now, we leave room to implement this extension at a later date without needing to change the API.

openvdb/openvdb/tree/RootNode.h
Idclip: > Why would we exclude the RootNode?

Because the point of this change was to take advantage of the space we know is in use? I don't know what 32bits of storage on the RootNode really gives us...

danrbailey: It gives us consistency across all node types in the tree hierarchy. One benefit is generic programming - you can write an algorithm that calls `NodeT::transientData()` without having to have one method for InternalNode and LeafNode and another for the RootNode which doesn't have this method. 

This change increases the memory footprint of a tree by just 4 bytes so I really don't see the problem.

Idclip: I'm not sure I agree with this - you already have to specialise RootNode iteration on iter types and things like node masks so the interface is not identical to Internal/Leaf Nodes. The size is irrelevant to me, could be 1 bit! Still seems redundant (to be honest I also don't see the need for it on the internal nodes, but if it's utilising used space then it doesn't matter). Perhaps @jmlait @kmuseth could comment.

danrbailey: Lots of things are different between the RootNode, InternalNode and LeafNode, but I frequently don't need to specialize for the RootNode because I'm only using the things that are the same. I ran into a case recently with some changes to the VDB Visualize SOP where I had to build and maintain an external array that would have been simpler and faster if I had a 32-bit integer per-node.

In my mind this is intended to be a flexible mechanism that we can't know exactly how people might wish to use it. Seeing as the cost (4 bytes) is so low, I see no reason not to offer full flexibility rather than just constrain this to leaf nodes. 

I don't think we have a zero redundancy policy for the tree hierarchy - after all the origin coordinates are technically redundant.

openvdb/openvdb/tree/InternalNode.h
kmuseth: Even if we use the three lower bits of the three 32 bit integers in mOrigin that's only 41 bits! The problem is that the current API suggests that points can be safely encoded (via a cast) as transient data, which is clearly not true. So for now it should be Index32 and if later on we change it to full 64 bit we can change it the API to Index64 since that will introduce an ABI change anyway.

danrbailey: (My maths was off there - fixed my above comments).

What you're suggesting is that we ignore those 9 available transient bits because they don't fit neatly into an API? As a user, I would much prefer to get access to those additional bits while knowing that calling `setTransientData(uint64_t)` will only use 41-bits of that uint64_t.

How about we expose an API like the following `setTransientData(uint64_t, bool checkBits=true)` and by default it checks that the upper 23 bits are zero at runtime? If you understand the risks, you can switch this check off for faster performance?

danrbailey: As agreed in our last meeting - we will use 32-bit indices as a change to 64-bit could be implemented as an API change only.

openvdb/openvdb/tree/RootNode.h
danrbailey: As agreed in our last meeting, we'll add the 32-bit index to the RootNode as well.

kmuseth (approved):

kmuseth:
@danrbailey  - thanks for addressing my review comments. I've approved. I have non-blocking request thought, What do you say to changing the name from "TransientData" to "BlindData"? It's much shorter and (IMO) more standard (Houdini uses the same terminology). Also, this is what the your PR was originally named :)

danrbailey:
> @danrbailey - thanks for addressing my review comments. I've approved. I have a non-blocking request thought, What do you say to changing the name from "TransientData" to "BlindData"? It's much shorter and (IMO) more standard (Houdini uses the same terminology). Also, this is what the your PR was originally named :)

Thanks @kmuseth! We actually discussed the name in one of our earlier TSC meetings and settled on "transientData()" because it was deemed clearer that the data is indeed "transient" and thus not serialized with the grid whereas "blindData()" is a bit more ambiguous in that sense. Meeting notes here:

https://github.com/AcademySoftwareFoundation/openvdb/blame/master/tsc/meetings/2021-05-04.md#L75

kmuseth:
okay, I bow to the majority decision

Issue #621, opened 2020-01-28
Don't require the python interpreter in openvdb/python when unit tests are disabled


e4lam:
In trying to build the openvdb python module for vcpkg, I ran into a problem with openvdb/python requiring the python interpreter even when unit tests are disabled. By default, the openvdb vcpkg doesn't build unit tests and so it doesn't have a dependency that provides the python interpreter. Since I didn't want to add an extraneous dependency just for building I just patched it to remove this requirement. However, I think a worthwhile improvement here is to only require the python interpreter when the option `OPENVDB_BUILD_UNITTESTS` is enabled.

I encountered this while trying to help out issue #603.

danrbailey:
I agree, I think this is a good suggestion, thanks @e4lam. I'm going to mark this as a good first issue as it doesn't seem too challenging to implement.

Idclip:
I'll be looking as this as part of the some upcoming windows improvements

Idclip:
https://github.com/AcademySoftwareFoundation/openvdb/pull/646

Idclip:
Fixed as of #646

e4lam:
Thanks!

PR #1055, opened 2021-05-07
Nacho/fix python path ubuntu

nachovizzo:
# Fix Python path

**WARNING:** I don't use any Windows or macOS systems, nor Houdinim Maya, or any of the software that uses OpenVDB. Thus, this method has been only tested under Linux (though I don't see why this will break other installations)

## Problem description

The `pyopenvdb.so` gets installed in (as example) `/usr/local/lib/python3.8`. This looks to be correct, but sadly most of the Linux-based distributions **does not** look for python packages there but rather under `/usr/local/lib/python3.8/dist-packages`. Therefore when one tries to `import pyopenvdb as vdb` it doesn't find the shared library.

To check this, one can run:
```sh
python -c 'import site; print(site.getsitepackages())'
```
that in my case outputs:
```sh
['/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages', '/usr/lib/python3.8/dist-packages']
```
## Possible solution

One possible solution is to update the PYTHONPATH as follow:
```sh
export PYTHONPATH=/usr/local/lib/python3.8/
```

But of course, this is no too optimal, since one needs to do it on all systems that are using the library. Plus, at least on my systems, there is no other library located in there.

## One word about FindPython

So the [FindPython](https://cmake.org/cmake/help/v3.12/module/FindPython.html) module provides great functionality but sadly nothing outputs the `/usr/local/<>` directory I need. Please check the cmake example below

## Minimal CMake Example

I made this rather small example so reviewers can look through the introduced variables in more detail in different systems

```cmake
cmake_minimum_required(VERSION 3.12)
project(test_find_python)

include(GNUInstallDirs)
find_package(Python)

# FindPython Variables
message(STATUS "Python_EXECUTABLE:${Python_EXECUTABLE}")
message(STATUS "Python_STDLIB:${Python_STDLIB}")
message(STATUS "Python_SITELIB:${Python_SITELIB}")
message(STATUS "Python_SITEARCH:${Python_SITEARCH}")
message(STATUS "Python_SITEARCH:${Python_SITEARCH}")

# Method used to get the dist-packages location
get_filename_component(Python_PACKAGES_DIR ${Python_SITELIB} NAME)
message(STATUS "pyopenvdb.so install dir:${CMAKE_INSTALL_FULL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/${Python_PACKAGES_DIR}")

# How to find this library later when we need pyopenvdb by other targets
set(OPENVDB_PYTHON_PATH_SUFFIXES
  lib64/python
  lib64/python2.7
  lib64/python3
  lib/python
  lib/python2.7
  lib/python3
)

file(GLOB PYTHON_SITE_PACKAGES ${CMAKE_INSTALL_FULL_LIBDIR}/python**/*)
foreach(_site_package_full_dir ${PYTHON_SITE_PACKAGES})
  string(REPLACE ${CMAKE_INSTALL_FULL_LIBDIR} "${CMAKE_INSTALL_LIBDIR}"
                 _site_package_dir ${_site_package_full_dir})
  list(APPEND OPENVDB_PYTHON_PATH_SUFFIXES ${_site_package_dir})
endforeach()

# Debug msg
message(STATUS "OPENVDB_PYTHON_PATH_SUFFIXES:")
foreach(OPENVDB_PYTHON_PATH_SUFFIX ${OPENVDB_PYTHON_PATH_SUFFIXES})
    message(STATUS "${OPENVDB_PYTHON_PATH_SUFFIX}")
endforeach()
```

In my case the output is:
```sh
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/local/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/local/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Python: /usr/bin/python3.8 (found version "3.8.5") found components: Interpreter 
-- Python_EXECUTABLE:/usr/bin/python3.8
-- Python_STDLIB:/usr/lib/python3.8
-- Python_SITELIB:/usr/lib/python3/dist-packages
-- Python_SITEARCH:/usr/lib/python3/dist-packages
-- Python_SITEARCH:/usr/lib/python3/dist-packages
-- pyopenvdb.so install dir:/usr/local/lib/python3.8/dist-packages
-- OPENVDB_PYTHON_PATH_SUFFIXES:
-- lib64/python
-- lib64/python2.7
-- lib64/python3
-- lib/python
-- lib/python2.7
-- lib/python3
-- lib/python2.7/dist-packages
-- lib/python2.7/site-packages
-- lib/python3.8/dist-packages
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ivizzo/dev/examples/FindPythonCmake/build
```

### DIFF
--- a/openvdb/openvdb/python/CMakeLists.txt
+++ b/openvdb/openvdb/python/CMakeLists.txt
@@ -69,12 +69,14 @@ endfunction()
 # To ensure consistent versions between components Interpreter, Compiler,
 # Development and NumPy, specify all components at the same time when using
 # FindPython.
-# @note  the Interpreter component is only required for the python test.
 # @note  the Python::Module target should be used over the Python::Python
 #   target but this was only added in CMake 3.15. See:
 #      https://github.com/AcademySoftwareFoundation/openvdb/issues/886
 set(OPENVDB_PYTHON_DEPS)
-set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development Interpreter)
+set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development)
+if(NOT DEFINED PYOPENVDB_INSTALL_DIRECTORY)
+    list(APPEND OPENVDB_PYTHON_REQUIRED_COMPONENTS Interpreter)
+endif()
 
 if(${CMAKE_VERSION} VERSION_LESS 3.14)
   find_package(Python QUIET COMPONENTS ${OPENVDB_PYTHON_REQUIRED_COMPONENTS})
--- a/openvdb/openvdb/python/CMakeLists.txt
+++ b/openvdb/openvdb/python/CMakeLists.txt
@@ -74,11 +74,7 @@ endfunction()
 #   target but this was only added in CMake 3.15. See:
 #      https://github.com/AcademySoftwareFoundation/openvdb/issues/886
 set(OPENVDB_PYTHON_DEPS)
-set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development)
-
-if(OPENVDB_BUILD_PYTHON_UNITTESTS)
-  list(APPEND OPENVDB_PYTHON_REQUIRED_COMPONENTS Interpreter)
-endif()
+set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development Interpreter)
 
 if(${CMAKE_VERSION} VERSION_LESS 3.14)
   find_package(Python QUIET COMPONENTS ${OPENVDB_PYTHON_REQUIRED_COMPONENTS})
@@ -210,8 +206,9 @@ set(OPENVDB_PYTHON_MODULE_SOURCE_FILES
 )
 
 if(NOT DEFINED PYOPENVDB_INSTALL_DIRECTORY)
+  get_filename_component(Python_PACKAGES_DIR ${Python_SITELIB} NAME)
   set(PYOPENVDB_INSTALL_DIRECTORY
-    ${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}
+    ${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/${Python_PACKAGES_DIR}
     CACHE STRING "The directory to install the pyopenvdb.so module."
   )
 endif()
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,8 +37,6 @@ jobs:
       run: ./ci/build.sh Release None OFF None "core,python,doc" -DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON -DDISABLE_DEPENDENCY_VERSION_CHECKS=ON
     - name: epydoc
       run: |
-        export LD_LIBRARY_PATH=/usr/local/lib64
-        export PYTHONPATH=/usr/local/lib64/python2.7
         epydoc --html -o /usr/local/share/doc/OpenVDB/html/python pyopenvdb
     - name: deploy
       # only deploy documentation to gh-pages on a manual workflow dispatch
--- a/cmake/FindOpenVDB.cmake
+++ b/cmake/FindOpenVDB.cmake
@@ -306,9 +306,6 @@ list(APPEND _OPENVDB_LIBRARYDIR_SEARCH_DIRS
 set(_OPENVDB_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
 
 set(OPENVDB_PYTHON_PATH_SUFFIXES
-  ${CMAKE_INSTALL_LIBDIR}/python
-  ${CMAKE_INSTALL_LIBDIR}/python2.7
-  ${CMAKE_INSTALL_LIBDIR}/python3
   lib64/python
   lib64/python2.7
   lib64/python3
@@ -316,6 +313,14 @@ set(OPENVDB_PYTHON_PATH_SUFFIXES
   lib/python2.7
   lib/python3
 )
+
+# Recurse through all the site-packages and dist-packages on the file system
+file(GLOB PYTHON_SITE_PACKAGES ${CMAKE_INSTALL_FULL_LIBDIR}/python**/*)
+foreach(_site_package_full_dir ${PYTHON_SITE_PACKAGES})
+  string(REPLACE ${CMAKE_INSTALL_FULL_LIBDIR} "${CMAKE_INSTALL_LIBDIR}"
+                 _site_package_dir ${_site_package_full_dir})
+  list(APPEND OPENVDB_PYTHON_PATH_SUFFIXES ${_site_package_dir})
+endforeach()
 
 set(OPENVDB_LIB_PATH_SUFFIXES
   ${CMAKE_INSTALL_LIBDIR}


nachovizzo:
@Idclip Sorry for the spam, shall I close this PR? I don't want to left garbage on the repository if it's not going to be used.
Thanks!

openvdb/openvdb/python/CMakeLists.txt
```diff
@@ -74,11 +74,7 @@ endfunction()
 #   target but this was only added in CMake 3.15. See:
 #      https://github.com/AcademySoftwareFoundation/openvdb/issues/886
 set(OPENVDB_PYTHON_DEPS)
-set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development)
-
-if(OPENVDB_BUILD_PYTHON_UNITTESTS)
-  list(APPEND OPENVDB_PYTHON_REQUIRED_COMPONENTS Interpreter)
-endif()
+set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development Interpreter)
```
Idclip, line 77: Was there a reason you changed this? Technically, the `Interpreter` should not be needed unless the unit tests are being run

openvdb/openvdb/python/CMakeLists.txt
```diff
@@ -210,8 +206,9 @@ set(OPENVDB_PYTHON_MODULE_SOURCE_FILES
 )
 
 if(NOT DEFINED PYOPENVDB_INSTALL_DIRECTORY)
+  get_filename_component(Python_PACKAGES_DIR ${Python_SITELIB} NAME)
   set(PYOPENVDB_INSTALL_DIRECTORY
-    ${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}
+    ${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/${Python_PACKAGES_DIR}
```
Idclip, line 211: Any reason why this isn't simply ` set(PYOPENVDB_INSTALL_DIRECTORY ${Python_SITELIB})` ?

openvdb/openvdb/python/CMakeLists.txt
nachovizzo: I totally agree, but the component name "Interpreter" is somehow misleading. The variable `${Python_SITELIB}` will only get populated if we include the `Intepreter` component. Is not about accessing the python interpreter per see. 

That variable gets created at https://github.com/Kitware/CMake/blob/859241d2bbaae83f06c44bc21ab0dbd38725a3fc/Modules/FindPython/Support.cmake#L1830

Idclip: Hmm we previously always searched for both but we changed it due to this issue:

https://github.com/AcademySoftwareFoundation/openvdb/issues/621

@e4lam do you remember what the exact side effects were? The issue refers to the fact that the `Interpreter` target forces another dependency, but I'm unclear how problematic this actually is. For context, this PR aims to improve the default installation of the python module by using the `Python_SITELIB` CMake variable... which is only set if you look for the `Interpreter` component

e4lam: Sorry, I don't really remember the exact error. AFAIR, building the openvdb vcpkg would fail until I removed `Interpreter` there. I don't recall now if vcpkg simply never built the python interpreter at the time. So when I tried building openvdb, cmake would fail because it couldn't find a working python interpreter since I didn't build anything in vcpkg that required the python interpreter. Looking this up now, I see https://github.com/microsoft/vcpkg/pull/14891 has since occurred, so maybe if we rejigged the openvdb vcpkg dependencies, it might be possible to add this back in again. Not sure if I can find time to try this though.

Idclip: Thanks a lot @e4lam! It's weird that we need the interpreter for this, but I guess it's running some python to figure out the correct installation path. If that's the best and most correct python-y way of getting it then seems like we'll need it - if the fix in vcpkg only went in for py3 then we can re-jig this so that it's only brought in for in py3 and later. 

Additionally, if a user has provided their own absolute file path for the vdb cmake variable `PYOPENVDB_INSTALL_DIRECTORY` (i.e. `-DPYOPENVDB_INSTALL_DIRECTORY=/my/custom/path`) then we can also avoid searching for `Interpreter`. That logic should help mitigate issues. @nachovizzo does that make sense to you too?

nachovizzo: It's weird indeed and I don't like it either. But looks like that's how it's done on the python world :P. To dive even deeper you can take a look at [this line@cmake v.3.12](https://github.com/Kitware/CMake/blob/f84c15ef2fa30dd074fcccafccec6b9b69266619/Modules/FindPython/Support.cmake#L329). Actually the one pointed out above it was the wrong line to look for.

If we decrypt a bit what they are running in python looks like:
```python
import sys
from distutils import sysconfig

sys.stdout.write(
    ";".join(
        [
            sysconfig.get_python_lib(plat_specific=False, standard_lib=True),
            sysconfig.get_python_lib(plat_specific=True, standard_lib=True),
            sysconfig.get_python_lib(plat_specific=False, standard_lib=False),
            sysconfig.get_python_lib(plat_specific=True, standard_lib=False),
        ]
    )
)
```

Which in my `Ubuntu 20.04` installation outputs:
```sh
/usr/lib/python3.8;/usr/lib/python3.8;/usr/lib/python3/dist-packages;/usr/lib/python3/dist-packages
```

But when using the docker image for the CI builds this is the output
```sh
$ docker run --rm -it aswf/ci-openvdb python -c '
import sys
from distutils import sysconfig

sys.stdout.write(
    ";".join(
        [
            sysconfig.get_python_lib(plat_specific=False, standard_lib=True),
            sysconfig.get_python_lib(plat_specific=True, standard_lib=True),
            sysconfig.get_python_lib(plat_specific=False, standard_lib=False),
            sysconfig.get_python_lib(plat_specific=True, standard_lib=False),
        ]
    )
)'
/usr/local/lib/python2.7;/usr/local/lib/python2.7;/usr/local/lib/python2.7/site-packages;/usr/local/lib/python2.7/site-packages
```

As you can see they are different locations, and **apparently** the only way of getting to know this is by launching a python shell and getting the `sysconfig` variables.

Aout the `PYOPENVDB_INSTALL_DIRECTORY`, I agree and fixed it in de283e4eada4465

Idclip (approved):

eloop:
This PR is of interest to me as I maintain a https://github.com/spack/ package for OpenVDB. I put in a post_install hook to move the pyopenvdb.so into the correct site-packages location, as well as handle the situation where the .so ends up under lib/ (Ubuntu) or lib64/ (RHEL). It would be great if I can remove the hack, it can be seen at the bottom of this file ...

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/openvdb/package.py

Idclip:
@eloop thanks a lot for commenting with this. From what I can tell, your install always moves the python module to the `site-packages` directory. This fix *may* move it there or, on some platforms (debian), it may be moved to the `dist-packages` sub directory. I think this is the correct behaviour but just a heads up.

This needs one more approval @danrbailey @jmlait @kmuseth @apradhana

nachovizzo:
@Idclip just checking in here. Will this get merged at some point? Thanks!

Idclip:
Yes, pinging @danrbailey @jmlait @kmuseth @apradhana - still needs one approval

danrbailey (approved):

PR #1056, opened 2021-05-07
AX improvements for LLVM 12

Idclip:


openvdb_ax/openvdb_ax/CMakeLists.txt
```diff
@@ -94,6 +94,12 @@ if(MINIMUM_LLVM_VERSION)
       "but required is at least \"${MINIMUM_LLVM_VERSION}\" (found ${LLVM_DIR})")
   endif()
 endif()
+if(OPENVDB_FUTURE_DEPRECATION AND FUTURE_MINIMUM_LLVM_VERSION)
+  if(LLVM_PACKAGE_VERSION VERSION_LESS FUTURE_MINIMUM_LLVM_VERSION)
+    message(DEPRECATION "Support for LLVM versions < ${FUTURE_MINIMUM_LLVM_VERSION} "
+      "is deprecated and will be removed.")
+  endif()
+endif()
```
Idclip, line 102: This PR doesn't introduce this, but I'd like to bump the minimum supported version of LLVM from 6->7. I haven't been promoting a warning so haven't removed it in this PR, but perhaps that's okay? It seems that `OPENVDB_FUTURE_DEPRECATION` is off by default which kind of defeats the purpose of it. @danrbailey I think you introduced this setting, any thoughts?

danrbailey: See this thread for context:

https://github.com/AcademySoftwareFoundation/openvdb/pull/532#discussion_r334128582

Reading back through this, I think we were in the situation where the "typical" configuration was deprecated because of where we were in the year and the general adoption of VFX Reference Platform at the time was slightly delayed, so the proposal was to flip this to ON only in the run-up to a new release. I don't believe we've ever remembered to do this and perhaps the situation has changed anyway so we should revisit this, I agree it does seems a bit pointless as-is.

danrbailey (approved):

openvdb_ax/openvdb_ax/CMakeLists.txt
Idclip: Okay I won't turn this variable on in this PR. But do you think its acceptable to remove support for LLVM 6 in 8.1? Or should we revisit this after.

PR #1066, opened 2021-05-17
Doxygen fixes and improvements

Idclip:


.github/workflows/docs.yml
```diff
@@ -20,7 +20,7 @@ on:
 
 jobs:
   doxygen:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
Idclip, line 23: drive by: https://github.com/actions/virtual-environments/issues/3287

danrbailey (approved):

PR #1067, opened 2021-05-18
Modify AX unit tests to build against CppUnit without sstream

danrbailey:
The AX unit tests failed for me because it looks like I've been building against a version of CppUnit that has sstream support disabled. This has worked for many years when OpenVDB was using CppUnit and the fix is pretty simple - to use `CPPUNIT_ASSERT(a == b)` instead of `CPPUNIT_ASSERT_EQUAL(a, b)` for enums.


Idclip:
@danrbailey you can delete the pendingchanges - all the code you've modified was introduced with the active tile streaming stuff which is new in 8.1.

danrbailey:
> @danrbailey you can delete the pendingchanges - all the code you've modified was introduced with the active tile streaming stuff which is new in 8.1.
> 
> edit: assuming this goes in to 8.1

Sure, I dropped the pending changes.

Idclip (approved):

PR #1070, opened 2021-05-21
macos TBB CI fixes

Idclip:
Signed-off-by: Nick Avramoussis <4256455+Idclip@users.noreply.github.com>

PR #1075, opened 2021-05-26
Guard the cmakedefines in case systems have -D options

jmlait:
We noticed our internal build was setting OPENVDB_USE_BLOSC, which then conflicted with the version.h.  The intent of the version.h was not to get in the way of people with explicit -D command lines, so thus have disabled it in this fashion.

I'm hesitant to remove from our build command line as then we'll fail if we revert openvdb at any point...


Idclip (approved):
lgtm - I did do this for the ABI define but didn't think it was necessary for these. Don't see any harm in it.

Note that #1070 fixes the CI issues

PR #1076, opened 2021-05-28
Rename VDB Vector Merge to VDB Vector from Scalar

jmlait:
Internal review found this the least objectionable new name, but it
is up to the TSC if it meets our standards.  Only the label is changed,
I believe I have preserved the internal names properly.


danrbailey (approved):
I'm fine with this name change. It does beg the question about whether to rename "VDB Vector Split" to "VDB Scalar from Vector" for completeness though.

jmlait:
We did discuss that as an option....  But Vector Split does seem like a more descriptive name.

Scalar from Vector might also be expected to have things like length computations built in.

PR #1087, opened 2021-06-03
Add optional CMake variable to build a subset of the Houdini DSOs

danrbailey:
This PR (#1025) has been really useful to me, so adding a similar option to control Houdini DSOs as well.

`cmake -DOPENVDB_DSO_NAMES=SOP_OpenVDB_Activate ..`
`cmake -DOPENVDB_DSO_NAMES="SOP_OpenVDB_Activate;SOP_OpenVDB_Resample" ..`

### DIFF
--- a/openvdb_houdini/openvdb_houdini/CMakeLists.txt
+++ b/openvdb_houdini/openvdb_houdini/CMakeLists.txt
@@ -241,6 +241,10 @@ endif()
 
 if(OPENVDB_DSO_NAMES)
   set(OPENVDB_USE_CUSTOM_DSO_NAMES 1)
+  # if AX SOP is explicitly enabled, also switch on the build variable
+  if (SOP_OpenVDB_AX IN_LIST OPENVDB_DSO_NAMES)
+    set(OPENVDB_BUILD_SOP_OpenVDB_AX ON)
+  endif()
 endif()
 
 if(NOT OPENVDB_USE_CUSTOM_DSO_NAMES)
--- a/openvdb_houdini/openvdb_houdini/CMakeLists.txt
+++ b/openvdb_houdini/openvdb_houdini/CMakeLists.txt
@@ -72,6 +72,12 @@ else()
   endif()
 endif()
 
+set(OPENVDB_DSO_NAMES "" CACHE STRING [=[
+  An optional list of Houdini DSOs to build. If not provided, all Houdini DSOs will be built.
+  DSOs should be listed without the .cc suffix, for example the string
+  \"SOP_OpenVDB_Activate;SOP_OpenVDB_Resample\" would build just the VDB Activate and
+  VDB Resample SOPs.]=])
+
 #########################################################################
 
 message(STATUS "----------------------------------------------------")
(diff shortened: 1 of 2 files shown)


Idclip:
This technique made sense with the tests because the tests all compiled to a single executable. Ideally the CMake build system should configure everything and you can then choose what to build. This is the case with the SOPs as they are individual DSOs. So they way that it works currently is simply:

```
cmake ..
make SOP_OpenVDB Active 
```

There is one issue here which is that the `install` target will still build everything.

danrbailey:
> This technique made sense with the tests because the tests all compiled to a single executable. Ideally the CMake build system should configure everything and you can then choose what to build. This is the case with the SOPs as they are individual DSOs. So they way that it works currently is simply:
> 
> ```
> cmake ..
> make SOP_OpenVDB Active 
> ```
> 
> There is one issue here which is that the `install` target will still build everything. Not sure how to necessarily solve that, but should this meet your requirements I don't think we need this.

I take your point here, but this approach is really not ideal. In addition to the install issue you mentioned which I would need a solution to, there is another problem - to build multiple DSOs concurrently is not possible, if you do something like this:

```
cmake ..
make SOP_OpenVDB_Activate SOP_OpenVDB_Resample -j32
```

It will build each SOP one-by-one and re-check the dependent targets (openvdb_shared, openvdb_houdini) each time, so throughput is _way_ slower.

Unless you can figure out a solution to these issues, I would prefer to adopt my proposed solution here as it's a relatively simple change.

Idclip:
> It will build each SOP one-by-one and re-check the dependent targets (openvdb_shared, openvdb_houdini) each time, so throughput is way slower.

FWIW, what you're describing here is a limitation of CMake/Make. Ninja generators should solve this problem for you.

> Unless you can figure out a solution to these issues, I would prefer to adopt my proposed solution here as it's a relatively simple change.

The `install` issue could actually be easy, we simply provide a `-DOPENVDB_NO_INSTALL_ALL` var which only installs built targets (i.e. `make SOP_OpenVDB_Activate SOP_OpenVDB_Resample install -j32`). 

But yes, it's fine, we can go with this. I did want to point out though that it's conceptually wrong.

danrbailey:
> > It will build each SOP one-by-one and re-check the dependent targets (openvdb_shared, openvdb_houdini) each time, so throughput is way slower.
> 
> FWIW, what you're describing here is a limitation of CMake/Make. Ninja generators should solve this problem for you.
> 
> > Unless you can figure out a solution to these issues, I would prefer to adopt my proposed solution here as it's a relatively simple change.
> 
> The `install` issue could actually be easy, we simply provide a `-DOPENVDB_NO_INSTALL_ALL` var which only installs built targets (i.e. `make SOP_OpenVDB_Activate SOP_OpenVDB_Resample install -j32`).
> 
> But yes, it's fine, we can go with this. I did want to point out though that it's conceptually wrong.

Yes, that all makes sense. Somewhat tangential to this discussion - but we use a single SOP_VDB.so library at ILM and I know that's how SideFX packages these SOPs for distribution too, perhaps it's worth discussing if we would like to officially support this?

Idclip:
> Somewhat tangential to this discussion - but we use a single SOP_VDB.so library at ILM and I know that's how SideFX packages these SOPs for distribution too, perhaps it's worth discussing if we would like to officially support this

Yeah if there's interest in building them this way then definitely, I'd be interested to hear the advantages/reasons for.

Idclip:
One slightly weird thing here is the existence of this:

**Note: Individual DSO components can be turned off with their own CMake  variable "OPENVDB_BUILD_{NAME}" i.e. OPENVDB_BUILD_SOP_OpenVDB_Scatter=OFF**

This was only introduced because it's useful to allow the AX SOP to be enabled if `OPENVDB_BUILD_AX` is off (i.e. search for AX on the system). Perhaps this system can be removed in favour of a more specific variable dedicated to the AX SOP? This is kind of the inverse of what the `OPENVDB_DSO_NAMES` variable is solving as it enables certain SOPs, where as `OPENVDB_BUILD_{NAME}` disables SOPs (except in the case of AX).

danrbailey:
> One slightly weird thing here is the existence of this:
> 
> **Note: Individual DSO components can be turned off with their own CMake variable "OPENVDB_BUILD_{NAME}" i.e. OPENVDB_BUILD_SOP_OpenVDB_Scatter=OFF**
> 
> This was only introduced because it's useful to allow the AX SOP to be enabled if `OPENVDB_BUILD_AX` is off (i.e. search for AX on the system). Perhaps this system can be removed in favour of a more specific variable dedicated to the AX SOP? This is kind of the inverse of what the `OPENVDB_DSO_NAMES` variable is solving as it enables certain SOPs, where as `OPENVDB_BUILD_{NAME}` disables SOPs (except in the case of AX).

Yeah, I wasn't totally sure how to handle the AX SOP in this change. I agree, a more specific variable for just the AX SOP would be clearer I think.

Idclip:
if  `-DOPENVDB_DSO_NAMES=SOP_OpenVDB_AX` and the core libs are not being build, I would expect `OPENVDB_BUILD_SOP_OpenVDB_AX` to be defined. Currently I don't think that's the case so the AX SOP won't be built? I think we just need to check to see if `SOP_OpenVDB_AX` has been included in the list (if the list has been provided) and set that build variable to ON

danrbailey:
> if `-DOPENVDB_DSO_NAMES=SOP_OpenVDB_AX` and the core libs are not being build, I would expect `OPENVDB_BUILD_SOP_OpenVDB_AX` to be defined. Currently I don't think that's the case so the AX SOP won't be built? I think we just need to check to see if `SOP_OpenVDB_AX` has been included in the list (if the list has been provided) and set that build variable to ON

I've addressed this feedback now. Hopefully this is the change you were after?

Idclip (approved):

PR #1088, opened 2021-06-04
Various CI improvements

Idclip:
- Improved the usage of various actions
 - Bumped ubuntu images to latest
 - Simplified the ./build scripts to not use arbitrary positional arguments
 - Introduced a `linux-extra` build matrix in `build.yml` which includes various jobs which are designed to test different build types outside the standard build types. This includes the existing `NO_BLOSC`, `SSE` and `USE_IMATH` builds. Added an `AVX` build. In an attempt to speed up builds, these builds only build weekly, not on commit.
 - Various tweaks to improve the speed of some of the configurations
 - Don't trigger any CI if `pendingchanges/**` changes

### DIFF
--- a/.github/workflows/ax.yml
+++ b/.github/workflows/ax.yml
@@ -10,6 +10,7 @@ on:
       - 'doc/**'
       - 'openvdb_maya/**'
       - 'openvdb_houdini/**'
+      - 'pendingchanges/**'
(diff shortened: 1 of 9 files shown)


danrbailey (approved):
Looks good, thanks

PR #1090, opened 2021-06-04
Add more counting tools

danrbailey:
Adding:

* tools::countActiveLeafVoxels() [+bbox version]
* tools::countInactiveVoxels()
* tools::countInactiveLeafVoxels()
* tools::countActiveTiles()

and changed the Tree methods to use these tools.

### DIFF
--- a/openvdb/openvdb/tools/Count.h
+++ b/openvdb/openvdb/tools/Count.h
@@ -12,6 +12,7 @@
 #define OPENVDB_TOOLS_COUNT_HAS_BEEN_INCLUDED
 
 #include <openvdb/version.h>
+#include <openvdb/tree/LeafManager.h>
 #include <openvdb/tree/NodeManager.h>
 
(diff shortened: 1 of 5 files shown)


danrbailey:
@Idclip - I keep getting a couple of failures in the AX jobs with this PR. It seems to be the two g++ AX jobs that fail but with no actual error as far as I can tell. Did you see this before?

Idclip:
@danrbailey no idea, I've just kicked the jobs off again. I've even tried running the github cli tool `gh` to download the logs but they don't exist. Will see if it happens again

Idclip:
CI looks fine now

danrbailey:
> ~CI looks fine now~ spoke too soon. I have no idea why those jobs are failing, impossible to tell when the logs vanish...

It's weird, might try and close out this PR and create a new one. It doesn't seem like other PRs are having this problem.

PR #1091, opened 2021-06-08
Add an option to specify the expansion pattern.

jmlait:
Add an option to specify the expansion pattern.
Rename the label of the expand voxels to match the current Houdini name.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
@@ -237,14 +237,14 @@ operation or setting of values.)"));
     Specifies which nearby voxels are considered neighbors for expansion.
 */
     parms.add(hutil::ParmFactory(PRM_STRING, "expansionpattern", "Expansion Pattern")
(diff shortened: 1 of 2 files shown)


openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Activate.cc
```diff
@@ -490,9 +504,9 @@ sopFillSDF(GridType &grid, int dummy)
 
 template <typename GridType>
 static void
-sopDilateVoxels(GridType& grid, exint count)
+sopDilateVoxels(GridType& grid, exint count, openvdb::tools::NearestNeighbors nn)
 {
-    openvdb::tools::dilateActiveValues(grid.tree(), static_cast<int>(count));
+    openvdb::tools::dilateActiveValues(grid.tree(), static_cast<int>(count), nn);
```
Idclip, line 509: worth passing in for the `sopErodeVoxels` method as well?

jmlait: Yes, I think that makes sense.

Idclip (approved):

PR #1092, opened 2021-06-09
Doxygen Improvements

Idclip:
- A variety of fixes (primary to AX \ syntax and in `math/Math.h`) for later versions of doxygen
 - Consolidated all `OPENVDB_<>_INTERNAL` doxygen clauses into a single `OPENVDB_DOCS_INTERNAL` variable
 - In an attempt to streamline the doxygen and remove unnecessary information, encapsulated the majority of internal namespaces with the `OPENVDB_DOCS_INTERNAL` variable
 - Added full path support to the VDB folder layouts
 - Added better deprecation listings

Old | New
-- | --
 | 
 | 

Deprecation listings example:


danrbailey (approved):
Agreed, the new version is much nicer, thanks!

PR #1094, opened 2021-06-09
Meeting notes for June 8th 2021.

jmlait:
Added a placeholder file to avoid the pendingchanges directory
from vanishing.

Did not integerate the change notes as we have to bump our version
number first.

### DIFF
--- a/pendingchanges/README
+++ b/pendingchanges/README
@@ -1,0 +1,11 @@
+This directory stores pending changes to our CHANGES and changes.txt
+files.  The idea is that PRs can have their proposed changes audited
+without triggering false merges.  We periodically collapse these into
+the CHANGES and changes.txt.
+
+Release versions should have this directory empty, but if you
+checkout the latest version you may see changes reflected here that
+haven't been merged into the CHANGES file yet.
+
+This README must be maintained or git removes the directory and we
+won't remember what directory to put changes into.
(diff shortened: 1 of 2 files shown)


Idclip (approved):

danrbailey (approved):

diiigle:
Glad to see a motion towards pybind11. 

In research, we often script everything in python, and only move the heavy lifting into C++ when necessary. This will allow us to write our own pybind11 modules that work on OpenVDB python types and are also compatible with the C++ API. The transition should hopefully not be that big of a pain, as pybind11 was mostly designed to be syntax compatible with Boost.Python. 

If you're looking for inspiration (ie projects that use pybind11) I can only recommend [mitsuba2](https://github.com/mitsuba-renderer/mitsuba2) ( and related projects, [enoki](https://github.com/mitsuba-renderer/enoki)) from Wenzel Jakob, the creator of pybind11. These projects are massively templated and they managed to squeeze that into python without writing enormous amounts of binding code. 

A few of the tricks I found there:
- neat CMake integration with distributed binding code (`python` subdir of the code it binds)
- heavy use of (nested) macros that generate the binding code for different variants
- (python) runtime attachment of functions to a set of types
- mixture of python & C++ code to form the python API

apradhana (approved):

Issue #1096, opened 2021-06-10
[BUG] Python API background value different from what's expected


nachovizzo:
### Environment

**Operating System:** Linux
**Version / Commit SHA:** master

### Describe the bug

Something it's wrong with the python wrappers for `FloatGrid`. I'm trying to debug this issue myself but so far I haven't reach anywhere. Looks like the precision of floating point numbers is messing around, or sometihng similar.

I manage to narrow down the error to a minimalistic example.

**NOTE** This problem only appears in the Python wrappers in `pyopenvdb.so`, not a C++ API issue.

#### C++ Example

**Code:**

```cpp
#include <openvdb/openvdb.h>

#include <iostream>

using std::cout;
using std::endl;

int main() {
    openvdb::initialize();
    const float background_value = 0.3;
    openvdb::FloatGrid::Ptr grid = openvdb::FloatGrid::create(background_value);

    grid->print();
    cout << "grid->background() = " << grid->background() << endl;
    cout << std::boolalpha << (grid->background() == background_value) << endl;
    return 0;
}
```

**Output:**

```sh
Information about Tree:
  Type: Tree_float_5_4_3
  Configuration:
    Root(0), Internal(32^3), Internal(16^3), Leaf(8^3)
  Background value: 0.3
Transform:
  voxel size: 1
  index to world:
     [1, 0, 0, 0]
     [0, 1, 0, 0]
     [0, 0, 1, 0]
     [0, 0, 0, 1]

grid->background() = 0.3
true
```

#### Python Example

**Code:**

```python
import pyopenvdb as vdb

background_value = float(0.3)
grid = vdb.FloatGrid(background_value)
print(grid.info())
print("grid.background =", grid.background)
print(grid.background == background_value)
```

**Output:**

```sh
Information about Tree:
  Type: Tree_float_5_4_3
  Configuration:
    Root(0), Internal(32^3), Internal(16^3), Leaf(8^3)
  Background value: 0.3
Transform:
  voxel size: 1
  index to world:
     [1, 0, 0, 0]
     [0, 1, 0, 0]
     [0, 0, 1, 0]
     [0, 0, 0, 1]


grid.background = 0.30000001192092896
False
```

### Expected behavior

As shown in the example, the value `0.3` gets interpreted as `0.30000001192092896` when using the Python API. Which seems strange...

Moreover, the comparission `grid.background == background_value` return `False`, **only** on the Python example, which confirms that something is "wrong".

### Additional context

[...]

PR #1105, opened 2021-06-15
Mark `-background` values as inactive in copyFromDense()

nachovizzo:
This solves the bugs encountered in #1096. If the voxel value is the
same as the background but with a negative sign, then that voxel should
also, mark as inactive.

Signed-off-by: Ignacio Vizzo <ignaciovizzo@gmail.com>

## Problem Description

- On `openvdb::FloatGrid`, some voxel values are `-background` and thus, marked as `inactive. Which makes total sense to me.
- When copying the data to a dense array, the mask for the voxels is lost, which is also not a problem to my eyes.
- When copying back from a dense array to a `openvdb::FloatGrid`, it's checked which values are within a certain `mTolerance` close to the `background` value, and thus, marked as inactive.

To showcase my problem I made a simple `VDB` -> `np` -> `VDB` pipeline, without changing any value or anything...

## Simple Example

<details>
<summary>Python Example</summary>
  
```python
import numpy as np
import pyopenvdb as vdb


def level_set_to_numpy(grid: vdb.FloatGrid) -> np.ndarray:
    shape = grid.evalActiveVoxelDim()
    start = grid.evalActiveVoxelBoundingBox()[0]
    sdf_volume = np.zeros(shape, dtype=np.float32)
    grid.copyToArray(sdf_volume, ijk=start)
    return sdf_volume


# Input VDB, a simple sphere
vdb_grid_in = vdb.createLevelSetSphere(2.0, voxelSize=0.1)
vdb_grid_in.name = "vdb_grid in"

# Covert to dense np array
np_grid = level_set_to_numpy(vdb_grid_in)

# Covert back to a VDB grid
vdb_grid_out = vdb.FloatGrid()
vdb_grid_out.transform = vdb_grid_in.transform
vdb_grid_out.background = vdb_grid_in.background
vdb_grid_out.gridClass = vdb_grid_in.gridClass
vdb_grid_out.name = "vdb_grid out"
vdb_grid_out.copyFromArray(
    np_grid,
    ijk=vdb_grid_in.evalActiveVoxelBoundingBox()[0],
    tolerance=0.0,
)

print("Stats:")
print("Same background?", vdb_grid_in.background == vdb_grid_out.background)
print("Same min value ?", vdb_grid_in.evalMinMax()[0] == vdb_grid_out.evalMinMax()[0])
print("Same max value ?", vdb_grid_in.evalMinMax()[1] == vdb_grid_out.evalMinMax()[1])
print("max[vdb  in] = {:.30}".format(vdb_grid_in.evalMinMax()[1]))
print("max[vdb out] = {:.30}".format(vdb_grid_out.evalMinMax()[1]))
print("min[vdb  in] = {:.30}".format(vdb_grid_in.evalMinMax()[0]))
print("min[vdb out] = {:.30}".format(vdb_grid_out.evalMinMax()[0]))

vdb.write("in.vdb", vdb_grid_in)
vdb.write("out.vdb", vdb_grid_out)
```
</details>

<details>
<summary>Output</summary>

```sh
Stats:
Same background? True
Same min value ? False
Same max value ? True
max[vdb  in] = 0.2978250682353973388671875
max[vdb out] = 0.2978250682353973388671875
min[vdb  in] = -0.2970613539218902587890625
min[vdb out] = -0.300000011920928955078125
```
</details>

### Input
image

### Output
image

### Cross-section Input
image

### Cross-section Output
image

## Analysis

While the volumes represent exactly the same geometry, due to an **unknown error to me** in the `copyFromArray` method (I suspect), some voxel values don't match the target `background` value, meaning that they won't get "pruned" and marked as inactive. In consequence, we end up with a "less sparse volume" since we need to also densely store all those values.

This is the exact same behavior I'm facing on my application, the `background value` is the exact same `32` bits representation value but there are more `active` voxels in the output grid.

Here is another comparison between the output grids, to showcase that the "copy error" gets reflected on the `min` value of the output grid...
image

## NOTE

Most of the time this doesn't present an issue, on my sphere example, if you **don't** cut the sphere to manually inspect the values that are inside, you will never notice that there are some errors.

Thanks and sorry for the long explanation!

### DIFF
--- a/openvdb/openvdb/tree/LeafNode.h
+++ b/openvdb/openvdb/tree/LeafNode.h
@@ -1265,9 +1265,9 @@ LeafNode<T, Log2Dim>::copyFromDense(const CoordBBox& bbox, const DenseT& dense,
             const DenseValueType* s2 = s1 + yStride * (y - min[1]);
             Int32 n2 = n1 + ((y & (DIM-1u)) << LOG2DIM);
             for (Int32 z = bbox.min()[2], ez = bbox.max()[2]+1; z < ez; ++z, ++n2, s2 += zStride) {
-                if (math::isApproxEqual(background, ValueType(*s2), tolerance)) {
+                if (math::isApproxEqual(background, math::Abs(ValueType(*s2)), tolerance)) {
                     mValueMask.setOff(n2);
-                    mBuffer[n2] = background;
+                    mBuffer[n2] = math::Sign(ValueType(*s2)) * background;
                 } else {
                     mValueMask.setOn(n2);
                     mBuffer[n2] = ValueType(*s2);


Idclip:
Thanks for diagnosing this @nachovizzo. I agree that the `copyFromDense` is behaving unexpectedly here, but I don't think this fix will work across the board. The issue is that the behaviour related to deactivating values of either sign is specific **only** to level sets. For example consider a velocity volume with a background of (1,2,3). With your fix, an active voxel with a value (-1,2,-3) will be incorrect deactivated by the `copyFromDense` code.

I need to have a think about this some more, but right now I would instead advocate for a change in `tools/Dense.h` `tools::copyFromDense` where we instead check the grid's class. If the grid is marked as a level set then the behaviour you're proposing here can be applied. In your example, as you're copying over the grid class `vdb_grid_out.gridClass = vdb_grid_in.gridClass` before running the copy, this would fix your issue but maintain consistent behaviour for non level-sets (fog-volumes, vector volumes etc).

@kmuseth thoughts?

Idclip:
I would also mention that these methods, `copyToDense` and `copyFromDense` don't give any specifics on active/inactive mask handling. You would, for example, see this behaviour for other grid types for inactive voxels which are **not** background. Given a fog volumes with a background of -1.0f that contains an **inactive** voxel with a value of -2.0f (i.e. different from the background), this voxel will end up as **active** after the copyTo/copyFrom workflow. I think this is expected behaviour. For your specific case though (with level-sets) it makes sense to handle the deactivation of negative background values.

nachovizzo:
@Idclip sorry for the late reply. That makes total sense. Haven't thought about that since I'm **only** using `FloatingGrid` and more specifically, levelsets... I'd add a grid check there

nachovizzo:
@Idclip Should I attempt to update this and add the fix for only level sets?

jmlait:
Thank you again for this PR... We looked at it again and realized that doing the change as suggested would have surprising results if you have a non-levelset grid.

We think the right course is to have a new copyFromDenseSDF method on the Dense.h.  The LeafNode code for copyFromDense should just be inlined in Dense.h and not pushed to the tree, as we are trying to reduce methods on the tree.

Then, the Python API for copyFromArray needs to be made aware of the grid class and select the copyFromDense or copyFromDenseSDF depending on the class.

PR #1107, opened 2021-06-15
Counting Tools

danrbailey:
Adding:

* tools::countActiveLeafVoxels() [+bbox version]
* tools::countInactiveVoxels()
* tools::countInactiveLeafVoxels()
* tools::countActiveTiles()

and changed the Tree methods to use these tools.

### DIFF
--- a/openvdb/openvdb/tools/Count.h
+++ b/openvdb/openvdb/tools/Count.h
@@ -12,6 +12,7 @@
 #define OPENVDB_TOOLS_COUNT_HAS_BEEN_INCLUDED
 
 #include <openvdb/version.h>
+#include <openvdb/tree/LeafManager.h>
 #include <openvdb/tree/NodeManager.h>
 
 namespace openvdb {
@@ -29,6 +30,32 @@ Index64 countActiveVoxels(const TreeT& tree, bool threaded = true);
 /// a bounding box.
 template <typename TreeT>
 Index64 countActiveVoxels(const TreeT& tree, const CoordBBox& bbox, bool threaded = true);
+
+
+/// @brief Return the total number of active voxels stored in leaf nodes.
+template <typename TreeT>
+Index64 countActiveLeafVoxels(const TreeT& tree, bool threaded = true);
+
+
+/// @brief Return the total number of active voxels stored in leaf nodes that intersects
+/// a bounding box.
+template <typename TreeT>
+Index64 countActiveLeafVoxels(const TreeT& tree, const CoordBBox& bbox, bool threaded = true);
+
+
+/// @brief Return the total number of inactive voxels in the tree.
+template <typename TreeT>
+Index64 countInactiveVoxels(const TreeT& tree, bool threaded = true);
+
+
+/// @brief Return the total number of inactive voxels stored in leaf nodes.
+template <typename TreeT>
+Index64 countInactiveLeafVoxels(const TreeT& tree, bool threaded = true);
+
+
+/// @brief Return the total number of active tiles in the tree.
+template <typename TreeT>
+Index64 countActiveTiles(const TreeT& tree, bool threaded = true);
 
 
 /// @brief Return the total amount of memory in bytes occupied by this tree.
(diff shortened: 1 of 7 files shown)


danrbailey:
@Idclip @richhones - I believe we're hitting some memory/resourcing limit when building the AX CI using two threads with GCC and adding in this change pushed it over some arbitrary threshold. I watched the build progress and it seemed to get to `VolumeExecutable.cc` and then hang when trying to build this translation unit then shortly after access to the log would drop. On a hunch, I added a commit to drop `CMAKE_BUILD_PARALLEL_LEVEL` from 2 down to 1 and the CI appears to have succeeded as a result. 

The simplest option looks to be just making that parameter configurable for this specific configuration although that will slow down the CI quite a bit. Possibly it might help to split out the `VolumeExecutable::execute()` and related registration methods to a distinct translation unit though to ease the compile pressure? What do you think?

Idclip:
> @Idclip @richhones - I believe we're hitting some memory/resourcing limit when building the AX CI using two threads with GCC and adding in this change pushed it over some arbitrary threshold.

Makes sense, that's rather annoying - though at least we now know that this results in the logs disappearing. For this PR I suggest we do what you're saying with `CMAKE_BUILD_PARALLEL_LEVEL`. It's an env variable so can't be passed to CMake, but the build script can do something like:

```
if [ -z $CMAKE_BUILD_PARALLEL_LEVEL ]; then
  export CMAKE_BUILD_PARALLEL_LEVEL=2
fi
```

and the AX builder can export it as `CMAKE_BUILD_PARALLEL_LEVEL=1`

danrbailey:
> > @Idclip @richhones - I believe we're hitting some memory/resourcing limit when building the AX CI using two threads with GCC and adding in this change pushed it over some arbitrary threshold.
> 
> Makes sense, that's rather annoying - though at least we now know that this results in the logs disappearing. For this PR I suggest we do what you're saying with `CMAKE_BUILD_PARALLEL_LEVEL`. It's an env variable so can't be passed to CMake, but the build script can do something like:
> 
> ```
> if [ -z $CMAKE_BUILD_PARALLEL_LEVEL ]; then
>   export CMAKE_BUILD_PARALLEL_LEVEL=2
> fi
> ```
> 
> and the AX builder can export it as `CMAKE_BUILD_PARALLEL_LEVEL=1`

Thanks. I've made the changes you've suggested and it looks like everything has passed now.

openvdb/openvdb/tools/Count.h
```diff
@@ -31,6 +32,32 @@ template <typename TreeT>
...
+
+
+/// @brief Return the total number of inactive voxels stored in leaf nodes.
+template <typename TreeT>
+Index64 countInactiveLeafVoxels(const TreeT& tree, bool threaded = true);
+
+
+/// @brief Return the total number of active tiles in the tree.
+template <typename TreeT>
+Index64 countActiveTiles(const TreeT& tree, bool threaded = true);
```
Idclip, line 58: Regarding the `threaded = true` idiom, have we ever discussed deferring this logic to the user's thread lib selection (in our cases, tbb). For example:

```
auto c = countActiveLeafVoxels(tree, false);
```

can instead be achieved by:

```
{
   tbb::task_arena arena(/* max_concurrency=*/1);
   c = countActiveLeafVoxels(tree);
}
```

We could further wrap this logic in some kind of utility:

```
{
  auto d = openvdb::thread::disable();
  c = countActiveLeafVoxels(tree);
}
```

I'm just conscious of the increasing amount of API methods that are introducing this flag and we don't seem to have a good solution in place for it.

openvdb/openvdb/tree/Tree.h
```diff
@@ -350,15 +350,15 @@ class Tree: public TreeBase
...
-    Index64 inactiveLeafVoxelCount() const override { return mRoot.offLeafVoxelCount(); }
+    Index64 inactiveLeafVoxelCount() const override { return tools::countInactiveLeafVoxels(*this); }
     /// Return the total number of active voxels.
     Index64 activeVoxelCount() const override { return tools::countActiveVoxels(*this); }
     /// Return the number of inactive voxels within the bounding box of all active voxels.
-    Index64 inactiveVoxelCount() const override;
+    Index64 inactiveVoxelCount() const override { return tools::countInactiveVoxels(*this); }
     /// Return the total number of active tiles.
-    Index64 activeTileCount() const override { return mRoot.onTileCount(); }
+    Index64 activeTileCount() const override { return tools::countActiveTiles(*this); }
```
Idclip, line 361: As these are virtual, do we ever see them being deprecating?

Idclip:
lgtm but it would be great if we can decide if the threading flag is really necessary here

openvdb/openvdb/tools/Count.h
danrbailey: There's quite a lot wrapped up in this suggestion. I'm reading more frustration at us collectively not having figured out a solution to this problem (which I share) as opposed to any realistic desire to try and use this PR as the vessel through which to actually drive this change. I count 131 instances of this pattern in the codebase already, so there would likely still need to be a targeted deprecation and refactoring effort to tackle this of which the few extra instances in this PR I would suggest are pretty negligible. With that in mind, here's my comments on your proposal.

As things stand, what you are suggesting would just do multi-threading with 1 thread which is distinct from and significantly worse than single-threading. We would need additional code paths and likely some extra `openvdb::thread` functionality to determine the number of available threads and to switch the logic between the two methods if it's just one available thread.

`tbb::task_arena` is not the ideal candidate here IMO, because the method that you're calling could just create an uncapped task arena and obtain access to the full thread pool. That's not to say that this is any different to the method ignoring the `threaded=false` flag you pass in and doing it's own thing, but the ideal behaviour here is to that this method and all sub-methods are sure to not use more than 1 thread.

My suggestion is we start by introducing something like an `openvdb::thread::ParallelFor()` and use this kind of branching:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/ValueTransformer.h#L323

Then instead of controlling it by a bool thread argument, we would control it by `openvdb::thread` logic . Ideally, this could either be something that could be defined at (1) compile-time if we were to make tbb an optional dependency or (2) at runtime if the host application were using global_control by querying `tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism)` or (3) some fallback global value that could be set in a similar way to `openvdb::initialize()`.

Either way, I agree this is something we should at least try and figure out a plan for so happy to have discussions about how we should move forward here towards solving this.

openvdb/openvdb/tree/Tree.h
danrbailey: I don't see this being much different to `Tree::activeVoxelCount()` which we agreed would be too disruptive to deprecate at this stage. Perhaps some of these other methods are less popular (`Tree::inactiveLeafVoxelCount()`?), but if I don't really see the point in removing some but not others.

What is more practical perhaps is to deprecate some of the intermediate calls such as `RootNode::onTileCount()`.

openvdb/openvdb/tools/Count.h
Idclip: Let's continue this discussion on the mailing lists, are you happy for me to respond there?

One quick thing relevant for this PR; I wasn't advocating a particular system to launch these tasks single threaded, more to stop *introducing* the `threaded=false` argument as there are other ways the client can achieve this.

Idclip (approved):

PR #1109, opened 2021-06-17
Introduced hsvtorgb and rgbtohsv functions to AX

Idclip:

### DIFF
--- a/openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
+++ b/openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
@@ -2735,6 +2735,13 @@ inline FunctionGroup::UniquePtr axhsvtorgb(const FunctionOptions& op)
         llvm::Value* zero = llvm::ConstantFP::get(precision, 0.0);
         llvm::Value* one = llvm::ConstantFP::get(precision, 1.0);
 
+        // wrap hue values to [0,1] domain, including negative values
+        // i.e. -0.1 -> 0.9, 4.5 -> 0.5
+        hsv[0] = axfloormod(op)->execute({hsv[0], one}, B);
+
+        // clamp saturation values to [0,1]
+        hsv[1] = axclamp(op)->execute({hsv[1], zero, one}, B);
+
         llvm::BasicBlock* then = llvm::BasicBlock::Create(B.getContext(), "then", base);
         llvm::BasicBlock* el = llvm::BasicBlock::Create(B.getContext(), "else", base);
         llvm::BasicBlock* post = llvm::BasicBlock::Create(B.getContext(), "post", base);
@@ -2861,6 +2868,8 @@ inline FunctionGroup::UniquePtr axhsvtorgb(const FunctionOptions& op)
         .addSignature<HSVtoRGB3F, true>(generate)
         .setArgumentNames({"input"} )
         .addDependency("floor")
+        .addDependency("floormod")
+        .addDependency("clamp")
         .addParameterAttribute(0, llvm::Attribute::NoAlias)
         .addParameterAttribute(1, llvm::Attribute::ReadOnly)
         .addFunctionAttribute(llvm::Attribute::NoUnwind)
--- a/openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
+++ b/openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
@@ -2876,7 +2876,9 @@ inline FunctionGroup::UniquePtr axhsvtorgb(const FunctionOptions& op)
         .addFunctionAttribute(llvm::Attribute::AlwaysInline)
         .setConstantFold(op.mConstantFoldCBindings)
         .setPreferredImpl(op.mPrioritiseIR ? FunctionBuilder::IR : FunctionBuilder::C)
-        .setDocumentation("Convert HSV color space into RGB color space.")
+        .setDocumentation("Convert HSV color space into RGB color space. Note "
+            "that the input hue is wrapped to its periodic [0,1] values and "
+            "the input saturation is clamped between [0,1].")
         .get();
 }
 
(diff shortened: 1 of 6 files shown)


Idclip:
@richhones

richhones:
Did some testing of hsvtorgb in VEX and looks like it is does a fmod(hueinput, 1) for the hue value, and clamps the saturation between 0 and 1. I think maybe it is worth doing that? I get confused by all the mod options we have in ax so maybe worth making it easier for users? Currently AX just returns a rgb of 0.0,0.0,0.0 if you use hue outside of 0,1. 

@jmlait do you have any thoughts on whether we should clamp the inputs of this function? Or leave it as it is now e.g. replicating the Imath implementation and require users to sanitise their inputs themselves?

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
```diff
@@ -2716,6 +2716,323 @@ inline FunctionGroup::UniquePtr axsort(const FunctionOptions& op)
...
 ///////////////////////////////////////////////////////////////////////////
 
+// Colour
+
+inline FunctionGroup::UniquePtr axhsvtorgb(const FunctionOptions& op)
+{
+    auto generate =
+        [op](const std::vector<llvm::Value*>& args,
+           llvm::IRBuilder<>& B) -> llvm::Value*
+    {
```
jmlait, line 2726: Okay.... This is a really, really, strong argument to either support functions already in AX, or to create a plain text scripts that drive this building.

Reading code that assembles an LLVM parse tree is very hard.

The history of vex was first to make .vrw files that were plain text that could be parsed to generate the correct LLVM sequences, written at exactly this sort of level.  Here for example is the code to find the maximum of an array:

```
max@F[F (0) VEX_OPVANILLA {
    arraylength@I[F $it $1
    set@II $ii i(0)
    set@FF $0 f(-1e8)
    bwhile
$label1
    lt@III $ic $ii $it
    while@LI $label2 $ic
        getcomp@F[FI $ft $1 $ii
        max@FFF $0 $0 $ft
        add@II $ii i(1)
    jump@L $label1
$label2
}
```
This is then translated into the LLVM sequence points.

But, we then added the ability to just define these intrinsics as functions in the language itself, which made life infinitely better....

This rant prompted because I thought I'd just look at what function you implemented, but I can't have any idea from this...

Please prefix with a block comment with the equation you intend this to do, either written in AX or in LLVM....

jmlait (commented):
You should sanitize input and not rely on users.  Very common to do an rgb2hsv -> operation -> hsv2rgb.  If operation is something like a scale or offset, you will commonly leave the valid interval.  (And hue in particular is periodic; only accepting 0..1 is like writing a sin() function that only accepts 0..2pi.)

I looked at our implementation for any gotchas:
1) wrapmod() is needed, to ensure -0.1 hue wraps to 0.9, not to -0.1
2) rgbtohsv defines hue = 0 for saturation < 0.5/255.  That looks like some old 8-bit assumption, but there is the argument that hue becomes unstable in that interval anyways so clamping isn't too unreasonable rather than remaining unstable.  Definitely need a clamp somewhere above zero to deal with zero and denormal, you don't want barely saturated colours generating INF or NAN.
3) The output of rgbtohsv() should generate hue in [0, 1).  We have code to detect if the case statements ended up with hue == 1 and set it to 0.  I'm guessing this may occur if there is some difference big enough to select the last hextant, but rounds off to 1 in application.   That code has been there since '96.

openvdb_ax/openvdb_ax/codegen/StandardFunctions.cc
Idclip: Yeah I'm completely aware of this - it's on the roadmap but isn't a small amount of work. After working so much with IR it's become a bit too easy to fall into this trap. Additionally, because we're lacking this higher level support that you're referring to, it's **essential** these are written in IR over C++ bindings for optimisation purposes. 

To combat this issue, AX is able to take both an IR implementation **and** a C/C++ binding. If you look through `StandardFunctions.cc` you'll see a number of cases where both a C++ function pointer and an IR implementation exist. We're starting to move away from this because direct IR (JIT) implementations **always** make more sense if they exist, but this "solves" the issues you're having because you can look at the C binding to infer what the IR is doing. Further more, for bindings to other libraries, we can just test that the result from the IR implementation matches. 

In summary, the C++ implementation is used in the unit tests to validate the IR and you can use that for a more readable implementation. See the diff for the unit tests: https://github.com/AcademySoftwareFoundation/openvdb/pull/1109/files

hsvtorgb: https://github.com/AcademySoftwareFoundation/openvdb/pull/1109/files#diff-a4df75455d40759f98e9279a80fffdc008a579eab7bb32af72ebccbe06e37e3bR663
rgbtohsv: https://github.com/AcademySoftwareFoundation/openvdb/pull/1109/files#diff-a4df75455d40759f98e9279a80fffdc008a579eab7bb32af72ebccbe06e37e3bR1181

And to confirm, I am working on frontend function support which is a re-requisite to what we're discussing here.

Idclip:
@jmlait @richhones I believe I've addressed the feedback - input hue is now wrapped to the periodic [0,1] and input saturation is clamped to [0,1] for `hsvtorgb`

richhones:
> @jmlait @richhones I believe I've addressed the feedback - input hue is now wrapped to the periodic [0,1] and input saturation is clamped to [0,1] for `hsvtorgb`
Think we should probably add the checks for small input saturation values and hue == 1 as suggested by Jeff above too?

Idclip:
> Think we should probably add the checks for small input saturation values and hue == 1 as suggested by Jeff above too?

I took a look at both of these and couldn't see how the current implementation could cause the issues mentioned - though may have missed something.

richhones:
> > Think we should probably add the checks for small input saturation values and hue == 1 as suggested by Jeff above too?
> 
> I took a look at both of these and couldn't see how the current implementation could cause the issues mentioned - though may have missed something.

Hmm yeah ok looking at the reference implementations in the unit tests, I don't either.

richhones (approved):

jmlait (approved):

PR #1110, opened 2021-06-17
Deprecate string grids

Idclip:
Based on the conversation here https://github.com/AcademySoftwareFoundation/openvdb/blob/master/tsc/meetings/2021-04-20.md, deprecate `StringTree` and `StringGrid` aliases. They current cause segmentation faults on read so the aim would be to eventually deprecate them as natively supported serializable types but continue to support the their template instantiation for intermediate storage.

### DIFF
--- a/openvdb/openvdb/openvdb.cc
+++ b/openvdb/openvdb/openvdb.cc
@@ -97,7 +97,12 @@ initialize()
     DoubleGrid::registerGrid();
     Int32Grid::registerGrid();
     Int64Grid::registerGrid();
+    // @note String grids types are deprecated but we still register them
+    //   as supported serializable types for backward compatibility. This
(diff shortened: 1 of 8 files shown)


danrbailey (approved):

jmlait (approved):

Idclip:
We discussed deprecating the `NodeUnion` specialisation that's hit for non trivially-copyable types. I'm afraid I'm not sure I can efficiently deprecate this class without significant amounts of compiler output. Not only that, but there's no easy way to silence that output for `StringGrid` types whilst allowing it to propagate for custom user types. Basically I think this PR may have to go in as is and once all non trivially-copyable Tree types (i.e. string grids) are removed from VDB we can introduce this deprecation warning... although it will still spit out a tonne of stuff.

PR #1122, opened 2021-07-12
Add -DUSE_PKGCONFIG=OFF when building on Windows

eliemichel:
When using vcpkg, pkg-config is not needed.
(The instruction as they are currently do not build on a fresh Windows since there is no pkg-config)


Idclip:
Thanks for this @eliemichel, we've made it so that this option is now OFF automatically on Windows (#1237)

PR #1128, opened 2021-07-16
Extend VDB Point conversion to support Vec3f as well as Vec3d

danrbailey:


kmuseth (approved):
Excellent - thanks for these updates Dan

danrbailey:
There was some ambiguity in the function signatures that was causing the CI to fail. I found a slightly improved way to do this that doesn't require deprecations or changing the API. In this implementation, the type it uses is `PointWrapperT::value_type` unless this has not been defined in which case it falls back to Vec3R. This should mean that all existing code continues to work and it just requires a `using value_type = Vec3f` in the wrapper to be able to change the value type.

PR #1138, opened 2021-07-28
VDB Visualize Tree SOP Improvements

danrbailey:
A fairly significant re-write of the VDB Visualize Tree SOP. Here are the changes summarized:

* Remove logic for DWA-specific surfacing and region of interest (parameters are now obsolete)
* Use multi-threading for the geometry generation
* Introduce slicing and color remapping
* Add gl_lit=0 detail attribute to disable gl lighting

**Note:** care has been taken to preserve the geometry and point ordering such that the output is identical to what it was before. For instance, there is still two-pass geometry generation where the leaf nodes and internal nodes are processed first and then the voxels and tiles (even though a simpler implementation would likely have been possible by combining them).

In terms of multi-threading performance, here's the full res Disney cloud asset visualized with leaf nodes, internal nodes and tiles as wireframe box and points as voxels:

Old implementation - 1 minute 9 seconds
New implementation - 13.8 seconds

The main change here is moving away from the "addPoint()" and "addBox()" calls to a method where the offsets are pre-computed and the point, vertex and polygon data generated in parallel. The only caveat is that the polygon generation is still single-threaded for now where the solid boxes and wire-frame boxes are mixed as part of the same pass (ie internal nodes shown as wireframe, leaf nodes as solid). 

The slicing has been pretty closely modeled on the Volume Slice SOP where I tried to keep the same parameter names and layout (see parameter pane below). There are however three modes for slicing - one where only geometry that intersects a plane is generated, one where all voxels in an intersecting leaf are generated (sometimes useful visually) and finally one where the geometry is "flattened" to the plane so that a 3D box becomes a 2D square more akin to the output of the Volume Slice SOP.

Here is a side-by-side of a slice through the full res Disney cloud asset:

volumeslice

The left is the Volume Slice SOP with regular point spacing, the right is the VDB Visualize Tree SOP with geometry that matches that precise structure of the tree. Performance for this example:

Volume Slice SOP - 1.64s
VDB Visualize Tree SOP - 0.09s

The huge performance increase is primarily from avoiding the need to interpolate the VDB. Even with some very large VDBs, it's now possible to change the offset slider and have it update in real-time.

The offset defaults to a relative -1..1 range, however there is a "use world-space" toggle which changes the value to Houdini units relative to the VDB origin. The offsets are stored in the Node cache so providing the geometry has been generated once, clicking the toggle will translate the value seamlessly between the two - the key advantage being that the world space mode maps to the same plane even when the bounds of the VDB is changing (the node cache approach is similar to one used already in VDB from Particles SOP).

Finally, here's the new parameter pane:

vdbvisualize

The new parameters are all added to the bottom, but I'm open to changing the layout to make it more usable.


jmlait:
/home/jlait/dev_green/src/houdini/custom/DWA/vdb/SOP_OpenVDB_Visualize.C:1954:43: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
                  style2 == STYLE_WIRE_BOX && (style1 != STYLE_SOLID_BOX)) {
                  ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jlait/dev_green/src/houdini/custom/DWA/vdb/SOP_OpenVDB_Visualize.C:1965:51: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
                         style2 == STYLE_SOLID_BOX && (style1 != STYLE_WIRE_BOX)) {


This seems suspicious, I think the warning is right and style2 should have had a preceding parenthesis?  At least from the structure it looks that way, I've not looked at the actual logic.

jmlait:
Note on GEO_Detail there are:
   bool        setDetailAttributeI(const UT_StringHolder &aname, int v);
These make very quick ways to just splat stuff like gl_lit without so much code.

jmlait:
Why is there
          "none", "No Mapping",
when that doesn't link to a valid UT_Color setting?  The switch to test this doesn't include "none".

jmlait:
Compiled and passes our reg tests after I suppressed gl_lit.  (I don't mind gl_lit being added, btw, just suppressed to make sure the rest matched.)

jmlait (approved):

danrbailey:
Thanks for those suggestions, I've addressed them now.

> This seems suspicious, I think the warning is right and style2 should have had a preceding parenthesis? At least from the structure it looks that way, I've not looked at the actual logic.

Yeah. I think the code is correct because && precedence is before || but I agree that it's not clear, so I've tidied this up.

> Why is there
> "none", "No Mapping",
> when that doesn't link to a valid UT_Color setting? The switch to test this doesn't include "none".

I don't have the source code of the Volume Slice SOP to look at here, but I was trying to match the implementation as closely as possible. There is not a "No Mapping" option under `UT_Color::getRampColor()` even though there is in the drop-down of the Volume Slice SOP, so instead the default value it uses (when "No Mapping" is selected) is UT_COLORRAMP_NUMMODES which appears to produce the correct output?

jmlait:
NONE is a special case to directly copy the value into Cd, which also implies no clamping.  NUMMODES triggers a hundred asserts for me as the lower level libraries are not happy, but try to fail-safe to something reasonable :>

                    // Find the colour correlating to this value.
                    if (sopparms.getVismode() == SOP_VolumeSliceEnums::Vismode::NONE)
                    {
                        cd = v;
                    }
                    else if (color_ramp)
                    {
                        float vals[4];
                        color_ramp->rampLookup(v, vals);
                        cd.assign(vals[0], vals[1], vals[2]);
                    }
                    else
                    {
                        UT_Color::getRampColor((UT_ColorRamp) (int(sopparms.getVismode())-1), v,
                                        &cd.x(), &cd.y(), &cd.z());
                    }

I'd actually suggest going straight to a Ramp and not passing UT_ColorRamp at all. We've since added all the standard vis as presets in the color ramp so it feels less necessary to have that specific drop down.  Volume Slice only has it because it predated having defaults on colour ramps.  But if a full ramp seems overkill, I won't press the point either.

danrbailey:
> NONE is a special case to directly copy the value into Cd, which also implies no clamping. NUMMODES triggers a hundred asserts for me as the lower level libraries are not happy, but try to fail-safe to something reasonable :>
> 
> ```
>                 // Find the colour correlating to this value.
>                 if (sopparms.getVismode() == SOP_VolumeSliceEnums::Vismode::NONE)
>                 {
>                     cd = v;
>                 }
>                 else if (color_ramp)
>                 {
>                     float vals[4];
>                     color_ramp->rampLookup(v, vals);
>                     cd.assign(vals[0], vals[1], vals[2]);
>                 }
>                 else
>                 {
>                     UT_Color::getRampColor((UT_ColorRamp) (int(sopparms.getVismode())-1), v,
>                                     &cd.x(), &cd.y(), &cd.z());
>                 }
> ```
> 
> I'd actually suggest going straight to a Ramp and not passing UT_ColorRamp at all. We've since added all the standard vis as presets in the color ramp so it feels less necessary to have that specific drop down. Volume Slice only has it because it predated having defaults on colour ramps. But if a full ramp seems overkill, I won't press the point either.

Full ramp sounds fine to me, I'll make that change. Thanks for explaining.

danrbailey:
@jmlait - I've made the changes to switch to a Ramp as you suggested, but I'm not sure that I've done it correctly, I couldn't find many examples of working with ramps in a verbified context. I also couldn't figure out how to apply the "infra-red" preset by default so that it matches the Volume Slice SOP, so I instead opted to just set the "rampcolordefault" spare data with infra-red values. Would you mind reviewing this change please?

https://github.com/AcademySoftwareFoundation/openvdb/pull/1138/commits/c098944d40de14c8dcd52e504b82490261e121d0

jmlait:
Regarding defaulting to infrared: to my knowledge that is the best way ATM.

PR #1139, opened 2021-08-03
removed nanovdb_viewer for now

kmuseth:


Idclip:
Some initial structure thoughts before I take a closer look: 

 - The CI is written using circle. It'll need to be ported to a `.github/workflows/nanovdb.yml` and made to work with github actions. We'll also need CI which tests building/configuring OpenVDB and NanoVDB together
 - The copied CMake files in nanovdb/cmake can be removed and we should use the ones in the root `cmake` folder. This includes the ones for `IlmBase`, `OpenVDB`, `TBB`, `Blosc`, `Utils` and `GLFW3`. Additionally is there a reason we have modules for finding OpenGL and EGL? CMake comes with these.
 - There's no build support for NanoVDB from the root directory of the repo. This should be added and needs to include things like the doxygen documentation (which should work with VDBs docs), unit tests and examples as well as the core NanoVDB "lib".
 - If the docker stuff is primarily designed for the CI then I would recommend we move it to the CI folder to only provide the public essentials in `/nanovdb`
 - I propose we nest the `nanovdb` directory in a similar fashion to all the other modules i.e. `nanovdb/nanovdb/<stuff>`. This will allow us to use NanoVDB in the OpenVDB Houdini and OpenVDB AX modules and will prevent issues down the line with ambiguous include paths should we add anything else that depends on `nanovdb`.

danrbailey:
Closing this PR to avoid confusion - feature/nanovdb will remain a separate branch for anyone already working from this branch, but NanoVDB will be merged to master separately (#1213) and then this feature branch subsequently updated.

PR #1146, opened 2021-08-09
CI Updates

Idclip:
- Consolidated all specific build scripts into a single shell script
 - Added workflow dispatch triggers for some workflows to allow for building of specific components
 - Turned off building the CI on push for all branches *except* master and feature/** branches
 - Added a CI build to allow for the generation and upload of AX grammar on unix

With regards to the launching of jobs on push, I found that it can be incredibly annoying when you're working on specific features and I very rarely wanted all the CI to run on odd branches. This change only enables CI on push for feature/** and master branches, but adds workflow dispatchers so you can manually trigger whatever jobs you like. The CI for PRs is unchanged.


danrbailey:
Some nice cleanup, thanks. Two questions regarding the CI:

* I like having CI jobs kicked off automatically when pushing branches to my fork as a precursor to submitting as a PR - do these changes now require me to manually trigger jobs?
* I like having the ability to run the full suite of CI tests on my local fork before submitting as a PR - does this change remove this by making the Houdini cache update CI only available for the ASWF repo?

Idclip:
> I like having CI jobs kicked off automatically when pushing branches to my fork as a precursor to submitting as a PR - do these changes now require me to manually trigger jobs?

Yeah same, but when I'm pushing up remotes with no intention to merge its very annoying! So the compromise I ended up with is allowing CI on push for `master` and for `feature/**` branches - and if you don't want to call your branch `feature/**` yes, you'll have to manually trigger the CI. Open to suggestions here though.

> I like having the ability to run the full suite of CI tests on my local fork before submitting as a PR - does this change remove this by making the Houdini cache update CI only available for the ASWF repo?

AFAIK the Houdini cache stuff only works on the ASWF repo anyway?

danrbailey:
> > I like having CI jobs kicked off automatically when pushing branches to my fork as a precursor to submitting as a PR - do these changes now require me to manually trigger jobs?
> 
> Yeah same, but when I'm pushing up remotes with no intention to merge its very annoying! So the compromise I ended up with is allowing CI on push for `master` and for `feature/**` branches - and if you don't want to call your branch `feature/**` yes, you'll have to manually trigger the CI. Open to suggestions here though.

I understand, I guess it's a small effort to get in the habit of prefixing the branches I intend to submit as a PR. What about we also allow `pr/**`? I find we use `feature/**` for a subtly different purpose on the aswf repo - to me it mainly means branches that we wish to collaborate or share on before they're ready to become a PR into master.

> > I like having the ability to run the full suite of CI tests on my local fork before submitting as a PR - does this change remove this by making the Houdini cache update CI only available for the ASWF repo?
> 
> AFAIK the Houdini cache stuff only works on the ASWF repo anyway?

No, you just have to sync your fork master with the ASWF master at least once every six months and it builds it on schedule. I do use it.

Idclip:
> I understand, I guess it's a small effort to get in the habit of prefixing the branches I intend to submit as a PR. What about we also allow pr/**? I find we use feature/** for a subtly different purpose on the aswf repo - to me it mainly means branches that we wish to collaborate or share on before they're ready to become a PR into master.

Fine by me!

> No, you just have to sync your fork master with the ASWF master at least once every six months and it builds it on schedule. I do use it.

Right, it's the "Houdini Cache" job that runs, not the "Houdini" job. The latter one skips. That's fine, I can revert the job skip in the `houdini_cache.yaml` and keep the one in the `houdini.yaml`. I was under the impression the EULA didn't allow that but must be misremembering.

danrbailey:
> > I understand, I guess it's a small effort to get in the habit of prefixing the branches I intend to submit as a PR. What about we also allow pr/**? I find we use feature/** for a subtly different purpose on the aswf repo - to me it mainly means branches that we wish to collaborate or share on before they're ready to become a PR into master.
> 
> Fine by me!
> 
> > No, you just have to sync your fork master with the ASWF master at least once every six months and it builds it on schedule. I do use it.
> 
> Right, it's the "Houdini Cache" job that runs, not the "Houdini" job. The latter one skips. That's fine, I can revert the job skip in the `houdini_cache.yaml` and keep the one in the `houdini.yaml`. I was under the impression the EULA didn't allow that but must be misremembering.

You still need to add the client and secret key to your fork settings for it to actually store the cache, so you can still sign your own EULA and provide your own client and secret key. I believe the way we set this up - the validation check for `houdini.yaml` is skipped for everything *except* the aswf repo - the intention being that it's not a fatal error for a fork not to have access to the Houdini cache, but if it does it should use it. 

On investigation, it hasn't been working correctly for me. I believe it hit the six month window recently, meaning it was storing an empty cache where the filename matches precisely the restore key requested. Not sure that I know the cleanest way to solve that, perhaps just creating a dummy hou.tar.gz would be sufficient to generate a valid hash. Either way, let's leave it enabled for now because it certainly has been working well for me and I'll try and fix this cache issue.

Idclip:
> You still need to add the client and secret key to your fork settings for it to actually store the cache, so you can still sign your own EULA and provide your own client and secret key. I believe the way we set this up - the validation check for houdini.yaml is skipped for everything except the aswf repo - the intention being that it's not a fatal error for a fork not to have access to the Houdini cache, but if it does it should use it.

Think I've fixed this now - I've added two new lightweight steps to the Houdini/Houdini Cache workflows that only run on forks. The prior checks for the existence of a cache and only runs the CI if it exists. The latter checks if secrets have been set and only runs if they have been. These don't run on the ASWF (and the subsequent jobs assume the necessary data exists). I think this is clearer, both from visualizing the CI and in the workflows. It's also safer as the ASWF job will now fail if anything is wrong with the cache (e.g. it could previously silently succeed which checks like `[[ -d hou ]]` or `[[ -d build ]]` skipping things).

Idclip:
> On investigation, it hasn't been working correctly for me. I believe it hit the six month window recently, meaning it was storing an empty cache where the filename matches precisely the restore key requested. Not sure that I know the cleanest way to solve that, perhaps just creating a dummy hou.tar.gz would be sufficient to generate a valid hash. Either way, let's leave it enabled for now because it certainly has been working well for me and I'll try and fix this cache issue.

I've solved this by bumping the cache version as I also ran into this problem. Seems so weird that the cache action doesn't have a dedicated push/pull and instead works the way it does... really difficult to work with.

Idclip:
Okay, I've decided the github actions cache workflow is pretty terrible (the action itself, not how we're using it). The only way I can get this to work is if we merge this with broken Houdini jobs. That will allow me to then run the Houdini Cache job, which will resync the cache. I can then add back in the skips to the Houdini job. Otherwise, if I do something like this:

```
    - id: fetch_houdini
      uses: actions/cache@v2
      with:
        path: hou
        key: vdb-v1-houdini${{ matrix.config.hou }}-${{ hashFiles('hou/hou.tar.gz') }}
        restore-keys: vdb-v1-houdini${{ matrix.config.hou }}-
    - if: steps.fetch_houdini.outputs.cache-hit == 'true'
       ....
```

The Houdini job completes and submits an empty 0mb cache with the restore key. That is then always prioritised.

In summary, I'll need to fix this with a subsequent PR once I'm able to re-run the Houdini Cache job on master.

danrbailey (approved):
This looks good, thanks Nick. I'm fine with up-versioning the Houdini cache, I know that this action is not great in many ways.

PR #1148, opened 2021-08-09
Fix bug with VDB LOD SOP Chaining

danrbailey:
The VDB LOD SOP can output grids with either int64 or float metadata with the same name ("MultiResGrid_Level") depending on the mode used. This fix is to address an issue reported in production where it's not possible to chain two VDB LOD SOPs with different modes. You can replicate by using one VDB LOD SOP with mode == "single level" and another connected to it with mode == "LOD Pyramid" or vice-versa.

In this case, it now checks whether the metadata already exists on the input grid and is of an incompatible type and then drops that metadata before calling the MultiResGrid code.

Note that I only fixed for ABI>=7 because we're soon to remove support for older ABIs / Houdini versions anyway.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_LOD.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_LOD.cc
@@ -182,12 +182,26 @@ struct MultiResGridFractionalOp
     template<typename GridType>
     void operator()(const GridType& grid)
     {
+        using TreeT = typename GridType::TreeType;
         if ( level <= 0.0f ) {
             outputGrid = typename GridType::Ptr( new GridType(grid) );
         } else {
             const size_t levels = openvdb::math::Ceil(level) + 1;
-            using TreeT = typename GridType::TreeType;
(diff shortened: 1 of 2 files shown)


Idclip (commented):
lgtm, I'd confirm with @jmlait 

> Note that I only fixed for ABI>=7 because we're soon to remove support for older ABIs / Houdini versions anyway.

To clarify, this compile time abi branching is used because the method `grid.copyReplacingMetadata` is ABI >= 7 (this confused me at first)

danrbailey:
> lgtm, I'd confirm with @jmlait
> 
> > Note that I only fixed for ABI>=7 because we're soon to remove support for older ABIs / Houdini versions anyway.
> 
> To clarify, this compile time abi branching is used because the method `grid.copyReplacingMetadata` is ABI >= 7 (this confused me at first)

Yes, sorry that's what I meant - the bug will still be present on ABI=6 / Houdini 18.0 because this convenient method to replace just the metadata was not available until OpenVDB 7.0.

jmlait (approved):

PR #1153, opened 2021-08-17
Explicit Instantiation Unit Test

danrbailey:
This PR adds a new unit test that performs explicit template instantiation for many of the most common math and tree classes. As the explicit template instantiation does a full instantiation of all methods, this catches errors in the VDB API that have not been noticed because they either haven't been instantiated by anything in the OpenVDB library or reported by any of our users who have attempted to call them. One simple example is `RootNode::stealChild(const MapCIter& i, ...)` - attempting to call this method results in a compile error as you cannot steal data from a const iterator. 

The biggest change here is introducing `ConstValueIter`, `ConstChildIter` and `ConstDenseIter`. Currently, non-const base iterators are used even for const iterators (such as `using ValueOnCIter = typename BaseLeaf::template ValueIter<...>`) with a comment not to call these methods (such as `// Note: setItem() can't be called on const iterators.`). This change could also be implemented using SFINAE, but I felt introducing new iterators was cleaner.

Here's the summary of changes to get this new unit test to compile:

* New ConstValueIter, ConstChildIter and ConstDenseIter classes - the same as the non-const iterators but with no `setItem()` or `modifyItem()` calls, rewire all RootNode, InternalNode and LeafNode iterator aliases for const iterators.
* Move `operator<<(ostream&)` method for LeafNode buffer into LeafBuffer and use public method to access data.
* Remove `RootNode::stealChild(const MapCIter& i, ...)` method.
* Add `RootNode::probeLeaf(const Coord&)` definition, which was previously declared but not defined.
* Fix compile error in Quat `read(ostream&)` and `write(istream&)` by using reinterpret_cast instead of static_cast.
* Fix compile error in `Mat3::setSkew()` by passing correct template parameter.
* Fix lots of float conversion warnings in Mat, Mat4 and Quat by using static_cast.
* Fix warning in `VecT::unitSafe()` where it was treating a float value as a bool instead of comparing with zero.

Ultimately, I expect we would probably remove this unit test (as it's not really a unit test, but a compile test) when we introduce proper explicit template instantiation as part of the library, but this at least tests that no new errors are introduced in the interim.

### DIFF
--- a/openvdb/openvdb/math/Mat.h
+++ b/openvdb/openvdb/math/Mat.h
@@ -278,8 +278,8 @@ rotation(const Vec3<typename MatType::value_type> &_axis, typename MatType::valu
     Vec3<T> axis(_axis.unit());
 
     // compute trig properties of angle:
-    T c(cos(double(angle)));
-    T s(sin(double(angle)));
+    T c(static_cast<T>(cos(double(angle))));
+    T s(static_cast<T>(sin(double(angle))));
     T t(1 - c);
 
     MatType result;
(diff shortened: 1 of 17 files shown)


danrbailey:
I'm going to close this PR and keep it on my own local branch. Given that this is no longer priority for 9.0 because we're not intending to explicitly instantiate anything in the OpenVDB hierarchy or math classes and that the changes are fairly extensive, seems like it would be good to just drop this PR for the time being and come back to it at a later date if needed.

PR #1163, opened 2021-09-01
Integrate Imath and OpenEXR v3.1

cary-ilm:
All components seem to compile and tests succeed, although there may be other constraints that aren't accounted for. I haven't built the Houdini or Maya plugins, so I'm not sure if they introduce other complexity.

Imath_ROOT/OpenEXR_ROOT and/or CMAKE_PREFIX_PATH point the build to specific installations, but in lieu of pre-installed dependencies, this uses CMake's FetchContent to automatically download and build Imath and OpenEXR from GitHub, so it's not essential for these dependencies to exist prior to building. This is the strategy that OpenEXR 3.x uses to build Imath.

This replaces all instances of "IlmBase" with "Imath". Imath has no "components", it is a single library that includes half.h along with everything else.

This removes entirely the local copy of Half.h and the USE_IMATH_HALF option. Ideally, it would be better to embrace this as a required dependency and be free of the extra complexity.

Even if this isn't the 100% solution, hopefully it serves as a starting point for discussion.

Signed-off-by: Cary Phillips <cary@ilm.com>


cary-ilm:
The FetchContent step effectively pulls a half.h into the build implicitly, with hopefully no extra work on the user's part. It's not a "dependency" in the sense of something the user has to have in place before the build. It's the Find module that does the fetching.

If you have use cases that require support for IlmBase 2.x, it's certainly possible to pull that back in, I just thought it best to present here only what's required for 3.x  I recall mention that you'd hoped to remove the use of OpenEXR completely, which would leave only Imath and half.h.

danrbailey:
Hi Cary,

Thanks very much for making a start on this. I echo Nick's comments, but I'll reply to your comments and add some extra context if I can.

> The FetchContent step effectively pulls a half.h into the build implicitly, with hopefully no extra work on the user's part. It's not a "dependency" in the sense of something the user has to have in place before the build. It's the Find module that does the fetching.

The problem with FetchContent is that it requires an internet connection and we've typically been operating under the assumption that many of our users will be building OpenVDB from behind a firewall. We want to avoid forcing people to download and build Imath/OpenEXR separately even if they have no internet access. The embedded half.h is intended to support this workflow, but we certainly should replace these headers with up-to-date ones pulled from Imath 3.1 in releasing v9.0.

> If you have use cases that require support for IlmBase 2.x, it's certainly possible to pull that back in, I just thought it best to present here only what's required for 3.x I recall mention that you'd hoped to remove the use of OpenEXR completely, which would leave only Imath and half.h.

Our current policy is to support current and last two years of the VFX Reference Platform (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/tsc/process/deprecation.md). This means that we would retain support for building against (an external) OpenEXR 2.4 in all releases until OpenVDB 11.0 (November 2024) when OpenEXR 3.1 would become our minimum. This is mainly to allow people a little grace period in migrating and so that our users can build the latest version of OpenVDB in such a way that it still works with vendor software shipped two years ago.

I don't believe we have any plans to remove support for OpenEXR entirely, but it has been made into an optional dependency now so that the vdb_render binary is the only tool that _requires_ OpenEXR.

cmake/FindImath.cmake
```diff
@@ -0,0 +1,121 @@
...
+include(GNUInstallDirs)
+
+set(IMATH_REPO "https://github.com/AcademySoftwareFoundation/Imath.git" CACHE STRING
+    "Repo for auto-build of Imath")
+set(IMATH_TAG "master" CACHE STRING
+  "Tag for auto-build of Imath (branch, tag, or SHA)")
+
+find_package(Imath CONFIG QUIET)
+
+if(NOT TARGET Imath::Imath AND NOT Imath_FOUND)
```
meshula, line 93: FetchContent is a huge convenience when developing. To address the concern in the main conversation thread about potential difficulties of reaching across a firewall to the internet, a simple fix, to accommodate both, would be to add a top-level option such as USE_LOCAL_IMATH, default ON. Such a defaulted option would non-disruptively preserve the existing functionality.

meshula (commented):
Thanks for taking this on Cary, very much appreciated!

meshula:
@ldclip, Since there's an expectation of being able to build with a minimum of OpenEXR 2.4, an introduction of Find is necessary, because we didn't introduce Config.cmake files until 2.5.something, if I recall correctly. To that end, the Find script would have to be able to detect 2.4 through current, correct? The script could be dropped once OpenVdb's minimum coincided with the introduction of the Config.

cmake/FindImath.cmake
danrbailey: I think we already have the option you're proposing - there's `USE_IMATH_HALF` which defaults to OFF so as to use the embedded half. When set to ON, it enables the workflow where you can use an externally built Imath.

I don't know if I see the benefit of having three ways to use Imath - (1) using the embedded half, (2) on-demand download through FetchContent and (3) by building against an externally built one. My preference for Half is that we only support (1) and (3) - with (1) remaining the default. Essentially, we're making the assumption that there are two main groups of users - individuals and studios. For the individual user, the embedded half is likely preferable whereas for the firewalled studio the externally built half is likely preferable, so I'm not clear on who would benefit most from the FetchContent workflow?

I should add that I don't necessarily object to introducing FetchContent in general - for dependencies that are too large or cumbersome to embed it might make a lot of sense, but as the Half dependency is so lightweight, I think the embedded option is even better.

danrbailey:
> @ldclip, Since there's an expectation of being able to build with a minimum of OpenEXR 2.4, an introduction of Find is necessary, because we didn't introduce Config.cmake files until 2.5.something, if I recall correctly. To that end, the Find script would have to be able to detect 2.4 through current, correct? The script could be dropped once OpenVdb's minimum coincided with the introduction of the Config.

I think you're right about this. It might be ok to add support for the config and an option to enable it's use, but it's probably simpler to do this only when we can drop 2.4 support.

danrbailey:
Closing this PR as we now have support for OpenEXR 3.1.

PR #1167, opened 2021-09-09
Prototype of support OpenEXR/Imath v3.x alongside v2.x

cary-ilm:
This is not almost certainly complete and may still be missing
critical features, but hopefully it serves as another reference in the
steps towards full support for OpenEXR/Imath 3.x.

This will look for Imath first (via -DImath_ROOT= if provided), and if
that fails, it will look for IlmBase (via -DIlmBase_ROOT= if
provided), then set the dependencies appropriately. If the
find_package(OpenEXR) finds v3.x it sets the dependencies
appropriately. Both these use $<TARGET_NAME_IF_EXIST:XXX> that
simultaneously pull both IlmBase and Imath libraries, assuming that
only one set of components will exist, with the other being expanded
to nothing. The same thing accounts for the OpenEXR libraries, which
change from "IlmImf" to "OpenEXR". And Iex and IlmThread have moved
from IlmBase to OpenEXR.

The USE_IMATH_HALF option and the local copy of half.h are unchanged
and should work just as before.

As a separate step, the copy of Half.h should be updated, that's
untouched here.

Signed-off-by: Cary Phillips <cary@ilm.com>


openvdb/openvdb/cmd/CMakeLists.txt
```diff
@@ -116,8 +117,12 @@ endif()
...
 
 if(OPENVDB_BUILD_VDB_RENDER)
-  find_package(IlmBase ${MINIMUM_ILMBASE_VERSION} REQUIRED COMPONENTS Half Iex IlmThread Imath)
-  find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED COMPONENTS IlmImf)
+  find_package(Imath CONFIG)
+  if (NOT TARGET Imath::Imath)
+    find_package(IlmBase ${MINIMUM_ILMBASE_VERSION} REQUIRED COMPONENTS Half Iex IlmThread Imath)
+  endif()
+
+  find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED)
```
Idclip, line 125: Does OpenEXR come with a config modue? Can we not do the same thing here as above (search for OpenEXR CONFIG and recover if the target doesn't exist with our Find scripts)

cmake/FindOpenEXR.cmake
```diff
@@ -198,6 +173,34 @@ if(EXISTS "${OpenEXR_INCLUDE_DIR}/OpenEXRConfig.h")
   set(OpenEXR_VERSION ${OpenEXR_VERSION_MAJOR}.${OpenEXR_VERSION_MINOR})
 endif()
 
+if(${OpenEXR_VERSION_MAJOR} MATCHES "3")
+  set(_OPENEXR_COMPONENT_LIST OpenEXR Iex IlmThread)
+else()
+  set(_OPENEXR_COMPONENT_LIST IlmImf)
```
Idclip, line 179: Any reason why IlmImfUtil was removed?

openvdb/openvdb/Types.h
```diff
@@ -9,7 +9,7 @@
 #include "TypeList.h" // backwards compat
 
 #ifdef OPENVDB_USE_IMATH_HALF
-#include <OpenEXR/half.h>
+#include <half.h>
```
Idclip, line 12: Is this correct? As far as I remember the path to half was `include/OpenEXR/half.h`

cmake/FindOpenVDB.cmake
```diff
@@ -619,10 +619,16 @@ if(OpenVDB_USES_LOG4CPLUS)
...
+  if (NOT TARGET Imath::Imath)
+    find_package(IlmBase REQUIRED COMPONENTS Half)
+  endif()
+
   if(WIN32)
     # @note OPENVDB_OPENEXR_STATICLIB is old functionality and should be removed
-    if(OPENEXR_USE_STATIC_LIBS OR (${ILMBASE_LIB_TYPE} STREQUAL STATIC_LIBRARY))
+    if(OPENEXR_USE_STATIC_LIBS OR
+        (${ILMBASE_LIB_TYPE} STREQUAL STATIC_LIBRARY)
+        (${IMATH_LIB_TYPE} STREQUAL STATIC_LIBRARY))
```
Idclip, line 631: Is there some kind of implicit `OR` between these two parenthesised statements? Looks a bit odd

Idclip:
@cary-ilm this looks great, I have a few questions above but I don't see any issues with this approach

openvdb/openvdb/cmd/CMakeLists.txt
meshula: OpenEXR 3 has a config module, but not OpenEXR 2.x. If I'm not mistaken the approach Cary is using here is following the pattern in the OpenEXR 3 migration guide written by @lgritz

cmake/FindOpenEXR.cmake
meshula: The architectural change of note in OpenEXR3 is that we reorganized all the libraries per community feedback.

First, we split out Imath to be a standalone library with no other dependencies; previously it depended on Iex for example.

Second, we consolidated the various OpenEXR image format libraries and the support they required, under an OpenEXR module.

openvdb/openvdb/Types.h
meshula: Since Half used to be a separate library, under Half/half, and is now in Imath, and is not part of OpenEXR, the forward looking approach is to refer to half.h directly, through having it discoverable in the include search paths.

cmake/FindOpenVDB.cmake
meshula: definitely need an OR in there.

meshula:
@cary-ilm @Idclip ~ I've tested this patch on Windows debug and release in an environment with both OpenEXR 2 and 3, and it successfully built with 3 as desired. So clean bill of health on Windows in a real world environment.

cmake/FindOpenEXR.cmake
```diff
@@ -198,6 +173,34 @@ if(EXISTS "${OpenEXR_INCLUDE_DIR}/OpenEXRConfig.h")
   set(OpenEXR_VERSION ${OpenEXR_VERSION_MAJOR}.${OpenEXR_VERSION_MINOR})
 endif()
 
+if(${OpenEXR_VERSION_MAJOR} MATCHES "3")
```
lgritz, line 176: Just as an aside, this is equivalent and perhaps more clear (or at least, more general, if you needed a divider at something other than a major version only):

    if (OpenEXR_VERSION VERSION_GREATER_EQUAL 3.0)

cmake/FindOpenEXR.cmake
Idclip: @meshula that makes sense thanks, but I'm referring to the branch where branch where `OpenEXR_VERSION_MAJOR` != 3. This used to import `IlmImfUtil` for EXR 2 and it's been removed for EXR 2. I don't think we're even using it but I don't see why it was removed. Just trying to minimise any behaviour changes if they're not necessary.

openvdb/openvdb/Types.h
Idclip: At the moment our FindOpenEXR module brings in both `-I/EXR_ROOT/include/OpenEXR -I/EXR_ROOT/include` because of the way IlmBase and OpenEXR used to be installed separately. I really don't like doing `-I/IMATH_ROOT/include/Imath` and then bringing in half with `#include <half.h>`, but seems the alternative is to introduce a header define (https://github.com/AcademySoftwareFoundation/Imath/blob/master/docs/PortingGuide2-3.md#accommodating-openexrimath-3x-or-openexr-2x-1) as there's no filepath alias... Does the Imath config module bring in both include paths? i.e. if I built with `find_package(Imath CONFIG) + target_link_libraries(myproj Imath)` does that bring in the right include paths such that writing `#include <half.h>` works? If not, we'll have to go with the define branch option.

openvdb/openvdb/cmd/CMakeLists.txt
Idclip: Yeah so it seems this pattern is being follows for Imath, but not for OpenEXR https://github.com/AcademySoftwareFoundation/Imath/blob/master/docs/PortingGuide2-3.md#openexrimath-3x-only

I'd expect:
```
find_package(Imath CONFIG)
if (NOT TARGET Imath::Imath)
  # use our modules for both ilmbase and openexr
else()
   find_package(OpenEXR CONFIG)
endif()
```

Basically only use our FindOpenEXR and FindIlmBase modules if we can't find CONFIG modules. Assuming the CONFIG modules were introduce in OpenEXR 3 then we don't even need to make any changes to our Find modules right?

cary-ilm: I just pushed a commit that follows your suggestion. I'm not entirely sure if it's right, but it seems to build properly with 2.x and 3.x in my tests.

cmake/FindOpenEXR.cmake
cary-ilm: I took it out originally because it wasn't used so seemed superfluous, but I've added it back now, along with the 3.x counterpart OpenEXRUtil.

openvdb/openvdb/Types.h
cary-ilm: I changed this to do a conditional include of "OpenEXR/half.h" or "Imath/half.h" based on the Imath_VERSION, now included in version.h. I think this is the best solution.

cmake/FindOpenVDB.cmake
cary-ilm: Typo.

cmake/FindOpenEXR.cmake
cary-ilm: I changed it to use VERSION_GREATER_EQUAL.

openvdb/openvdb/cmd/CMakeLists.txt
Idclip: My thought process here is as follows: We ideally shouldn't have FindOpenEXR and FindIlmBase modules in OpenVDB, we should exclusively use your CONFIG modules. As EXR 3 comes with these, we can first try searching for libs with the CONFIG signature and, if that fails, we can revert back to the MODULE signature. This has the effect of looking for EXR 3 first and falling back to EXR 2.X with our own modules.

We needed these modules for 2 reasons a) EXR didn't install CONFIG files and b) we needed custom location of exr libs from 3rdparty packages such as Houdini. I don't think we need the latter anymore as the EXR symbols provided by Houdini namespaced so we need our own copy anyway (I think this is correct).

lgritz: A complicating factor is that some OpenEXR 2.x releases also had config files, to differing degrees of brokenness, and in any case, the names of the targets of the OpenEXR 2.x config files are different from those of the OpenEXR 3.x config files.

I've found that what you should do is look for `find_package(Imath CONFIG)` first, if that succeeds, it's definitely 3.x you're dealing with (because there was no separate Imath package for 2.x. I think this is what you're leaning toward above as well.

cary-ilm: I just pushed another commit that fixes the code block in question here to add find_package(OpenEXR) alongside the find_package(IlmBase) in the case where it can't find Imath, i.e. version 2.x, which I believe retains the previous behavior. I'd accidentally removed that, so the build checks failed. Hopefully this gets things back to working order.

cmake/FindOpenEXR.cmake
cary-ilm: The houdini CI build is failing, and apparently it can't find IlmImfUtil.  Is it possible this is linking against OpenEXR built into Houdini that doesn't include IlmImfUtil? I'm not sure what's different in that build vs the others. I can't easily build the Houdini dependencies, if someone who can could possibly look into what's going on.

danrbailey: Yes, I think that is the problem. As far as I can tell - the find_package in openvdb/cmd/CMakeLists did not previously request the IlmImfUtil component so adding it in is almost certainly what is causing the Houdini CI to fail. However, all I think Nick was suggesting was to leave IlmImfUtil as a possible component in cmake/FindOpenEXR.cmake, even though it doesn't look like anything is actually using it.

cary-ilm:
I pushed a commit that removes IlmImfUtil from the components line in openvdb/cmd/CMakeLists.txt, and it looks like the Houdini CI tests build but are failing the test with a link error that seems unrelated:

2: Test command: /usr/local/bin/python2.7 "/__w/openvdb/openvdb/openvdb/openvdb/python/test/TestOpenVDB.py" "-v"
2: Environment variables: 
2:  PYTHONPATH=/usr/local/lib/python2.7/site-packages:/usr/local/lib64/python2.7/site-packages:/usr/local/lib/python::/__w/openvdb/openvdb/build/openvdb/openvdb/python
2: Test timeout computed to be: 10000000
2: Traceback (most recent call last):
2:   File "/__w/openvdb/openvdb/openvdb/openvdb/python/test/TestOpenVDB.py", line 18, in <module>
2:     import pyopenvdb as openvdb
2: ImportError: libblosc.so.1: cannot open shared object file: No such file or directory

Any clues what might be lead to this?

Idclip:
Hey @cary-ilm yeah it's unrelated, your changes look good to me

cary-ilm:
Good to know, thanks.  We'll leave it in your hands at this point, but obviously don't hesitate to contact us if there are any further issues.

danrbailey:
Thanks @cary-ilm, this is a big help. I'm going to merge this into a new feature/openexr3 branch and then submit a new PR into master to apply the additional changes to adapt to the RPATH changes Nick has made and we can resolve any additional issues from there.

PR #1173, opened 2021-09-18
Fix description of indexToWorld method

3Descape:
"index" and "world" should be exchanged with each other


danrbailey:
Thanks for submitting. In addition to the signed CLA (which you've done), you also need to sign off all your commits so that we can consider accepting this. It's as simple as added a "Signed-off-by" statement in your commit message, more info here (https://probot.github.io/apps/dco/).

3Descape:
Hey!
I signed it now and I think it should be ready to go. 
Thanks ;)

danrbailey (approved):

kmuseth:
I have added this fix and will make it available in the feature/nanovdb branch shortly

PR #1174, opened 2021-09-18
Fix: Memory alignment in nanovdb_GridData

3Descape:
The fields mGridIndex and mGridCount are missing in the struct, which leads to an 8 bit memory missalignment with the C++ data in the buffer of gridhandler->data().
This caused invalid data in mMap, as it is defined, after the missalignmented happened. Artefacts of this bug could be seen in the OpenCL version of the viewer.

For reference see the struct GridData in NanoVDB.h(line 2131-2148).

### DIFF
--- a/nanovdb/cmd/viewer/code/CNanoVDB.h
+++ b/nanovdb/cmd/viewer/code/CNanoVDB.h
@@ -424,6 +424,8 @@ CNANOVDB_DECLARE_STRUCT_BEGIN(nanovdb_GridData)
     uint64_t         mChecksum; // 8B. Checksum of grid buffer.
     uint32_t         mMajor;// 4B. major version number.
     uint32_t         mFlags; // 4B. flags for grid.
+    uint32_t         mGridIndex;// 4B. Index of this grid in the buffer
+    uint32_t         mGridCount; // 4B. Total number of grids in the buffer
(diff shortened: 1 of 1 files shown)


3Descape:
There seems to be another memory alignment issue somewhere, that causes incorrect values when reading from the grid via TreeContext and nanovdb_ReadAccessor.
I'm currently investigating this issue and will let you know, once I've found it.

3Descape:
The problem seems to be in the nanovdb_RootData_* struct. 
In C++ e.g. the mBackground member starts at byte offset 28, while in C it starts at 36. Also, the members in itself are quite different. I'm not a 100% sure, how to best resolve this yet. 

grafik

3Descape:
(sorry for the many comments, but I will just document the needed changes here for a start, and if it works, I'll update the pullrequest)(at least from what I think should be updated.. I'm by far no expert, just trying out if it solves it)

CNANOVDB_DECLARE_STRUCT_BEGIN(nanovdb_Node##LEVEL##_##SUFFIX)

grafik

3Descape:
CNANOVDB_DECLARE_STRUCT_BEGIN(nanovdb_RootData_Tile_##SUFFIX)
childID should be int64_t and not int32_t

3Descape:
All changes combined will be submitted in a new pull-request

Issue #1172, opened 2021-09-16
[BUILD] Auto-detect OpenVDB ABI when building Houdini plugin


plattfot:
### Environment
**Operating System:** Linux CentOS 7.7.1908 
**Version / Commit SHA:** 8.1.0
**CMake Version:** 3.18
**Compiler:** gcc 9.4.0

### Describe the problem
When building `openvdb_houdini` for `houdini-19.0.376` cmake errors out with:

``` sh
CMake Error at CMakeLists.txt:655 (message):
  Located OpenVDB installation is not ABI compatible with Houdini Version
  19.0.376.  Requires ABI 7, found ABI 8.


-- Configuring incomplete, errors occurred!
```

`houdini-19.0.376` does ship with `openvdb-8.1.0` with ABI 8 according to the version file:
`<houdini_root>/toolkit/include/openvdb/version.h`

``` c++
// Auto generated build configurations from CMake

///@{
/// @brief Library major, minor and patch version numbers
/// @hideinitializer
#define OPENVDB_LIBRARY_MAJOR_VERSION_NUMBER 8
/// @hideinitializer
#define OPENVDB_LIBRARY_MINOR_VERSION_NUMBER 1
/// @hideinitializer
#define OPENVDB_LIBRARY_PATCH_VERSION_NUMBER 0
///@}

/// @note  This ifndef exists for compatibility with older versions of OpenVDB.
///   This value should never be different from the value configured when
///   OpenVDB was built, but this previously needed to be defined by downstream
///   software. Redefining it here would cause build failures, so this allows
///   users to transition and remove the define in their build systems.
#ifndef OPENVDB_ABI_VERSION_NUMBER
/// @brief The ABI version that OpenVDB was built with
/// @hideinitializer
#define OPENVDB_ABI_VERSION_NUMBER 8
#endif

/// @brief Library version number string of the form "<major>.<minor>.<patch>"
/// @details This is a macro rather than a static constant because we typically
/// want the compile-time version number, not the runtime version number
/// (although the two are usually the same).
/// @hideinitializer
#define OPENVDB_LIBRARY_VERSION_STRING "8.1.0"

/// @brief Library version number string of the form "<major>.<minor>.<patch>abi<abi>"
/// @details This is a macro rather than a static constant because we typically
/// want the compile-time version number, not the runtime version number
/// (although the two are usually the same).
/// @hideinitializer
[...]

PR #1176, opened 2021-09-20
Set ABI=8 as default for Houdini and lock H18.5 to ABI=7

danrbailey:
As reported by @plattfot in issue #1172.

I've slightly tweaked the logic here so that it compares with the major.minor version of Houdini and thus doesn't require us to make any assumptions about which version of Houdini that SideFX might ship next, so that we can opt to extend this as soon as we enter a new calendar year.

### DIFF
--- a/cmake/OpenVDBHoudiniSetup.cmake
+++ b/cmake/OpenVDBHoudiniSetup.cmake
@@ -145,6 +145,8 @@ elseif(MINIMUM_HOUDINI_VERSION)
   endif()
 endif()
 
+set(Houdini_VERSION_MAJOR_MINOR "${Houdini_VERSION_MAJOR}.${Houdini_VERSION_MINOR}")
+
 find_package(PackageHandleStandardArgs)
 find_package_handle_standard_args(Houdini
   REQUIRED_VARS _houdini_install_root Houdini_FOUND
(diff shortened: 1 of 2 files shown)


Idclip:
This instead now assumes the ABI will always be bumped with a half version? I'm fine with this, just checking.

Can we also change the error messages? We have a maximum known ABI->Houdini Version map and if a Houdini Version is detected outside (read: higher/later) than this we can error that the ABI check _may_ not be correct. I think `-DOPENVDB_HOUDINI_ABI=<>` can be provided as an override in these cases.

danrbailey:
> This instead now assumes the ABI will always be bumped with a half version? I'm fine with this, just checking.

I wouldn't put it quite like that, I think this merely assumes that the next version of Houdini after 18.5 will be ABI=8. The point is that expressing it like this doesn't provide any explicit information about what the next version of Houdini will be or what the timeline is, so we can change this logic sooner than needing to know (and be allowed to make public) the precise major.minor release number.

> Can we also change the error messages? We have a maximum known ABI->Houdini Version map and if a Houdini Version is detected outside (read: higher/later) than this we can error that the ABI check _may_ not be correct. I think `-DOPENVDB_HOUDINI_ABI=<>` can be provided as an override in these cases.

This is a good idea, I added this.

Idclip (approved):

Issue #1175, opened 2021-09-20
[BUILD] VDB AX SOP not included in build with other SOPs by default when building AX


richhones:
When building both OpenVDB AX and the OpenVDB Houdini Plugin (without specifying any custom dso list), it no longer automatically builds the OpenVDB AX SOP. Instead, in order to build the OpenVDB AX plugin you need to specify both OPENVDB_BUILD_HOUDINI_PLUGIN=ON and OPENVDB_BUILD_SOP_OpenVDB_AX=ON. 

This may be intentional but highlighting it is a change in functionality of the build system to how it used to work for building the OpenVDB AX SOP.

danrbailey:
Hi @richhones - I agree that I would have thought the latter flag would not be required (ie OPENVDB_BUILD_HOUDINI_PLUGIN =ON and OPENVDB_BUILD_AX=ON would build the AX SOP?), but I'm a bit confused by all the AX options.

It is no doubt this PR that has changed the behaviour, see the discussion between @Idclip and myself (https://github.com/AcademySoftwareFoundation/openvdb/pull/1087#issuecomment-869641362). Would one of you be able to propose / submit a change to address this?

richhones:
Ok @danrbailey thanks for verifying this wasnt intended behaviour. Have submitted a PR to fix this now.

PR #1177, opened 2021-09-21
Bugfix/axsopdefault

richhones:
This is to address the issue in [#1175](https://github.com/AcademySoftwareFoundation/openvdb/issues/1175). The AX SOP will now be included if AX and the Houdini SOPs are being built.


Idclip (approved):

danrbailey (approved):
Thanks @richhones!

PR #1178, opened 2021-09-21
AX Attribute Bindings

richhones:
Added attribute bindings to AX to allow point attributes and volumes to be accessed by different AX attribute names. This has been added to the VDB AX SOP, and ax::run() functions but is not yet exposed in the command line binary.


openvdb_ax/openvdb_ax/compiler/AttributeBindings.h
```diff
@@ -0,0 +1,158 @@
...
+///   axname: a   ->  dataname: b
+///   axname: b   ->  dataname: a
+class AttributeBindings
+{
+public:
+
+    AttributeBindings() {}
+
+    AttributeBindings(const AttributeBindings& other)
+        :  mAXToDataMap(other.mAXToDataMap), mDataToAXMap(other.mDataToAXMap) {}
```
Idclip, line 43: I think we can remove this copy constructor and let the compiler init it. Do we need the empty constructor too? i.e. is this class ever created without a bindings vector?

openvdb_ax/openvdb_ax/compiler/PointExecutable.cc
```diff
@@ -343,8 +346,11 @@ struct PointExecuterOp
 
         // add attributes based on the order and existence in the attribute registry
         for (const auto& iter : mAttributeRegistry.data()) {
-            const std::string& name = (iter.name() == "P" ? mPositionAttribute : iter.name());
-            addAttributeHandle(args, leaf, name, iter.type(), iter.writes());
+            const std::string* name = mAttributeBindings.dataNameBoundTo(iter.name());
+            // all persistent attributes are in bindings
+            assert(name);
+            if (*name == "P") name = &this->mPositionAttribute;
+            addAttributeHandle(args, leaf, *name, iter.type(), iter.writes());
```
Idclip, line 353: Does this need to be done per leaf? Can't we do it once in execute()?

openvdb_ax/openvdb_ax/compiler/PointExecutable.h
```diff
@@ -121,6 +122,13 @@ class PointExecutable
     /// @return  The current grain size
     size_t getGrainSize() const;
 
+    /// @brief  Set attribute bindings.
+    /// @param attributeBindings A map of attribute bindings to expected names on
+    ///   the geometry to be executed over
+    void setAttributeBindings(const AttributeBindings& bindings);
```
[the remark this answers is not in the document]
Idclip: Same deal in the VolumeExecutable

openvdb_ax/openvdb_ax/compiler/AttributeBindings.h
```diff
@@ -0,0 +1,158 @@
...
+
+    /// @brief Returns the map of data attribute names to AX attribute names
+    inline const std::map<Name, Name>& dataToAXMap() const {
+        return mDataToAXMap;
+    }
+
+private:
+
+    std::map<Name, Name> mAXToDataMap;
+    std::map<Name, Name> mDataToAXMap;
```
Idclip, line 149: I think we should replace all uses of `openvdb::Name` with `std::string`

Idclip:
This looks great @richhones I only have minor comments - can you also add a file in `pendingchanges` with the change description

openvdb_ax/openvdb_ax/compiler/AttributeBindings.h
richhones: in the sop it currently does create a bindings vector and then populate it, this could obviously be changed to use some temporary vector if needed. I can remove these though and let the compiler do it for me.

openvdb_ax/openvdb_ax/compiler/PointExecutable.cc
richhones: This has been updated, made some minor changes to PointExecuterOp and where the P stuff gets handled to only need to use the bindings map in one place( processAttributes())

openvdb_ax/openvdb_ax/ax.h
```diff
@@ -63,7 +63,13 @@ void uninitialize();
...
 /// @param  ax    The null terminated AX code string to parse and compile
 /// @param  grid  The grid to which to apply the compiled AX function
-void run(const char* ax, openvdb::GridBase& grid);
+/// @param  bindings A vector of attribute mappings from the AX string to the
+///                  names of the point attributes/grids (points/volumes resp.).
+///                  These should be pairs of strings e.g.
+///                  {{"axname0","dataname0"}, {"axname1","dataname1"}} see
+///                  AttributeBindings.h for details.
+void run(const char* ax, openvdb::GridBase& grid,
+         const std::vector<std::pair<openvdb::Name, openvdb::Name>>& bindings = {});
```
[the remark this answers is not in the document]
richhones: As discussed online, to do this, and allow default arg of no bindings requires AttributeBindings* arg which will not allow use of openvdb::ax::run(str, grid, {{"axname0", "dataname0"}, {"axname1","dataname1"}}) one liner.

richhones:
All changes except the fwd declaration of bindings in ax.h have been made now.

Idclip (approved):

PR #1179, opened 2021-09-21
Fix typo in cmake/FindOpenVDB.cmake to handle the QUIET argument of find_package() correctly

sterin:
I've changed `OpenVDB_FIND_QUIET` to `OpenVDB_FIND_QUIETLY`. It fixes a typo, as the [docs](https://cmake.org/cmake/help/latest/command/find_package.html) say that:

> When loading a find module or package configuration file find_package defines variables to provide information about the call arguments (and restores their original state before returning):
> 
> ...
> 
> \<PackageName\>_FIND_QUIETLY
> True if QUIET option was given

### DIFF
--- a/cmake/FindOpenVDB.cmake
+++ b/cmake/FindOpenVDB.cmake
@@ -433,7 +433,7 @@ if(NOT _OPENVDB_HAS_NEW_VERSION_HEADER)
   endif()
 endif()
 
-if(NOT OpenVDB_FIND_QUIET)
+if(NOT OpenVDB_FIND_QUIETLY)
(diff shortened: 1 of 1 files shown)


Idclip:
@sterin thanks for this, could you modify your commit with a DCO signoff for this check (link has instructions):

https://github.com/AcademySoftwareFoundation/openvdb/pull/1179/checks?check_run_id=3668602067

sterin:
@Idclip I've signed off on the commit, at least I think I did.

Idclip (approved):

jmlait (approved):

PR #1181, opened 2021-09-22
Exposed const accessors data buffers of AttributeArray as of ABI >= 9

Idclip:

### DIFF
--- a/openvdb/openvdb/points/AttributeArray.h
+++ b/openvdb/openvdb/points/AttributeArray.h
@@ -344,6 +344,11 @@ public:
     bool operator==(const AttributeArray& other) const;
     bool operator!=(const AttributeArray& other) const { return !this->operator==(other); }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Virtual function to retrieve the data buffer cast to a char byte array
+    virtual const char* dataAsByteArray() const = 0;
+#endif
+
 private:
     friend class ::TestAttributeArray;
 
@@ -353,7 +358,9 @@ private:
 
     /// Virtual function to retrieve the data buffer cast to a char byte array
     virtual char* dataAsByteArray() = 0;
+#if OPENVDB_ABI_VERSION_NUMBER < 9
     virtual const char* dataAsByteArray() const = 0;
+#endif
 
     /// Private implementation for copyValues/copyValuesUnsafe
     template <typename IterT>
@@ -739,12 +746,22 @@ public:
     /// Return @c true if all data has been loaded
     bool isDataLoaded() const override;
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Return the raw data buffer
+    inline const StorageType* data() const { assert(validData()); return mData.get(); }
+
+    /// Virtual function to retrieve the data buffer from the derived class cast to a char byte array
+    const char* dataAsByteArray() const override;
+#endif
+
 protected:
     AccessorBasePtr getAccessor() const override;
 
     /// Return the raw data buffer
     inline StorageType* data() { assert(validData()); return mData.get(); }
+#if OPENVDB_ABI_VERSION_NUMBER < 9
     inline const StorageType* data() const { assert(validData()); return mData.get(); }
+#endif
 
     /// Verify that data is not out-of-core or in a partially-read state
     inline bool validData() const { return !(isOutOfCore() || (flags() & PARTIALREAD)); }
@@ -772,7 +789,9 @@ private:
 
     /// Virtual function to retrieve the data buffer from the derived class cast to a char byte array
     char* dataAsByteArray() override;
+#if OPENVDB_ABI_VERSION_NUMBER < 9
     const char* dataAsByteArray() const override;
+#endif
 
     size_t arrayMemUsage() const;
     void allocate();


Idclip:
@danrbailey the only slightly annoying thing doing it this way is that you have to explicitly call the const methods:

```
AttributeArray* a = new TypedAttributeArray<float>(...);
const char* buffer = a->dataAsByteArray(); // fails
const char* buffer = const_cast<const AttributeArray*>(a)->dataAsByteArray(); // make const, works
```

I see no issues exposing the non-const accessors (I need them and with this addition have to const_cast to non-const anyway) but let me know your thoughts

danrbailey:
> @danrbailey the only slightly annoying thing doing it this way is that you have to explicitly call the const methods:
> 
> ```
> AttributeArray* a = new TypedAttributeArray<float>(...);
> const char* buffer = a->dataAsByteArray(); // fails
> const char* buffer = const_cast<const AttributeArray*>(a)->dataAsByteArray(); // make const, works
> ```
> 
> I see no issues exposing the non-const accessors (I need them and with this addition have to const_cast to non-const anyway) but let me know your thoughts

What we've done in other places to solve this problem is to expose a uniquely named method that only provides the "const" accessor such as https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/points/PointDataGrid.h#L398.

I'm fine with providing const access, because if you choose to cast away the const-ness, you're knowingly bypassing the API and thus we are not responsible if we break that code by making changes here. I would not want to have code in the repo here that does this though.

We have discussed this aspect before and I think Jeff's suggestion was the best option, I'll post his comment from that discussion:

> With PageArrays, we added the GA_PageHandle, which is similar to that Buffered object above. It does a zero-copy in-place edit if the types match and the block of data is not constant, but buffers into a flat array if not. The algorithm then can operate in place on its native type as if it were uncompressed. Again, this is one page (1k data elements) at a time so can stream nicely.
> 
> This nicely amortizes all the codec decision code paths and allows the codec to run efficiently on a block; so in most algorithms that are doing a dense traversal it has very little overhead even when it has to copy the data. The local buffer will be hot in the cache, and the reading from the source is done as a burst so doesn't have to indirect every access.

We should try and do something like this so that we can offer modify-in-place with AttributeArrays.

Idclip:
> I would not want to have code in the repo here that does this though.

To clarify, you don't want to add non-const accessors but are okay with me const_casting in AX? I'm a bit worried that this reads that you don't want any const_casting in the repo - I already do this for the NodeUnion hook which we previously discussed and this behaves in the same way (it's pretty integral so some of the things I'm working on)

danrbailey:
> > I would not want to have code in the repo here that does this though.
> 
> To clarify, you don't want to add non-const accessors but are okay with me const_casting in AX? I'm a bit worried that this reads that you don't want any const_casting in the repo - I already do this for the NodeTable hook which we previously discussed and this behaves in the same way (it's pretty integral so some of the things I'm working on)

I don't want to add non-const accessors because it offers an officially supported option in the API to bypass the compression which means we can easily end up in a situation where code written against this API can no longer support new compression codecs being added. I would rather we don't add const_casts in our own code, but as AX is a bit of a special case, I'm happy to talk about it on a case-by-case basis to be sure that we're not making the kind of assumptions that we'll come to regret later.

The proposed solution to this problem is the one I shared in my comment and I don't expect this to need an ABI change so I think this can be added at any time if the need becomes great enough.

Idclip:
>  I'm happy to talk about it on a case-by-case basis

Cool, sounds good.

So for this PR, you propose I should add new const methods with unique names that just calls the private const methods?

Idclip:
@danrbailey bump

> So for this PR, you propose I should add new const methods with unique names that just calls the private const methods?

can you confirm this is what you'd like to see here?

danrbailey:
> @danrbailey bump
> 
> > So for this PR, you propose I should add new const methods with unique names that just calls the private const methods?
> 
> can you confirm this is what you'd like to see here?

Sorry for slow reply, yes exactly. Something like `constDataAsByteArray()` would be fine with me.

openvdb/openvdb/points/AttributeArray.h
```diff
@@ -344,6 +344,11 @@ class OPENVDB_API AttributeArray
     bool operator==(const AttributeArray& other) const;
     bool operator!=(const AttributeArray& other) const { return !this->operator==(other); }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 9
+    /// Indirect virtual function to retrieve the data buffer cast to a char byte array
+    const char* constDataAsByteArray() const { return this->dataAsByteArray(); }
```
danrbailey, line 349: Any good reason why this isn't also a virtual function?

Idclip: Well it's redundant no? `Typed` array will have access to both and has a function which returns a `const StorageType*` i.e:

```
const TypedAttributeArray<>* a = ...
const char* cbuf = a->constDataAsByteArray();
auto buf = a->constData();
```

danrbailey: I just meant if you wanted to do this:
```
const AttributeArray* a = ...
const char* cbuf = a->constDataAsByteArray();
```
This doesn't really apply for the `constData()` method as you'd need to the know the storage type. However for the `constDataAsByteArray()` method, if we're going to expose it, we might as well expose it on the AttributeArray as a virtual method?

Idclip:
> However for the constDataAsByteArray() method, if we're going to expose it, we might as well expose it on the AttributeArray as a virtual method?

Yeah that's what I've done right? So we've ended up with:

```
class AttributeArray
{
public:
  const char* constDataAsByteArray() const { this->dataAsByteArray(); }
private:
  virtual const char* dataAsByteArray() const =0;
};

class TypedArray
{

private:
  const char* dataAsByteArray() const override { ... }
};
```

rather than:

```
class AttributeArray
{
public:
  virtual const char* constDataAsByteArray() const  =0;
private:
  virtual const char* dataAsByteArray() const =0;
};

class TypedArray
{
  const char* constDataAsByteArray() const override { this->dataAsByteArray(); }
private:
  const char* dataAsByteArray() const override { ... }
};
```

PR #1182, opened 2021-09-23
Add Explicit Template Instantiation Support

danrbailey:
This PR adds explicit template instantiation support for `tools::activate()` and `tools::deactivate()`. Once the new framework and CMake changes are approved, I'll add support to more methods.

The essence of these changes are as we discussed in our TSC meetings, although the implementation is a little different. It might help to start with an example of how to add explicit template instantiation support to a header file. This is an example from `tools/Activate.h`:

```
#define _ACTIVATE(GridOrTreeT) \
    void activate(GridOrTreeT&, const GridOrTreeT::ValueType&, const GridOrTreeT::ValueType&, const bool)
OPENVDB_TREE_INSTANTIATE(_ACTIVATE)
```

The macro `OPENVDB_TREE_INSTANTIATE` expects another macro as an argument which takes a tree type as its argument. In the header file, this gets expanded to:

```
extern template void activate(FloatTree&, ...);
extern template void activate(DoubleTree&, ...);
```

which suppresses all template instantiation for this method and tree type.

There are also new CMake-generated source files, here's how one such example (`ActivateFloatTree.cc`) would look after expansion:

```
#include <openvdb/openvdb.h>
#include <openvdb/tools/Activate.h>
extern template void activate(FloatTree&, ...);
template void activate(FloatTree&, ...);
```

(Technically the template instantiation statements would live at the footer of the Activate.h file, but just for demonstration purposes).

There is one of these CMake-generated source files for each header and tree type so as to provide the greatest opportunity for parallelism (`ActivateFloatTree.cc`, `ActivateDoubleTree.cc`, etc).

There is a CMake option to disable this feature (it's on by default) - `USE_TEMPLATE_INSTANTIATION` and the inline statements have been removed from these templated methods to allow the explicit template instantiation to work. As we discussed, this is understood to not be an issue for ODR because this rule does not apply to function templates.

Finally, some numbers, I've only included `TestActivate.cc` in the vdb_test binary here. With explicit template instantiation disabled:

libopenvdb.so - 7.3MB
vdb_test - 3.7MB

Time to compile TestActivate.o - 16.9s

With explicit instantiation enabled:

libopenvdb.so - 9.4MB
vdb_test - 2.8MB

Time to compile TestActivate.o - 12.6s
Time to compile ActivateFloatTree.o - 1.8s

Obviously, the benefits for this relatively simple example are not particularly strong, but some of the other tools I'll move onto after this will provide much more significant improvements.

@e4lam


e4lam:
Alas, looks like Windows is still failing!

danrbailey:
> Alas, looks like Windows is still failing!

Indeed. It took me a couple of attempts to fix the escaped characters for the auto-generation on Windows, but I think I've resolved that now. This is why we have Windows CI. :)

danrbailey:
I've just pushed up the two improvements mentioned in our TSC meeting today:

* added an `#undef` to the macros created in the headers
* moved source file auto-generation up to CMake level instead of build time

Much nicer, thank you for the feedback! :)

openvdb/openvdb/CMakeLists.txt
```diff
@@ -443,6 +469,52 @@ set(OPENVDB_LIBRARY_THREAD_INCLUDE_FILES
...
+      )
+
+      # The auto-generated file defines the OPENVDB_TREE_TEMPLATE macro for a single tree type
+      # and then the header file itself which defines all functions to be explicitly instantiated.
+      execute_process(
+        COMMAND echo
+          "#define OPENVDB_TREE_TEMPLATE(Function)  template OPENVDB_TEMPLATE_API Function(${INSTANTIATION_TREE});\n"
+          "#include <openvdb/${INSTANTIATION_FILE}>"
+        OUTPUT_FILE ${INSTANTIATION_CMAKE_SOURCE}
+      )
```
Idclip, line 508: You might even be able to replace this with `file(WRITE)`:
https://cmake.org/cmake/help/latest/command/file.html#write

danrbailey: Yes, that's better, thanks. I made that change now.

danrbailey:
I've rebuilt this PR with some additional changes. This adds support for explicit template instantiation for all the classes and functions under tools for which this makes sense (excluding those that require an Interrupter for now).

I've re-done the CMake so that the configuration is largely driven by macros that get written into version.h. I've also combined all the instantiations into a single translation unit per-header for simplicity. Finally, I fixed an issue where nested headers were also being instantiated where they shouldn't have been.

e4lam:
@danrbailey Just curious about your Windows fix in [ddcce5a](https://github.com/AcademySoftwareFoundation/openvdb/pull/1182/commits/ddcce5a9c26a1f6a4174dfa4e5ec68377ff9c286)

It should be unnecessary for min/max issues because we already define `NOMINMAX` when building on Windows. But perhaps your issue was something else.
```
./CMakeLists.txt:  list(APPEND OPENVDB_CORE_PUBLIC_DEFINES -D_WIN32 -DNOMINMAX)
```

danrbailey:
> > @danrbailey Just curious about your Windows fix in [ddcce5a](https://github.com/AcademySoftwareFoundation/openvdb/pull/1182/commits/ddcce5a9c26a1f6a4174dfa4e5ec68377ff9c286)
> > It should be unnecessary for min/max issues because we already define `NOMINMAX` when building on Windows. But perhaps your issue was something else.
> > ```
> > ./CMakeLists.txt:  list(APPEND OPENVDB_CORE_PUBLIC_DEFINES -D_WIN32 -DNOMINMAX)
> > ```
> 
> I'm still trying to track this issue down. It looks like we moved the NOMINMAX definition from CMake to `openvdb/Platform.h` (#1040) and I suspect this is somehow a header include ordering issue, but I haven't confirmed that yet.

This was indeed a header ordering issue. Even without explicit template instantiation, attempting to compile an executable with a single line `#include <openvdb/tools/FastSweeping.h>` on Windows currently fails. This is due to the fact that the headers that pull in min/max macros are included before `openvdb/Platform.h` where the logic that disables them lives. I've reordered the includes in the FastSweeping header file and the Windows build is now working again.

There was another small issue in that the additional disk space required for all the instantiation object files when building the Windows static library and all unit tests in debug mode exceeded the available disk space that GitHub Actions provides, so I've disabled explicit instantiation in this configuration for the time being.

I believe this feature is ready to be merged. The implementation here follows largely the same pattern for each of the headers, but mainly looking for feedback from @idclip regarding the CMake integration.

PR #1184, opened 2021-09-25
Fix: CNanoVDB.c/h for OpenCL

3Descape:
This commit updates a few structs and some of the functionality, such that the memory layout of the C version matches the one from C++ again. This mismatch caused the kernel of the OpenCL implementation of the renderer to fail.

Since the new version uses base + offset instead of a array access based on a childID,
I tried to implement the childIndex calculation based on memory addresses as generic as possible, such that it can be used for OpenGL/GLSL as well.(see e.g. nanovdb_Node##LEVEL##_##SUFFIX##_getChild ).
Although I have to admit, I have no experience with GLSL and, in general, am quite new to the OpenCL/OpenGL world.
I couldn't get the OpenGL backend fully working, as it throws some errors about some implicit casts and similar. Sadly I currently don't have the time to investigate further.


jmlait:
Thank you for the PR!

I was a bit confused because I thought we had fixed the offsets in CNanoVDB, as I use that myself.  But looking closer I realize that this was in the viewer subdirectory.  I didn't realize we had cloned the CNanoVDB header there so when I fixed the offsets in nanovdb/CNanoVDB.h I didn't fix this version.  You should be able to also copy that file over to get a fix.

There currently is directory re-org coming in (PR1171) so we may need to re-do this after that.   But we may do it by just copying the nanovdb/nanovdb/CNanoVDB.h over.

PR #1185, opened 2021-09-25
Synchronization of these files with H19.

jmlait:
This includes CE grid support (From 18.5 but last sync Nano wasn't official
yet)

The one real change (which isn't just FYI/reference) is to GEO_VDBTranslator
to ignore empty vdbs when building bounding boxes.  We were enlarging
the final stat by the min/max of the bbox, but for invalid boxes that
obviously will be wrong.  This showed up as the File SOP load-as-info
giving huge bounding boxes if an empty vdb was present.


danrbailey (approved):

jmlait:
Are these failing tests known to fail occasionally?  Not sure how my changes triggered them.

danrbailey:
> Are these failing tests known to fail occasionally? Not sure how my changes triggered them.

These failures happen consistently and repeatedly since 1 week ago - I investigated this issue here #1186

PR #1186, opened 2021-09-28
Move find_package for Blosc, TBB, IlmBase and Log4CPlus up to root CMakeLists

danrbailey:
Our Houdini CI has been failing for almost a week now and along with it any other PRs we've submitted since then:

https://github.com/AcademySoftwareFoundation/openvdb/actions/workflows/houdini.yml?query=event%3Aschedule

I would guess it's somehow related to #1055 based on the timing of when other PRs have been merged, but it's not clear to me what in this PR might have caused these failures to suddenly start to happen.

On investigating, I noticed that the RPATH for the python module pyopenvdb.so did not include the Houdini dsolib directories where libblosc.so was intended to have been retrieved from. A little more digging led me to realize that the variable `Blosc_LIBRARY_DIRS` was being read from (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/python/CMakeLists.txt#L264) but only set in the find_package call in the core library (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/CMakeLists.txt#L155). As variables in CMake only propagate from parent to child and these two CMakeLists are essentially siblings (they're both invoked by an add_subdirectory call in the root CMakeLists), this wasn't working.

The change I'm making here is simply to move the find_package calls for Blosc, TBB, IlmBase and Log4CPlus up into the root CMakeLists which fixes the problem. I couldn't move the Boost find_package call because it relies on the `OPENVDB_CORE_SHARED` variable that comes from the core library CMakeLists. This is also a problem for the cmd and unittests CMakeLists hence why I opted to move the find_package instead of calling it again from each CMakeLists that requires it, though I'm open to better solutions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -742,6 +742,68 @@ determine the ABI number. You may set this to force a given ABI number.]=] FORCE
 
 ##########################################################################
 
+# Collect and configure lib dependencies
+
+if (OPENVDB_BUILD_CORE)
+  find_package(TBB ${MINIMUM_TBB_VERSION} REQUIRED COMPONENTS tbb)
+  if(OPENVDB_FUTURE_DEPRECATION AND FUTURE_MINIMUM_TBB_VERSION)
+    if(${Tbb_VERSION} VERSION_LESS FUTURE_MINIMUM_TBB_VERSION)
+      message(DEPRECATION "Support for TBB versions < ${FUTURE_MINIMUM_TBB_VERSION} "
+        "is deprecated and will be removed.")
+    endif()
+  endif()
+
+  if(USE_IMATH_HALF)
+    find_package(IlmBase ${MINIMUM_ILMBASE_VERSION} REQUIRED COMPONENTS Half)
+    if(OPENVDB_FUTURE_DEPRECATION AND FUTURE_MINIMUM_ILMBASE_VERSION)
+      if(${IlmBase_VERSION} VERSION_LESS FUTURE_MINIMUM_ILMBASE_VERSION)
+        message(DEPRECATION "Support for IlmBase versions < ${FUTURE_MINIMUM_ILMBASE_VERSION} "
+          "is deprecated and will be removed.")
+      endif()
+    endif()
+  endif()
+
+  if(USE_LOG4CPLUS)
+    # Find Log4CPlus libraries
+    find_package(Log4cplus ${MINIMUM_LOG4CPLUS_VERSION} REQUIRED)
+  endif()
+
+  if(USE_BLOSC)
+    # Find Blosc libraries
+    find_package(Blosc ${MINIMUM_BLOSC_VERSION} REQUIRED)
+    if(Blosc_FOUND AND Blosc_VERSION VERSION_GREATER MINIMUM_BLOSC_VERSION)
+      message(WARNING "The version of Blosc located is greater than ${MINIMUM_BLOSC_VERSION}. "
+        "There have been reported issues with using later versions of Blosc with OpenVDB. "
+        "OpenVDB has been tested fully against Blosc ${MINIMUM_BLOSC_VERSION}, it is "
+        "recommended that you use this version where possible."
+      )
+    endif()
+  else()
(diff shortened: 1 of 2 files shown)


danrbailey:
@Idclip - do you have thoughts about this one? I'm hoping to try and get the CI fixed soon. We can also do a stop-gap of disabling the python binding being built in the Houdini CI while we try and figure out what to do here?

Idclip:
Yeah I'm trying to understand this more will update asap

Idclip:
This has been a nightmare to debug. There's a number of things wrong but here are my findings:

 - 1) The reason we're noticing this now is that, now that the python module is installed correctly (i.e. `/usr/local/lib/python/site-packages`) it is being prioritised over the python module in the CMake binary directory during the tests. On the one hand this is useful, on the other it's perhaps unexpected when running the tests in the binary directory. 
 - 2) Our include/library paths with Houdini are completely broken with the Houdini CI because it uses the ASWF image with installed dependencies to `/usr/local/include`. We can fix this, but the only reason this PR works is because, at runtime, it's picking up some libraries (e.g. tbb) from `/usr/local/lib` which is not good. This is why this only manifested for blosc which only exists in the houdini dso directory.
 - 3) You're correct in your diagnosis in that the reason this fails is due to missing RPATHs. The thing is we shouldn't even need to manually specify RPATHs ourselves, CMake should figure this out for us and re-link the rpaths when `make install` is run. It simply **refuses** to work. There is an option for this `CMAKE_INSTALL_RPATH_USE_LINK_PATH` which has always been set to on. I've tried completely removing all manual settings of RPATHs but CMake always strips all rpath information during `make install`. I cannot replicate this behaviour locally so I'm at a total loss as to why it's insistent on doing it on the github runner.

In summary, we still need to explicitly set the RPATHs unless we can figure out why CMake refuses to add them automatically. This means your suggestion here works but I'm trying to consider how VDB might be embedded into other projects and what the nicest way to work would be. Additionally we need to fix the Houdini CI paths either by changing the order so TBB/Blosc are included first or by removing the ASWF docker image and installing boost/exr ourselves (we should probably do both). 

I'll continue working on this, if we want the CI to pass for the time being I suggest we disable the python tests but leave this PR open.

Idclip:
Okay I held off commenting and then, in typical fashion, immediately figured something out... We currently unpack the houdini install into the checkout copy of VDB on the github runner like so:

```
| openvdb.                   // repo checkout 
| --- houdini_install/... // the downloaded houdini tar
| --- ....
| --- CMakeLists.txt    // Root CMakeLists.txt of VDB
```

I *think* that this is causing CMake to ignore the libraries when relinking the rpaths because it thinks they're part of the build tree. I was already aware of this behaviour and because the houdini install is NOT in the binary directory I didn't think it would be a problem. But it seems that it also considers libraries in the root source to be part of the build 👎 I've simply moved the extracted Houdini libraries out of the vdb root source directory and RPATHs are now appearing without manually specifying them. Will tidy this up to confirm (and still need to fix the include dirs).

danrbailey:
> Okay I held off commenting and then, in typical fashion, immediately figured something out... We currently unpack the houdini install into the checkout copy of VDB on the github runner like so:
> 
> ```
> | openvdb.                   // repo checkout 
> | --- .build/                   // binary dir
> | --- houdini_install/... // the downloaded houdini tar
> | --- ....
> | --- CMakeLists.txt    // Root CMakeLists.txt of VDB
> ```
> 
> I _think_ that this is causing CMake to ignore the libraries when relinking the rpaths because it thinks they're part of the build tree. I was already aware of this behaviour and because the houdini install is NOT in the binary directory I didn't think it would be a problem. But it seems that it also considers libraries in the root source to be part of the build 👎 I've simply moved the extracted Houdini libraries out of the vdb root source directory and RPATHs are now appearing without manually specifying them. Will tidy this up to confirm (and still need to fix the include dirs).

Oh that is surprising. Good find!

danrbailey:
Cancelling in favour of #1188

PR #1187, opened 2021-09-28
Bump to CY2020 as Minimum VFX Ref Platform

danrbailey:
I created a new v9.0 branch and took this opportunity to bump some of the other versions we have been preparing to do for some time now such as CMake 3.12 => 3.15.

This also means that Houdini 18.5 is now the minimum supported version of Houdini by this version of OpenVDB. As we've done in the past and as dictated by our deprecation policy, we still support ABI=6 until v9.1 is released.

I've also set the future minimum GCC version to 9.3.1 as the next major release will support C++17 as the minimum version.


.github/workflows/ax.yml
```diff
@@ -52,27 +52,22 @@ jobs:
       CXX: ${{ matrix.compiler }}
     strategy:
       matrix:
-        image: ['2019-clang8', '2020-clang7', '2021-clang10']
```
Idclip, line 55: Think 2020-clang7 should stay here no? At least we still test against 2020 I think

ci/install_cmake.sh
```diff
@@ -13,6 +13,6 @@ fi
 
 mkdir build
 cd build
-cmake ../.
+cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ ../.
```
Idclip, line 16: huh, cmake doesn't like clang?

openvdb/openvdb/openvdb.cc
```diff
@@ -28,14 +28,19 @@
         #error ABI = 6 is deprecated, CMake option OPENVDB_USE_DEPRECATED_ABI_6 suppresses this error
     #endif
 #endif
+#ifndef OPENVDB_USE_DEPRECATED_ABI_7
+    #if OPENVDB_ABI_VERSION_NUMBER == 7
+        #error ABI = 7 is deprecated, CMake option OPENVDB_USE_DEPRECATED_ABI_7 suppresses this error
+    #endif
+#endif
```
Idclip, line 35: We're should remove the ABI6 branching right? Think we're dropping that in 9. Same in the .cc

Idclip:
This looks great, thanks for doing this @danrbailey

.github/workflows/ax.yml
danrbailey: Well - it doesn't work because with these changes the minimum LLVM version becomes 8 so we can't use Clang 7? Unfortunately we don't have a '2020-clang8' we can build against.

ci/install_cmake.sh
danrbailey: I believe the issue is that Clang7 doesn't support string_view, which the newest CMake 3.15 requires. It does seem to work with GCC6 though, so I just opted to build CMake using GCC instead.

openvdb/openvdb/openvdb.cc
danrbailey: Our current deprecation policy is to leave ABI=6 in place until 9.1 when we would drop it?

https://github.com/AcademySoftwareFoundation/openvdb/blame/master/tsc/process/deprecation.md#L20

.github/workflows/ax.yml
Idclip: Hmm that's unfortunate... lets continue like this for now, I'll have to consider what to do here

danrbailey: We can discuss adding a Clang 8 version to the 2020 container, but what I suggest might be simpler is to just accept LLVM 7 as the minimum version this year and then set LLVM 10 as the future minimum when we bump to CY2021. That's what I've implemented here for now.

Tagging @richhones.

Idclip: Hmm I guess testing against 2020 is considerably more important that dropping support for LLVM 7. I think this would be okay

PR #1188, opened 2021-09-30
RPATH and Houdini improvements

Idclip:
- Removed explicit RPATHs in favour of CMake's automatic behaviour
 - Changed the Houdini CI to extract the Houdini install outside the OpenVDB root source to fix CMake install RPATHs being ignored
 - Changed the include order for VDB dependencies to better resolve issues where system libraries would conflict
 - Download CMake rather than building from source (saves ~5min)


danrbailey (approved):
This looks good to me, thanks Nick!

PR #1190, opened 2021-10-01
initial commit of NanoVDB merge candidate

kmuseth:


danrbailey (approved):
This looks good to me as an initial merge candidate. I think we should get this brought into feature/nanovdb_merge_candidate as soon as possible and then iterate on getting that branch ready for master.

This is not a big deal as the files in question look small and I imagine will change infrequently, but wondered if it might be better to put the binary PDF data into the openvdb-website repo (and leave the .tex files here)?

apradhana (approved):
I agree with @danrbailey with regards to taking out the pdf from this repository, but leaving the tex doc here. I'll put my comments and questions regarding the build system when we want to merge into master.

Idclip:
My only comment is that there's seems to be an unnecessary layer for the `nanovdb` headers (they are currently 3 folders deep, it only needs to be 2) eg:

```
nanovdb
└── nanovdb
    ├── nanovdb
    │   ├── CNanoVDB.h
    │   ├── NanoVDB.h
    │   ├── PNanoVDB.h
```

vs

```
nanovdb
└── nanovdb
    ├── CNanoVDB.h
    ├── NanoVDB.h
    ├── PNanoVDB.h
```

but if we prefer it like this (as there's other stuff in there which I agree shouldn't necessarily be picked up with an include) then it's fine!

kmuseth:
removed the third nanovdb directory

PR #1192, opened 2021-10-03
Update NanoVDB Merge Candidate with all other v9.0 Changes

danrbailey:
This contains changes to bump all the versions ready for v9 and adds support for OpenEXR 3.0.


Idclip:
We want to get the merge candidate into master, we shouldn't be merging v9 changes. We should just merge current master.

danrbailey:
> We want to get the merge candidate into master, we shouldn't be merging v9 changes. We should just merge current master.

Wouldn't it be better to get the merge candidate into v9? We're not planning to make NanoVDB available for v8.x so would make sense to me to target the actual CI configuration we'll use for v9.

At some point, we should probably create a v8.2 branch and merge v9 to master, so perhaps we should just talk about whether to switch which version the master points to.

Idclip:
Yeah we can discuss this in TSC. To me though it makes sense for master to be separated from any v9 changes until we are ready to release. Alternatively we can do the 8.2 release or make an 8.2 branch and then iterate v9 on master, but I figured we'd simply merge master->feature/nano_merge->master.

danrbailey:
Closed in favour of #1197 as discussed in TSC (merging v9 -> master -> feature/nanovdb_merge_candidate).

PR #1196, opened 2021-10-06
Made EXR optional for vdb_render and added 8bit PNG support

Idclip:


openvdb/openvdb/cmd/CMakeLists.txt
```diff
@@ -81,17 +81,40 @@ endif()
...
-  find_package(Imath CONFIG)
-  if (NOT TARGET Imath::Imath)
-    find_package(IlmBase ${MINIMUM_ILMBASE_VERSION} REQUIRED COMPONENTS Half Iex IlmThread Imath)
-    find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED COMPONENTS IlmImf)
-  else()
-    find_package(OpenEXR CONFIG)
+
+  if(USE_PNG)
+    find_package(PNG REQUIRED)
+  endif()
```
Idclip, line 87: Kind of annoying you have to manually enable this (i.e. `-DUSE_PNG=ON`). @jmlait I know we discussed auto enabling dependencies if they're found and you were against it. I agree for the core library but perhaps for binary features this would be nice?

openvdb/openvdb/CMakeLists.txt
```diff
@@ -505,8 +505,7 @@ list(APPEND OPENVDB_CORE_PRIVATE_DEFINES -DOPENVDB_PRIVATE)
 if(WIN32)
   if(USE_IMATH_HALF)
     # @note OPENVDB_OPENEXR_STATICLIB is old functionality and should be removed
-    if(ILMBASE_USE_STATIC_LIBS OR (${ILMBASE_Half_LIB_TYPE} STREQUAL STATIC_LIBRARY) OR
-       IMATH_USE_STATIC_LIBS OR (${IMATH_LIB_TYPE} STREQUAL STATIC_LIBRARY))
```
Idclip, line 509: So this failed for me on Windows because `IMATH_LIB_TYPE` isn't a think. @cary-ilm I see that Imath checks for a `Imath_DLL` define. Is this automatically defined on Windows for us if Imath is built as a DLL? Or do downstream projects still need to define this themselves?

PR #1242, opened 2021-11-04
Fix potential issue in nanovdb HostBuffer.h

kmuseth:
Fixed a move assignment inconsistency in nanovdb/util/HostBuffer.h.
This could potentially be a problem on Windows debug builds.


kmuseth:
I would like in the future that we label bugs as "critical" or "non-critical", where the latter refers to fixes to issues that are unconfirmed, i.e. could in theory result in errors which have notably not been observed or maybe even reproduced. In other words, they are not critical but more preventive in nature.

Idclip:
@kmuseth I'm confused, we fixed this with the version 9 release but you've added changes for 9.0.1? If this fix went out with 9 its not worth documenting as NanoVDB wasn't officially part of VDB until 9 anyway

kmuseth:
> @kmuseth I'm confused, we fixed this with the version 9 release but you've added changes for 9.0.1? If this fix went out with 9 its not worth documenting as NanoVDB wasn't officially part of VDB until 9 anyway

I don't blame you for the confusion! The failing unit-test you reported was related to a bug in the move **constructor** of the HostBuffer class. This fixed went into v9.0.0 as you correctly points out and need no further mentioning. The issue fixed in the PR is however related to the move **assignment operator** of HostBuffer, which might suffer from the same issue as the move constructor (almost identical logic which undoubtedly caused your confusion). However, I'm unable to confirm that it's an actual issue and obviously the unit-tests are not failing - even on Windows debug builds, which is why the CHANGES documentation for version 9.0.1 tag this as a non-critical bug fix.

Idclip:
> > @kmuseth I'm confused, we fixed this with the version 9 release but you've added changes for 9.0.1? If this fix went out with 9 its not worth documenting as NanoVDB wasn't officially part of VDB until 9 anyway
> 
> @Idclip I don't blame you for the confusion! The failing unit-test you reported was related to a bug in the move **constructor** of the HostBuffer class. This fixed went into v9.0.0 as you correctly points out and need no further mentioning. The issue fixed in this PR is however related to the move **assignment operator** of HostBuffer, which might suffer from the same issue as the move constructor (almost identical logic which undoubtedly caused your confusion). However, I'm unable to confirm that it's an actual issue and obviously the unit-tests are not failing - even with Windows debug builds, which is why the CHANGES documentation for version 9.0.1 tag this bug as a non-critical.

Ahhh my apologies, that makes total sense! Originally thought this was the same fix

Idclip (approved):

kmuseth:
No worries, I should have documented this PR better :)

PR #1243, opened 2021-11-04
Auto detect OpenVDB API for houdini-19+

plattfot:
During cmake configuration if `Houdini_VERSION_MAJOR_MINOR` is greater than or equal to 19.0 it will extract the ABI version number from `openvdb/version.h` shipped with Houdini.

If it fails it will use the similar fallback as before, except it will use `OpenVDB_MAJOR_VERSION` instead of hardcoded `8`.

Signed-off-by: Fredrik Salomonsson <fredriks@d2.com>

### DIFF
--- a/cmake/OpenVDBHoudiniSetup.cmake
+++ b/cmake/OpenVDBHoudiniSetup.cmake
@@ -335,12 +335,31 @@ if(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 18.0)
   set(OPENVDB_HOUDINI_ABI 6)
 elseif(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 18.5)
   set(OPENVDB_HOUDINI_ABI 7)
-elseif(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 19.0)
-  set(OPENVDB_HOUDINI_ABI 8)
 else()
-  message(WARNING "Unknown version of Houdini, assuming OpenVDB ABI=8, "
-    "but if this not correct, the CMake flag -DOPENVDB_HOUDINI_ABI=<N> can override this value.")
-  set(OPENVDB_HOUDINI_ABI 8)
+  find_file(_houdini_openvdb_version_file "openvdb/version.h"
+    PATHS ${HOUDINI_INCLUDE_DIR}
+    NO_DEFAULT_PATH
+    NO_PACKAGE_ROOT_PATH
+    NO_CMAKE_PATH
+    NO_CMAKE_ENVIRONMENT_PATH
+    NO_SYSTEM_ENVIRONMENT_PATH
+    NO_CMAKE_SYSTEM_PATH)
+  if(NOT ${_houdini_openvdb_version_file} EQUAL "_houdini_openvdb_version_file-NOTFOUND")
+    file(STRINGS ${_houdini_openvdb_version_file} _houdini_openvdb_version)
+    foreach(line ${_houdini_openvdb_version})
+      if(line MATCHES "^#define OPENVDB_ABI_VERSION_NUMBER ([0-9]+)$")
+        set(OPENVDB_HOUDINI_ABI ${CMAKE_MATCH_1})
+        break()
+      endif()
+    endforeach()
+    unset(_houdini_openvdb_version)
+  endif()
+  unset(_houdini_openvdb_version_file)
+  if(NOT OPENVDB_HOUDINI_ABI)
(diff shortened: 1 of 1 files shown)


cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -335,12 +335,31 @@ if(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 18.0)
...
-    "but if this not correct, the CMake flag -DOPENVDB_HOUDINI_ABI=<N> can override this value.")
-  set(OPENVDB_HOUDINI_ABI 8)
+  find_file(_houdini_openvdb_version_file "openvdb/version.h"
+    PATHS ${HOUDINI_INCLUDE_DIR}
+    NO_DEFAULT_PATH
+    NO_PACKAGE_ROOT_PATH
+    NO_CMAKE_PATH
+    NO_CMAKE_ENVIRONMENT_PATH
+    NO_SYSTEM_ENVIRONMENT_PATH
+    NO_CMAKE_SYSTEM_PATH)
```
Idclip, line 346: I think the only option you need to provide here is `NO_DEFAULT_PATH`

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -335,12 +335,31 @@ if(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 18.0)
...
+  if(NOT ${_houdini_openvdb_version_file} EQUAL "_houdini_openvdb_version_file-NOTFOUND")
```
Idclip, line 347: Think this can just be `if(NOT _houdini_openvdb_version_file)`

cmake/OpenVDBHoudiniSetup.cmake
```diff
@@ -335,12 +335,31 @@ if(Houdini_VERSION_MAJOR_MINOR VERSION_EQUAL 18.0)
...
+    file(STRINGS ${_houdini_openvdb_version_file} _houdini_openvdb_version)
+    foreach(line ${_houdini_openvdb_version})
+      if(line MATCHES "^#define OPENVDB_ABI_VERSION_NUMBER ([0-9]+)$")
+        set(OPENVDB_HOUDINI_ABI ${CMAKE_MATCH_1})
+        break()
+      endif()
+    endforeach()
```
Idclip, line 354: There exists a `OPENVDB_VERSION_FROM_HEADER` function in OpenVDBUtils.cmake - You should just be able to include that file and call that to get the ABI. Instead of looping through each line it performs a single regex match call

Idclip (approved):
Minor comments which can be addressed later. Looks good to me!

cmake/OpenVDBHoudiniSetup.cmake
plattfot: Yes, you are correct. I completely missed that in the docs
> If NO_DEFAULT_PATH is specified, then no additional paths are added to the search.

---https://cmake.org/cmake/help/latest/command/find_file.html

I'll submit a fix for this shortly.

cmake/OpenVDBHoudiniSetup.cmake
plattfot: That's correct

> or ends in the suffix -NOTFOUND

---https://cmake.org/cmake/help/latest/command/if.html

I'll fix this as well.

cmake/OpenVDBHoudiniSetup.cmake
plattfot: Fixed in c81aa6e.

cmake/OpenVDBHoudiniSetup.cmake
plattfot: Fixed in 81a9434. Although I still needed to dereference the variable. Testing this with `cmake-3.18`.

Idclip: oops, my apologies, i think this should be `if(_houdini_openvdb_version_file)` without the `NOT` (mixed up the original statement)

cmake/OpenVDBHoudiniSetup.cmake
plattfot: Ah, thanks. I should've looked around the `cmake` files a bit more. The `ABI` field is not in the docs for the function though, I had to look in the source code to double check that it existed.

Switched to using that in  c7ce97f.

cmake/OpenVDBHoudiniSetup.cmake
plattfot: Right, that makes perfect sense. I'll fix that right away.

jmlait (approved):

PR #1246, opened 2021-11-12
Add H19 download to weekly build and rely on timestamps instead of tarball hashes

danrbailey:


Idclip (approved):
I'd consider combining these jobs into a matrix (reduces duplication and errors) but looks good either way

danrbailey:
> I'd consider combining these jobs into a matrix (reduces duplication and errors) but looks good either way

Yes that would be worth doing. Just to mention that I couldn't get H19 to pass when using VFX 2021 images due to a link issue with lzma and boost 1.73 so opted to put this in with VFX 2020 for the time being (hence why the jobs are split). Think it's something to do with needing a 5.2.2 version of LZMA instead of 5.2.3 which is what is shipped with Houdini. Perhaps this is an underlying issue in the docker images or some other configuration we need to change, but I didn't spend enough time to figure that out. This at least adds H19 support, so worth adding as-is, but we may need to resolve this to switch on the Houdini 19 commitly builds.

Issue #1127, opened 2021-07-16
[BUG] Understanding combineExtended on grids


frapit:
### Environment
**Operating System:** ubuntu18.04
**Version / Commit SHA:** 8.1
**Other:** g++ --std=c++17

### Describe the bug
I don't know if I am getting the concept but in the [combineExtended](https://www.openvdb.org/documentation/doxygen/classopenvdb_1_1v8__0_1_1tree_1_1Tree.html#a71140d5e67f0fce2c7fe0ae119db9711) method of a grid, I am not able do deactivate the voxel in the combine functor using the [CombineArgs](https://www.openvdb.org/documentation/doxygen/classopenvdb_1_1v8__0_1_1CombineArgs.html). `args.setResultIsActive(false)` should do this as per documentation.

When I use the [combine2Extended](https://www.openvdb.org/documentation/doxygen/classopenvdb_1_1v8__0_1_1tree_1_1Tree.html#a878302f18a7f23adac7da7a9fc5f7441) instead, The behavior is as I would expect. 

Consider this minimal example:
```
#include <openvdb/openvdb.h>
#include <iostream>
int main()
{
    openvdb::initialize();
    openvdb::FloatGrid::Ptr grid = openvdb::FloatGrid::create();
    openvdb::FloatGrid::Ptr grid_merge = openvdb::FloatGrid::create();
    openvdb::FloatGrid::Accessor accessor = grid->getAccessor();

    accessor.setValue({0, 0, 5}, 3.0f);
    accessor.setValue({0, 0, 4}, 4.0f);

    auto set_inactive = [](openvdb::CombineArgs<float>& args) {
        args.setResultIsActive(false);
        // args.setResult(1.0); // That works
    };
  
    std::cout << "Active voxel count grid: " << grid->activeVoxelCount() << "\n"; // 2
    std::cout << "Active voxel count merge: " << grid_merge->activeVoxelCount() << "\n"; // 0

    grid_merge->tree().combineExtended(grid->tree(), set_inactive);
    // grid_merge->tree().combine2Extended(grid_merge->tree(), grid->tree(), set_inactive);

    std::cout << "Active voxel count grid: " << grid->activeVoxelCount() << "\n"; // 0
    std::cout << "Active voxel count merge after: " << grid_merge->activeVoxelCount() << "\n"; // 2, but should be zero.
    return 0;
}

```

[...]

PR #1247, opened 2021-11-13
Fixed a bug in Tree::combineExtended

Idclip:

Fixes https://github.com/AcademySoftwareFoundation/openvdb/issues/1127


danrbailey (approved):
Looks fine. Just to note that `Tree::combineExtended` is one method we are looking to deprecate when we have a suitable alternative.

PR #1255, opened 2021-11-18
Fix SOP OpenVDB Extrapolate: expand SDF narrow band with dilate value…

apradhana:
Cherry-pick SOP Extrapolate to do no operation when dilation value less than 1.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -53,6 +53,8 @@ Version 8.2.0 - In development
     - The default OpenVDB ABI is now 8 for Houdini versions > 18.5.
     - VDB Visualize Tree SOP is now multi-threaded and provides slicing and
       color remapping.
+    - Add a sanitizer in SOP OpenVDB Extrapolate when expanding a narrow-band
+      level-set with a dilation value of 0, which will result in no operation.
 
     Build:
     - Added support for TBB 2021.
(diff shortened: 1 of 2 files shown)


danrbailey:
Hi Andre, if you update this branch by merging in master, it should pass the CI. I recently bumped the cache key to v4 and I believe v3 has been evicted now.

apradhana:
@danrbailey : if I merge-in master, won't that incorporate all the changes in master to this branch, including NanoVDB? I think we don't want that to be included in the v8.2 release, correct?

danrbailey:
> @danrbailey : if I merge-in master, won't that incorporate all the changes in master to this branch, including NanoVDB? I think we don't want that to be included in the v8.2 release, correct?

Oh I hadn't spotted you were trying to merge this to v8.2. Just realized I missed a change from master so that it re-saves the Houdini cache as an empty dummy cache so added that here - #1256.

I've then taken all the cache related changes and applied them in a new PR that's targeted towards the v8.2 branch - #1257. We're in a bit of a transition phase, but I think this is the minimum needed to build this branch against the weekly Houdini caches.

danrbailey:
> > @danrbailey : if I merge-in master, won't that incorporate all the changes in master to this branch, including NanoVDB? I think we don't want that to be included in the v8.2 release, correct?
> 
> Oh I hadn't spotted you were trying to merge this to v8.2. Just realized I missed a change from master so that it re-saves the Houdini cache as an empty dummy cache so added that here - #1256.
> 
> I've then taken all the cache related changes and applied them in a new PR that's targeted towards the v8.2 branch - #1257. We're in a bit of a transition phase, but I think this is the minimum needed to build this branch against the weekly Houdini caches.

I merged that PR now, so you should be able to update this one by merging in the v8.2 branch and then the CI will pass.

danrbailey (approved):

PR #1270, opened 2021-11-27
Fix Unit Test Failures with Blosc >= 1.19

danrbailey:


Idclip:
There's a test in `weekly.yml` which tests blosc versions - might be worth adding 1.19/1.20/1.21 there!

danrbailey:
> There's a test in `weekly.yml` which tests blosc versions - might be worth adding 1.19.0/1.20.0/1.21.0 there!

Good idea, mind if we do that as a separate PR?

jmlait (approved):

Idclip:
> > There's a test in `weekly.yml` which tests blosc versions - might be worth adding 1.19.0/1.20.0/1.21.0 there!
> 
> Good idea, mind if we do that as a separate PR?

Sounds like we should add this now if there are concerns with 1.21?

danrbailey:
> > > There's a test in `weekly.yml` which tests blosc versions - might be worth adding 1.19.0/1.20.0/1.21.0 there!
> > 
> > 
> > Good idea, mind if we do that as a separate PR?
> 
> Sounds like we should add this now if there are concerns with 1.21?

Yeah, sounds good. I'll merge this now.

PR #1271, opened 2021-11-27
Add Mac OS 11 CI

danrbailey:


Idclip:
We already have a `macos-latest` test in `weekly.yml`

Idclip:
Ah turns out:

> The macos-latest label currently uses the macOS 10.15 runner image.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

maybe we should just switch that one to 11

danrbailey:
> https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Yes, but think it perhaps makes sense to use the explicit versions instead of latest so our CI doesn't quietly change underneath us.

Idclip:
That's what I suggested we do.

danrbailey:
@Idclip - so should we have both 10.15 and 11.0 in the weekly?

Idclip:
Nah I think we just change this to 11, the other is already tested elsewhere if I remember correctly

danrbailey:
> Nah I think we just change this to 11, the other is already tested elsewhere if I remember correctly

Ok, I've updated this branch to change `macos-latest` to `macos-11` for all the mac CI jobs and added a `macos-1015` job to the weekly build as with these other changes, only the AX build was running on 10.15.

Idclip (approved):

PR #1273, opened 2021-12-01
Added support for OpenVDB AX on Windows

Idclip:


danrbailey (approved):
Looks good. It would be great to find some time to convert the AX unit tests to GoogleTest.

Idclip:
Thanks @danrbailey, agreed about CppUnit, though I realistically don't think I'll be able to prioritise that anytime soon :(

openvdb_ax/openvdb_ax/math/OpenSimplexNoise.h
```diff
@@ -101,7 +101,7 @@ namespace OSN
  */
 
 // 3D Implementation of the OpenSimplexNoise generator.
-class OSNoise
+class OPENVDB_AX_API OSNoise
```
richhones, line 104: was gonna comment about whether this is needed, i.e. whether we should export these symbols but i guess for backwards compatibility we should regardless of whether we want other people to use these noise functions outside of our needs in AX

Idclip: Anything that's tested in the unit tests needs to be exported. I suspect we'll end up deprecating these classes anyway

PR #1274, opened 2021-12-04
cmake: unconditionally search for Python interpreter

thesamesam:
When setting PYOPENVDB_INSTALL_DIRECTORY, CMake would fail with:
```
-- Could NOT find Python (missing: Python_LIBRARIES Development Development.Module Development.Embed) (found version "3.9.9")
CMake Error at openvdb/openvdb/python/CMakeLists.txt:65 (message):
  Could NOT find Python::Module (Required is at least version "2.7")
Call Stack (most recent call first):
  openvdb/openvdb/python/CMakeLists.txt:112 (openvdb_check_python_version)
```

It seems like we always need to search for the interpreter.

cc @waebbl

Bug: https://bugs.gentoo.org/790350
Signed-off-by: Sam James <sam@gentoo.org>


Idclip:
Thanks @thesamesam - we've had issues with this before, we changed it due to some problems on Windows:

https://github.com/AcademySoftwareFoundation/openvdb/pull/1055#discussion_r650279365

I wonder if this is an issue with newer versions of CMake where the `Development` component specifies both `Development.Module` and `Development.Embed`? These were added in CMake `3.18` - I think we only need `Development.Module`. Instead of this change could you try changing:

line 100:
```
set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development)
```

to
```
set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development.Module)
```

and see if that also works?

waebbl:
Configuration works with this change for 8.2.0 and 9.0.0 packages from Gentoo. The patch won't work for the 8.1.0 sources, so I haven't tried for it. We also don't have a live package for OpenVDB in Gentoo, so I haven't tested against the git master HEAD either. If the latter two are necessary, I can make a patch for 8.1.0 and a live package as well and test against these versions.

thesamesam:
@kmuseth Thank you for the background. I should've looked into a bit more, to be honest, as it seemed quite specific. Appreciate you getting back to us so quickly with a fix.

@waebbl Thanks for beating me to it!

Idclip:
Here's the proposed fix for master:

https://github.com/AcademySoftwareFoundation/openvdb/pull/1275

If this looks good to you I can close this

thesamesam:
Works for me, thanks @Idclip!

Idclip:
Merged! Thanks again for reporting folks

PR #1275, opened 2021-12-08
Fixed an issue with find_python and CMake 3.18

Idclip:


richhones (approved):
Looks good to me.

PR #1277, opened 2021-12-16
AX binary updates

Idclip:
- Added a thread CLI argument
 - Added multi-file reading
 - Added a ci script to download vdbs
 - Added tests for the vdb_ax binary


openvdb_ax/openvdb_ax/cmd/openvdb_ax.cc
```diff
@@ -462,26 +576,37 @@ main(int argc, char *argv[])
...
+        for (const auto& filename : opts.mInputVDBFiles) {
+            openvdb::io::File file(filename);
+            try {
+                axlog("[INFO] | \"" << filename << "\"");
+                axtimer();
+                file.open();
+                auto in = file.getGrids();
+                grids.insert(grids.end(), in->begin(), in->end());
+                // choose the first files metadata
+                if (!meta) meta = file.getMetadata();
```
richhones, line 602: maybe the metadata should combine all the metadatas? take the first version of entry?

richhones: should also document this somewhere

openvdb_ax/openvdb_ax/test/cmd/vdb_ax_test_pass_2
```diff
@@ -0,0 +1,14 @@
...
+         vdb_ax analyze -s "@density += 1;" --try-compile points  // compile code for points
+
+    -s [code], -f [file]  input code as a string or from a file.
+    --ast-print           print the generated abstract syntax tree
+    --re-print            re-interpret print of the code post ast traversal
+    --reg-print           print the attribute registry (name, types, access, dependencies)
+    --try-compile <points | volumes>
+                          attempt to compile code for points, volumes or both if no
+                          option is provided, reporting any failures or success.
+
```
richhones, line 14: as there is a test to check the extended help works for analyze, execute and functions extended help should be tested too

pendingchanges/ax_bin.txt
```diff
@@ -0,0 +1,3 @@
+Improvements:
+ - Added a --thread [n] argument to the vdb_ax binary.
+ - Changed the positional arguments for the vdb_ax binary in execute mode. All positional arguments are now input grids which allows vdb_ax to read multiple files. Output files are specified with -o.
```
richhones, line 3: do we need to make more of a big deal of the fact that this is changing, e.g. may need people to update any scripts using it?

Idclip: Oh I could rework this. Maybe we use `-i`? for example: 
`vdb_ax -i in1.vdb -i in2.vdb ...` 
`vdb_ax -i in1.vdb in2.vdb ...`
etc.  

Would prefer positional but maybe this is too breaking for now? We can then continue to support `in out` args with deprecation warning and transition it to an error.

openvdb_ax/openvdb_ax/cmd/openvdb_ax.cc
Idclip: You run into a new set of problems with combining them regarding duplicate entries. We should add options for this down the line. Perhaps we don't even copy any file level metadata and just output a new file? That might make the most sense but not sure what this was doing before

richhones: Yeah sorry I meant take the first version of each entry that occurs, to handle duplicates. Maybe as you say if before it would not copy the metadata, we just don't do that for now? Not sure what the options would be apart from first or last version of each entry.

pendingchanges/ax_bin.txt
richhones: Would it be too unintuitive to do something like, require -i/-o and then if no file is input rather than execute do analyze --try-compile or something? This is probably tangential to this issue.

openvdb_ax/openvdb_ax/cmd/openvdb_ax.cc
Idclip: So the old behaviour was to copy the metadata of the input file which is what this currently does. Probably makes more sense not to do that (for example file metadata from Houdini has the "creator" info in it, it's most likely not expected to be propagated)

pendingchanges/ax_bin.txt
Idclip: Don't quite understand what you're suggesting

Idclip: I've added support for -i, currently works as `-i file.vdb -i file.vdb` etc. Added back support for `input.vdb output.vdb` but if this is used then -i/-o can't be used. e.g:

`input.vdb output.vdb` - works with a warning
`input.vdb` - works with a warning
`input.vdb -o output.vdb` - errors
`-i input.vdb output.vdb` - errors
`-i input.vdb -o output.vdb` - works, no warning

openvdb_ax/openvdb_ax/test/cmd/vdb_ax_test_pass_2
Idclip: done

openvdb_ax/openvdb_ax/cmd/openvdb_ax.cc
Idclip: I've added this option and removed the file meta copy by default

Idclip:
@richhones I think I've addressed your feedback but if you could build/test that would be great

richhones (approved):
This looks good to go now! Gave it a build and test and seems working fine.

PR #1279, opened 2021-12-28
Fix windows compilation issues

3Descape:
Current problem: When using the most recent version of oneTBB on windows with MSVC(2019), linking results in an LNK1104 "cannot open file tbb12.lib" error. 
Cause: the target e.g. openvdb_shared references tbb.lib. This on the other hand is just a reference to tbb12.lib, but the directory containing that file is not included in additional library directories. Therefore the linker can't find it.
This pull request includes the directory into the additional link directories and therefore lets the linker find the missing library files.


Idclip (commented):
Thanks a lot @3Descape, this looks fine though haven't tested - could you confirm the exact version of TBB you had to add this for?

CI is complaining about whitespace (most likely tabs or trailing spaces)

3Descape:
@Idclip Seems to be 2021.6.0 if I'm not mistaken.

3Descape:
I've now changed it to setting the INTERFACE_LINK_DIRECTORIES on the imported target itself, as it seems target_include_directories() only works on compilable target types and the IMPORTED target type is no compilable.

Idclip (approved):
Nice, thanks @3Descape.

For reference: https://gitlab.kitware.com/cmake/cmake/-/issues/19434 - seems to be fixed from cmake 3.16

jmlait (approved):

PR #1282, opened 2022-01-09
addLeaf and addTile functions don't work correctly for non-standart trees

zakirovandrey:
If the tree depth is larger than 3, than addLeaf() and addTile() functions will not add any new nodes actually. This is true only for non-standart predefined trees (FloatTree, DoubleTree,etc).

The corresponding test **TestDepth.cc** is added to check this behavior -- test fails and the last commit fixes this bug.


openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -604,7 +604,10 @@ class CacheItem
     void addLeaf(LeafNodeType* leaf)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't add a node to a const tree");
-        if (NodeType::LEVEL == 0) return;
+        if (NodeType::LEVEL == 0) {
+            mNext.addLeaf(leaf);
+            return;
+        }
```
kmuseth, line 610: please change CacheItem::addLeaf to: 

```
void addLeaf(LeafNodeType* leaf)
{
        static_assert(!TreeCacheT::IsConstTree, "can't add a node to a const tree");
        if (NodeType::LEVEL>0 && this->isHashed(leaf->origin())) {
            assert(mNode);
            return const_cast<NodeType*>(mNode)->addLeafAndCache(leaf, *mParent);
        }
        mNext.addLeaf(leaf);
}
```

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -615,7 +618,10 @@ class CacheItem
     void addTile(Index level, const Coord& xyz, const ValueType& value, bool state)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't add a tile to a const tree");
-        if (NodeType::LEVEL < level) return;
+        if (NodeType::LEVEL < level) {
+            mNext.addTile(level, xyz, value, state);
+            return;
+        }
```
kmuseth, line 624: Please change CacheItem::addTile to:

```
void addTile(Index level, const Coord& xyz, const ValueType& value, bool state)
{
        static_assert(!TreeCacheT::IsConstTree, "can't add a tile to a const tree");
        if (NodeType::LEVEL >= level && this->isHashed(xyz)) {
            assert(mNode);
            return const_cast<NodeType*>(mNode)->addTileAndCache(
                level, xyz, value, state, *mParent);
        }
        mNext.addTile(level, xyz, value, state);
}
```

openvdb/openvdb/unittest/CMakeLists.txt
```diff
@@ -85,6 +85,7 @@ else()
     TestDelayedLoadMetadata.cc
     TestDense.cc
     TestDenseSparseTools.cc
+    TestDepth.cc
```
kmuseth, line 88: this can be removed (see comment below)

openvdb/openvdb/unittest/TestDepth.cc
```diff
@@ -0,0 +1,74 @@
+// Copyright Contributors to the OpenVDB Project
```
kmuseth, line 1: it's great that you add a unit-test for this change, but instead of created an entirely new file please add these tests to the existing TestValueAccessor.cc file

kmuseth (changes requested):
sorry for the slow review :)  Overall your changes look fine (there clearly was a bug in CacheItem::addLeaf) but I have a few minor requests for changes that I have inlingned as comments. They should be easy to address after which I can approve your PR. Please also double check that my logic for the new implementations of CacheItem::addLeaf and CacheItem::addTile hold and of course passes your unit tests.

jmlait (approved):

openvdb/openvdb/tree/ValueAccessor.h
ghurstunither: @kmuseth I can verify that OP's changes fix the issues I had recently came across. However, the changes you propose are throwing errors such as

```malloc: *** error for object 0x7f95a2848400: pointer being freed was not allocated.```

Coverting a `Tree5<float, 5, 4, 3, 3>` sphere to a fog volume should be enough to reproduce this.

kmuseth (approved):
Let's just get this in, then I can make minor tweaks myself :)

Idclip:
Hey @zakirovandrey thanks a lot for these changes - they have since been incorporated (#1452) and are part of the latest v10 release. See here https://www.openvdb.org/documentation/doxygen/changes.html#v10_0_0_changes

Closing this.

PR #1289, opened 2022-01-15
Remove unused tbb/task.h include to avoid deprecation warnings

e4lam:

### DIFF
--- a/openvdb/openvdb/tools/PointAdvect.h
+++ b/openvdb/openvdb/tools/PointAdvect.h
@@ -21,7 +21,6 @@
 
 #include <tbb/blocked_range.h>             // threading
 #include <tbb/parallel_for.h>              // threading
-#include <tbb/task.h>                      // for cancel
 
 #include <vector>
 


Idclip (approved):

e4lam:
@Idclip Note that it looks like from the build log, these warnings are still showing up in the explicit instantiation cases for some reason (eg. when compiling `D:\a\openvdb\openvdb\build\openvdb\openvdb\instantiations\VolumeToSpheres.cc` in the openvdb_static Windows build in the CI) . But I don't get these because it seems like we don't currently build with explicit instantiations. Or perhaps there's some other build difference. The motivation for this commit was to fix the warning we got when building the VDB SOPs themselves in the Houdini code.

Idclip:
@e4lam thanks for pointing that out - looks like they are still showing up in the regular CI builds too:

Windows Release (with EI): https://github.com/AcademySoftwareFoundation/openvdb/runs/4824722768?check_suite_focus=true
Windows Debug (without EI): https://github.com/AcademySoftwareFoundation/openvdb/runs/4824722777?check_suite_focus=true

I suspect it's being included by tbb itself...

e4lam:
@Idclip Ah I didn't look at the regular CI builds. We're also building with TBB 2020 U3 but I just noticed that we're a bit behind having only merged from upstream back in [Nov 1](https://github.com/sideeffects/openvdb_dev/tree/e65d3554b5fd64502f3cf402065e0aa6c3ac1b03) so maybe other things have changed since then.

e4lam:
@Idclip Ah yes, confirmed that something in upstream added these new warnings. If I build in our build system with upstream master instead, I'm now also seeing a lot more TBB Warnings.

e4lam:
PS. Also getting a build error now as well:
```2>C:\cygwin\home\edward\dev_tools\src\openvdb\openvdb-9.0\openvdb_dev\openvdb\op
envdb\io\Archive.cc(1005,25): error C2065: 'OPENVDB_FILE_VERSION_HBOOST_UUID': u
ndeclared identifier
```

e4lam:
@Idclip It looks like we lost the changes in upstream master needed to suppress these warnings. Here's a comparison on `Threading.h` that we have compared to upstream master:
```
$ diff $HFS/custom/include/openvdb/thread/Threading.h Threading.h 
11,16d10
< #ifndef __TBB_show_deprecation_message_task_H
<     #define __TBB_show_deprecation_message_task_H
<     #define OPENVDB_THREAD_THREADING_RESTORE_DEPRECATION_MESSAGE_TASK
< #endif
< 
< 
59,64d52
< 
< #ifdef OPENVDB_THREAD_THREADING_RESTORE_DEPRECATION_MESSAGE_TASK
<     #undef __TBB_show_deprecation_message_task_H
<     #undef OPENVDB_THREAD_THREADING_RESTORE_DEPRECATION_MESSAGE_TASK
< #endif
< 
```

e4lam:
Or perhaps we never managed to merge [this commit](https://github.com/sideeffects/openvdb_dev/commit/e9e097aa6cfc840a9f994ae3e61bb4d260955dcf) upstream. :(

e4lam:
I've now issued PR #1290 for this.

PR #1295, opened 2022-01-24
Fixed a bug in new minMax()

Idclip:
A thread that doesn't visit a valid value would run an invalid join() op

### DIFF
--- a/openvdb/openvdb/tools/Count.h
+++ b/openvdb/openvdb/tools/Count.h
@@ -368,14 +368,20 @@ struct MinMaxValuesOp
 
     bool join(const MinMaxValuesOp& other)
     {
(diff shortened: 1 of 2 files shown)


danrbailey:
Changes looks good, but would be good to get a unit test for this case?

Tagging @ghurstunither.

ghurstunither:
Thanks for fixing this. We can certainly add a unit test for this.

Idclip:
> Changes looks good, but would be good to get a unit test for this case?

As with testing any parallel code there's no guarantee it'll be hit - but I can just update the current test to add an empty leaf to the tree.

danrbailey (approved):

PR #1297, opened 2022-01-27
Remove unnecessary boost system dependency

nachovizzo:
This library dependency was not being used anywhere, and therefore
adding an additional dependency to the whole build system for no reason.

Signed-off-by: Ignacio Vizzo <ignaciovizzo@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ The following provides basic installation examples for the core OpenVDB library.
 
 ```bash
 apt-get install -y libboost-iostreams-dev
-apt-get install -y libboost-system-dev
 apt-get install -y libtbb-dev
 apt-get install -y libblosc-dev
 ```
@@ -84,7 +83,6 @@ vcpkg install zlib:x64-windows
 vcpkg install blosc:x64-windows
 vcpkg install tbb:x64-windows
 vcpkg install boost-iostreams:x64-windows
-vcpkg install boost-system:x64-windows
 vcpkg install boost-any:x64-windows
 vcpkg install boost-algorithm:x64-windows
 vcpkg install boost-uuid:x64-windows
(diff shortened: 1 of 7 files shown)


Idclip (approved):
Thanks @nachovizzo, good spot! Looks good to me. Note that I think iostreams still depends on system (https://pdimov.github.io/boostdep-report/boost-1.66.0/iostreams.html) but we don't list these secondary dependencies anyway

nachovizzo:
> Thanks @nachovizzo, good spot! Looks good to me. Note that I think iostreams still depends on system (https://pdimov.github.io/boostdep-report/boost-1.66.0/iostreams.html) but we don't list these secondary dependencies anyway

Agree.

By the way... I'd love to get rid of the `boost::iostreams` dependency for real. Since is the only library I need to install in order to build my application. And depending on the target system(CentOS, Debian, etc).... this varies a lot. Getting rid of the iostreams dependency would simplify my life quite a lot...

Is there any interest in addressing this?

Idclip:
> By the way... I'd love to get rid of the boost::iostreams dependency for real. Since is the only library I need to install in order to build my application. And depending on the target system(CentOS, Debian, etc).... this varies a lot. Getting rid of the iostreams dependency would simplify my life quite a lot...

We're totally on board with this and it's something we've been slowly working towards. The main complexity involved in this is with boost::interprocess. Other components such as uuid and filesystem aren't as tricky, but until we have a replacement for interprocess there's not much point acting on the other components. We also have other work which adds other boost dependencies (see the feature/sharpening branch) and of course, boost::python. 

In summary, absolutely, and any work contributed to help remove boost would be welcomed!

jmlait (approved):

Issue #612, opened 2020-01-22
Use Relative Paths in #includes


danrbailey:
Angle-bracket #includes like <openvdb/math/Math.h> run the risk of pulling in headers from nonlocal distributions, such as those installed by package managers. Use relative includes for cases where "../" is not introduced into the paths, such as `openvdb/math/Finite.h` using `#include "Math.h"` directly.

mridullpandey:
hi @danrbailey i wan to work on it kindly elaborate

kubaroth:
@mridullpandey there's nothing really to be added here. All the [bracket includes](https://github.com/AcademySoftwareFoundation/openvdb/blob/f8d318581635a2c90c1924aa9973e1ae8a4d9d7f/openvdb/math/Maps.h#L15) need to be substituted with relative version. 
@danrbailey would this include unittest directory as well?

rksharma95:
hi, @danrbailey Is anyone up to this issue? If not, I can work on this issue if that's okay?

gaida-exe:
This issue should have been resolved with #1418.

PR #1300, opened 2022-01-30
Issue #612 Use Relative Paths in #includes

lach9859:
Signed-off-by: lach9859 <73550815+lach9859@users.noreply.github.com>

### DIFF
--- a/openvdb/openvdb/Grid.cc
+++ b/openvdb/openvdb/Grid.cc
@@ -3,7 +3,7 @@
 
 #include "Grid.h"
 
-#include <openvdb/Metadata.h>
+#include "Metadata.h"
 #include <boost/algorithm/string/case_conv.hpp>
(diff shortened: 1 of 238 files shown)


danrbailey (changes requested):
We should stick to commonly accepted practices of never using upward relativity. [Google's Style Guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) puts it more precisely as:

> All of a project's header files should be listed as descendants of the project's source directory without use of UNIX directory aliases . (the current directory) or .. (the parent directory).

I'd be happy with the slightly looser definition of just not allowing "../" in the #include paths.

Edit: this is in contradiction to the original issue description, so I've re-written that description it to be more correct.

lach9859:
Oh, okay. My updates all used "../" in the include paths, so I'll just close this.

PR #1304, opened 2022-02-16
The vdbPrimsToRemove in MergeOp is shared across threads,

jmlait:
The vdbPrimsToRemove in MergeOp is shared across threads,
so calling push_back resulted in race conditions and the prims
occasionally not being deleted or the std::dequeue becoming
corrupt and crashing.

Made the simplest change of just switching to a UT_ConcurrentVector
that wraps the tbb structure.

A more correct change would be to return vdbPrimsToRemove like we
do the outputgrids so we could explicitly gather them without locking,
this would also avoid this unseen action at distance.


danrbailey (approved):
Good catch, thank you. Agree with a bit of refactoring, we could avoid the cost of the concurrent vector.

PR #1317, opened 2022-03-01
openvdb_tool: a new command-line tool that exposes most of the high-level functionalities in openvdb/tools

kmuseth:

A new command-line tool, dubbed vdb_tool, that can combine any sequence of the of high-level tools available in openvdb.
For instance, it can convert a sequence of polygon meshes and particles to level sets, and perform a large number of operations on these level set surfaces. It can also generate adaptive polygon meshes from level sets, render images and write particles, meshes or VDBs to disk or stream VDBs to stdout so other tools can render them (using pipelining).


richhones:
One general comment on the structure, looking a bit into the future if we want to use this more widely. Is there any way to pull out the actual functional stuff from the Tool.h, and have a registration system instead? This would avoid people having to modify the Tool header to add new functionality etc.

kmuseth:
> One general comment on the structure, looking a bit into the future if we want to use this more widely. Is there any way to pull out the actual functional stuff from the Tool.h, and have a registration system instead? This would avoid people having to modify the Tool header to add new functionality etc.

I'm not sure I understand your proposal/comment/concern. Adding new actions is as simple as I can imagine - add a single call-back and documentation strings. Why is it important that it's not in Tools.h? Are you suggesting run-time loading of modules? Clearly I'm missing something so please elaborate

Idclip:
How did you merge in master? It doesn't look like a merge commit to me and the diff in github is showing 200+ file changes

Idclip:
I've just re-merged master to fix the diff

openvdb_cmd/vdb_tool/include/Parser.h
```diff
@@ -0,0 +1,1037 @@
...
+        add("ln","natural log of value, e.g. {1:ln} -> {0.0}",
+            [&](){this->set(std::log(str2float(mCallStack.top())));});
+        add("log","10 base log of value, e.g. {1:log} -> {0.0}",
+            [&](){this->set(std::log10(str2float(mCallStack.top())));});
+        add("sqrt","squareroot of value, e.g. {2:sqrt} -> {1.414214}",
+            [&](){this->set(std::sqrt(str2float(mCallStack.top())));});
+        add("to_int","convert value to integer, e.g. {1.2:to_int} -> {1}",
+            [&](){this->set(int(str2float(mCallStack.top())));});
+        add("to_float","convert value to floating point, e.g. {1:to_float} -> {1.0}",
+            [&](){this->set(str2float(mCallStack.top()));});
```
Idclip, line 360: Why do you need all these functions? As far as I understand it you can only use these with loop variables? I can't see any examples using math operations though maybe I've missed them

openvdb_cmd/vdb_tool/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4244>")
+# It's not possible to use STL types in DLL interfaces in a portable and
+# reliable way so disable this warning
+add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4251>")
+# Conversion from size_t to uLong
+add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4267>")
+# Non dll-interface class used as base for dll-interface class
+add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4275>")
+# Truncation from 'int' to 'bool'
+add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/wd4305>")
```
Idclip, line 134: All these options should either be removed, moved to the core build system or only applied to the vdb_tool target. As they are they will influence all build components

openvdb_cmd/vdb_tool/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+
+list(APPEND CMAKE_MODULE_PATH ${OPENVDB_CMAKE_PATH})
+
+# Pseudo library to manage dependencies
+add_library(vdb_tool_common INTERFACE)
+
+# Optional components
+option(BUILD_TEST "Build unit tests" OFF)
+
+option(USE_NANO "Compile with NanoVDB support" ON)
```
Idclip, line 31: I can't see any attempt to get this to work with AX. Have you considered how this might clash with the math operations you've added?

Idclip:
Are the `.txt` files instead to be read by the tool or are they just examples? is plain text the best format for this?

openvdb_cmd/vdb_tool/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+  cmake_policy(SET CMP0091 NEW)
+endif()
+
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)
+endif()
+
+project(vdb_tool LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 14)
```
Idclip, line 15: Remove

kmuseth:
> I've just re-merged master to fix the diff

great thanks! However, I'm not sure how to merge your changes into my clone. The second command fails for me:

git fetch upstream
git merge upstream/feature/vdb_tool

openvdb_cmd/vdb_tool/include/Parser.h
kmuseth: you can define your own variables using "@", e.g. "{1:@x}" is semantically equivalent to "x=1". See "# String evaluation:" in README.md for an (extreme) examples where it's used to solve a quadratic equation. So in short you can define your own local variables that can then be used to set any of the options for the various actions. Very useful indeed

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: please add any changes to the CMakeLists/txt - it was originally written by someone else (Greg Klar) and I have to admit I still don't know how the cmake system works :)

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: no you I understand your concern but let's discuss. You're definitely right that I haven't attempted to add AX support yet but I'll need your help do define exactly what that means

openvdb_cmd/vdb_tool/CMakeLists.txt
Idclip: Would you be willing to attempt these changes so you can maintain the builds for components you submit?

Idclip:
> > I've just re-merged master to fix the diff
> 
> great thanks! However, I'm not sure how to merge your changes into my clone. The second command fails for me:
> 
> git fetch upstream git merge upstream/feature/vdb_tool

I merged the current master into your branch. Your fork should already by up to date

openvdb_cmd/vdb_tool/include/Parser.h
kmuseth: let's sat you want to use some procedure/function to define the value of an option, like the voxel-size of a grid, or radius or center of level set sphere, or even the frame number of a file you want to load. There are tons of example - think of all the expressions you can write in Houdini to define parameters in the GUI. This is the same but applied to option of the command-line actions :)

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: I can initially comment them out

kmuseth: did you also mean line 125, 128 and 130 or just 132 and 134?

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: I'll comment these out (and remove them entirely later)

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -0,0 +1,2767 @@
...
+     {"shader", "diffuse", "diffuse|normal|position|matte", "shader type; either \"diffuse\", \"matte\", \"normal\" or \"position\""},
+     {"light", "(0.3,0.3,0.0),(0.7,0.7,0.7)", "(0.3,0.3,0.0),(0.7,0.7,0.7)", "light source direction and optional color"},
+     {"frame", "1.0", "1.0", "orthographic camera frame width in world units"},
+     {"cutoff", "0.005", "0.005", "density and transmittance cutoff value (ignored for level sets)"},
+     {"gain", "0.2", "0.2", "amount of scatter along the shadow ray (ignored for level sets)"},
+     {"absorb", "(0.1,0.1,0.1)", "(0.1,0.1,0.1)", "absorption coefficients for RGB (ignored for level sets)"},
+     {"scatter", "(1.5,1.5,1.5)", "(1.5,1.5,1.5)", "scattering coefficients for RGB (ignored for level sets)"},
+     {"step", "1.0,3.0", "1.0,3.0", "step size in voxels for integration along the primary ray (ignored for level sets)"},
+     {"colorgrid", "-1", "1", "age of a vec3s VDB grid to be used to set material colors. Defaults to -1, i.e. disabled"}},
+     [&](){mParser.setDefaults();}, [&](){this->render();}, 0);
```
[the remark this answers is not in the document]
kmuseth: indeed, all the option for vdb_render should be back-ported into vdb_tool (let me know if I missed any)

openvdb_cmd/vdb_tool/include/Parser.h
Idclip: Why not just use the systems python or something... This is also exactly what AX can do and doesn't require new syntax (which looks pretty difficult to understand)

kmuseth:
I don't understand why the DCO test fails! I know for a fact that I used "--signoff" with my commits.  More to the point, I don't know how to resolve this issue - the guidelines for correcting this (git replace HEAD~27) fail!

kmuseth:
was replace by #1472 due to signoff issues in this PR

PR #1319, opened 2022-03-02
Relocate vdb binaries to openvdb_cmd root directory

danrbailey:
As discussed, this moves all the binaries out into a new root directory - I've called this `openvdb_cmd` to match the previous name, but I don't mind changing to `openvdb_binaries`, `openvdb_bin` or something else entirely.

I also dropped the 'open' prefix in the file names and renamed the files main.cc (ie `openvdb_print.cc` => `vdb_print/main.cc`). Now all the viewer files live in the same directory as the main.cc.


Idclip:
lgtm! A few related test failures need fixing

Idclip (approved):
Approved - is there any reason you didn't move the `vdb_ax` command line?

danrbailey:
> Approved - is there any reason you didn't move the `vdb_ax` command line?

No good reason, it was just a little more complicated to move. I've moved it now as part of this PR, but if you can just verify that the (small) change in behaviour is acceptable? Now, the ax binary is only built if `OPENVDB_BUILD_BINARIES=ON` and `OPENVDB_BUILD_AX_BINARIES=ON`, whereas previously it only required the latter.

Idclip:
Looks good to me. If you feel like it, you could deprecate `OPENVDB_BUILD_AX_BINARIES` and alias it to a new  `OPENVDB_BUILD_VDB_AX` which matches the format of the others.

Looks like the mac tests are failing as the AX builds now build all the binaries (they're all always on at the bottom of `build.sh` and "bin" just turns on the parent option). No easy solution here - probably the most correct one is to add each binary as its own component in `build.sh`. Maybe we should deprecate the `OPENVDB_BUILD_BINARIES` option as well? It dosn't really do much now that `vdb_print` is the only one enabled.

Idclip:
Yeah that seems okay to change to me too

danrbailey:
Updated this to pull in latest changes to master, this should be good to go in as-is, but here are the issues I ran into when attempting to remove `OPENVDB_BUILD_BINARIES`:

* Early exit when nothing to build - either needs to check each binary in turn or we remove this check altogether.  (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/CMakeLists.txt#L429).
* If neither binaries or unit tests are built, can still fail on looking for a custom allocator that is not required (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/CMakeLists.txt#L501).
* May need to check all the available binaries to know if we need to find openvdb instead of building core lib ourselves (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/CMakeLists.txt#L552).

Perhaps one way around these issues is to make it so that the `vdb_print` binary cannot be disabled (or otherwise add a trivial vdb hello world binary for this purpose?). It's small and self-contained but means the answer to the above three questions is simple. We are always building something, a custom allocator is always required and we always need a core lib to be built or available?

Idclip:
> Early exit when nothing to build - either needs to check each binary in turn or we remove this check altogether. 

We need to early exit if you just want to build the docs 

> If neither binaries or unit tests are built, can still fail on looking for a custom allocator that is not required 
> May need to check all the available binaries to know if we need to find openvdb instead of building core lib ourselves

If we checked for every binary this would be fine, correct? If we don't want to do this manually we could do something like:

```
get_cmake_property(ALL_VARS VARIABLES)
list(FILTER ALL_VARS INCLUDE REGEX "OPENVDB_BUILD_VDB_.*")
if(ALL_VARS)
  foreach(VAR ${ALL_VARS})
    message(STATUS "${VAR} ${${VAR}}")
    if(${${VAR}})
      set(SOME_VDB_BINARY_COMMAND_HAS_BEEN_ENABLED ON)
    endif()
  endforeach()
endif()
```

Otherwise (and I think this is more correct albiet more duplication) we should move the malloc/find(OPENVDB) logic into all sub components that needs it. That means the python, houdini and ax core libs would all do something like:

```
# if openvdb has not been built and the imported OpenVDB has not already been found, find it
if(NOT TARGET openvdb AND NOT TARGET OpenVDB)
  find_package(OpenVDB REQUIRED)
endif()
```

Not ideal, I even sort of prefer the first one though that has its own problems (`-DOPENVDB_BUILD_VDB_NON_EXISTANT_BINARY=ON` will trigger it)

ci/build.sh
```diff
@@ -35,6 +35,8 @@ COMPONENTS['core']='OPENVDB_BUILD_CORE'
 COMPONENTS['python']='OPENVDB_BUILD_PYTHON_MODULE'
 COMPONENTS['test']='OPENVDB_BUILD_UNITTESTS'
 COMPONENTS['bin']='OPENVDB_BUILD_BINARIES'
+COMPONENTS['view']='DOPENVDB_BUILD_VDB_VIEW'
+COMPONENTS['render']='DOPENVDB_BUILD_VDB_RENDER'
```
Idclip, line 39: Minor mistake here - remove the preceding `D`, otherwise it's being passed as `-DDOPENVDB_BUILD_VDB_VIEW`

danrbailey: Good catch, thanks

danrbailey:
Merged as-is. We can come back to trying to remove `OPENVDB_BUILD_BINARIES` in a separate effort.

PR #1330, opened 2022-03-17
Updated UBSAN builds to fail on errors

Idclip:
Updated the UBSAN builds to actually be useful - i.e. they now fail if they find an issue. I removed NanoVDB as part of this as NanoVDB had a lot of warnings.

As part of this I fixed a few issues and suppressed the rest. Most of the suppressions are required in tbb, the ones in VDB as predominately due to left shifts of negative numbers.


Idclip:
I should note that this is only testing on linux - I had slightly different issues on macos (in upstream deps like tbb) and a whole world of pain on mac M1 which I gave up with. But it's a start for now

danrbailey (approved):

PR #1336, opened 2022-03-23
Exposed TypeLists of the default set of OpenVDB types

Idclip:

Due to interest in the last TSC, I've split this out into a more complete and isolated PR. Thoughts welcome

### DIFF
--- a/openvdb/openvdb/TypeList.h
+++ b/openvdb/openvdb/TypeList.h
@@ -112,36 +112,85 @@ struct TSHasTypeImpl<TypeList<T, Ts...>, T, Idx>
 };
 
 
-/// @brief    Remove any duplicate types from a @c TypeList.
-/// @details  This implementation effectively rebuilds a @c TypeList by starting
-///           with an empty @c TypeList and recursively defining an expanded
-///           @c TypeList for every type (first to last), only if the type does
-///           not already exist in the new @c TypeList. This has the effect of
-///           dropping all but the first of duplicate types.
-/// @note     Each type must define a new instantiation of this object.
-/// @tparam ListT The starting @c TypeList, usually (but not limited to) an
-///               empty @c TypeList
-/// @tparam Ts    The list of types to make unique
-template <typename ListT, typename... Ts>
-struct TSMakeUniqueImpl {
-    using type = ListT;
+/// @brief  Similar to TsAppendImpl but only appends types to a list if the
+///   type does not alreay exist in the list.
+/// @details Defines a new @c TypeList with non-unique types appended
+/// @tparam U      Type to append
+/// @tparam ListT  The @c TypeList to append to
+template <typename U, typename ListT,
+  bool ListContainsType = TSHasTypeImpl<ListT, U>::Value>
+struct TSAppendUniqueImpl;
+
+/// @brief  Partial specialization where the currently evaluating type @c U in
+///   a @c TypeList already exists in the list. Returns the unmodified list.
+/// @tparam U  Type to append
+/// @tparam Ts Other types within the @c TypeList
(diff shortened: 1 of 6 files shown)


Idclip:
@jmlait @danrbailey @kmuseth @apradhana @richhones

Idclip:
As discussed in the TSC, will remove the `Native` prefix and merge after next TSC

danrbailey (approved):
This is awesome, thanks Nick!

Only minor thing is that I think it's (a little) less clear how to register custom grid types now. It's not obvious to me whether people should append to `GridTypes` or whether they should instead follow the example of StringGrid and do `StringGrid::registerGrid()` in `openvdb::initialize()`. I know there are users already adding custom grid types which will likely trigger conflicts when attempting to update after this change, so perhaps some comments to explain what to do in `openvdb::initialize()` would help?

danrbailey:
To summarize the different categories in use (I dropped "grid", "tree" and "types" from these to make the differences clearer):

TypeList:

```
Float = [Float, Double]
Integer = [Int32, Int64]
Scalar = [Float, Double, Int32, Int64]
Vec3 = [Vec3I, Vec3S, Vec3D]
 = [Mask, Bool, Float, Double, Int32, Int64, Vec3I, Vec3S, Vec3D, PointIndex, PointData]
```

Explicit Instantiation:

```
Real = [Float, Double]
Numeric = [Float, Double, Int32, Int64]
Vec3 = [Vec3I, Vec3S, Vec3D]
Volume = [Bool, Float, Double, Int32, Int64, Vec3I, Vec3S, Vec3D]
All = [Mask, Bool, Float, Double, Int32, Int64, Vec3I, Vec3S, Vec3D, PointIndex, PointData]
```

Houdini:

```
Real = [Float, Double]
Scalar = [Float, Double, Int32, Int64]
Vec3 = [Vec3I, Vec3S, Vec3D]
Point = [PointIndex, PointData]]
Topology = [Bool, Float, Double, Int32, Int64, Vec3I, Vec3S, Vec3D]
 = [Float, Double, Int32, Int64, Vec3I, Vec3S, Vec3D]
```

Idclip:
Thanks @danrbailey! Are you advocating I change the categories here? I'm not sure what would be better. In Houdini's (or other applications) cases though, I would expect them to reconstruct their own TypeLists based on their custom requirements. For example, in the houdini plugin, I would expect to see something like:

```
using HoudiniTopologyGridTypes =
   openvdb::ScalarGridTypes::
      Append<openvdb::Vec3GridTypes>::
      Append<openvdb::BoolGrid>;
```

Rather than us arbitrarily defining this.

danrbailey:
> Thanks @danrbailey! Are you advocating I change the categories here? I'm not sure what would be better. In Houdini's (or other applications) cases though, I would expect them to reconstruct their own TypeLists based on their custom requirements. For example, in the houdini plugin, I would expect to see something like:
> 
> ```
> using HoudiniTopologyGridTypes =
>    openvdb::ScalarGridTypes::
>       Append<openvdb::Vec3GridTypes>::
>       Append<openvdb::BoolGrid>;
> ```
> 
> Rather than us arbitrarily defining this.

Houdini has more baggage because it's evolved from an older point in time and is probably least important of the three be in sync, but it's still useful to see quite how different each of these categories are though.

My proposal is to change this in your categories:

Float -> Real
Scalar -> Numeric

Then add any new ones as and when we need them in the future (volume, point, etc).

Idclip:
Think I prefer Scalar to Numeric but don't feel strongly about it - I will make this change

Idclip:
I also applied `Real` and `Numeric` to the Attribute list counteparts for VDB Points

PR #1341, opened 2022-03-31
Fixed guard for NanoVDB CountOn for Windows builds

shinjin-cr:
The header for `__popcnt64` is not included unless `NANOVDB_USE_INTRINSICS` is on


kmuseth:
Thanks for this PR. Unfortunately the CLA was not signed so I cannot merge it. Considering the fact that it's a one-line change I'll just add this fix manually and close this PR

kmuseth:
Turns out this issue was actually fixed in my private branch with:

```
// __popcnt* intrinsic support was added in VS 2019 16.8
#if defined(_MSC_VER) && defined(_M_X64) && (_MSC_VER >= 1928)
```

I'll sync this to the public version

PR #1343, opened 2022-04-01
CI: Setup vcpkg caching on GHA

MehdiChinoune:
See: https://github.com/microsoft/vcpkg/blob/master/docs/users/binarycaching.md#github-packages

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,6 +117,7 @@ jobs:
     name: windows-vc:${{ matrix.config.vc }}-type:${{ matrix.config.build }}
     env:
       VCPKG_DEFAULT_TRIPLET: ${{ matrix.config.vc }}
+      VCPKG_BINARY_SOURCES: 'clear;nuget,GitHub,readwrite'
     strategy:
       matrix:
         config:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -138,6 +139,14 @@ jobs:
         # note: system path must be modified in a previous step to it's use
         echo "$Env:VCPKG_INSTALLATION_ROOT\installed\${{ matrix.config.vc }}\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
         echo "${{github.workspace}}\build\openvdb\openvdb\${{ matrix.config.build }}" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+    - name: Setup NuGet Credentials
+      run: |
+        & $(vcpkg fetch nuget | select -last 1) sources add `
+        -source "https://nuget.pkg.github.com/${{ github.repository_owner }}/index.json" `
+        -storepasswordincleartext `
```
Idclip, line 146: what is `-storepasswordincleartext` doing here?

Idclip, line 146: what is `-storepasswordincleartext` doing here?

MehdiChinoune: I don't know I have just copied/pasted the recommended setup from the link mentioned above

Idclip: Thanks, will investigate but might take some time

danrbailey: Not an expert on this, but after a quick look, I believe this flag only needs to be set on Linux/MacOS as dotnet only supports encryption on Windows.

MehdiChinoune: Removed!
Although the password stored is `"${{ secrets.GITHUB_TOKEN }}"`, nothing private is passed/stored.

PR #1346, opened 2022-04-03
Various AX improvements

Idclip:
- Fixed a regression in ax::run which wouldn't propagate exceptions
 - Improved logger exit handling in the compiler
 - Improved the behaviour of ax::parse for invalid ASTs

### DIFF
--- a/openvdb_ax/openvdb_ax/ast/Parse.cc
+++ b/openvdb_ax/openvdb_ax/ast/Parse.cc
@@ -36,11 +36,14 @@ extern void axerror (openvdb::ax::ast::Tree**, char const *s) {
 }
 
 openvdb::ax::ast::Tree::ConstPtr
-openvdb::ax::ast::parse(const char* code, openvdb::ax::Logger& logger)
+openvdb::ax::ast::parse(const char* code,
+    openvdb::ax::Logger& logger)
 {
     std::lock_guard<std::mutex> lock(sInitMutex);
     axlog = &logger; // for lexer errs
     logger.setSourceCode(code);
+
+    const size_t err = logger.errors();
 
     // reset all locations
     axlloc.first_line = axlloc.last_line = 1;
@@ -55,6 +58,8 @@ openvdb::ax::ast::parse(const char* code, openvdb::ax::Logger& logger)
     openvdb::ax::ast::Tree::ConstPtr ptr(const_cast<const openvdb::ax::ast::Tree*>(tree));
 
     ax_delete_buffer(buffer);
+
+    if (logger.errors() > err) ptr.reset();
 
     logger.setSourceTree(ptr);
     return ptr;
--- a/openvdb_ax/openvdb_ax/ast/Parse.h
+++ b/openvdb_ax/openvdb_ax/ast/Parse.h
@@ -23,14 +23,19 @@ namespace OPENVDB_VERSION_NAME {
 namespace ax {
 namespace ast {
 
-/// @brief  Construct an abstract syntax tree from a code snippet. If the code is
-///         not well formed, as defined by the AX grammar, this will simply return
-///         nullptr, with the logger collecting the errors.
+/// @brief   Construct an abstract syntax tree from a code snippet.
+/// @details This method parses the provided null terminated code snippet and
+///   attempts to construct a complete abstract syntax tree (AST) which can be
+///   passed to the AX Compiler. If the code is not well formed (as defined by
+///   the AX grammar) a nullptr is returned and instances of any errors
+///   encoutered are stored to the provided logger.
 /// @note   The returned AST is const as the logger uses this to determine line
-///         and column numbers of errors/warnings in later stages. If you need to
-///         modify the tree, take a copy.
+///   and column numbers of errors/warnings in later stages. If you need to
+///   modify the tree, take a copy.
 ///
-/// @return A shared pointer to a valid const AST, or nullptr if errored.
+/// @return A shared pointer to a valid const AST. Can be a nullptr on error.
+/// @todo  In the future it may be useful for ::parse to return as much of
+///   the valid AST that exists.
 ///
 /// @param code    The code to parse
 /// @param logger  The logger to collect syntax errors
@@ -38,14 +43,15 @@ namespace ast {
 OPENVDB_AX_API openvdb::ax::ast::Tree::ConstPtr
 parse(const char* code, ax::Logger& logger);
 
-/// @brief  Construct an abstract syntax tree from a code snippet.
-///         A runtime exception will be thrown with the first syntax error.
+/// @brief   Construct an abstract syntax tree from a code snippet.
+/// @details A runtime exception will be thrown with the first syntax error.
 ///
 /// @return A shared pointer to a valid AST.
 ///
 /// @param code The code to parse
 ///
-OPENVDB_AX_API openvdb::ax::ast::Tree::Ptr parse(const char* code);
+OPENVDB_AX_API openvdb::ax::ast::Tree::Ptr
+parse(const char* code);
 
 } // namespace ast
 } // namespace ax
--- a/openvdb_ax/openvdb_ax/codegen/VolumeComputeGenerator.cc
+++ b/openvdb_ax/openvdb_ax/codegen/VolumeComputeGenerator.cc
@@ -362,7 +362,8 @@ AttributeRegistry::Ptr VolumeComputeGenerator::generate(const ast::Tree& tree)
     // full code generation
     // errors can stop traversal, but dont always, so check the log
 
-    if (!this->traverse(&tree) || mLog.hasError()) return nullptr;
+    const size_t err = mLog.errors();
+    if (!this->traverse(&tree) || (mLog.errors() > err)) return nullptr;
 
     // insert free calls for any strings
 
--- a/openvdb_ax/openvdb_ax/compiler/Compiler.cc
+++ b/openvdb_ax/openvdb_ax/compiler/Compiler.cc
@@ -463,13 +463,15 @@ bool initializeGlobalFunctions(const codegen::FunctionRegistry& registry,
     return count == logger.errors();
 }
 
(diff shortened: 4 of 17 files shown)


openvdb_ax/openvdb_ax/compiler/Compiler.cc
```diff
@@ -682,6 +686,12 @@ Compiler::compile(const ast::Tree& tree,
         CustomData::Ptr data,
         Logger& logger)
 {
+    // @todo  Not technically necessary for volumes but does the
+    //   executer/bindings handle this?
+    if (!verifyTypedAccesses(tree, logger)) {
+        return nullptr;
+    }
```
richhones, line 693: pretty sure the bindings does not as it doesn't care about type, only name.

openvdb_ax/openvdb_ax/compiler/Logger.cc
```diff
@@ -187,19 +187,19 @@ void Logger::setSourceCode(const char* code)
 bool Logger::error(const std::string& message,
                    const Logger::CodeLocation& lineCol)
 {
-    // already exceeded the error limit
-    if (this->atErrorLimit()) return false;
+    // check if we've already exceeded the error limit
+    const bool limit = this->atErrorLimit();
+    // Always increment the error counter
+    ++mNumErrors;
+    if (limit) return false;
```
richhones, line 194: Isnt the whole point of max erroring to limit the error counter? If i now query this for some print out in my application it will say one more than my max error?

openvdb_ax/openvdb_ax/compiler/Logger.h
```diff
@@ -130,7 +133,9 @@ class OPENVDB_AX_API Logger
     bool getWarningsAsErrors() const;
 
     /// @brief Sets the maximum number of errors that are allowed before
-    ///   compilation should exit
+    ///   compilation should exit.
+    /// @note The logger will continue to increment the error counter beyond
+    ///   this value but, once reached, it will not invoke the error callback.
```
richhones, line 138: as above, I don't know if it should do this.

richhones (commented):
Most of this looks good, main question around why you chose to make it so calling error() ups the count even after max errors?  Now if I return logger->errors() I get one too many?

openvdb_ax/openvdb_ax/compiler/Logger.h
Idclip: Replied above

openvdb_ax/openvdb_ax/compiler/Compiler.cc
Idclip: I don't think I changed the behaviour of this here, not sure what you mean?

richhones: this is a comment on the comment

openvdb_ax/openvdb_ax/compiler/Logger.cc
richhones: This is kind of the point of the max errors limit though. I think it is weirder to have the ability to have errors() that doesnt match the number of entries that have been logged. errors() is there to be the way to determine how many things have been logged, e.g. if I'm printing messages to std::cerr and I want to say afterwards say ''x' errors found'. Once the Logger has hit maxErrors it shouldnt carry on with compilation, and should exit.

Idclip: Okay, then the logger can't be reliably used for error termination and should never have been - each individual method will need to track this alongside the logger.

richhones: I was unaware that any syntax error from the parser could occur and still return an AST, so when it goes into the compiler the assumption was if it hadn't errored yet, you would be working with a legit AST. I think the only other place where this was actually being the sole marker for erroring was in verifyTypedAccesses however I think a failure there will NOT stop a compile so that actually doesn't need the change above. eg. I might actually want my snippet to try and compile if the typed accesses are different, and get an error saying the type is wrong in various places.

openvdb_ax/openvdb_ax/compiler/Compiler.cc
richhones: As commented elsewhere, do we want to terminate compilation here? As this would still get through generate() even if the typed accesses are wrong no?

openvdb_ax/openvdb_ax/compiler/Logger.cc
Idclip: What if you use a logger which previously errored in compialtion in a new call to `parse`? This is totally valid, but now the logger has an error state. This paradigm dosesn't work. Checking `hasError()` is not reliable - the object could literally have any state. If the class cannot continue to track errors it cannot be used for reliable termination unless you add some other infrastructure - which seems excesive.

openvdb_ax/openvdb_ax/ast/Parse.cc
```diff
@@ -56,6 +59,8 @@ openvdb::ax::ast::parse(const char* code, openvdb::ax::Logger& logger)
 
     ax_delete_buffer(buffer);
 
+    if (logger.errors() > err) ptr.reset();
```
richhones, line 62: doesnt axparse return 1 if it errors during parsing (https://www.ibm.com/docs/en/zos/2.4.0?topic=works-yyparse-yylex) or is that not 1 in your example case? this could be used to determine erroring instead

openvdb_ax/openvdb_ax/compiler/Logger.cc
Idclip: If this isn't clear, consider the following:
```
Logger lg.
lg.setMaxErrors(1);
auto ast = parse(tree);
if (usesFunction(ast, "bad")) lg.error("you shouldn't use this function");
compile(ast, lg); // what do you want this to do - it **needs** to be able to compile and track if COMPILATION errored
```

openvdb_ax/openvdb_ax/codegen/VolumeComputeGenerator.cc
```diff
@@ -362,7 +362,8 @@ AttributeRegistry::Ptr VolumeComputeGenerator::generate(const ast::Tree& tree)
     // full code generation
     // errors can stop traversal, but dont always, so check the log
 
-    if (!this->traverse(&tree) || mLog.hasError()) return nullptr;
+    const size_t err = mLog.errors();
+    if (!this->traverse(&tree) || (mLog.errors() > err)) return nullptr;
```
richhones, line 366: These changes to compare the value of errors() obviously rely on it being incrementable beyond maxErrors. If we want to keep errors() in line with the logged number,  we could wrap the mLog.error() calls on the compute generator and set some error bool member of ComputeGenerator or something? I feel like it is probably worth it to make it so that logger.errors() is the same as the logged ones? I know this isn't ideal and should get looked at again. But such a change would allow this to catch if a new error has been found in this generate call, even if an error existed on the logger beforehand? The use of the logger here was assuming a very simple linear workflow but thats not too robust.

openvdb_ax/openvdb_ax/ast/Parse.cc
Idclip: It does in every case except for right precedence operators - I'm not entirely sure why yet, I suspect its due to error recovering being able to fall back and create the partial AST.

openvdb_ax/openvdb_ax/codegen/VolumeComputeGenerator.cc
Idclip: This also needs to be doen in Compile.cc in a bunch of methods too. I don't feel that the extra code/complexity necessary to solve this is worth it but I'll take another look.

openvdb_ax/openvdb_ax/compiler/Logger.cc
richhones: Because the logger has some non trivial state e.g. its AST node to Code Location map, it relies on users calling clear() between parse calls, in fact parse() should probably do that (I'm not sure why it doesnt). I would imagine in this scenario it should check if the maxErrors has been hit and then subsequently call or not call compile? then max errors is doing its thing of allowing early exit.

openvdb_ax/openvdb_ax/compiler/Compiler.cc
richhones: I think that this last comment still stands. Even if the typed accesses are wrong (and so should log errors) I dont think it _needs_ to early exit from compilation here. Is it useful to continue compilation and log all subsequent errors that might occur due to these wrong type accesses?

openvdb_ax/openvdb_ax/compiler/Logger.cc
richhones: Discussed offline, we should allow the counter to increase beyond maxerrors but add a comment stating how to get the number of actually reported errors in Logger.h

openvdb_ax/openvdb_ax/ast/Parse.h
```diff
@@ -23,29 +23,35 @@ namespace OPENVDB_VERSION_NAME {
...
+/// @brief   Construct an abstract syntax tree from a code snippet.
+/// @details A runtime exception will be thrown with the first syntax error.
 ///
 /// @return A shared pointer to a valid AST.
 ///
 /// @param code The code to parse
 ///
-OPENVDB_AX_API openvdb::ax::ast::Tree::Ptr parse(const char* code);
+OPENVDB_AX_API openvdb::ax::ast::Tree::Ptr
+parse(const char* code);
```
richhones, line 54: Can we update this to call clear() on the logger? Seeing as the Loggers internal tree ptr and SourceCode* are reset in this call already, to avoid the CodeLocationMap growing with repeated calls to parse. This will change behaviour slightly to reset the number of errors/warnings and so such behaviour should be commented here.

richhones:
In the spirit of not holding this up, think this is fine to go in. Any further documentation improvements can be made in a future PR.

richhones (approved):

PR #1349, opened 2022-04-06
FindTBB: also search for `tbb/version.h`

mathstuf:
The `tbb_stddef.h` header no longer exists. In environments with older
TBB installations laying around beside newer TBB installations, the
newer TBB ends up being found except for the include directory because
the `find_file` finds the old TBB's `tbb_stddef.h` header which doesn't
match the new prefix. Instead, also search for `tbb/version.h` to allow
the fallback `oneapi/` version extraction end up working.

### DIFF
--- a/cmake/FindTBB.cmake
+++ b/cmake/FindTBB.cmake
@@ -163,7 +163,7 @@ list(APPEND _TBB_INCLUDE_SEARCH_DIRS
 )
 
 # Look for a standard tbb header file.
-find_path(Tbb_INCLUDE_DIR tbb/tbb_stddef.h
+find_path(Tbb_INCLUDE_DIR NAMES tbb/version.h tbb/tbb_stddef.h
   ${_FIND_TBB_ADDITIONAL_OPTIONS}
   PATHS ${_TBB_INCLUDE_SEARCH_DIRS}
   PATH_SUFFIXES ${CMAKE_INSTALL_INCLUDEDIR} include


Idclip:
This is actually already handled by the FindTBB module:

```
# Look for a standard tbb header file.
find_path(Tbb_INCLUDE_DIR tbb/tbb_stddef.h
  ${_FIND_TBB_ADDITIONAL_OPTIONS}
  PATHS ${_TBB_INCLUDE_SEARCH_DIRS}
  PATH_SUFFIXES ${CMAKE_INSTALL_INCLUDEDIR} include
)

set(_tbb_version_file "${Tbb_INCLUDE_DIR}/tbb/tbb_stddef.h")

if(NOT EXISTS ${_tbb_version_file})
  # From TBB 2021, tbb_stddef is removed and the directory include/tbb is
  # simply an alias for include/oneapi/tbb. Try and find the version header
  # in oneapi/tbb
  find_path(Tbb_INCLUDE_DIR oneapi/tbb/version.h
    ${_FIND_TBB_ADDITIONAL_OPTIONS}
    PATHS ${_TBB_INCLUDE_SEARCH_DIRS}
    PATH_SUFFIXES ${CMAKE_INSTALL_INCLUDEDIR} include
  )
  set(_tbb_version_file "${Tbb_INCLUDE_DIR}/oneapi/tbb/version.h")
endif()
```

But, as you point out, it prioritises finding the older tbb installation format with `stddef.h` and only falls back to `version.h` if that's not found. If you have two versions of TBB installed on your PATH place then you can work around this by manually providing `-D TBB_INCLUDEDIR=` to control the search or `-D Tbb_INCLUDE_DIR=` to explicitly set the result.

We can change this to prioritise the newer installs but that will simply reverse this behaviour and break backwards compatibility.

Idclip:
I've incorporated these changes into a new PR which attempts to also warn if multiple installations are detected: https://github.com/AcademySoftwareFoundation/openvdb/pull/1376

mathstuf:
Thanks; that looks OK at a first glance at least.

PR #1358, opened 2022-04-19
Boost python CMake changes

Idclip:
This addresses an issue where:

```
  find_package(Boost COMPONENTS pythonXY)
```

Creates a broken `Boost::python` target on some implementation which subsequently stops other searches. This breaks our technique of progressively searching for boost targets should one fail i.e:

```
  find_package(Boost COMPONENTS pythonXY)
  find_package(Boost COMPONENTS pythonX)
  find_package(Boost COMPONENTS python)
```

These changes should continue to support some of this behaviour but also exposes options for users to control how boost python is found. That being said, the boost cmake situation is a mess and I'd rather spend any time in this area trying to transition to pybind as fast as possible.


Idclip:
Fixes #1353

danrbailey (approved):

PR #1371, opened 2022-04-27
CI - Blosc & ABI checks

Idclip:
- Added blosc CI for 1.19, 1.20 and 1.21
 - Added ABI CI for VDB major versions
 
I added a new virtual method and a new `int` member to `Grid`. Here's an example of the ABI run and subsquent report: https://github.com/Idclip/openvdb/actions/runs/2234898611

image


danrbailey (approved):
Looks good to me. Would it be possible to add a link to that report to the `README.md` too so it shows up alongside all the other CI badges?

Idclip:
> Looks good to me. Would it be possible to add a link to that report to the `README.md` too so it shows up alongside all the other CI badges?

I think the badges can only be created per workflow not per job - so I'd have to split this out into a separate `abi_compliance.yml`

danrbailey:
> > Looks good to me. Would it be possible to add a link to that report to the `README.md` too so it shows up alongside all the other CI badges?
> 
> I think the badges can only be created per workflow not per job - so I'd have to split this out into a separate `abi_compliance.yml`

Not a deal-breaker. Also, a quick mention that there's also this simple ABI check that I added a while back:

https://github.com/AcademySoftwareFoundation/openvdb/tree/master/openvdb_houdini/openvdb_houdini/abitest

Idclip:
@danrbailey is that a go ahead? Would you like me to move these into their own workflow? Happy to do so.

danrbailey:
> @danrbailey is that a go ahead? Would you like me to move these into their own workflow? Happy to do so.

No, that's ok, go ahead. Just if it was easy to make it show up as a badge.

PR #1373, opened 2022-04-29
Auto deploy doxygen to openvdb.org

Idclip:
Includes various doxygen fixes

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -69,7 +69,7 @@ jobs:
           UserKnownHostsFile=/dev/null
         " > /etc/ssh/ssh_config
     - name: deploy
-      # only deploy documentation to gh-pages on a manual workflow dispatch
+      # only deploy documentation on a manual workflow dispatch
       if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -49,17 +49,39 @@ jobs:
...
         github.event_name == 'workflow_dispatch' &&
         github.event.inputs.deploy == 'true'
       uses: peaceiris/actions-gh-pages@v3
       with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
         publish_dir: /usr/local/share/doc/OpenVDB/html
+        destination_dir: documentation/doxygen
+        external_repository: AcademySoftwareFoundation/openvdb-website
+        publish_branch: doxygen
```
Idclip, line 83: This could also be the master branch to avoid having to do another PR we we allowed direct pushing to master

danrbailey (approved):

Issue #161, opened 2017-07-05
Publish Python library on PyPi?


ataber:
Hello! I'm very excited about using OpenVDB, but I'm looking for a cross-platform, easy-to-install way of interacting with it. I have successfully compiled but it would be great to be able to install the python library like `pip install pyopenvdb`. I'm not sure of the challenges to do that, but I thought I would bring it up because having python packages available via `pip` decreases the startup cost by an order of magnitude, easily. Thoughts?

peytondmurray:
Hey all, sorry to dig up such an old issue, but I'm interested in being able to install openvdb's python bindings as a package from PYPI as well. I'd be happy to submit a PR with the packaging necessary to publish on PYPI, if the maintainers would accept it. I can also write the github workflows to automate building and publishing the releases. Would this be something the maintainers be okay with?

Idclip:
Hi @peytondmurray - I think this would be great to see! I'm just not sure what exactly this looks like as I haven't worked too much with PYPI. My understanding is that you would be contributing a bunch of scripts to allow us to automate and test the publishing process of a PYPI package from within the OpenVDB repo? I see a number of `pyopenvdb` packages already published by the community - do you see any issues with us publishing our own with the same name?

peytondmurray:
> My understanding is that you would be contributing a bunch of scripts to allow us to automate and test the publishing process of a PYPI package from within the OpenVDB repo?

Yeah, that's right. Actually, I think there's a way to set up Github actions to automatically publish (not just test) a new version of pyopenvdb on PYPI whenever you make a new release. If we can get that working then the process of publishing to PYPI won't require any extra effort at all when it comes time to make new releases of OpenVDB.

> I see a number of `pyopenvdb` packages already published by the community - do you see any issues with us publishing our own with the same name?

There are indeed issues with publishing our own with the same name, but I think the effort in overcoming them will be worth it. For reference, there are currently two separate projects on PYPI which [show up if you search "pyopenvdb"](https://pypi.org/search/?q=pyopenvdb), but they're both really the same:
1. [pyopenvdb 0.1.4](https://pypi.org/project/pyopenvdb/) is a project for building and packaging pyopenvdb in a docker container. [There's a github page](https://github.com/theNewFlesh/docker_pyopenvdb) which has commits from January 2021, but no PYPI release has been made in 2 years.
[...]

PR #1377, opened 2022-05-11
Add python packaging support

peytondmurray:
## Summary
This PR adds python packaging support, allowing the python OpenVDB bindings to be installed with `pip`. It also includes github actions to automatically publish releases to the [Python Package Index (PyPI)](https://pypi.org/). See #161 for original discussion.

## Changes
* Added `pyproject.toml`, which specifies build-time dependencies for the python package.
* Added `setup.py`, which makes `pyopenvdb` installable by `pip`, and specifies the python side of the build recipe.
* Added a new github action to build a source distribution of the pyopenvdb package, and upload it to PyPI. The build can be triggered either by publishing a new OpenVDB release, or manually by navigating on GitHub to `Actions->Build and Release Python Bindings to PyPI->Run Workflow` and providing a branch and ref to build.

## Additional Info
I explored a couple of options for generating binary python package distributions (wheels), but eventually opted only to generate source distributions (sdists). Here's what I tried:

1. [cibuildwheel](https://github.com/pypa/cibuildwheel), a project by the Python Packaging Authority, allows a user to easily build wheels for python packages for a range of python versions, platforms, and system architectures. It does this by using docker containers to provide build environments. These [manylinux](https://github.com/pypa/manylinux) docker containers can generate wheels for a wide group of linux distributions, but in opting for widespread compatibility they sacrifice being up to date:
 - We can't easily use `manylinux2014` because it's based on CentOS 7, which only has boost 1.53 available in the official repositories, so we'd have to build it from scratch
 - We can't easily use `manylinux_2_24` because it's based on Debian 9, and uses GCC 6.3 which can't build our dependencies.

I eventually gave up on cibuildwheel to explore other options.
2. Using `ubuntu:22.04` to build wheels for Python 3.8, 3.9, and 3.10 seems like a good idea, and in fact I think it should in principle work - except we still need to build OpenVDB from source, which takes long enough on the github actions runners that the job times out.
 
So for this reason that I think it is best just to release an sdist to PyPI, which means that when a user installs the package with `pip install pyopenvdb`, `pip` will trigger a build of the shared library and copy it to the appropriate `site-packages`. sdists are commonly uploaded as "fallback" distributions on PyPI anyway, because they should be compatible with any system that has the appropriate dependencies, and are compiled at install time rather than at the time the package is published.

## Update 2023-02-03

I've reworked this PR so that it now works with the new pybind11 bindings.

### Changes

* Added a workflow to build and publish an sdist package to either `test.pypi.org` (if a `workflow_dispatch` event is used ad the appropriate checkbox is clicked) or by default to `pypi.org`, the main python package index. Publishing requires the presence of a github secret with a PYPI token, named `PYPI_API_TOKEN`. If publishing to `test.pypi.org`, the secret that is used should be called `TEST_PYPI_API_TOKEN` I can meet to help out with this if the process is unclear. When a new release is made, this workflow will automatically build and publish the release to PYPI.
* At the moment, there's no way to dynamically get the version with the build backend we are using; see https://github.com/scikit-build/scikit-build-core/issues/116 for more info. So whoever does the release will have to update the version number in `pyproject.toml` before publishing :/ hopefully this feature gets added to the build backend soon.
* The python bindings will be published under the `openvdb` package name, as discussed before. To install, you'd just do `pip install openvdb` and to use them you'd do `import openvdb`.

## Todo
- [x] CLA
- [x] Sign-off the commits
- [x] Document requirements around using the github action to publish to PyPI (need to set up a PyPI token as a secret on the repository. OpenVDB maintainers should have access to the secrets on my fork, so you can see how to set it up.)


Idclip:
Hey @peytondmurray thanks a lot for your work on this, this is looking great! I need to spend a bit more time looking at what you have here but here's some initial comments:

 - Interesting that the `cibuildwheel` approach is locked to those older versions, though I see the reasons behind it. Out of interest what dependencies failed to build with GCC 6.3 with the `manylinux_2_24` container?
 - Regrading the CI timeouts for `ubuntu:22.04` that you were experiencing; we should be able to solve this. You can try turning `-D USE_EXPLICIT_INSTANTIATION=OFF` as this isn't really needed for the python bindings. Additionally, our actions support the use of a shared `ccache` to speed up subsequent builds. IIRC each job has a 6hr timeout limit? If we had a build matrix of a wheel per job (if such a workflow make sense with the wheel approach), combined with the above, I'd very much hope we wouldn't hit timeout limits!

I do like the idea of the `sdist` approach, though the idea of deploying binaries was pretty appealing to reduce the barrier to entry w.r.t build issues. I'm not sure if this is still enough of a reason to not go the `sdist` approach. Are their genral guidelines to choose one or the other? Do most pypi packages support both types of installations?

Finally we can absolutely credit Alex for their work! In case I haven't mentioned, the TSC have weekly catch-ups every Tuesday (10:00am to 11:00am (UTC-07:00)). They're minuted, but informal and open to anyone should you or Alex wish to attend and discuss anything here. Reminders are posted here: https://lists.aswf.io/g/openvdb-dev/topics. Either way I will feedback your progress, thanks again!

peytondmurray:
@Idclip Thanks for the info - I think based on your feedback I'll try once more to build wheels using `ubuntu:22.04`, knowing that it may lower the barrier to entry for python users.

> Out of interest what dependencies failed to build with GCC 6.3 with the manylinux_2_24 container?

[OpenVDB itself depends on GCC>=6.3.1](https://github.com/AcademySoftwareFoundation/openvdb/blob/2561c253d052aa27c5405d7fef4fcec1b2923ba0/cmake/config/OpenVDBVersions.cmake#L27).

> Are their general guidelines to choose one or the other? Do most pypi packages support both types of installations?

Most large PyPI projects publish both wheels and sdists; the wheels are faster and smaller to install, but the sdists have better compatibility because the compilation step happens at install time. `pip` has flags for letting users choose whether to install sdists (`--no-binary`) or bdists (`--only-binary`) if they have a preference.

> In case I haven't mentioned, the TSC have weekly catch-ups every Tuesday (10:00am to 11:00am (UTC-07:00)). They're minuted, but informal and open to anyone should you or Alex wish to attend and discuss anything here.

Hey, thanks for the invite! I saw that these meetings were on the calendar, but wasn't sure whether anyone would have questions about all this if I showed up. I'm traveling next week, but I think I might be able to make the meeting anyway. Even if nobody has any questions about this packaging stuff I'll come and say hello.

peytondmurray:
Okay, so I spent some time fiddling about getting wheels to build on `ubuntu:22.04`, and I'm getting close but I'm still not exactly sure how long it could take. There are still some difficulties getting cmake to find dependencies :upside_down_face: so if sdists are acceptable for now, I would like to break off support for building wheels into a separate branch to be added in a later PR. For now, I'll mark this as ready for review.

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  build-sdist:
+    name: Build sdist
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo 
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.tag }}
```
Idclip, line 21: does this just checkout master if no tag is provided?

openvdb/openvdb/python/CMakeLists.txt
```diff
@@ -114,6 +114,14 @@ if(USE_NUMPY)
   list(APPEND OPENVDB_PYTHON_REQUIRED_COMPONENTS NumPy)
 endif()
 
+# scikit-build doesn't yet support FindPython, but this workaround was
+# supplied by one of the scikit-build devs. See
+# https://github.com/pypa/cibuildwheel/issues/727 for details.
+if(SKBUILD)
```
Idclip, line 120: is skbuild always guaranteed to set this CMake variable?

openvdb/openvdb/python/setup.py
```diff
@@ -0,0 +1,44 @@
...
+    license="Mozilla Public License 2.0",
+    long_description=long_description,
+    long_description_content_type='text/markdown',
+    url="https://github.com/AcademySoftwareFoundation/openvdb",
+    cmake_args=[
+        '-DUSE_NUMPY=ON',
+        '-DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON',
+        '-DOPENVDB_BUILD_CORE=ON',
+        f'-DNUMPY_INCLUDE_DIR={np.get_include()}',
+        f'-DBoost_PYTHON_VERSION={sys.version_info[0]}.{sys.version_info[1]}',
```
Idclip, line 40: Doe this build from the root of OpenVDB or from openvdb/openvdb/python? Our CMake files aren't really setup to be built from any sub-directory so I'm surprised this works if it's the latter?

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Upload artifact ⇑
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: ./openvdb/openvdb/python/dist/
+
+  publish:
+    name: Publish Python packages on PyPI
+    needs: [build-sdist, build-wheels]
```
Idclip, line 49: is `build-wheels` redundant now?

.github/workflows/pypi_release.yml
peytondmurray: Actually, when checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event.

openvdb/openvdb/python/CMakeLists.txt
peytondmurray: It is - see [this doc](https://scikit-build.readthedocs.io/en/latest/usage.html#how-to-test-if-scikit-build-is-driving-the-compilation) for the reference.

openvdb/openvdb/python/setup.py
peytondmurray: This builds from `openvdb/openvdb/python`. In fact, when the `sdist` is downloaded by a user from PYPI, they only get the contents of `openvdb/openvdb/python`. It's convenient that this can successfully build (for now), but it also sounds like there's potentially some pitfalls here if we don't build from the root of the repo...?

.github/workflows/pypi_release.yml
peytondmurray: :100: nice catch

openvdb/openvdb/python/setup.py
Idclip: How is it finding the core library target? In normal cmake builds, the python module will either use the generated `openvdb` target or an existing `OpenVDB::openvdb` target which is imported from the system:

```
if(NOT OPENVDB_BUILD_CORE)
  set(OPENVDB_LIB OpenVDB::openvdb)  # we're not building vdb - this target is defined by the rot cmake which tries to look for vdb on the system
else()
  set(OPENVDB_LIB openvdb) # we're building the core lib so link and use that
endif()
```

But these targets are defined outside of the CMake in `openvdb/openvdb/python` -

peytondmurray:
Okay, after looking more closely at this I've found the issue:

```cmake
# openvdb/openvdb/openvdb/python/CMakeLists.txt
...
# @note  explicitly only search for Development.Module from 3.18 as searching
#   Development.Embed can cause issues on linux systems where it doesn't exist
if(${CMAKE_VERSION} VERSION_LESS 3.18)
  set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development)
else()
  set(OPENVDB_PYTHON_REQUIRED_COMPONENTS Development.Module)
endif()
```

I'm building with cmake 3.23.2 on my local machine, but it fails unless I force it to search for `Development` rather than `Development.Module`. @Idclip How would you like to proceed here? Checking the git log it looks like you might have run across some of these issues in the past, do you have any advice?

Idclip:
@peytondmurray I'm pretty confident that searching for Development.Module only is the right thing to do. I'm building with CMake 3.22.2 and have no issues. The fact that you're getting `Could NOT find Python (missing: Development.Module Interpreter NumPy)` to me implies something is wrong with your Python installation?

peytondmurray:
@Andrej730 I'm really sorry to have left this in a partially completed state for so long, so let me give a status update with some information:

1. I reached out to the owner of the openvdb package on PyPI, but have not heard back. If we are going to publish a package on PyPI for this, it feels like it could be useful to choose another name to publish it under; perhaps just `openvdb`?
2. It feels like a good time to start this work back up. I know there have been some discussions around using `pybind11` and it would be useful to have CI set up for if/when that work moves forward.

Let me see if I can rebase onto the latest version and rebuild the python package. I'll spend a bit of time updating this PR, as I think a few of the github actions are now a bit out of date.

Andrej730:
> 1. I reached out to the owner of the `pyopenvdb` package on PyPI, but have not heard back. If we are going to publish a package on PyPI for this, it feels like it could be useful to choose another name to publish it under; perhaps just `openvdb`?

If it would be possible to publish it as just `openvdb` it would be like the best outcome since that's how people would people usually go looking for the package after seing `import openvdb` somewhere in code.

peytondmurray:
Agreed, but it's really up to the maintainers. If I remember correctly the current python bindings are still being built to `pyopenvdb`, but it would be a really simple change to make this work. @Idclip thoughts?

openvdb/openvdb/python/setup.py
peytondmurray: When users install this, they won't be getting the whole repository - just the python bindings. When the user builds the sdist during installation, it will need to link to the system version of openvdb - just as `numpy` does with `LAPACK`.

peytondmurray:
Hi all, after a discussion with the maintainers at today's meeting, we have agreed to publish the python bindings under the `openvdb` name. However, this effort is currently blocked pending the merge of #1515. Once that PR is merged, this PR will be reworked to add the automation needed to build and publish the python bindings, which will no longer be dependent on `Boost.Python` :relieved: Thank you for your patience while this gets sorted out!

Andrej730:
PR #1515 moved to #1571

peytondmurray:
Thanks @kmuseth and everyone else for the pybind11 effort, I'll rework this PR to get the build automation in place!

peytondmurray:
Well, the upshot of this taking so long is that the pybind11/cmake/scikit-build ecosystem has become significantly simpler to make use of since I was last working on this. There's now a [scikit-build-core](https://github.com/scikit-build/scikit-build-core) build backend that is used in a great new pybind11 [example](https://github.com/pybind/scikit_build_example) that made this straightforward. Thanks to all who are working on making those projects so well documented and easy to use!

I tested and was able to publish an sdist on `test.pypi.org`, and was able to install it with `pip install`. However, I kept getting a segfault whenever I tried to import the bindings - I assume this is some pybind11 issue, but anyway I didn't investigate too much yet.

peytondmurray:
Currently blocked by https://github.com/AcademySoftwareFoundation/openvdb/issues/1583.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[build-system]
+requires = [
+    "scikit_build_core",
+    "pybind11",
+    "numpy"
```
JeanChristopheMorinPerso, line 5: If OpenVDB uses the Numpy C-API, you might want to take a look at https://github.com/scipy/oldest-supported-numpy to avoid incompatibilities with newer versions and older versions.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[build-system]
+requires = [
+    "scikit_build_core",
```
JeanChristopheMorinPerso, line 3: `scikit_build_core` is not yet stable, I would recommend pinning the minor version until it becomes stable.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+    "pybind11",
```
JeanChristopheMorinPerso, line 4: This could probably be pinned to avoid any future surprises.

openvdb/openvdb/python/CMakeLists.txt
JeanChristopheMorinPerso: I think this might not be needed anymore. scikit-build-core has supported FindPython since the beginning. Also as a side note, scikit-build-core != scikit-build.

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name:   Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: dist
+          path: dist
+
+      - name: 🧪 Publish to PyPI Testing
+        uses: pypa/gh-action-pypi-publish@master
```
JeanChristopheMorinPerso, line 60: I would recommend to not use `@master` and at least use `@release/v1`

openvdb/openvdb/python/README.md
```diff
@@ -0,0 +1,87 @@
...
+installation.
+
+### Using pip
+
+The python bindings for OpenVDB can be installed using `pip install .`. In this
+case, the build build toolchain executes in the following order:
+
+1. The project's `pyproject.toml` is read, invoking the build backend. The
+   `openvdb` python bindings use `scikit-build-core` as a build backend.
+   `scikit-build` acts as [a wrapper for `cmake`][scikit-build].
```
JeanChristopheMorinPerso, line 26: `scikit-build` isn't used not required from what I see in this PR.

openvdb/openvdb/python/README.md
```diff
@@ -0,0 +1,87 @@
+# `openvdb`
+
+This project contains python bindings for OpenVDB.
+
+## Dependencies
+
+| Name    | Version |
+| ---     | ---     |
+| cmake   | >=3.15  |
+| numpy   | >=1.14  |
```
JeanChristopheMorinPerso, line 10: This should be defined in in the `pyproject.toml`'s `project.dependencies` attribute.

openvdb/openvdb/python/README.md
```diff
@@ -0,0 +1,87 @@
+# `openvdb`
+
+This project contains python bindings for OpenVDB.
+
+## Dependencies
+
+| Name    | Version |
+| ---     | ---     |
+| cmake   | >=3.15  |
```
JeanChristopheMorinPerso, line 9: The CMakeLists defines 3.18 as the minimum version.

openvdb/openvdb/python/README.md
```diff
@@ -0,0 +1,87 @@
...
+| numpy   | >=1.14  |
+| openvdb | >=9.0.0 |
```
JeanChristopheMorinPerso, line 11: I find this confusing. Aren't the bindings made specifically for a specific OpenVDB version?

JeanChristopheMorinPerso (commented):
Hey, I left some comments that are worth looking at.

Additionally, it doesn't compile on my machine. I get this:
```
[jcmorin@arch01 python]$ VERBOSE=1 python -m build . -w
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (numpy, pybind11, scikit_build_core)
* Getting build dependencies for wheel...
* Installing packages in isolated environment... (pathspec, pyproject_metadata)
* Building wheel...
*** scikit-build-core 0.2.1 using CMake 3.25.2
*** Configurating CMake...
loading initial cache file /tmp/tmp8rp2irzx/build/CMakeInit.txt
Re-run cmake no build system arguments
-- The CXX compiler identification is GNU 12.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- ----------------------------------------------------
-- ------------ Configuring OpenVDBPython -------------
-- ----------------------------------------------------
-- Found Python: /tmp/build-env-u3rgd8l4/bin/python (found version "3.10.9") found components: Development.Module Interpreter NumPy 
-- Performing Test HAS_FLTO
-- Performing Test HAS_FLTO - Success
-- Found pybind11: /tmp/build-env-u3rgd8l4/lib/python3.10/site-packages/pybind11/include (found version "2.10.3")
-- Found Python::Module: /usr/include/python3.10) (found suitable version "3.10.9", minimum required is "")
-- Found Python::NumPy: /tmp/build-env-u3rgd8l4/lib/python3.10/site-packages/numpy/core/include) (found suitable version "1.24.2", minimum required is "")
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/tmp8rp2irzx/build
*** Building project with Unknown...
/usr/bin/cmake -S/home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python -B/tmp/tmp8rp2irzx/build --check-build-system CMakeFiles/Makefile.cmake 0
/usr/bin/cmake -E cmake_progress_start /tmp/tmp8rp2irzx/build/CMakeFiles /tmp/tmp8rp2irzx/build//CMakeFiles/progress.marks
/usr/bin/make  -f CMakeFiles/Makefile2 all
make[1]: Entering directory '/tmp/tmp8rp2irzx/build'
/usr/bin/make  -f CMakeFiles/_openvdb.dir/build.make CMakeFiles/_openvdb.dir/depend
make[2]: Entering directory '/tmp/tmp8rp2irzx/build'
cd /tmp/tmp8rp2irzx/build && /usr/bin/cmake -E cmake_depends "Unix Makefiles" /home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python /home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python /tmp/tmp8rp2irzx/build /tmp/tmp8rp2irzx/build /tmp/tmp8rp2irzx/build/CMakeFiles/_openvdb.dir/DependInfo.cmake --color=
make[2]: Leaving directory '/tmp/tmp8rp2irzx/build'
/usr/bin/make  -f CMakeFiles/_openvdb.dir/build.make CMakeFiles/_openvdb.dir/build
make[2]: Entering directory '/tmp/tmp8rp2irzx/build'
[ 11%] Building CXX object CMakeFiles/_openvdb.dir/pyFloatGrid.cc.o
/usr/bin/c++ -DPY_OPENVDB_USE_NUMPY -D_openvdb_EXPORTS -isystem /usr/include/python3.10 -isystem /tmp/build-env-u3rgd8l4/lib/python3.10/site-packages/pybind11/include -isystem /tmp/build-env-u3rgd8l4/lib/python3.10/site-packages/numpy/core/include -O3 -DNDEBUG -fPIC -fvisibility=hidden -flto -fno-fat-lto-objects -MD -MT CMakeFiles/_openvdb.dir/pyFloatGrid.cc.o -MF CMakeFiles/_openvdb.dir/pyFloatGrid.cc.o.d -o CMakeFiles/_openvdb.dir/pyFloatGrid.cc.o -c /home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python/pyFloatGrid.cc
In file included from /home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python/pyFloatGrid.cc:8:
/home/jcmorin/jcmenv/aswf/openvdb/openvdb/openvdb/python/pyGrid.h:15:10: fatal error: openvdb/tools/MeshToVolume.h: No such file or directory
   15 | #include <openvdb/tools/MeshToVolume.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/_openvdb.dir/build.make:76: CMakeFiles/_openvdb.dir/pyFloatGrid.cc.o] Error 1
make[2]: Leaving directory '/tmp/tmp8rp2irzx/build'
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/_openvdb.dir/all] Error 2
make[1]: Leaving directory '/tmp/tmp8rp2irzx/build'
make: *** [Makefile:136: all] Error 2

*** CMake build failed

ERROR Backend subprocess exited when trying to invoke build_wheel
```

I unfortunately don't have time to investigate what's wrong there. I will guess that the same will happen on others computer too.

And ironically, that's why projects usually publish pre-compiled libraries in PyPI. I would highly recommend providing pre-compiled libraries. My experience with other ASWF projects is that OpenVDB users will constantly create GitHub issues because they can't compile the project for a multitude of reasons, like on Windows where it's pretty common for users to not have any build toolchain installed, or also because they expect to be able to simply `pip install` any project and not have to worry about compiling.

> We can't easily use manylinux2014 because it's based on CentOS 7, which only has boost 1.53 available in the official repositories, so we'd have to build it from scratch

The ASWF wg-ci workgroup provides Conan boost packages at https://linuxfoundation.jfrog.io/ui/repos/tree/General/aswf-conan/aswf/boost. I believe these could be used.

Let me know if you have any questions or comments on the comments I made!

JeanChristopheMorinPerso:
Oh, and I forgot to mention that it would be great if the package's metadata included some [urls](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls).

openvdb/openvdb/python/pyOpenVDBModule.cc
```diff
@@ -326,13 +326,8 @@ struct VecTypeDescr
 ////////////////////////////////////////
 
 
-#ifdef DWA_OPENVDB
 #define PY_OPENVDB_MODULE_NAME  _openvdb
```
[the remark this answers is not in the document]
peytondmurray: Hmm, the last commit I made failed to build due to a name collision in `CMakeLists.txt` when I used `openvdb` as the package name here, which is why I used `_openvdb` as the module name here originally.

> Minor: If _openvdb was renamed to openvdb, the __init__.py file wouldn't be needed.

If we don't include an `__init__.py`, the import statement for the `openvdb` bindings looks like either

```python
from openvdb import _openvdb
...
_openvdb.foo(...)
```

Or 

```python
import openvdb._openvdb
...
openvdb._openvdb.foo(....)
```

Which is why I opted to include the `__init__.py` - it makes the `openvdb` site-packages directory a module. If this is really something that people want, I'm happy to make the change, though.

openvdb/openvdb/python/pyproject.toml
peytondmurray: Excellent idea - I'll make this change.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+    "numpy",
+]
+readme = "README.md"
+authors = [
+  { name = "OpenVDB Developer Team", email = "openvdb-dev@lists.aswf.io" },
+]
+requires-python = ">=3.7"
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "License :: OSI Approved :: MIT License",
```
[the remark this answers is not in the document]
peytondmurray: Thanks, I've updated this accordingly.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+dependencies = [
+    "numpy",
+]
+readme = "README.md"
+authors = [
+  { name = "OpenVDB Developer Team", email = "openvdb-dev@lists.aswf.io" },
+]
+requires-python = ">=3.7"
+classifiers = [
+  "Development Status :: 4 - Beta",
```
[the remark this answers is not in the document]
peytondmurray: No - sorry, this was left over from testing. Removed!

openvdb/openvdb/python/pyproject.toml
peytondmurray: Agreed - doing this now.

openvdb/openvdb/python/pyproject.toml
peytondmurray: Sure, that makes sense. Pinning to 2.10.3 unless others feel differently.

.github/workflows/pypi_release.yml
peytondmurray: :+1:

openvdb/openvdb/python/README.md
peytondmurray: Nice catch, changing now.

openvdb/openvdb/python/README.md
peytondmurray: Removing this now :+1:

openvdb/openvdb/python/README.md
peytondmurray: Yep, this is now outdated. Removed this section.

openvdb/openvdb/python/README.md
peytondmurray: This should be removed anyway, but pybind11 support was only merged a little while ago. You won't be able to build this without a recent version. Anyway, this section is removed now.

openvdb/openvdb/python/pyOpenVDBModule.cc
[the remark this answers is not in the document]
JeanChristopheMorinPerso: Umm, that's curious. Probably because of the mismatch between the package name and the module name. Anyway, the `__init__.py` file doesn't hurt in any way.

peytondmurray:
@JeanChristopheMorinPerso Thanks for taking a look at this. URLs have been added.

About building precompiled binaries: you can read through the comment history above, but the reason I didn't do this before was effectively because of the dependency on `Boost.Python`; `openvdb` required a relatively newer version which wasn't available on the CentOS 7 (?) docker images that everyone uses for manylinux builds. This resulted in dependency hell trying to get the right version of boost with the right version of gcc, and everything needed to be compiled which ran up against CI runner timeout limits, and so on and so on.

The upshot of dropping `Boost.Python` as a dependency and swapping to `pybind11` is that we don't have to worry about this. I'll get to work building this with cibuildwheel, which should make the job pretty simple.

I'm also seeing the issue with the include you saw above. This wasn't happening last time I worked on this, so I assume it's part of the growing pains associated with adding `pybind11` support. I don't think it's related to the build system changes introduced here.

Thanks again for the input :rocket:

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+]
+
+[project.urls]
+homepage = "https://www.openvdb.org/"
+documentation = "https://www.openvdb.org/documentation/doxygen"
```
JeanChristopheMorinPerso, line 32: I'm not an OpenVDB user, so sorry if I misunderstand something here, but wouldn't https://www.openvdb.org/documentation be a better URL than the C++ docs?

JeanChristopheMorinPerso:
Great @peytondmurray! I wasn't sure if boost was used for something else than the Python bindings (previously). It's good to know it was only used for that and that it's no more required.

Feel free to poke me when you'll have wheels built by GitHub Actions. I'll happily review the changes and the produced wheels on all platforms (I usually download the wheels and inspect the binaries manually, and then install on all platforms in different scenarios to see if everything holds up together).

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+    "oldest-supported-numpy"
+]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "openvdb"
+version = "10.0.1"
+description= "Python bindings for OpenVDB: sparse volume data structure and tools."
+dependencies = [
+    "oldest-supported-numpy",
```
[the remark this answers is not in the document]
peytondmurray: Yep, as long as the most recent numpy is compatible, which it should be except maybe for numpy 2.0 next year? In any case I'll make this `numpy` for now.

openvdb/openvdb/python/pyproject.toml
peytondmurray: Sure - this was just the link from the `README.md`, but I can point to the docs page one level up.

peytondmurray:
@JeanChristopheMorinPerso as expected the pybind11 dependency makes everything way easier to set up with `cibuildwheel`. Right now I'm blocked on the build failure you saw above, but I'll also see if I can't find a fix for that in a little bit - then we should be able to merge this.

jdumas:
Hi. Has there been any progress on this issue?

I've looked into it a bit, and the problem I see is that you're trying to build the openvdb libs + bindings directly using the `openvdb/openvdb/python/CMakeLists.txt` as the entry point in the GitHub actions script. However, when reaching [this line](https://github.com/AcademySoftwareFoundation/openvdb/blob/9a57c7e68e5de9ecfd05a98364dcfb8063f71186/openvdb/openvdb/python/CMakeLists.txt#L39) of the build script, `openvdb` as a CMake target is not yet defined. That's why one should _always_ link against [namepaced targets](https://cmake.org/cmake/help/latest/policy/CMP0028.html) in CMake. Indeed, in this case, CMake did not find any target named `openvdb` in scope, so it will simply add `-lopenvdb` to the link flags for the `_openvdb` target and fail at link time. Here we fail sooner because the transitive include directories for the `openvdb` target are not passed down to the client lib `_openvdb`. Again this is not a problem when building the bindings from the root `CMakeLists.txt`, because by the time we reach this line `openvdb` is a well-defined CMake target, and will bring its transitive list of include directories necessary for compilation.

So my recommendation would be twofolds:
1. Use an namespaced ALIAS target [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/9a57c7e68e5de9ecfd05a98364dcfb8063f71186/openvdb/openvdb/CMakeLists.txt#L630-L634), e.g. `OpenVDB::OpenVDB`, and replace all usages of `openvdb` with this new namespaced target.
2. Add the following code snippet in [this line](https://github.com/AcademySoftwareFoundation/openvdb/blob/9a57c7e68e5de9ecfd05a98364dcfb8063f71186/openvdb/openvdb/python/CMakeLists.txt#L35):
   ```cmake
   if(NOT TARGET openvdb)
     add_subdirectory(../../.. openvdb_build)
   endif()
   ```
   This will allow using `openvdb/openvdb/python/CMakeLists.txt` as the entry point for the build script.


This still doesn't work for me, as it's looking for BLOSC and doesn't find it, but I'm guessing by disabling the corresponding options you could get it to work eventually.

Hope this helps!

jdumas:
@peytondmurray The CI is now complaining because the commit is not signed off. Also I think a maintainer needs to approve the workflow for the build process to be triggered...

peytondmurray:
@jdumas I meant to post this yesterday, but thanks for looking into this. I committed your suggested changes yesterday, then tried to build the python bindings. It was taking too long on my laptop, so I pushed the changes with the intent to build when I got home, but never got around to taking care of this. I'll test it now, then sign the commit in a few minutes.

jdumas:
I was able to build it on my machine with the following changes to your branch:
```patch
diff --git i/openvdb/openvdb/python/pyproject.toml w/openvdb/openvdb/python/pyproject.toml
index 9876bd04..bd8bc50b 100644
--- i/openvdb/openvdb/python/pyproject.toml
+++ w/openvdb/openvdb/python/pyproject.toml
@@ -40,3 +40,4 @@ cmake.minimum-version = "3.18"
 [tool.scikit-build.cmake.define]
 OPENVDB_BUILD_CORE = "ON"
 USE_NUMPY = "ON"
+USE_BLOSC = "OFF"
diff --git i/openvdb_cmd/CMakeLists.txt w/openvdb_cmd/CMakeLists.txt
index c4351f49..f5cd7cda 100644
--- i/openvdb_cmd/CMakeLists.txt
+++ w/openvdb_cmd/CMakeLists.txt
@@ -42,7 +42,7 @@ if(NOT OPENVDB_BUILD_CORE)
   # both cases
   set(OPENVDB_LIB OpenVDB::openvdb)
 else()
-  set(OPENVDB_LIB openvdb)
+  set(OPENVDB_LIB OpenVDB::OpenVDB)
 endif()

 set(OPENVDB_BINARIES_DEPENDENT_LIBS
 ```

**Comment by peytondmurray** [comment]
Yep, I think I just missed a few places which needed the new alias (I didn't need to fiddle the BLOSC flag). I was able to successfully build it, so I've signed off now and pushed the new version.

**Comment by JeanChristopheMorinPerso** [comment]
@peytondmurray let me know when you are ready for another code review.

**Comment by jdumas** [comment]
Thinking about this PR again I think we might want to use the alias name `OpenVDB::openvdb` to be consistent with the result of the `FindOpenVDB.cmake` provided in this repository. Although ideally we shouldn't write any `FindFoo.cmake` by hand, and we should instead rely on exported `FooConfig.cmake` generated by CMake, such a change would be beyond the scope of this PR.

It'd be great if we could:
- Update the alias name from `OpenVDB::OpenVDB` to `OpenVDB::openvdb` to be consistent with the current `FindOpenVDB.cmake`.
- Have a maintainer approve the workflow to make sure it works properly.

I'd like to help move the needle on this, so please let me know if there are other ways I can help.

**Comment by peytondmurray** [comment]
@jdumas Thanks for following up about this. Following your suggestion, I've changed the alias to `OpenVDB::openvdb`. Now we need to start running the CI to make sure this works. @JeanChristopheMorinPerso could you please trigger a workflow run to see if cibuildwheel succeeds?

**Comment by JeanChristopheMorinPerso** [comment]
I don't have the permissions to trigger the workflows. I'm only a bystander that is helping doing code review on this PR.

**Comment by peytondmurray** [comment]
Ahh :sweat_smile: sorry! @Idclip Would it be possible to retrigger CI?

File: .github/workflows/pypi_release.yml [hunk 18]
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          python -m build --sdist
+
+      - name:   Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: ./openvdb/openvdb/python/dist/
+
+  build-wheels:
```
JeanChristopheMorinPerso, line 48: I would recommend to avoid building wheels at release time. It's better to build in PRs and when merging PRs and to release the artifacts generated by these builds at release time. The main reason being that lots of things can change between the time a PR is merged and a release is made on the GitHub Actions side. The OS images used change all the time.

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[build-system]
+requires = [
+    "scikit_build_core==0.2.1",
```
JeanChristopheMorinPerso, line 3: There is more recent releases. We could update to get new features (like editable installs)

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+    "pybind11==2.10.3",
```
JeanChristopheMorinPerso, line 4: FYI, pybind11 2.10.4 is the latest available version.

.github/workflows/pypi_release.yml
peytondmurray: Okay, following your advice I've split the workflow into two: `wheels.yml` will build python bindings when a PR is made, and the second will release those wheels when a release is made.

JeanChristopheMorinPerso: Great! I'll review this soon (in he upcoming day(s)).

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name:   Checkout the repo
+        uses: actions/checkout@v3
+
+      - name: 🛞 Build wheels
+        uses: pypa/cibuildwheel@v2.13.1
+        with:
+          package-dir: ./openvdb/openvdb/python
+          config-file: ./openvdb/openvdb/python/pyproject.toml
+        env:
+          CIBW_ARCHS_MACOS: auto universal2
```
JeanChristopheMorinPerso, line 51: I think you should set the manylinux versions to make sure you target CentOS 7.

openvdb/openvdb/python/CMakeLists.txt
```diff
@@ -105,6 +102,14 @@ if(USE_NUMPY)
...
+    set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
```
JeanChristopheMorinPerso, line 109: Is this still needed?

openvdb/openvdb/python/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+forum = "https://github.com/AcademySoftwareFoundation/openvdb/discussions"
+repository = "https://github.com/AcademySoftwareFoundation/openvdb"
+slack = "https://slack.aswf.io/"
+
+[tool.scikit-build]
+cmake.minimum-version = "3.18"
+
+[tool.scikit-build.cmake.define]
+OPENVDB_BUILD_CORE = "ON"
+USE_NUMPY = "ON"
```
JeanChristopheMorinPerso, line 41: Is the default to link dynamically or statically?

.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,42 @@
...
+          user: __token__
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+          repository_url: https://test.pypi.org/legacy/
+          packages_dir: dist
+
+      - name: 🎉 Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        if: ${{ !inputs.test_pypi }}
+        with:
+          user: __token__
```
JeanChristopheMorinPerso, line 40: Instead of this, trusted publishers should be used, which are now preferred and simplify this quite a bit. https://docs.pypi.org/trusted-publishers/

.github/workflows/wheels.yml
peytondmurray: What do you think about targeting something `manylinux_2_28`? I'd rather not target CentOS 7 because although in many projects it is still the default build platform, it's discontinued, and basically the entire reason this PR didn't get merged with wheels before was because it uses an old build toolchain that OpenVDB didn't support, and there was no reasonable way to update it.

openvdb/openvdb/python/CMakeLists.txt
peytondmurray: Really not sure, it's been a long time since this was put in but I recall it being necessary to get the bindings to build. I'll double check now and report back.

.github/workflows/wheels.yml
JeanChristopheMorinPerso: 28 is way too recent. The manylinux_2014 container I believe has a newer compiler with the CentOS 7 glibc. It's all about glibc.

jdumas:
@peytondmurray have you beem able to address @JeanChristopheMorinPerso's lastest comments? We still need a maintainer approval to run the github action workflows for this PR.

openvdb/openvdb/python/pyproject.toml
peytondmurray: Looking through the `FindPython` that is part of scikit-build-core I believe the default is static linking.

.github/workflows/wheels.yml
peytondmurray: Sure, we can use that - it's the default for cibuildwheel currently, but setting the appropriate options in `pyproject.toml` will ensure it stays that way.

.github/workflows/pypi_release.yml
peytondmurray: 100% agree - resolved.

JeanChristopheMorinPerso:
@AcademySoftwareFoundation/openvdb

openvdb/openvdb/python/CMakeLists.txt
peytondmurray: Doesn't look like it anymore. Removing.

peytondmurray:
@jdumas Yep, I think the comments above are addressed.

chrisflesher:
@Idclip Can you re-trigger CI please?

Idclip:
> @Idclip Can you re-trigger CI please?

Worflow flie seems to be invalid: https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/5639306194/workflow

jdumas:
It looks like the github action file is missing a `jobs:` section before the `publish:` job.

jdumas:
@peytondmurray sorry to ping you again, but would you be able to update your github action file to correct the syntax by adding the `jobs:` section?

peytondmurray:
Yep! Sorry, the last notification got buried, I'll do it now!

jdumas:
Argh it looks like it needs approval again from a maintainer to run :/. Hope it works this time.

peytondmurray:
Yeah, I wish I'd made a smaller/easier to merge PR before submitting this one so we didn't have to wait for the workflow checks - I think they're only guarded against first time contributors. Thanks for your patience!

jdumas:
I'm not sure that it needs to be approved again for every commit of the same PR though, that seems a bit too much.

Idclip:
> I'm not sure that it needs to be approved again for every commit of the same PR though, that seems a bit too much.

Not sure we can change this (?) but I'll try to remember to press the button when I see notifications

peytondmurray:
Hey folks, it's been a while since I updated this PR. At this point I'm open to PRs against this branch, or if someone wants to take over pushing this forward. At the moment I've been preoccupied with other work, and don't have much time to press this forward much more. What's the best way to tie this up? I'd be happy to do something here if it's small, but otherwise would it be best to close it out? I'm not sure how much need there is for this right now.

I'd also like to point out that [meson-python](https://meson-python.readthedocs.io/en/latest/) has been really useful in a couple of projects that I've worked on in building compiled python extensions; it's much easier to deal with than cmake IMO. It might be worth looking into if fiddling with cmake is a challenge here.

jdumas:
@peytondmurray could you give me write access to your fork? I don't have a lot of time right now, but I'd be happy to contribute when I can. One thing that makes it slow to iterate on this PR is that each commit requires approval from a maintainer before running GitHub actions. However, I think it only does that for first-time contributors, and I already have existing contributions to this repo, so it's possible that any of my commit triggers GitHub action right away (haven't been able to test that on @peytondmurray's branch since I don't have write access to it).

peytondmurray:
Invite sent!

PR #1381, opened 2022-05-16
Add a feature flag to the raw CsgDifferenceOp to immediately prune degenerate tiles.

jmlait:
When the same tile subtracted from itself a background tile is created.
greatly reduces the generated ghost borders when similar objects are
differenced.

Provided for discussion.   Is a template parameter correct way for this?  I want this in VDB Combine as a toggle parameter, so do we also add parameters to higher level functions?  This also can extend to Union/Intersect.

And is this even a good thing?

### DIFF
--- a/openvdb/openvdb/tools/Merge.h
+++ b/openvdb/openvdb/tools/Merge.h
@@ -228,6 +228,9 @@ struct CsgUnionOrIntersectionOp
     /// @brief Return the number of trees being merged
     size_t size() const { return mTreesToMerge.size(); }
 
+    /// Enables immediate pruning of tiles that cancel each other out.
+    void setPruneCancelledTiles(bool doprune) { mPruneCancelledTiles = doprune; }
+
     // Processes the root node. Required by the NodeManager
     bool operator()(RootT& root, size_t idx) const;
 
@@ -245,6 +248,7 @@ private:
 
     mutable std::vector<TreeToMerge<TreeT>> mTreesToMerge;
     mutable const ValueT* mBackground = nullptr;
+    bool mPruneCancelledTiles = false;
 }; // struct CsgUnionOrIntersectionOp
 
 
@@ -261,7 +265,7 @@ using CsgIntersectionOp = CsgUnionOrIntersectionOp<TreeT, /*Union=*/false>;
 /// PruneCancelledTiles will set to background any leaf tile that matches
 /// in the two trees, thus minimizing ghost banding when common borders
 /// are differenced.
-template<typename TreeT, bool PruneCancelledTiles = false>
+template<typename TreeT>
 struct CsgDifferenceOp
 {
     using ValueT = typename TreeT::ValueType;
@@ -281,6 +285,9 @@ struct CsgDifferenceOp
     /// @brief Constructor to CSG difference the tree in a TreeToMerge object
     /// from another.
     explicit CsgDifferenceOp(TreeToMerge<TreeT>& tree) : mTree(tree) { }
+
+    /// Enables immediate pruning of tiles that cancel each other out.
+    void setPruneCancelledTiles(bool doprune) { mPruneCancelledTiles = doprune; }
 
     /// @brief Return the number of trees being merged (only ever 1)
     size_t size() const { return 1; }
@@ -306,6 +313,7 @@ private:
     mutable TreeToMerge<TreeT> mTree;
     mutable const ValueT* mBackground = nullptr;
     mutable const ValueT* mOtherBackground = nullptr;
+    bool mPruneCancelledTiles = false;
 }; // struct CsgDifferenceOp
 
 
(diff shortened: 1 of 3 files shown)


openvdb/openvdb/tools/Merge.h
```diff
@@ -1158,30 +1161,40 @@ bool CsgDifferenceOp<TreeT>::operator()(LeafT& leaf, size_t) const
         return false;
     }
 
+    bool allequal = true;
     for (Index i = 0 ; i < LeafT::SIZE; i++) {
         const ValueT& aValue = leaf.getValue(i);
-        ValueT bValue = math::negative(mergeLeaf->getValue(i));
+        ValueT bValue = mergeLeaf->getValue(i);
+        if (PruneCancelledTiles)
```
danrbailey, line 1168: Probably worth doing `PruneCancelledTiles && allequal` here so the first non-match disables subsequent value checks.

danrbailey (commented):
> When the same tile subtracted from itself a background tile is created. greatly reduces the generated ghost borders when similar objects are differenced.
> 
> Provided for discussion. Is a template parameter correct way for this? I want this in VDB Combine as a toggle parameter, so do we also add parameters to higher level functions? This also can extend to Union/Intersect.
> 
> And is this even a good thing?

I don't see an ideal solution here as we can't easily modify the constructors due to the tag dispatch class. I think the simplest approach is to just add a runtime modifier such as:

`CsgDifferenceOp::setPruneCancelledTiles(true);`

then if we want to maintain the performance of the non-pruned leaf method (which we most likely do), then we either add a private member function that is templated to optimize out the value equality checking or for readability what I'd probably do in this case is to add a lambda for the `aValue < bValue` clause then live with a little code duplication for the rest of the loop. That at least avoids performing redundant template instantiation for the other methods that don't use this parameter.

openvdb/openvdb/tools/Merge.h
jmlait: I'd sort of hope the compiler would deal with
allequal &= foo;
implying
if (allequal) allequal = foo;
for boolean foo/allequal and no side effects on foo...  We also don't get to avoid value checks because we have to load aValue and bValue in any case, so it is mostly if the cost of the branch is worth eliding a comparison and logic operation...   My random estimate is branches are more expensive, so the branch free seems best to me, but with the potential the compiler will ignore this at its discretion.

jmlait:
> I don't see an ideal solution here as we can't easily modify the constructors due to the tag dispatch class. I think the simplest approach is to just add a runtime modifier such as:
> 
> `CsgDifferenceOp::setPruneCancelledTiles(true);`
> 
> then if we want to maintain the performance of the non-pruned leaf method (which we most likely do), then we either add a private member function that is templated to optimize out the value equality checking or for readability what I'd probably do in this case is to add a lambda for the `aValue < bValue` clause then live with a little code duplication for the rest of the loop. That at least avoids performing redundant template instantiation for the other methods that don't use this parameter.

This is a good idea....  I think, however, the correct solution is just to duplicate the loop.  We have a 512:1 amortization rate with the leaf iteration, and it is just a few lines duplicated.  I think setting up the template or other approach would generate more lines of code than we end up actually repeating here.  I also like that the allequal stack variable can be skipped in the other loop making things clearer.

jmlait:
I've added support for union/intersection, though it is somewhat difficult to setup these cases in practice.  Two touching cubes tend to convert slightly differently so don't trigger it unless I round off distance values artificially.

Also switched to the member variable to avoid template explosion.

openvdb/openvdb/tools/Merge.h
```diff
@@ -966,12 +977,35 @@ bool CsgUnionOrIntersectionOp<TreeT, Union>::operator()(LeafT& leaf, size_t) con
...
+                    leaf.setValueOnly(i, newValue);
+                    leaf.setActiveState(i, mergeLeaf->isValueOn(i));
+                }
+            }
+            if (allnegequal) {
+                // If two diffed tiles have the same values of opposite signs,
+                // we know they have both the same distances and gradients.
+                // Thus they will cancel out.
+                if (Union) { leaf.fill(math::negative(this->background()), false); }
+                else { leaf.fill(this->background(), false); }
```
jmlait, line 997: I don't like this->background(), but this was done at the start of this method!
const ValueT background = Union ? this->background() : -this->background();

openvdb/openvdb/tools/Merge.h
danrbailey: You're probably right about that.

openvdb/openvdb/tools/Merge.h
danrbailey: I don't understand, why can't you just do:
```
leaf.fill(background, false);
```
?

danrbailey:
This looks fine to me, though there is perhaps a little ambiguity on what whether `prune=false` will override `prunecancelledtiles=true`. There's a missing explicit template instantiation parameter that is causing the CI to fail. Also, minor quibble, I think we tend to prefer lower camel case for parameters (ie `pruneCancelledTiles` instead of `prunecancelledtiles`).

openvdb/openvdb/tools/Merge.h
jmlait: That background is opposite sign to what we want.   So it could be
leaf.fill(math::negative(background), false);
but that would be rather confusing... I like the background always meaning positive background.

danrbailey: Ah yes, makes sense.

danrbailey (approved):
Looks good to me! It appears that the CI is now failing for unrelated Windows reasons.

PR #1382, opened 2022-05-18
Add a placeholder in version.h for the URL and revision to be stored

jmlait:
When distributing an OpenVDB package that is forked from the official it can be useful to provide the URL and commit hash that was used so people can re-create it or derive from it if they wish.

What we do for USD is have a define that we replace when we download the library from github with our particular URL and revision.  This seems like a useful thing to have in the master branch with the placeholder values, so others can follow a similar protocol if they wish.  If we eventually distribute binaries we may also ensure our binary build process likewise patches this.

### DIFF
--- a/openvdb/openvdb/version.h.in
+++ b/openvdb/openvdb/version.h.in
@@ -92,6 +92,11 @@
 /// @hideinitializer
 #define OPENVDB_LIBRARY_VERSION_NUMBER ${OPENVDB_PACKED_VERSION}
 
+/// @brief Where this version was compiled from if it comes from a
+/// git repo.
+#define OPENVDB_PACKAGE_URL "@OPENVDB_PACKAGE_URL@"
+#define OPENVDB_PACKAGE_REVISION "@OPENVDB_PACKAGE_REVISION@"
+
 /// @brief The version namespace name for this library version
 /// @hideinitializer
 ///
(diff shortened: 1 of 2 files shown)


Idclip:
We should update the CMake to try and automatically populate these. Something like the below (untested, typed into comment):

```
set(OPENVDB_PACKAGE_URL "unknown")
set(OPENVDB_PACKAGE_REVISION  "unknown")
find_package(Git)

if(TARGET Git::Git)
  execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
  ERROR_QUIET 
  RESULT_VARIABLE code
  OUTPUT_VARIABLE result
  )
  if(NOT ${code} AND ${result})
     set(OPENVDB_PACKAGE_REVISION ${result})
  endif()

  execute_process(COMMAND ${GIT_EXECUTABLE} remote get-url origin
  ERROR_QUIET 
  RESULT_VARIABLE code
  OUTPUT_VARIABLE result
  )
  if(NOT ${code} AND ${result})
     set(OPENVDB_PACKAGE_URL ${result})
  endif()
endif()
```

jmlait:
I like the idea of it being built into the CMake, but then it occurred to me that this might be a security issue?  It might not be expected that the git URL is auto-published by our cmake, so someone could leak their URL unexpectedly?

Idclip:
Very good point - we could add a parent option which turns on this behaviour which is off by default.

Idclip:
@jmlait did you mention you were going to try and solve this with options in the CMake build?

danrbailey (approved):

jmlait:
As mentioned, we'd ideally have options in Cmake, but we can make that a separate PR if we get there.

PR #1390, opened 2022-05-24
Dummy Point Surfacing SOP for testing purposes

Idclip:

@jmlait I haven't tried to use this SOP since the initial PR which was a while ago - it still builds and looks okay but let me know if you have any issues

### DIFF
--- a/openvdb_houdini/openvdb_houdini/CMakeLists.txt
+++ b/openvdb_houdini/openvdb_houdini/CMakeLists.txt
@@ -257,6 +257,7 @@ if(NOT OPENVDB_DSO_NAMES)
     SOP_OpenVDB_Points_Convert
     SOP_OpenVDB_Points_Delete
     SOP_OpenVDB_Points_Group
+    SOP_OpenVDB_Points_Surfacer
     SOP_OpenVDB_Potential_Flow
     SOP_OpenVDB_Prune
     SOP_OpenVDB_Rasterize_Points
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Points_Surfacer.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Points_Surfacer.cc
@@ -1,0 +1,485 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
+//
+/// @author Nick Avramoussis
+///
+/// @file SOP_OpenVDB_Points_Surfacer.cpp
+///
+/// @brief Surface VDB Points into a VDB Level Set using a variety of methods
+
+#include <houdini_utils/ParmFactory.h>
+#include <openvdb_houdini/Utils.h>
+#include <openvdb_houdini/PointUtils.h>
+#include <openvdb_houdini/SOP_NodeVDB.h>
+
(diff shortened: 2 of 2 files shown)


Idclip:
I spoke to soon, it doesn't build... I'll fix these soon

danrbailey:
Looks good. My main request is to add support for on-the-fly converting of Houdini points to VDB points and then rename this VDB Surfacer, then I'd be good for this to be merged more or less as-is. There are a few other small things - renaming the methods to make them a bit more descriptive (rasterZbP), removing the redundant inline statement on the templated methods, tidying up some the parameters and tooltips and ideally making this a compiled SOP instead of using duplicateSourceStealable().

This can be done as a later change after this has been merged, but it would also be good to support rasterizing multiple VDB points grids into a single level set and to add streaming support.

Idclip:
> Looks good. My main request is to add support for on-the-fly converting of Houdini points to VDB points and then rename this VDB Surfacer, then I'd be good for this to be merged more or less as-is. There are a few other small things - renaming the methods to make them a bit more descriptive (rasterZbP), removing the redundant inline statement on the templated methods, tidying up some the parameters and tooltips and ideally making this a compiled SOP instead of using duplicateSourceStealable().
> 
> This can be done as a later change after this has been merged, but it would also be good to support rasterizing multiple VDB points grids into a single level set and to add streaming support.

Thanks for the feedback - I agree with everything you've said _but_ this was never designed to be a feature ready SOP, only to provide the necessary hooks in Houdini so people could test. If you think this is actually worth making feature complete (i.e. native SideFX support) then I'd want to go back and re-write this/re-work the UI entirely - but I'm also not sure that's worth doing until we've figured out how best to expose all these raster tools in Houdini?

danrbailey:
> > Looks good. My main request is to add support for on-the-fly converting of Houdini points to VDB points and then rename this VDB Surfacer, then I'd be good for this to be merged more or less as-is. There are a few other small things - renaming the methods to make them a bit more descriptive (rasterZbP), removing the redundant inline statement on the templated methods, tidying up some the parameters and tooltips and ideally making this a compiled SOP instead of using duplicateSourceStealable().
> > This can be done as a later change after this has been merged, but it would also be good to support rasterizing multiple VDB points grids into a single level set and to add streaming support.
> 
> Thanks for the feedback - I agree with everything you've said _but_ this was never designed to be a feature ready SOP, only to provide the necessary hooks in Houdini so people could test. If you think this is actually worth making feature complete (i.e. native SideFX support) then I'd want to go back and re-write this/re-work the UI entirely - but I'm also not sure that's worth doing until we've figured out how best to expose all these raster tools in Houdini?

I think it absolutely could be a feature-ready SOP, there's a real gap in our toolset for a SOP like this. I think to make progress here and side-step the issue of rasterization, it would be worth introducing a Surfacing SOP like this that has a narrower scope and specializes in turning points into an SDF and not _also_ rasterizing arbitrary attributes or generating fog volumes or masks. As I see it, it's an open source equivalent to the VDB from Particle Fluid SOP (which I believe also does not do arbitrary attribute rasterization) and it's through this node that we can introduce any other surfacing algorithms like anisotropic surfacing. After all, there's still the VDB from Particles SOP which has broader scope, like doing arbitrary attribute rasterization.

PR #1400, opened 2022-06-04
TopologyToLevelSet active tile fix

### LINKED ISSUE #1397: [BUG] tools::topologyToLevelSet ignores tiles
### Environment
**Operating System:** Windows 10
**Version / Commit SHA:** VDB 8.1 / 9.1 / master
**Other:** -

### Describe the bug
When creating a level set from the topology of another grid (tools/TopologyToLevelSet.h), active tile values are being ignored.
Areas with only active tile values (not voxels in the LeafNode) will not contribute to the level set creation and thus leave holes. 

Responsible code
https://github.com/AcademySoftwareFoundation/openvdb/blob/9cc0e32f282c2d11d20404f0861b78c2db4d2fed/openvdb/openvdb/tools/TopologyToLevelSet.h#L215-L216

image

### To Reproduce
1. Take any **sparse** grid
2. 
```C++
  auto levelset = openvdb::tools::topologyToLevelSet(*grid);
  openvdb::io::File('reproducer.vdb').write({levelset});
```
4. `vdb_view reproducer.vdb` and see the holes in the distance field and reconstructed mesh. 

Setting `closingSteps=0` is also not a hotfix for the problem.

### Expected behavior
I would expect the behavior of `EXPAND_TILES`, i.e. that tiles are being densified (explicit leaf node instantiation) and considered during the level set creation. 
In the image above, this would create one solid cube without the missing bits in the foreground. 

### Recommendation
Judging by git blame, this is code that hasn't been touched in a long time and this behavior is at least 6+ years present. 
  - I would change all occurrences of `IGNORE_TILES` with `EXPAND_TILES` in TopologyToLevelSet.h. 
  - or: If the current behavior was intended, I'd recommend to add a parameter `ignoreTiles` with default value `true` (to retain the current behavior by default). This would allow a choice for the user.
  - At the very least I'd like to see an improvement for the documentation that hints at the ignorance of tiles and provides a workaround (eg. using `grid.tree().voxelizeActiveTiles()`). Generally speaking `voxelizeActiveTiles()` is hard to find in the code/docs, maybe it should be mentioned as the opposite of `Grid::pruneGrid()`.

Idclip:
Fixes #1397

### DIFF
--- a/openvdb/openvdb/tools/TopologyToLevelSet.h
+++ b/openvdb/openvdb/tools/TopologyToLevelSet.h
@@ -14,7 +14,6 @@
 
 #include "LevelSetFilter.h"
 #include "Morphology.h" // for erodeActiveValues and dilateActiveValues
-#include "SignedFloodFill.h"
 
 #include <openvdb/Grid.h>
 #include <openvdb/Types.h>
@@ -22,7 +21,9 @@
 #include <openvdb/util/NullInterrupter.h>
 #include <openvdb/openvdb.h>
 #include <openvdb/points/PointDataGrid.h>
+
 #include <tbb/task_group.h>
+
 #include <algorithm> // for std::min(), std::max()
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


danrbailey (approved):
Nice!

Issue #1299, opened 2022-01-30
[static analyzer warning]


andreys7:
### Environment
**Operating System:** (--)
**Version / Commit SHA:** --)
**CMake Version:** (--)
**Compiler:** (--)

### Describe the problem
Static analyzer complains:
 
openvdb/math/Math.h line 520, col 26: absolute value function 'labs' given an argument of type 'long long' but has parameter of type 'long' which may cause truncation of value 
      note: use function 'std::abs' instead

Line:     int64_t difference = labs(longLeft - longRight); 

Is it bug or trick?

### To Reproduce
1. Checkout repo '...'
2. Run CMake '....'
3. Build with '....'

### Additional context
(Add any other context about the problem here.)

Idclip:
Hey @andreys7 thanks - was this on Windows by any chance?

andreys7:
Yes, it was in Windows, Visual Studio 2019

PR #1414, opened 2022-06-12
Sweep out use of labs() from openvdb/Math.h

e4lam:
On Windows, labs() uses long which is int32_t instead of int64_t so we
were silently casting down in math::isUlpsEqual(). Since C++11, there
have been proper overloads of std::abs() for both long and long long
provided by `<cstdlib>` so just use that instead.


Idclip (approved):
Thanks @e4lam - also fixes https://github.com/AcademySoftwareFoundation/openvdb/issues/1299

jmlait (approved):

e4lam:
Sorry folks, I just noticed that there were still some minor consistencies here so I've just pushed another commit for this.

PR #1420, opened 2022-06-18
Minor fix to ccache path used when ccache is found

Idclip:


danrbailey:
Probably makes sense to remove the `RULE_LAUNCH_LINK` line given that CCache does nothing for linking?

Also, am I right in saying you'd just specify `CCACHE_ROOT` to get CMake to find ccache in a specific location like you would with any other dependency?

Idclip:
> Probably makes sense to remove the RULE_LAUNCH_LINK line given that CCache does nothing for linking?

Good point, removed.

> Also, am I right in saying you'd just specify CCACHE_ROOT to get CMake to find ccache in a specific location like you would with any other dependency?

I'm not sure about `<Name>_ROOT` variables for `find_program` - the first search order step in the CMake docs seems to have been copy/pasted from the `find_package` search order and I at least couldn't get `find_program` to respect `CCACHE_ROOT` or `Ccache_ROOT`. It does however read from things like `CMAKE_PROGRAM_PATH` which you can provide. So `ccache` doesn't necessarily need to be available on your PATH (hence this fix).

https://cmake.org/cmake/help/latest/command/find_program.html

danrbailey (approved):

PR #1421, opened 2022-06-18
Initial round of improvements to volumeToMesh

Idclip:
A very quick round of improvements/modernisations to volumeToMesh. Tested on many of the typical vdb cached level sets (crawler/iss/etc). 

Results in 10-15% improved performance.

Tested on a M1 Max (64GB/10CPU/system malloc/AppleClang 13.1.6)
Tested on Ubuntu x86/64 (64GB/12CPU E5-1650 v4 @ 3.60GHz/jemalloc 5/GCC 7.5)

There are many more additional (but more complicated) improvements that can be made in future PRs.

Signed-off-by: Nick Avramoussis <4256455+Idclip@users.noreply.github.com>

PR #1430, opened 2022-06-29
Python module VolumeToMesh compile fix

Idclip:
Related to: https://github.com/AcademySoftwareFoundation/openvdb/pull/1421#issuecomment-1169773997

### DIFF
--- a/openvdb/openvdb/python/pyGrid.h
+++ b/openvdb/openvdb/python/pyGrid.h
@@ -1395,8 +1395,16 @@ meshToLevelSet(py::object pointsObj, py::object trianglesObj, py::object quadsOb
     return tools::meshToLevelSet<GridType>(*xform, points, triangles, quads, halfWidth);
 }
 
+template<typename GridType,
+    typename std::enable_if<!std::is_scalar<typename GridType::ValueType>::value>::type* = nullptr>
+inline py::object
+volumeToQuadMesh(const GridType&, py::object)
+{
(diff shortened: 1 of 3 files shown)


openvdb/openvdb/python/pyGrid.h
```diff
@@ -1395,8 +1395,16 @@ meshToLevelSet(py::object pointsObj, py::object trianglesObj, py::object quadsOb
     return tools::meshToLevelSet<GridType>(*xform, points, triangles, quads, halfWidth);
 }
 
+template<typename GridType,
+    typename std::enable_if<!std::is_scalar<typename GridType::ValueType>::value>::type* = nullptr>
+inline py::object
+volumeToQuadMesh(const GridType&, py::object)
+{
+    OPENVDB_THROW(TypeError, "volume to mesh conversion is supported only for scalar grids");
+}
```
Idclip, line 1404: I wonder if I should re-introduce this behaviour to `volumeToMesh`? Anyone that was previously instantiating this code with non-scalar grids but relying on it not being called will now encouter this compilation error. It's more correct but potentially disruptive? @danrbailey any thoughts?

danrbailey (approved):
Changes here look good so I'm approving, however the changes detailed in pending changes look unrelated?

Idclip:
> Changes here look good so I'm approving, however the changes detailed in pending changes look unrelated?

Yes the changes refrenced in the notes have already been merged but I forgot to add notes with them. This change fixes a related issue on master that was never released, so no need to document

PR #1432, opened 2022-06-29
Code to do convex-hull based activation.

jmlait:
Unfortunately this requires GU_ConvexHull3D changes from future
houdini, so has version guards to restrict use.

Code for faster sdf-activation using dynamic node manager.


Idclip (approved):

Idclip:
Made some minor comments - haven't tested but looks good

PR #1439, opened 2022-07-17
Fix clamp bug in VDB Visualize SOP

danrbailey:
When displaying voxels or points as colors, the input value is not being clamped to the range, so it wraps around. This fixes that by applying the clamp before evaluating the UT_Ramp.


jmlait (approved):
19.5 also changes Ramps in Houdini to always clamp rather than wrap around, so this was also fixed on our end :>

danrbailey:
> 19.5 also changes Ramps in Houdini to always clamp rather than wrap around, so this was also fixed on our end :>

Oh great, good to know!

PR #1440, opened 2022-07-17
Fix issues in merge tool when using sum mode

danrbailey:
There are two issues being addressed here - one produces incorrect results when deep-copying nodes from the source tree as the mask tree that stores the updated topology of the const input tree is being pruned at the wrong level, the other produces incorrect results for grids which have non-zero background values.

There's also some minor refactoring and a couple of additions to the merge API in preparation for adding maximum mode.


Idclip (approved):
Haven't tested but changes look good

PR #1448, opened 2022-08-17
Various build fixes

### LINKED ISSUE #1447: [BUILD/Windows] openvdb_ax.dll does not export any symbols.
### Environment
**Operating System: Windows 10
**Version / Commit SHA: VDB 9.1
**CMake Version: 3.23.1
**Compiler: msvc 2019 - v16.9.16

### Describe the problem

When building with these options

```
  -DOPENVDB_BUILD_PYTHON_MODULE=ON
  -DOPENVDB_BUILD_AX=ON
  -DOPENVDB_AX_SHARED=ON
  -DOPENVDB_AX_STATIC=OFF
```
The generated `openvdb_ax.dll` does not export any symbols (and the linker will not emit an import library .lib because of that) leading to the following issues: 

- A shared library that doesn't export any symbols isn't very useful :) 
- Due to no import library being generated the python module will have a linker error

### To Reproduce
1. Build with above build options on windows
2. See error

### Additional context
The root cause is while the regular openvdb library gets `OPENVDB_DLL`  defined [over here](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/CMakeLists.txt#L687) no such supporting code exists for openvdb_ax, hence it doesn't get `OPENVDB_AX_DLL` defined and the code [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/Platform.h#L268) in `platform.h` decides to use [this](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/Platform.h#L278) define, leading to no exports due to a lacking `dllexport` attribute

Adding `-DOPENVDB_AX_DLL` to the CMAKE_CXX_FLAGS to my cmake invocation, solves the issue short term, but a proper fix is likely the way to go here. 

I don't have a CLA signed, so cannot send a pull that would have added a define right around [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb_ax/openvdb_ax/CMakeLists.txt#L309) 

### Additional question

Reporting this feels strange, I can't shake the eerie feeling I'm missing something super obvious here, as CI should have caught this right out of the gate...what did i miss?

Idclip:
Incorporatres some changes from #1445 but in summary:

 - Fixes #1447, where AX symbols were not being exported
 - Added python tests for AX
 - Fixed some CMake issues with TBB and Boost
 - CI and doc updates

### DIFF
--- a/openvdb/openvdb/python/test/TestOpenVDB.py
+++ b/openvdb/openvdb/python/test/TestOpenVDB.py
@@ -115,13 +115,14 @@ class TestOpenVDB(unittest.TestCase):
 
         vec3_grid.fill((0, 0, 0), (7, 7, 7), (-1,-1,-1), active=True)
         openvdb.ax('v@vec_grid = @float_grid;', [vec3_grid, float_grid])
+        acc = vec3_grid.getAccessor()
 
         for i in range(0, 8):
             for j in range(0, 8):
                 for k in range(0, 8):
                     ijk = (i, j, k)
                     lsq = (i*i)+(j*j)+(k*k)
-                    self.assertEqual(acc.getValue(ijk), float(lsq))
+                    self.assertEqual(acc.getValue(ijk), (lsq,lsq,lsq))
 
 
     def testTransform(self):
--- a/openvdb/openvdb/python/test/TestOpenVDB.py
+++ b/openvdb/openvdb/python/test/TestOpenVDB.py
@@ -12,6 +12,17 @@ C++-to-Python bindings, not the OpenVDB library itself.
 import os, os.path
 import sys
 import unittest
+import inspect
+
+# If on Windows, add required dll directories from our binary build tree
+if 'add_dll_directory' in dir(os):
+    # Should be something like Release, Debug, etc
+    config = os.path.basename(os.getcwd())
+    os.add_dll_directory(os.getcwd() + '\\..\\..\\' + config)
+    if os.getenv('OPENVDB_TEST_PYTHON_AX'):
+        os.add_dll_directory(os.getcwd() +
+            '\\..\\..\\..\\..\\openvdb_ax\\openvdb_ax\\' + config)
+
 import pyopenvdb as openvdb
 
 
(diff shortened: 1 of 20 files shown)


openvdb/openvdb/python/test/TestOpenVDB.py
```diff
@@ -60,6 +84,46 @@ def testModule(self):
...
+
+        vec3_grid.fill((0, 0, 0), (7, 7, 7), (-1,-1,-1), active=True)
+        openvdb.ax('v@vec_grid = @float_grid;', [vec3_grid, float_grid])
+
+        for i in range(0, 8):
+            for j in range(0, 8):
+                for k in range(0, 8):
+                    ijk = (i, j, k)
+                    lsq = (i*i)+(j*j)+(k*k)
+                    self.assertEqual(acc.getValue(ijk), float(lsq))
```
richhones, line 124: should this be checking the vec3_grid values?

openvdb/openvdb/unittest/TestCoord.cc
```diff
@@ -277,6 +277,7 @@ TEST_F(TestCoord, testCoordBBox)
         const size_t count = b.volume();
         size_t n = 0;
         openvdb::Coord::ValueType unused = 0;
+        (void)unused;
```
richhones, line 280: [[maybe unused]] exists from C++17

openvdb_ax/openvdb_ax/compiler/VolumeExecutable.cc
```diff
@@ -1387,7 +1387,7 @@ VolumeExecutable::getActiveTileStreaming(const std::string& name,
                      const ast::tokens::CoreType& type) const
 {
     assert(mAttributeRegistry);
-    if (mSettings->mActiveTileStreaming == VolumeExecutable::Streaming::AUTO) {
+    if (mSettings->mActiveTileStreaming.get() == VolumeExecutable::Streaming::AUTO) {
```
richhones, line 1390: assume these were causing build issues too? (although not yet released hence no mention in release notes)

openvdb/openvdb/unittest/TestCoord.cc
Idclip: This PR doesn't enable C++17

openvdb_ax/openvdb_ax/compiler/VolumeExecutable.cc
Idclip: Yes for some reason MSVC doesn't like this

https://github.com/AcademySoftwareFoundation/openvdb/runs/7822777048?check_suite_focus=true

openvdb/openvdb/python/test/TestOpenVDB.py
Idclip: Oh great catch, will fix

Idclip: Fixed

richhones (approved):

PR #1452, opened 2022-08-17
ValueAccessor Buffer Cache

Idclip:
```
Improvements:
    - Significantly improved the performance of all ValueAccessor methods which
    access LeafNode value buffer data. This improvement applies to any type which
    is delay load compatible (all default types except for bool and mask grids).

Bug Fixes:
    - Fixed a bug with ValueAccessor::addLeaf which wouldn't add the provided
    leaf nodes to the underlying tree. This bug did NOT affect the specialized
    accessors which are used by the default tree configuration.
    - Fixed a bug where ValueAccessor::probeNode<NodeT> and ValueAccessor:probeConstNode<NodeT>
    would return a nullptr if the NodeT type was not explicitly being cached
    by the accessor but the node existed in the tree.
```

Test 1 (Volume Advect)
Before: 1 minute, 35 seconds and 860.3 milliseconds (95860.3ms)
After: 1 minute, 28 seconds and 372.2 milliseconds (88372.2ms)
**8% Faster**

Test 2 (Volume To Mesh)
Before: 872.2 milliseconds
After: 840.7 milliseconds
**4% Faster**

Test 3 (Fast Sweeping)
Before: 32 seconds and 328.6 milliseconds (32328.6ms)
After: 28 seconds and 859.0 milliseconds (28859.0ms)
**11% Faster**

### DIFF
--- a/openvdb/openvdb/tree/ValueAccessor.h
+++ b/openvdb/openvdb/tree/ValueAccessor.h
@@ -641,11 +641,10 @@ public:
     void addTile(Index level, const Coord& xyz, const ValueType& value, bool state)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't add a tile to a const tree");
-        if (NodeType::LEVEL < level) return;
-        if (this->isHashed(xyz)) {
+        if (NodeType::LEVEL >= level && this->isHashed(xyz)) {
             assert(mNode);
-            return const_cast<NodeType*>(mNode)->addTileAndCache(
-                level, xyz, value, state, *mParent);
+            const_cast<NodeType*>(mNode)->addTileAndCache(level, xyz, value, state, *mParent);
+            return;
         }
         mNext.addTile(level, xyz, value, state);
     }
@@ -683,31 +682,41 @@ public:
     NodeT* probeNode(const Coord& xyz)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't get a non-const node from a const tree");
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_BEGIN
-        if (this->isHashed(xyz)) {
-            if ((std::is_same<NodeT, NodeType>::value)) {
+        if constexpr ((std::is_same<NodeT, NodeType>::value)) {
+            if (this->isHashed(xyz)) {
                 assert(mNode);
                 return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
             }
+        }
+
+        if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
+            // don't need to ascend the chain, descend the tree
             return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
         }
-        return mNext.template probeNode<NodeT>(xyz);
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        else {
+            // ascend
+            return mNext.template probeNode<NodeT>(xyz);
+        }
     }
 
     template<typename NodeT>
     const NodeT* probeConstNode(const Coord& xyz)
     {
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_BEGIN
-        if (this->isHashed(xyz)) {
-            if ((std::is_same<NodeT, NodeType>::value)) {
+        if constexpr ((std::is_same<NodeT, NodeType>::value)) {
+            if (this->isHashed(xyz)) {
                 assert(mNode);
                 return reinterpret_cast<const NodeT*>(mNode);
             }
+        }
+
+        if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
+            // don't need to ascend the chain, descend the tree
             return mNode->template probeConstNodeAndCache<NodeT>(xyz, *mParent);
         }
-        return mNext.template probeConstNode<NodeT>(xyz);
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        else {
+            // ascend
+            return mNext.template probeConstNode<NodeT>(xyz);
+        }
     }
 
     /// Return the active state of the voxel at the given coordinates.
@@ -870,9 +879,7 @@ public:
         , mHash(CoordLimits::max())
         , mNode(nullptr)
         , mNext(parent)
-        , mBuffer(nullptr)
-    {
-    }
+        , mBuffer(nullptr) {}
 
     //@{
     /// Copy another CacheItem's node pointers and hash keys, but not its parent pointer.
@@ -881,9 +888,7 @@ public:
         , mHash(other.mHash)
         , mNode(other.mNode)
         , mNext(parent, other.mNext)
-        , mBuffer(other.mBuffer)
-    {
-    }
+        , mBuffer(other.mBuffer) {}
 
     CacheItem& copy(TreeCacheT& parent, const CacheItem& other)
     {
@@ -953,11 +958,10 @@ public:
     void addTile(Index level, const Coord& xyz, const ValueType& value, bool state)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't add a tile to a const tree");
-        if (NodeType::LEVEL < level) return;
-        if (this->isHashed(xyz)) {
+        if (NodeType::LEVEL >= level && this->isHashed(xyz)) {
             assert(mNode);
-            return const_cast<NodeType*>(mNode)->addTileAndCache(
-                level, xyz, value, state, *mParent);
+            const_cast<NodeType*>(mNode)->addTileAndCache(level, xyz, value, state, *mParent);
+            return;
         }
         mNext.addTile(level, xyz, value, state);
     }
@@ -995,27 +999,41 @@ public:
     NodeT* probeNode(const Coord& xyz)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't get a non-const node from a const tree");
-        if (this->isHashed(xyz)) {
-            if ((std::is_same<NodeT, NodeType>::value)) {
+        if constexpr ((std::is_same<NodeT, NodeType>::value)) {
+            if (this->isHashed(xyz)) {
                 assert(mNode);
                 return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
             }
+        }
+
+        if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
+            // don't need to ascend the chain, descend the tree
             return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
         }
-        return mNext.template probeNode<NodeT>(xyz);
+        else {
+            // ascend
+            return mNext.template probeNode<NodeT>(xyz);
+        }
     }
 
     template<typename NodeT>
     const NodeT* probeConstNode(const Coord& xyz)
     {
-        if (this->isHashed(xyz)) {
-            if ((std::is_same<NodeT, NodeType>::value)) {
+        if constexpr ((std::is_same<NodeT, NodeType>::value)) {
+            if (this->isHashed(xyz)) {
                 assert(mNode);
                 return reinterpret_cast<const NodeT*>(mNode);
             }
+        }
+
+        if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
+            // don't need to ascend the chain, descend the tree
             return mNode->template probeConstNodeAndCache<NodeT>(xyz, *mParent);
         }
-        return mNext.template probeConstNode<NodeT>(xyz);
+        else {
+            // ascend
+            return mNext.template probeConstNode<NodeT>(xyz);
+        }
     }
 
     /// Return the active state of the voxel at the given coordinates.
@@ -1895,15 +1913,21 @@ public:
     {
         assert(BaseT::mTree);
         static_assert(!BaseT::IsConstTree, "can't get a non-const node from a const tree");
-        if ((std::is_same<NodeT, NodeT0>::value)) {
+        if constexpr ((std::is_same<NodeT, NodeT0>::value)) {
             if (this->isHashed(xyz)) {
                 assert(mNode0);
                 return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
             }
             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
+        } else if constexpr (NodeT::LEVEL < NodeT0::LEVEL) {
+            // Might still be worth caching this path if a NodeT0
+            // is accessed in the future
+            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
+        } else {
+            // If we're accessing a node above our top most cache level then
+            // there's no point trying to cache it
+            return BaseT::mTree->root().template probeNode<NodeT>(xyz);
         }
-        // No point trying to cache as this NodeT isn't part of this accessors chain
-        return BaseT::mTree->root().template probeNode<NodeT>(xyz);
     }
     LeafNodeT* probeLeaf(const Coord& xyz)
     {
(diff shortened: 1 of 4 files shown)


jmlait:
Looks good, modulo the notes you have made above.

The ABI concern would be that an accessor is now longer if it comes from a new version.  So if accessors are passed through a void* we'd be in trouble.  Our requirement is that you can pass grids through void *, however, not accessors, so I think this is safe...

The grid may have a pointer to foreign accessors, but they trigger the clear() method on them, so should also be safe as they'd use the foreign vtable so respect the size.

All that said, I think I will still vote we ABI guard this new member.  We're moving to 10 anyways, and this ensures there are no unpleasant surprises.

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -189,15 +196,23 @@ class ValueAccessor: public ValueAccessorBase<_TreeType, IsSafe>
     using LeafNodeT = typename TreeType::LeafNodeType;
     using ValueType = typename RootNodeT::ValueType;
     using BaseT = ValueAccessorBase<TreeType, IsSafe>;
-    using LockT = typename MutexType::scoped_lock;
+    using LockT = std::scoped_lock<MutexType>;
```
Idclip, line 199: This is technically an API change but necessary for use with `std` mutex types.

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -604,12 +623,18 @@ class CacheItem
     void addLeaf(LeafNodeType* leaf)
     {
         static_assert(!TreeCacheT::IsConstTree, "can't add a node to a const tree");
-        if (NodeType::LEVEL == 0) return;
+        if (NodeType::LEVEL == 0) {
+            mNext.addLeaf(leaf);
+            return;
+        }
```
Idclip, line 629: This was a bug - calling `addLeaf` here wouldn't do anything and the leaf ptr would leak

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -801,6 +826,324 @@ class CacheItem
...
+    const NodeType* mNode;
+    using RestT = typename NodeVecT::PopFront;
+    CacheItem<TreeCacheT, RestT, /*AtRoot=*/RestT::Size == 1, false> mNext;
+};// end of CacheItem
+
+
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+
+template<typename TreeCacheT, typename NodeVecT>
+class CacheItem<TreeCacheT, NodeVecT, /*AtRoot=*/false, true>
```
Idclip, line 851: We unfortunately need an entire new specialization of `CacheItem` for LeafNodes for the general accessor case, otherwise we end up adding an extra pointer to every `CacheItem` unnecessarily.

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -1471,16 +1888,15 @@ class ValueAccessor1 : public ValueAccessorBase<_TreeType, IsSafe>
...
             if (this->isHashed(xyz)) {
                 assert(mNode0);
                 return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
             }
             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeNode<NodeT>(xyz);
```
Idclip, line 1899: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -1493,16 +1909,15 @@ class ValueAccessor1 : public ValueAccessorBase<_TreeType, IsSafe>
...
             if (this->isHashed(xyz)) {
                 assert(mNode0);
                 return reinterpret_cast<const NodeT*>(mNode0);
             }
             return BaseT::mTree->root().template probeConstNodeAndCache<NodeT>(xyz, this->self());
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeConstNode<NodeT>(xyz);
```
Idclip, line 1920: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -1908,19 +2412,16 @@ class ValueAccessor2 : public ValueAccessorBase<_TreeType, IsSafe>
             }
             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeNode<NodeT>(xyz);
```
Idclip, line 2416: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -1937,42 +2438,20 @@ class ValueAccessor2 : public ValueAccessorBase<_TreeType, IsSafe>
             }
             return BaseT::mTree->root().template probeConstNodeAndCache<NodeT>(xyz, this->self());
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeConstNode<NodeT>(xyz);
```
Idclip, line 2442: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -2454,8 +3018,8 @@ class ValueAccessor3 : public ValueAccessorBase<_TreeType, IsSafe>
             }
             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeNode<NodeT>(xyz);
```
Idclip, line 3022: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -2496,8 +3059,8 @@ class ValueAccessor3 : public ValueAccessorBase<_TreeType, IsSafe>
             }
             return BaseT::mTree->root().template probeConstNodeAndCache<NodeT>(xyz, this->self());
         }
-        return nullptr;
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
+        // No point trying to cache as this NodeT isn't part of this accessors chain
+        return BaseT::mTree->root().template probeConstNode<NodeT>(xyz);
```
Idclip, line 3063: Bugfix - even if `NodeT` isn't being cached it might exist and a nullptr will always be returned

openvdb/openvdb/unittest/TestValueAccessor.cc
```diff
@@ -39,76 +39,10 @@ class TestValueAccessor: public ::testing::Test
...
-    {
-        accessorTest<ValueAccessor1<Tree4Type, true,  2> >();
-        accessorTest<ValueAccessor1<Tree4Type, false, 2> >();
-    }
-    //cache node level 1 and 3
-    void testTree5Accessor213()
-    {
-        accessorTest<ValueAccessor2<Tree5Type, true, 1,3> >();
-        accessorTest<ValueAccessor2<Tree5Type, false, 1,3> >();
-    }
```
Idclip, line 107: There wern't even being run...

jmlait (approved):

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -54,7 +56,13 @@ template<typename TreeType, bool IsSafe = true, Index L0 = 0, Index L1 = 1>
 class ValueAccessor2;
 template<typename TreeType, bool IsSafe = true, Index L0 = 0, Index L1 = 1, Index L2 = 2>
 class ValueAccessor3;
-template<typename TreeCacheT, typename NodeVecT, bool AtRoot> class CacheItem;
+template<typename TreeCacheT, typename NodeVecT, bool AtRoot
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+, bool BypassLeafAPI =
+        std::is_same<typename NodeVecT::Front, typename NodeVecT::Front::LeafNodeType>::value &&
+        std::is_same<typename NodeVecT::Front::LeafNodeType::Buffer::StorageType, typename NodeVecT::Front::ValueType>::value
```
danrbailey, line 63: If I'm reading this correctly - this disables this optimization for Mask Grids where the BuildType != ValueType as expected, but don't we also wish to disable this for Bool Grids where BuildType == ValueType but delayed-loading is still disabled, so nothing to gain in using this optimization and it could even be more expensive?

danrbailey (approved):
Thanks Nick, this looks good to me. Shame we have to add this extra specialization, but seems unavoidable. Just one questions about Bool Grids, otherwise I'm happy for this to go in.

openvdb/openvdb/tree/ValueAccessor.h
Idclip: LeafBuffer StorageType for bool grids isn't bools, it's the same type as mask grids i.e. `NodeMaskType::Word` - so this is disabled for bool grids too. There's a small static assert test I've added which checks this as well

danrbailey: Got it, thanks for correcting me there.

Idclip:
@jmlait @danrbailey I've updated this and added the tile fix. I've added some additional tests - we need more but this is all I have time for for now unfortunately.

Could you both give this another once over? The main thing to check now is the behaviour of `probeNode` and `probeConstNode` methods. These should now always return the desired node if it exists which is what I believe to be the desired result.

For the general case:

Before:
```c++
        if (this->isHashed(xyz)) {  // This is only true if NodeT == NodeType anyway...
            if ((std::is_same<NodeT, NodeType>::value)) {
                assert(mNode);
                return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
            }
            // This can never be hit?
            return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
        }
        return mNext.template probeNode<NodeT>(xyz);
```

After:

```c++
        if constexpr ((std::is_same<NodeT, NodeType>::value)) {
            if (this->isHashed(xyz)) {
                assert(mNode);
                return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
            }
        }

        if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
            // don't need to ascend the chain, descend the tree
            return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
        }
        else {
            // ascend
            return mNext.template probeNode<NodeT>(xyz);
        }
```

And an example for specialized accessors:

Before, for ValueAccessor2:

```c++
        if ((std::is_same<NodeT, NodeT0>::value)) {
            if (this->isHashed0(xyz)) {
                assert(mNode0);
                return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
            } else if (this->isHashed1(xyz)) {
                assert(mNode1);
                return const_cast<NodeT1*>(mNode1)->template probeNodeAndCache<NodeT>(xyz, *this);
            }
            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
        } else if ((std::is_same<NodeT, NodeT1>::value)) {
            if (this->isHashed1(xyz)) {
                assert(mNode1);
                return reinterpret_cast<NodeT*>(const_cast<NodeT1*>(mNode1));
            }
            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
        }

        // What if we are using a ValueAccessor2 on a Tree5??
        return nullptr;
```


After:

```c++
        if constexpr ((std::is_same<NodeT, NodeT0>::value)) {
            if (this->isHashed0(xyz)) {
                assert(mNode0);
                return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
            } else if (this->isHashed1(xyz)) {
                assert(mNode1);
                return const_cast<NodeT1*>(mNode1)->template probeNodeAndCache<NodeT>(xyz, *this);
            }
            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
        } else if constexpr ((std::is_same<NodeT, NodeT1>::value)) {
            if (this->isHashed1(xyz)) {
                assert(mNode1);
                return reinterpret_cast<NodeT*>(const_cast<NodeT1*>(mNode1));
            }
            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
        } else if constexpr (NodeT::LEVEL < NodeT1::LEVEL) {
            // Might still be worth caching this path if a NodeT0 or NodeT1
            // are accessed in the future
            return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
        } else {
            // If we're accessing a node above our top most cache level then
            // there's no point trying to cache it
            return BaseT::mTree->root().template probeNode<NodeT>(xyz);
        }
```

danrbailey:
> @jmlait @danrbailey I've updated this and added the tile fix. I've added some additional tests - we need more but this is all I have time for for now unfortunately.
> 
> Could you both give this another once over? The main thing to check now is the behaviour of `probeNode` and `probeConstNode` methods. These should now always return the desired node if it exists which is what I believe to be the desired result.
> 
> For the general case:
> 
> Before:
> 
> ```c++
>         if (this->isHashed(xyz)) {  // This is only true if NodeT == NodeType anyway...
>             if ((std::is_same<NodeT, NodeType>::value)) {
>                 assert(mNode);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
>             }
>             // This can never be hit?
>             return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
>         }
>         return mNext.template probeNode<NodeT>(xyz);
> ```
> 
> After:
> 
> ```c++
>         if constexpr ((std::is_same<NodeT, NodeType>::value)) {
>             if (this->isHashed(xyz)) {
>                 assert(mNode);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeType*>(mNode));
>             }
>         }
> 
>         if (NodeT::LEVEL < NodeType::LEVEL && mNode) {
>             // don't need to ascend the chain, descend the tree
>             return const_cast<NodeType*>(mNode)->template probeNodeAndCache<NodeT>(xyz, *mParent);
>         }
>         else {
>             // ascend
>             return mNext.template probeNode<NodeT>(xyz);
>         }
> ```
> 
> And an example for specialized accessors:
> 
> Before, for ValueAccessor2:
> 
> ```c++
>         if ((std::is_same<NodeT, NodeT0>::value)) {
>             if (this->isHashed0(xyz)) {
>                 assert(mNode0);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
>             } else if (this->isHashed1(xyz)) {
>                 assert(mNode1);
>                 return const_cast<NodeT1*>(mNode1)->template probeNodeAndCache<NodeT>(xyz, *this);
>             }
>             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
>         } else if ((std::is_same<NodeT, NodeT1>::value)) {
>             if (this->isHashed1(xyz)) {
>                 assert(mNode1);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeT1*>(mNode1));
>             }
>             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
>         }
> 
>         // What if we are using a ValueAccessor2 on a Tree5??
>         return nullptr;
> ```
> 
> After:
> 
> ```c++
>         if constexpr ((std::is_same<NodeT, NodeT0>::value)) {
>             if (this->isHashed0(xyz)) {
>                 assert(mNode0);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeT0*>(mNode0));
>             } else if (this->isHashed1(xyz)) {
>                 assert(mNode1);
>                 return const_cast<NodeT1*>(mNode1)->template probeNodeAndCache<NodeT>(xyz, *this);
>             }
>             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
>         } else if constexpr ((std::is_same<NodeT, NodeT1>::value)) {
>             if (this->isHashed1(xyz)) {
>                 assert(mNode1);
>                 return reinterpret_cast<NodeT*>(const_cast<NodeT1*>(mNode1));
>             }
>             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
>         } else if constexpr (NodeT::LEVEL < NodeT1::LEVEL) {
>             // Might still be worth caching this path if a NodeT0 or NodeT1
>             // are accessed in the future
>             return BaseT::mTree->root().template probeNodeAndCache<NodeT>(xyz, *this);
>         } else {
>             // If we're accessing a node above our top most cache level then
>             // there's no point trying to cache it
>             return BaseT::mTree->root().template probeNode<NodeT>(xyz);
>         }
> ```

Yes, this seems correct to me, thanks Nick! :+1:

PR #1460, opened 2022-08-29
Fix VS2019/VS2022 compiler error in Morphology.h

e4lam:
For some unknown reason MSVC /permissive- has problems with correctly
inferring the LeafT here due to it being used with different types in
the same function. Changing it from LeafT to auto instead allows it
to compile.


e4lam:
For reference, I've also logged an MSVC bug for this:
https://developercommunity.visualstudio.com/t/openvdb-fails-to-compile-with-permissiv/10134549

Idclip (approved):

jmlait (approved):

PR #1461, opened 2022-08-29
Remove obsolete #ifndef _WIN32 code paths.

e4lam:
This is no longer needed in VS2019, and out right errors with
/permissive- enabled.


Idclip (approved):

e4lam:
Added similar changes to `tools/ValueTransformer.h` as well which I noticed belatedly.

jmlait (approved):

PR #1462, opened 2022-08-30
Sweep out use of M_PI* constants

e4lam:
Replace M_PI* constants with openvdb::math::pi<T>(). This allows us to
remove the reliance on having the _USE_MATH_DEFINES macro being defined
prior to any inclusion of cmath/math.h on Windows. This was easily
broken by bad include orders inside openvdb or by users of openvdb.

### DIFF
--- a/doc/examplecode.txt
+++ b/doc/examplecode.txt
@@ -921,7 +921,7 @@ openvdb::Vec3SGrid::Ptr grid = ...;
 
 // Construct the rotation matrix.
 openvdb::math::Mat3s rot45 =
-    openvdb::math::rotation<openvdb::math::Mat3s>(openvdb::math::Y_AXIS, M_PI_4);
+    openvdb::math::rotation<openvdb::math::Mat3s>(openvdb::math::Y_AXIS, openvdb::math::pi<double>()/4.0);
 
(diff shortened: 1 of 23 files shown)


Idclip (approved):
This is great @e4lam thank you. 

Once merged we should add a release note that mentions that define is no longer there on Windows.

e4lam:
> Once merged we should add a release note that mentions that define is no longer there on Windows.

Where is that? Should I add a commit that does this in `openvdb/CHANGES` and `openvdb/docs/changes.txt` ? I just noticed that the "In development" section is for Version 9.1.1 but I guess I could add a note there now and then someone else will change it to 10.0.0 ?

Idclip:
> Where is that? Should I add a commit that does this in `openvdb/CHANGES` and `openvdb/docs/changes.txt` ? I just noticed that the "In development" section is for Version 9.1.1 but I guess I could add a note there now and then someone else will change it to 10.0.0 ?

The way we do this now is that, for each PR where we want to log a release note, we create a `.txt` file in `/pendingchanges` which describes the change. We then periodically merge them into the main change log. If you could add a file there with a description of this change that would be great!

jmlait (approved):

PR #1474, opened 2022-09-20
add .envrc to .gitignore

marksisson:
direnv (https://github.com/direnv/direnv) can be used for local developer workflow enhancements.  direnv uses .envrc file for configuration.  Because .envrc is usually specific to a developer workflow, it is generally not committed.

Signed-off-by: Mark Sisson <5761292+marksisson@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /build/*
+.envrc


marksisson:
Instead of adding .envrc to .gitignore, I am able to use a global ignore, which solves the case where the project does not commit .envrc.  And I can use `git add -f .envrc` for projects that do commit the .envrc.

So this patch is no longer needed.

PR #1477, opened 2022-09-27
Offset root node

kmuseth:
Changing ABI to allow for future re-alignment of the root node


openvdb/openvdb/tree/RootNode.h
```diff
@@ -897,6 +897,17 @@ class RootNode
     template<typename OtherRootNodeType, typename VisitorOp>
     void visit2(OtherRootNodeType& other, VisitorOp&) const;
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+    /// Return the grid index coordinates of this node's local origin.
+    const Coord& origin() const { return mOrigin; }
+    /// @brief change the origin on this root node
+    /// @param origin the index coordinated of the new alignment
```
danrbailey, line 904: I think this is supposed to be "index coordinates" with an s?

openvdb/openvdb/tree/RootNode.h
```diff
@@ -1057,6 +1084,9 @@ inline
 RootNode<ChildT>::RootNode(const RootNode<OtherChildType>& other,
     const ValueType& backgd, const ValueType& foregd, TopologyCopy)
     : mBackground(backgd)
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+    , mOrigin(0, 0, 0)
```
danrbailey, line 1088: Should there be some kind of check to confirm that other.origin() == Coord(0, 0, 0)?

In a hypothetical world where VDB 10.0 only supports zero origins and 10.1 adds support for non-zero origins, it should be possible for a VDB with a non-zero origin to be reinterpret casted from 10.1 to 10.0 and this copy-constructor run that quietly adjusts the origin that could be unexpected for the user? Throwing here would at least inform them that they need to re-root _before_ trying to copy-construct from this grid.

openvdb/openvdb/tree/RootNode.h
```diff
@@ -1143,6 +1176,9 @@ struct RootNodeCopyHelper<RootT, OtherRootT, /*Compatible=*/true>
         };
 
         self.mBackground = Local::convertValue(other.mBackground);
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+        self.mOrigin = other.mOrigin;
```
danrbailey, line 1180: This seems odd to me? Shouldn't the copy constructor and the assignment operator have identical behavior in that they both set the origin to zero because this is all that is supported in this version (and ideally warn or error if the incoming grid is non-zero) ?

openvdb/openvdb/tree/RootNode.h
```diff
@@ -2629,6 +2668,18 @@ RootNode<ChildT>::addChild(ChildT* child)
...
 }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+template<typename ChildT>
+inline void
+RootNode<ChildT>::setOrigin(const Coord &origin)
+{
+    mOrigin = origin;
+    if (mOrigin != Coord(0,0,0)) {
+        OPENVDB_THROW(ValueError, "RootNode::setOrigin: non-zero offsets are still not supported");
```
danrbailey, line 2678: Maybe we use "currently" instead of "still"? Or otherwise, we could add "can you believe it?" to the end? :)

openvdb/openvdb/tree/RootNode.h
```diff
@@ -1172,6 +1216,9 @@ RootNode<ChildT>::operator=(const RootNode& other)
 {
     if (&other != this) {
         mBackground = other.mBackground;
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+        mOrigin = other.mOrigin;
+#endif
```
danrbailey, line 1221: Missing the zero check here?

openvdb/openvdb/tree/RootNode.h
```diff
@@ -1082,6 +1117,9 @@ inline
 RootNode<ChildT>::RootNode(const RootNode<OtherChildType>& other,
     const ValueType& backgd, TopologyCopy)
     : mBackground(backgd)
+#if OPENVDB_ABI_VERSION_NUMBER >= 10
+    , mOrigin(0, 0, 0)
```
danrbailey, line 1121: This should match the approach you took above - ie set origin to the origin of the other grid, then error if it is non-zero.

openvdb/openvdb/tree/RootNode.h
kmuseth: good catch - resolved

danrbailey (approved):

PR #1480, opened 2022-09-29
Removed unnecessary null pointer checks issue

Luc35b:
Looked through every file to remove unnecessary null pointer checks. Was unsure about change made in first commit in ".../Viewer.cc" so change was reverted in second commit. Signed-off-by: Lucas Baraya <lucasbarayarw@gmail.com>

### DIFF
--- a/openvdb/openvdb/tree/LeafBuffer.h
+++ b/openvdb/openvdb/tree/LeafBuffer.h
@@ -330,7 +330,7 @@ template<typename T, Index Log2Dim>
 inline bool
 LeafBuffer<T, Log2Dim>::deallocate()
 {
-    if (mData != nullptr && !this->isOutOfCore()) {
+    if (!this->isOutOfCore()) {
         delete[] mData;
         mData = nullptr;
         return true;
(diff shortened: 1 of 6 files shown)


openvdb/openvdb/tree/LeafBuffer.h
```diff
@@ -330,7 +330,7 @@ template<typename T, Index Log2Dim>
 inline bool
 LeafBuffer<T, Log2Dim>::deallocate()
 {
-    if (mData != nullptr && !this->isOutOfCore()) {
```
jmlait, line 333: This is not the same as we will return false for nullptr and in-core in the previous  code, but now are returning true..... 

I'm not sure of the implications though.

jmlait (commented):
Thanks for the PR....  Bad memories of the bad-old days when free(0) crashed :>

openvdb/openvdb/tree/LeafBuffer.h
Luc35b: If the implication is unknown I'll change it back thanks for reviewing.

kmuseth (approved):
Looks good to me

jmlait (approved):

PR #1469, opened 2022-09-05
Use Larger GHA instances from ASWF for Linux/Windows

danrbailey:

PR #1481, opened 2022-09-29
Fix for finding versioned TBB libraries as of TBB 2021.5 and Blosc external libs

Idclip:

### DIFF
--- a/cmake/FindBlosc.cmake
+++ b/cmake/FindBlosc.cmake
@@ -376,9 +376,74 @@ if(NOT TARGET Blosc::blosc)
     set_target_properties(Blosc::blosc PROPERTIES
       INTERFACE_LINK_DIRECTORIES
          "\$<\$<CONFIG:Release>:${Blosc_RELEASE_LIBRARY_DIRS}>;\$<\$<CONFIG:Debug>:${Blosc_DEBUG_LIBRARY_DIRS}>")
-    target_link_libraries(Blosc::blosc INTERFACE
-      $<$<CONFIG:Release>:lz4;snappy;zlib;zstd>
-      $<$<CONFIG:Debug>:lz4d;snappyd;zlibd;zstd>)
+
+    set(BLOSC_EXTERNAL_LIBRARIES lz4 snappy zlib zstd)
+
+    foreach(BLOSC_EXTERNAL_LIB ${BLOSC_EXTERNAL_LIBRARIES})
+
+      foreach(BUILD_TYPE ${BLOSC_BUILD_TYPES})
+        set(_BLOSC_CMAKE_IGNORE_PATH ${CMAKE_IGNORE_PATH})
+
+        if(VCPKG_TOOLCHAIN)
+          if(BUILD_TYPE STREQUAL RELEASE)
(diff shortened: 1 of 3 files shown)


Idclip:
I am so tired of vcpkg/library maintainers changing the names of their libraries whenever they feel like it with absolutely zero backwards compatibility. We've gone from `snappy` to `snappyd` back to `snappy` in vcpkg. Same with `zstd`. `tbb` now suffixes its libraries with an ABI number... but only on Windows... and only `tbb`, not `tbbmalloc` or others :|

I guess we should try to move to `find_package(<PACKAGE> CONFIG)`, though that assumes upstream config files are correct (spoilers: they won't always be). I don't want to have to support both our MODULE ones and CONFIG ones and relying solely on the upstream CONFIG files feels pretty restrictive, especially for VFX studios. Maybe that's less true now? Guess I might just need to accept that CMake Config files are now the norm as it's getting increasily more painful to update our modules every other week.

Idclip:
This should be good to go now. Remaining failure is due to disk space again:

> D:\a\openvdb\openvdb\openvdb\openvdb\unittest\TestTreeCombine.cc : fatal error C1085: Cannot write compiler generated file: 'D:\a\openvdb\openvdb\build\openvdb\openvdb\unittest\vdb_test.dir\Debug\TestTreeCombine.obj': No space left on device [D:\a\openvdb\openvdb\build\openvdb\openvdb\unittest\vdb_test.vcxproj]

danrbailey:
Thanks @idclip, the disk space issue will be solved by this one - https://github.com/AcademySoftwareFoundation/openvdb/pull/1469. I've brought across your tbb fix that had me stumped for a bit and it seemed to address the issue there, I think I'm now missing your blosc fix so will bring that across to test as well. I ran into the snappyd.lib not being found too.

I was going to ask about using `find_package(module CONFIG)`. In particular for TBB, I was wondering if there was a point in time when we could switch, perhaps when oneTBB becomes the minimum in VFX Ref Platform in CY2024? Or perhaps we try and switch sooner? Is it possible that we use it and fallback to using `FindTBB.cmake` if it doesn't work?

danrbailey (approved):

danrbailey:
> Thanks @Idclip, the disk space issue will be solved by this one - #1469. I've brought across your tbb fix that had me stumped for a bit and it seemed to address the issue there, I think I'm now missing your blosc fix so will bring that across to test as well. I ran into the snappyd.lib not being found too.
> 
> I was going to ask about using `find_package(module CONFIG)`. In particular for TBB, I was wondering if there was a point in time when we could switch, perhaps when oneTBB becomes the minimum in VFX Ref Platform in CY2024? Or perhaps we try and switch sooner? Is it possible that we use it and fallback to using `FindTBB.cmake` if it doesn't work (at least while we transition)?

I pulled across your Blosc changes and everything built fine on my MR, so these changes look good to me. 👍

Idclip:
> I was going to ask about using `find_package(module CONFIG)`. In particular for TBB, I was wondering if there was a point in time when we could switch, perhaps when oneTBB becomes the minimum in VFX Ref Platform in CY2024? Or perhaps we try and switch sooner? Is it possible that we use it and fallback to using `FindTBB.cmake` if it doesn't work (at least while we transition)?

@danrbailey lets discuss in TSC

PR #1482, opened 2022-09-30
Prune SDF value if there is a float max or float min.

apradhana:
Fix fast sweeping: prune SDF value if there is a voxel containing float max or float min.


openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -923,8 +930,12 @@ struct FastSweeping<SdfGridT, ExtValueT>::MinMaxKernel
         for (auto leafIter = r.begin(); leafIter; ++leafIter) {
             for (auto voxelIter = leafIter->beginValueOn(); voxelIter; ++voxelIter) {
                 const SdfValueT v = *voxelIter;
-                if (v < mMin) mMin = v;
-                if (v > mMax) mMax = v;
+                const SdfValueT fltMinDif = std::abs(v + std::numeric_limits<SdfValueT>::max());
+                const SdfValueT fltMaxDif = std::abs(v - std::numeric_limits<SdfValueT>::max());
+                if (v < mMin && fltMinDif >= 1.0e-4f) mMin = v;
+                if (v > mMax && fltMaxDif >= 1.0e-4f) mMax = v;
```
jmlait, line 936: I'm a bit confused by the 1e-4, I'd expect the fltFooDiff to always saturate to zero?

apradhana: You are right, the problem is not caused by a value that is near float-max. Rather, from a value that is untouched by the sweep algorithm when it was set to float-max originally. I addressed this issue in a new commit.

openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -930,12 +930,12 @@ struct FastSweeping<SdfGridT, ExtValueT>::MinMaxKernel
         for (auto leafIter = r.begin(); leafIter; ++leafIter) {
             for (auto voxelIter = leafIter->beginValueOn(); voxelIter; ++voxelIter) {
                 const SdfValueT v = *voxelIter;
-                const SdfValueT fltMinDif = std::abs(v + std::numeric_limits<SdfValueT>::max());
-                const SdfValueT fltMaxDif = std::abs(v - std::numeric_limits<SdfValueT>::max());
-                if (v < mMin && fltMinDif >= 1.0e-4f) mMin = v;
-                if (v > mMax && fltMaxDif >= 1.0e-4f) mMax = v;
-                if (fltMinDif < 1.0e-4f) mFltMinExists = true;
-                if (fltMaxDif < 1.0e-4f) mFltMaxExists = true;
+                const bool vEqFltMin = v == - std::numeric_limits<SdfValueT>::max();
```
richhones, line 933: std::numeric_limits<SdfValueT>::lowest()?

openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -933,11 +944,33 @@ struct FastSweeping<SdfGridT, ExtValueT>::MinMaxKernel
...
+template <typename SdfGridT, typename ExtValueT>
+struct FastSweeping<SdfGridT, ExtValueT>::PruneMinMaxFltKernel {
+
+  PruneMinMaxFltKernel(SdfValueT min, SdfValueT max) : mMin(min), mMax(max) {}
+
+  inline void operator()(const typename SdfGridT::ValueAllIter &iter) const {
+    if (std::abs(*iter + std::numeric_limits<SdfValueT>::max()) < 1.0e-4f) {
+      iter.setValue(mMin);
+    }
+    if (std::abs(*iter - std::numeric_limits<SdfValueT>::max()) < 1.0e-4f) {
```
richhones, line 966: these supposed to be equality too?

openvdb/openvdb/tools/FastSweeping.h
apradhana: I think it should be `-std::numeric_limits<SdfValueT>::max()` because this file uses the convention of defining min/max that way. See, e.g. [line 911](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/FastSweeping.h#L911) or [line 1104](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tools/FastSweeping.h#L1104).

openvdb/openvdb/tools/FastSweeping.h
apradhana: Thanks for catching this. I have fixed it in a new commit.

openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -934,11 +947,52 @@ struct FastSweeping<SdfGridT, ExtValueT>::MinMaxKernel
...
 };// FastSweeping::MinMaxKernel
 
+/// Private class of FastSweeping to prune sdf value that is equal to float max or
+/// float min.
+template <typename SdfGridT, typename ExtValueT>
+struct FastSweeping<SdfGridT, ExtValueT>::PruneMinMaxFltKernel {
+    PruneMinMaxFltKernel(SdfValueT min, SdfValueT max) : mMin(min), mMax(max) {}
+
+    // Do nothing for the root node
+    void operator()(typename SdfTreeT::RootNodeType&, size_t = 1) const { }
```
danrbailey, line 965: Doesn't this need to handle active tiles of the root node here?

jmlait (approved):

openvdb/openvdb/tools/FastSweeping.h
```diff
@@ -934,11 +947,61 @@ struct FastSweeping<SdfGridT, ExtValueT>::MinMaxKernel
...
 };// FastSweeping::MinMaxKernel
 
+/// Private class of FastSweeping to prune sdf value that is equal to float max or
+/// float min.
+template <typename SdfGridT, typename ExtValueT>
+struct FastSweeping<SdfGridT, ExtValueT>::PruneMinMaxFltKernel {
+    PruneMinMaxFltKernel(SdfValueT min, SdfValueT max) : mMin(min), mMax(max) {}
+
+    // Root node
+    void operator()(typename SdfTreeT::RootNodeType& node, size_t = 1) const {
```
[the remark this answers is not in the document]
apradhana: Thank you, Dan. I'll update it for v10.1, so that we don't overload the CI today.

PR #1483, opened 2022-10-05
Enable Houdini CI

danrbailey:
@Idclip - this is a resubmission of #1458 but with one small change on top.

The issue here is the ABI test which works by building a single binary against libopenvdb.so and libopenvdb_sesi.so and attempting to create VDB grids in one library then pass them to the other and vice-versa to detect any ABI issues. This is less rigorous than the various ABI checker tools around, but is intended to replicate the exact situation that we run into if there's an ABI issue.

The libopenvdb.so library builds against libboost_iostream.so.1.73 which in turn builds against the system liblzma.so.5.2.2 (RHEL/CentOS 7).
The libopenvdb_sesi.so library builds against libhboost_iostreams.so.1.72 which in turn builds against liblzma.so.5.2.3 which is shipped in the dsolib directory of Houdini.

The problem is that both liblzma.so.5.2.2 and liblzma.so.5.2.3 have the same soname (liblzma.so.5) which means regardless of which one you end up linking against in this ABI test, you are missing symbols that are provided in the other one.

The "proper" solution is to build boost against the same version of lzma as is used in Houdini or perhaps to add a suffix to lzma in one of them. However, I think the pragmatic solution here is to continue with making delayed loading optional (#823) and then to switch it off in these ABI tests because the presence of boost_iostreams is known to not affect ABI.

I should note that because the Houdini CI hasn't been run in a while, it is likely that this will fail until the Weekly build runs successfully (but this change needs to be merged for the cache to be shared with this branch).

(@e4lam)

### DIFF
--- a/.github/workflows/houdini.yml
+++ b/.github/workflows/houdini.yml
@@ -108,13 +108,11 @@ jobs:
         cd $HOME/houdini_install && tar -xzf hou.tar.gz && cd -
     - name: build
       shell: bash
-      run: >
-        cd $HOME/houdini_install/hou && source houdini_setup_bash && cd - &&
-        ./ci/build.sh -v
-        -j ${{ matrix.config.j }}
-        --build-type=${{ matrix.config.build }}
-        --components="${{ matrix.config.components }}"
-        --cargs=\"-DDISABLE_CMAKE_SEARCH_PATHS=ON -DOPENVDB_BUILD_HOUDINI_ABITESTS=ON -DOPENVDB_HOUDINI_INSTALL_PREFIX=/tmp -DDISABLE_DEPENDENCY_VERSION_CHECKS=${{ matrix.config.disable_checks }}\"
+      run: |
+        export HFS="$HOME/houdini_install/hou"
+        export HDSO="${HFS}/dsolib"
+        export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/lib64:${HDSO}"
+        ./ci/build.sh -v --build-type=Release --components="core,hou" --cargs=\"-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON -DDISABLE_CMAKE_SEARCH_PATHS=ON -DOPENVDB_BUILD_HOUDINI_ABITESTS=OFF -DOPENVDB_HOUDINI_INSTALL_PREFIX=/tmp\ -DTBB_INCLUDEDIR=/usr/local/include -DTBB_LIBRARYDIR=/usr/local/lib\"
     - name: test
       run: cd build && ctest -V
     # Keep ccache light by stripping out any caches not accessed in the last day
--- a/.github/workflows/houdini.yml
+++ b/.github/workflows/houdini.yml
@@ -56,10 +56,9 @@ jobs:
 
   linux-vfx-houdini:
     needs: [checksecret]
-#    if: >
-#      ${{ needs.checksecret.outputs.HOUDINI_SECRETS == 'true' ||
-#          github.repository_owner == 'AcademySoftwareFoundation' }}
-    if: ${{ false }}  # disable for now
+    if: >
+      ${{ needs.checksecret.outputs.HOUDINI_SECRETS == 'true' ||
+          github.repository_owner == 'AcademySoftwareFoundation' }}
     runs-on: ubuntu-20.04-8c-32g-300h
     name: hou:${{ matrix.config.hou }}-vfx:${{ matrix.config.image }}-cxx:${{ matrix.config.cxx }}
     container:
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -44,8 +44,8 @@ jobs:
         run: echo "HOUDINI_CLIENT_ID and HOUDINI_SECRET_KEY GitHub Action Secrets needs to be set to install Houdini builds"
 
   # download the latest production version of Houdini X, strip out headers,
-  # libraries and binaries required for building OpenVDB and if the build
-  # succeeds, put it into the GitHub Actions cache
+  # libraries and binaries required for building OpenVDB and put it into
+  # the GitHub Actions cache
   linux_houdini:
     needs: [checksecret]
     if: |
@@ -81,16 +81,6 @@ jobs:
         mkdir $HOME/houdini_install
         cp hou/hou.tar.gz $HOME/houdini_install/hou.tar.gz
         cd $HOME/houdini_install && tar -xzf hou.tar.gz && cd -
-    - name: build
-      shell: bash
-      run: >
-        cd $HOME/houdini_install/hou && source houdini_setup_bash && cd - &&
-        ./ci/build.sh -v
-        --build-type=Release
-        --components="core,hou,bin,view,render,python,test,axcore,axbin,axtest"
-        --cargs=\"-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON -DDISABLE_CMAKE_SEARCH_PATHS=ON -DOPENVDB_BUILD_HOUDINI_ABITESTS=ON -DOPENVDB_HOUDINI_INSTALL_PREFIX=/tmp\"
-    - name: test
-      run: cd build && ctest -V
     - name: write_houdini_cache
       uses: actions/cache@v2
       with:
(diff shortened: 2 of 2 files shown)


Idclip:
@danrbailey thanks for this. A tad confused, are you implying only the houdini abi builds were failing? From my old logs I see all executables failing to link; for example I was iterating on linking `vdb_print`.

danrbailey:
> @danrbailey thanks for this. A tad confused, are you implying only the houdini abi builds were failing? From my old logs I see all executables failing to link; for example I was iterating on linking `vdb_print`.
> 
> old logs: https://github.com/Idclip/openvdb/actions/workflows/houdini.yml?query=branch%3Afeature%2Fvdb10 example of vdb_print failing: https://github.com/Idclip/openvdb/actions/runs/2817025934/jobs/4447593039
> 
> ```
> [ 41%] Linking CXX executable vdb_print
> cd /__w/openvdb/openvdb/build/openvdb_cmd/vdb_print && /usr/local/bin/cmake -E cmake_link_script CMakeFiles/vdb_print.dir/link.txt --verbose=1
> /usr/local/bin/clang++ -O3 -DNDEBUG CMakeFiles/vdb_print.dir/main.cc.o -o vdb_print   -L/github/home/houdini_install/hou/dsolib  -Wl,-rpath,/github/home/houdini_install/hou/dsolib:/__w/openvdb/openvdb/build/openvdb/openvdb: ../../openvdb/openvdb/libopenvdb.so.10.0.0 /github/home/houdini_install/hou/dsolib/libjemalloc.so /github/home/houdini_install/hou/dsolib/libtbb.so.2 /github/home/houdini_install/hou/dsolib/libz.so -pthread /usr/local/lib/libboost_iostreams.so.1.73.0 
> /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: /usr/local/lib/libboost_iostreams.so.1.73.0: undefined reference to `lzma_cputhreads@XZ_5.2.2'
> /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: /usr/local/lib/libboost_iostreams.so.1.73.0: undefined reference to `lzma_stream_encoder_mt@XZ_5.1.2alpha'
> clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
> gmake[2]: *** [openvdb_cmd/vdb_print/CMakeFiles/vdb_print.dir/build.make:111: openvdb_cmd/vdb_print/vdb_print] Error 1
> gmake[2]: Leaving directory '/__w/openvdb/openvdb/build'
> gmake[1]: *** [CMakeFiles/Makefile2:670: openvdb_cmd/vdb_print/CMakeFiles/vdb_print.dir/all] Error 2
> gmake[1]: *** Waiting for unfinished jobs....
> [ 41%] Linking CXX executable vdb_lod
> cd /__w/openvdb/openvdb/build/openvdb_cmd/vdb_lod && /usr/local/bin/cmake -E cmake_link_script CMakeFiles/vdb_lod.dir/link.txt --verbose=1
> /usr/local/bin/clang++ -O3 -DNDEBUG CMakeFiles/vdb_lod.dir/main.cc.o -o vdb_lod   -L/github/home/houdini_install/hou/dsolib  -Wl,-rpath,/github/home/houdini_install/hou/dsolib:/__w/openvdb/openvdb/build/openvdb/openvdb: ../../openvdb/openvdb/libopenvdb.so.10.0.0 /github/home/houdini_install/hou/dsolib/libjemalloc.so /github/home/houdini_install/hou/dsolib/libtbb.so.2 /github/home/houdini_install/hou/dsolib/libz.so -pthread /usr/local/lib/libboost_iostreams.so.1.73.0 
> /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: /usr/local/lib/libboost_iostreams.so.1.73.0: undefined reference to `lzma_cputhreads@XZ_5.2.2'
> /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: /usr/local/lib/libboost_iostreams.so.1.73.0: undefined reference to `lzma_stream_encoder_mt@XZ_5.1.2alpha'
> clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> 
> Note that I also tested removing the explicit linker directories which did not help

Thanks for pointing that out, I had not realized that the binaries were also failing. It is the same issue - linking against two different versions of boost which link against two different versions of lzma but with the same soname. I will remove the binaries from these houdini builds for the time being and when we have a version where we can build openvdb without boost::iostreams, we can re-enable everything again. We can of course fix this at source - by rebuilding boost in the container to match the version used in Houdini, but that's always going to be a moving target based on which version SideFX ship with, so I don't really see the return on investment of going doing that particular rabbit hole.

In the longer term, the better option is a request to SideFX (@jmlait, @e4lam) - to namespace and add an soname suffix to the liblzma.so that ships with Houdini so that we can link against two different versions concurrently. That's really the root of the issue here - it's great to have libhboost namespaced and suffixed, but we also need all the libraries that hboost is dependent on and shipped with Houdini to be namespaced and suffixed too.

Idclip:
> Thanks for pointing that out, I had not realized that the binaries were also failing. It is the same issue - linking against two different versions of boost which link against two different versions of lzma but with the same soname

We don't link against two different versions of boost though, right? For the core library and vdb binaries we only use three parts of Houdini - `tbb`, `blosc` and `zlib` (and `jemalloc` for binaries). We should be picking up the system boost and there should be no leakage of hboost symbols into VDB (any that are are namespaced anyway) so there should be no way that either of these components are reliant on lzma from Houdini... at least that's my understanding. Take the above job for instance:

VDB link command (object files removed):

```
/usr/local/bin/clang++ -fPIC -O3 -DNDEBUG -shared -Wl,-soname,libopenvdb.so.10.0 -o libopenvdb.so.10.0.0 -L/github/home/houdini_install/hou/dsolib -Wl,-rpath,/github/home/houdini_install/hou/dsolib::::::::::::::::: /github/home/houdini_install/hou/dsolib/libtbb.so.2 /github/home/houdini_install/hou/dsolib/libz.so /usr/local/lib/libboost_iostreams.so.1.73.0 -pthread
```

vdb_print link command:

```
/usr/local/bin/clang++ -O3 -DNDEBUG -o vdb_print -L/github/home/houdini_install/hou/dsolib -Wl,-rpath,/github/home/houdini_install/hou/dsolib:/__w/openvdb/openvdb/build/openvdb/openvdb: ../../openvdb/openvdb/libopenvdb.so.10.0.0 /github/home/houdini_install/hou/dsolib/libjemalloc.so /github/home/houdini_install/hou/dsolib/libtbb.so.2 /github/home/houdini_install/hou/dsolib/libz.so -pthread /usr/local/lib/libboost_iostreams.so.1.73.0 
```

I dont see where we're picking up both lzma's. I thought it could be the explicit linker dirs `-L` to `dsolib` but I removed these and had the same issue. My last through was that we're picking up Houdini headers which are overriding lmza symbols names in some way but I don't see how this is possible either (we don't read from `hboost`). What have I missed?

danrbailey:
> 

I see what you mean and I agree - removing the dsolib linker path should have fixed the issue. Did you try setting LD_DEBUG?

Idclip:
> > 
> 
> I see what you mean and I agree - removing the dsolib linker path should have fixed the issue. Did you try setting LD_DEBUG?

er, during linking? I did not - I usually only use that for checking runtime loading of libraries during execution, not sure how useful it will be in this instance? What I ideally need is a version of the runner/image I can ssh into and look around (so we can ldd/nm the .so's) but I do not have `gh act`/`docker` setup

Idclip:
Also just for completeness, see this comment to disable the `-L` directories (if you're testing that too):

https://github.com/AcademySoftwareFoundation/openvdb/pull/1445#issuecomment-1217178621

danrbailey:
> Also just for completeness, see this comment to disable the `-L` directories (if you're testing that too):
> 
> [#1445 (comment)](https://github.com/AcademySoftwareFoundation/openvdb/pull/1445#issuecomment-1217178621)

Ok, I'll take a look. Also can you approve this one (#1469) so we can use the faster builds? This PR also includes the fix that is causing recent Windows builds to fail:

https://github.com/AcademySoftwareFoundation/openvdb/pull/1469/files#diff-c7f73fa00a0035a901886287a05980046c7c9c84f8a244bffb7a8895d4b9ce76

PR #1491, opened 2022-10-13
More ABI-related cleanup

danrbailey:


jmlait (approved):

openvdb/openvdb/math/Maps.h
```diff
@@ -293,14 +293,9 @@ class OPENVDB_API MapRegistry
...
 
-/// @note Macro to use a final specifier from ABI=8 onwards.
-#define OPENVDB_MAP_CLASS_SPECIFIER final
-#define OPENVDB_MAP_FUNC_SPECIFIER final
-
 /// @brief A general linear transform using homogeneous coordinates to perform
 /// rotation, scaling, shear and translation
-/// @note This class is marked final with ABI=8
-class OPENVDB_API AffineMap OPENVDB_MAP_CLASS_SPECIFIER: public MapBase
+class OPENVDB_API AffineMap final: public MapBase
```
kmuseth, line 298: why is it important to prevent inheriting from AffineMap?

kmuseth (approved):
seems fine but please see comment

kmuseth (commented):
Actually, I  question the wisdom of removing the NodeUnion implementation that supports value types that are not trivially copyable. Just because std::string doesn't supported anymore doesn't means someone isn't using this implementation for another value type that as a non-trivial copy constructor

openvdb/openvdb/math/Maps.h
danrbailey: Performance. By preventing inheritance, you can ensure that when you have an AffineMap object, calling one of the methods on it can be de-virtualized which eliminates the virtual function cost and allows for inlining. 

This feature has been enabled for the last couple of years incidentally - this is just tidying up and removing the macro.

danrbailey:
> Actually, I question the wisdom of removing the NodeUnion implementation that supports value types that are not trivially copyable. Just because std::string doesn't supported anymore doesn't means someone isn't using this implementation for another value type that as a non-trivial copy constructor

There is a new implementation that @Idclip wrote which does support value types that are not trivially copied. This has been in use for the last two years, this change is just removing the old one. Using VDB grids with std::string is still supported - it's just not going to be registered by default going forward.

PR #1492, opened 2022-10-13
Remove string grids

danrbailey:
This was deprecated in VDB 9.0.


jmlait (approved):

Idclip:
I'd like to keep the AX string test for now - just replace the type with a manual instantiation and that should be fine

Idclip (approved):

kmuseth (approved):
Looks good to me

PR #1493, opened 2022-10-13
Remove all functionality deprecated prior to VDB 9.0

danrbailey:
This includes:
    * FindActiveValues::any() / FindActiveValues::none()
    * tools::dilateVoxels() / tools::erodeVoxels()
    * All visitor methods throughout Tree hierarchy - visit(), visitActiveBBox(), visit2(), visit2Node()
    * LeafManager::getNodes()
    * PagedArray::push_back()


jmlait (approved):
Sigh, guess who has a lot of visit() calls still with OPENVDB_NO_DEPRECATION_WARNING around them?

Can't say I haven't been warned, and Dynamic Node Manager has proved itself in the meantime.

jmlait (commented):
There's no pendingchanges?  We should add a pending changes listing the APIs that are now gone (vist, visitActive, etc)

danrbailey:
> Sigh, guess who has a lot of visit() calls still with OPENVDB_NO_DEPRECATION_WARNING around them?
> 
> Can't say I haven't been warned, and Dynamic Node Manager has proved itself in the meantime.

Sorry about that! :/ I did come across this old comment from our meeting notes while I was scanning for things to remove that seems relevant here:

> Consensus that we should be a bit more aggressive with deprecations, users often ignore deprecation warnings anyway. Still nice to give users as much warning as possible. New major version is a good time to drop API support and force
> users to change their code.

Do let me know if you run into any further edge cases we haven't come across yet though...

danrbailey:
> There's no pendingchanges? We should add a pending changes listing the APIs that are now gone (vist, visitActive, etc)

Added now, thanks!

jmlait:
> > Sigh, guess who has a lot of visit() calls still with OPENVDB_NO_DEPRECATION_WARNING around them?
> > Can't say I haven't been warned, and Dynamic Node Manager has proved itself in the meantime.
> 
> Sorry about that! :/ I did come across this old comment from our meeting notes while I was scanning for things to remove that seems relevant here:
> 
> > Consensus that we should be a bit more aggressive with deprecations, users often ignore deprecation warnings anyway. Still nice to give users as much warning as possible. New major version is a good time to drop API support and force
> > users to change their code.

Yep, and I stand by that consensus.  We need to remove at some point, and this is a good time to do the forcing as we've had the deprecation in for a while.  I really should have addressed these in the past year :>

openvdb/openvdb/tree/LeafManager.h
```diff
@@ -535,42 +535,6 @@ class LeafManager
...
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_BEGIN
-        if (std::is_same<T, const LeafType*>::value) {
-            array.resize(mLeafCount);
-            for (size_t i=0; i<mLeafCount; ++i) array[i] = reinterpret_cast<T>(mLeafs[i]);
-        } else {
-            mTree->getNodes(array);
-        }
-        OPENVDB_NO_UNREACHABLE_CODE_WARNING_END
-    }
-
```
kmuseth, line 573: what are these methods removed? Tree::getNodes() is clearly slower since they always involve tree-traversals where these methods make use the existing linear caching of nodes

kmuseth (commented):
Please see comments

openvdb/openvdb/tree/LeafManager.h
danrbailey: The simple answer is that they're being removed because they have been marked as deprecated for over two years now. :)

The longer answer is this - you are right that `Tree::getNodes()` is slightly slower for the specific use case of populating an array of LeafNode pointers from an existing LeafManager, but I hope you'll agree that this is a rare use case. In fact, none of the tools in the OpenVDB core library use this method nor any of the auxiliary plugins. In addition, the previous rationale for using this method has also now gone, because it used to be recommended to build a NodeManager from a LeafManager for improved performance. Now, construction of both LeafManagers and NodeManagers have been threaded such that it is faster to build them from the Tree directly. Finally, there is no `getNodes()` method on the NodeManager, so this method is inconsistent, unused and as I said marked deprecated.

openvdb/openvdb/tree/InternalNode.h
```diff
@@ -2886,173 +2856,6 @@ InternalNode<ChildT, Log2Dim>::combine2(const InternalNode& other, const OtherVa
...
-
-        if (child != nullptr) child->visit2(otherIter, op, otherIsLHS);
-    }
-}
-
-
-////////////////////////////////////////
-
-
 template<typename ChildT, Index Log2Dim>
```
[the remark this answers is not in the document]
danrbailey: They have been replaced by the DynamicNodeManager. This class offers all the same functionality and benefits as the visitor classes, but is significantly more performant and doesn't pollute the Tree hierarchy with lots of additional class members that slows down compile times. We have worked through all the edge cases and concerns that have been raised since we introduced this class more than two years ago (const trees, grain sizes, sequential visitor patterns -addressed in adding `tools::visitNodesDepthFirst`) and lots of our existing tools have long since been rewritten or adapted to make use of the DynamicNodeManager (this includes activate, merge, count, evalMinMax, CSG, fast sweeping, etc) so the visitor classes have been deprecated and unused since at least v9.0. The decision was taken a while ago to drop these visitor methods on the next major API bump which is v10.0.

PR #1510, opened 2022-10-28
Use modern GA_Detail::createDetached*Group() API

e4lam:
... instead of the unsafe newDetachedPrimitiveGroup() that will be deprecated in a future Houdini version.


jmlait (approved):
UPtr was added in 18.5, which was ABI 7.   Our protocol is 3 simultaneous ABI, so 8, 9, 10 now, so this is definitely in the window.

danrbailey (approved):

PR #1513, opened 2022-11-03
Fix terse static assert error when building with C++14 on Windows

brechtvl:
USD defaults to C++14, which leads to a build error when using OpenVDB 10 because terse static asserts are a C++17 feature.

Signed-off-by: Brecht Van Lommel <brecht@blender.org>

### DIFF
--- a/openvdb/openvdb/tree/ValueAccessor.h
+++ b/openvdb/openvdb/tree/ValueAccessor.h
@@ -872,7 +872,8 @@ public:
     using LeafNodeType = typename NodeType::LeafNodeType;
     using CoordLimits = std::numeric_limits<Int32>;
 
-    static_assert(std::is_same<NodeType, LeafNodeType>::value);
+    static_assert(std::is_same<NodeType, LeafNodeType>::value,
+        "cache item node type is not leaf node type");
 
     CacheItem(TreeCacheT& parent)
         : mParent(&parent)


Idclip:
Hi @brechtvl thanks, but as of VDB 10 we only support C++17 as our minimum standard. I realise that this patch might be all that is needed currently to get VDB working with C++14, but going forward towards VDB 10.X.X we'll be using C++17 feature more regularly. If you need C++14 support I suggest using VDB 9.X. You can also patch it on your end but that will become a progressively less sustainable solution

brechtvl:
Fair enough. We can of course patch it, but I imagine others will run into the same issue. I'll try to get it fixed on the USD side then.

MehcanicalEnjiny:
@brechtvl Brecht, stop making lies about real  Engineering. 
You are not an engineer, never were, and never will be close. Engineering has nothing to do with sitting behind keyboard and writing code. There is no such thing as "software" "engineer". That is the stupidest, most pathetic & disrespectful lie ever. 
Stop being disrespectful to real engineering (mechanical engineering is the ONLY engineering).
Engineering is a real subject.
Engineering means designing and building real mechanical systems, like vehicular engines and turbojets. 
Software dev is easy and has absolutely nothing to do with engineering, you are just being stupid liars. 
You daft software kiddos will never be engineers, not even close. 
Mechanical Engineering is the ONLY engineering. Period.

PR #1522, opened 2022-11-16
Switch to using the static asserts.

jmlait:
The comment mentions that they have compiler warnings, but with clang 15 we now get compiler warnings for trying to do differences with nullptr.


jmlait:
I'd be all for deleting the other case, but I'm not sure what triggered this, we may need it around for people in that platform?

jmlait:
Discussion at meeting is to go through with this; so when we discover what triggered the other case we can restore it.  Since NanoVDB is going C++17 that other case hopefully doesn't exist so this may eventually be removed.

Idclip (approved):
Approving and merging based on TSC discussion

PR #1523, opened 2022-11-17
Fix uninitialized pointflags

jmlait:
The pointflags in VolumeToMesh were unintialized if we took one of the computeCellPoints branches as we weren't testing the size of the weight array (as it had turned into a std::array)

### DIFF
--- a/openvdb/openvdb/tools/VolumeToMesh.h
+++ b/openvdb/openvdb/tools/VolumeToMesh.h
@@ -1614,10 +1614,11 @@ ComputePoints<InputTreeType>::operator()(const tbb::blocked_range<size_t>& range
             if (inclusiveCell) getCellVertexValues(*inputNode, offset, values);
             else               getCellVertexValues(inputAcc, ijk, values);
 
-            size_t count;
+            size_t count, weightcount;
 
             if (refSigns == 0) {
                 count = computeCellPoints(points, values, signs, iso);
+                weightcount = 0;
             } else {
                 if (inclusiveCell && refInputNode) {
                     getCellVertexValues(*refInputNode, offset, refValues);
@@ -1626,6 +1627,7 @@ ComputePoints<InputTreeType>::operator()(const tbb::blocked_range<size_t>& range
                 }
                 count = computeCellPoints(points, weightedPointMask, values, refValues, signs, refSigns,
                     iso, refPointIndexNode->getValue(offset), mQuantizedSeamLinePoints);
+                weightcount = count;
             }
 
             xyz = ijk.asVec3d();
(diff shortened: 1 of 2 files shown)


openvdb/openvdb/tools/VolumeToMesh.h
```diff
@@ -1614,10 +1614,11 @@ ComputePoints<InputTreeType>::operator()(const tbb::blocked_range<size_t>& range
             if (inclusiveCell) getCellVertexValues(*inputNode, offset, values);
             else               getCellVertexValues(inputAcc, ijk, values);
 
-            size_t count;
+            size_t count, weightcount;
 
             if (refSigns == 0) {
                 count = computeCellPoints(points, values, signs, iso);
+                weightcount = 0;
```
Idclip, line 1621: Minor: maybe instead just do `weightedPointMask.fill(false)` ? as the `weightcount = count` in the second branch is confusing (essentially all we want to know is if `weightedPointMask` is initialized).

pendingchanges/fixseampoint.txt
```diff
@@ -0,0 +1,2 @@
+Bug Fixes:
+- Fix uninitialized point flags in tools::VolumeToMesh
```
Idclip, line 2: Did you get a compiler warning with this or have you seen UB in prod? We should probably do a patch release if so

openvdb/openvdb/tools/VolumeToMesh.h
jmlait: My logic was to mimic the original code - ie, we have the idea that weightcount is actually a vector that is being filled in. 

I would claim count itself is very confusing in this code - I much prefered the old code with the count tied to the vectors....  Internally we have UT_SmallArray that allows stack-based vectors with a size....

pendingchanges/fixseampoint.txt
jmlait: UB in prod.  The Convert VDB SOP randomly initializes seampoints and showed up in our reg tests.

A patch is probably a good idea, maybe wait a bit to see if any other shows up?

openvdb/openvdb/tools/VolumeToMesh.h
Idclip: It made more sense when it was a vector yeah, a smallarray impl would be good but probably overkill even here when the max size is alwasy 4. Anyway this was a minor thing

pendingchanges/fixseampoint.txt
Idclip: Eh, it's clearly UB and if the general case exhibits it I would advocate for a `10.0.1` release now.

Idclip (approved):

danrbailey (approved):

PR #1525, opened 2022-11-18
Support for hollow VDBs

jmlait:
From Tomas Skrivan comes this PR for providing hollow vdb support.

The volume-to-mesh gains an optional oracle to determine sign.   Rebuild level set uses this to ensure the rebuilt levelset always matches sign with the input.

From Polygons SOP gains an option to use the winding number to determine sign, enabled in H20 where the winding number is exposed.

### DIFF
--- a/openvdb/openvdb/tools/LevelSetRebuild.h
+++ b/openvdb/openvdb/tools/LevelSetRebuild.h
@@ -8,6 +8,9 @@
 #include <openvdb/Exceptions.h>
 #include <openvdb/math/Math.h>
 #include <openvdb/math/Transform.h>
+#include <openvdb/tools/VolumeToMesh.h>
+#include <openvdb/tools/MeshToVolume.h>
+#include <openvdb/tools/Interpolation.h>
 #include <openvdb/util/NullInterrupter.h>
 #include <openvdb/util/Util.h>
 #include <openvdb/openvdb.h>
@@ -250,13 +253,26 @@ doLevelSetRebuild(const GridType& grid, typename GridType::ValueType iso,
 
     QuadAndTriangleDataAdapter<Vec3s, Vec4I> mesh(points, primitives);
 
+    auto backToOldGrid = [&xform, &grid](const Coord& coord) -> openvdb::math::Vec3d {
+        return grid.transform().worldToIndex(xform->indexToWorld(coord));
+    };
+
+    auto interiorTest = [acc = grid.getConstAccessor(), &backToOldGrid, &xform](const Coord& coord) -> bool {
+        if (xform == nullptr) {
+            return acc.getValue(coord) <= 0 ? true : false;
+        } else {
+            float value = openvdb::tools::BoxSampler::sample(acc, backToOldGrid(coord));
+            return value <= 0 ? true : false;
+        }
+    };
+
     if (interrupter) {
         return meshToVolume<GridType>(*interrupter, mesh, *transform, exBandWidth, inBandWidth,
-            DISABLE_RENORMALIZATION, nullptr);
+            DISABLE_RENORMALIZATION, nullptr, interiorTest, EVAL_EVERY_VOXEL);
     }
 
     return meshToVolume<GridType>(mesh, *transform, exBandWidth, inBandWidth,
-        DISABLE_RENORMALIZATION, nullptr);
+        DISABLE_RENORMALIZATION, nullptr, interiorTest, EVAL_EVERY_VOXEL);
 }
 
 
--- a/openvdb/openvdb/tools/LevelSetRebuild.h
+++ b/openvdb/openvdb/tools/LevelSetRebuild.h
@@ -4,13 +4,25 @@
 #ifndef OPENVDB_TOOLS_LEVELSETREBUILD_HAS_BEEN_INCLUDED
 #define OPENVDB_TOOLS_LEVELSETREBUILD_HAS_BEEN_INCLUDED
 
+/// When rebuilding we convert from grid -> mesh -> grid.
+/// The conversion from mesh -> grid will close any internal bubbles.
+/// By using the original grid as an oracle we can set the sign of the
+/// bubbles properly.  Unfortunately, when increasing resolution the 
+/// rasterization of the mesh diverges too far from interpolated original
+/// grid and results in incorrect sign choices.   The likely correct solution
+/// is to use a different approach for rebuilding when resolution is
+/// increasing.
+#define OPENVDB_USE_ORACLE_IN_REBUILD 0
+
 #include <openvdb/Grid.h>
 #include <openvdb/Exceptions.h>
 #include <openvdb/math/Math.h>
 #include <openvdb/math/Transform.h>
+#if OPENVDB_USE_ORACLE_IN_REBUILD
 #include <openvdb/tools/VolumeToMesh.h>
 #include <openvdb/tools/MeshToVolume.h>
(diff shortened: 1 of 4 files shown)


jmlait:
I don't like the unused parameter "warning" in the best of days, but in this case the problem is it is unused because we have a if constexpr which is causing it to not be used!

error: parameter 'interiorTest' set but not used [-Werror=unused-but-set-parameter]

jmlait:
Current failures seem to be explicit instantiation, which makes sense as I think Tomas added a template for the oracle....  Ideas of how to resolve this are appreciated.

danrbailey:
> Current failures seem to be explicit instantiation, which makes sense as I think Tomas added a template for the oracle.... Ideas of how to resolve this are appreciated.

The simplest resolution to this is to provide explicit values for the new arguments (ie nullptr, etc) here:

https://github.com/AcademySoftwareFoundation/openvdb/pull/1525/files#diff-cb3368eb16b1ce350d0e14c29b73757572b90b22b8e390f6ec2038d60a087d47R4449

That will fix the failures. However, by adding this new oracle, any time you provide a custom oracle, this pretty heavy method will need to be instantiated. A better solution might be to split this method up to call a few sub-methods where we could explicitly instantiate all but the oracle sub-method to minimize the impact on compile-times. I'm not sure how worth the effort that is, however this tool is one of the largest at ~4500 lines so it might still be worth doing this.

Do you have any unit tests you could add to this PR? Given how extensively used this tool is and how subtle some of the bugs in this area can be, it would be good to have some strong unit tests here.

jmlait:
Yeah, the implications on explicit instantiation worry me.  I'd like some feedback on the right way forward.

Note in particular this change has affected RebuildLevelSet.  This does change behaviour even without hollow meshes - I think the new behaviour is better, but we still have to analyse the differences carefully.

jmlait:
I've taking out the change to rebuildLevelSet as that doesn't work for resolution increases.  I'd still like the oracle support, however, as it can at least fix the mesh case.

I've made it explicitly instantiate the std::null_ptr path, which I believe means any use of the oracle won't be explicitly instantiated.  This is okay, I think, as if a user has their own oracle they can't use an instantiation anyways as we won't have their oracle.  And in internal oracles, like if rebuildLevelSet were go through, they'll be instantiated when/if the containing rebuildLevelSet is explicitly instantiated.

openvdb/openvdb/tools/LevelSetRebuild.h
```diff
@@ -250,13 +265,36 @@ doLevelSetRebuild(const GridType& grid, typename GridType::ValueType iso,
 
     QuadAndTriangleDataAdapter<Vec3s, Vec4I> mesh(points, primitives);
 
+#if OPENVDB_USE_ORACLE_IN_REBUILD
+    auto backToOldGrid = [&xform, &grid](const Coord& coord) -> openvdb::math::Vec3d {
+        return grid.transform().worldToIndex(xform->indexToWorld(coord));
+    };
+
+    auto interiorTest = [acc = grid.getConstAccessor(), &backToOldGrid, &xform](const Coord& coord) -> bool {
```
kmuseth, line 273: This looks sub-optimal to me; a ValueAccessor is created, which is only used once. Can this be improved or is it not a performance concern? Or is the sampler using it for all the stencil points?

openvdb/openvdb/tools/MeshToVolume.h
```diff
@@ -3116,10 +3139,168 @@ traceExteriorBoundaries(FloatTreeT& tree)
...
+
+    const auto DIM = leafNode.DIM;
+    const auto SIZE = leafNode.SIZE;
+
+    std::vector<VoxelState> voxelState(SIZE, NOT_VISITED);
+
+    std::vector<std::pair<Index, VoxelState>> offsetStack;
+    offsetStack.reserve(SIZE);
+
+    for (Index offset=0; offset<SIZE; offset++) {
```
kmuseth, line 3163: I would have liked to see a short paragraph describing the high-level idea behind this new flood-filling algorithm

kmuseth (approved):
Looks good, but I added two very minor comments

openvdb/openvdb/tools/LevelSetRebuild.h
jmlait: If I understand properly, the 
[acc = grid.getConstAccessor, ... ]
will add a new by-value const accessor to the capture of the interiorTest lambda.  This is done at the time of creating this lambda, so this is not created on each call to the lambda, but instead on the creation.
Likewise, the interiorTest doesn't have to be thread safe so long as it is copy-constructable and copy-constructed versions of it can run in separate threads.   So we should be doing copy-construction of the accessor before going to separate threads.

openvdb/openvdb/tools/MeshToVolume.h
jmlait: Additional comments have been added!

openvdb/openvdb/tools/MeshToVolume.h
```diff
@@ -3118,8 +3141,178 @@ traceExteriorBoundaries(FloatTreeT& tree)
...
+        const auto value = leafNode.getValue(offset);
+
+        // We do not assign anything for voxel close to the mesh
+        // This condition is aligned with the condition in traceVoxelLine
+        if (abs(value) <= 0.75) {
```
apradhana, line 3180: std::abs(value). But I'm also curious, is there a significance of the constant 0.75 here?

openvdb/openvdb/tools/MeshToVolume.h
```diff
@@ -3118,8 +3141,178 @@ traceExteriorBoundaries(FloatTreeT& tree)
...
+    }
+}
+
+////////////////////////////////////////
+
+/// @brief Sets the sign of voxel values of `tree` based on the `interiorTest`
+///
+/// Inside is set to positive and outside to negative. This is in reverse to the usual
+/// level set convention, but `meshToVolume` uses the opposite convention at certain point.
+///
```
apradhana, line 3236: Thank you for this comment :)

openvdb/openvdb/tools/MeshToVolume.h
jmlait: 0.75 is a magic constant from traceVoxelLine.
                // Boundary voxel check. (Voxel that intersects the surface)
                if (!(dist > ValueType(0.75))) isOutside = false;

Idclip: This has caught me out in the past too - we should really constexpr inline this constant in `MeshToVolume.h` and use a variable for this

apradhana (approved):

openvdb/openvdb/tools/MeshToVolume.h
```diff
@@ -3118,8 +3141,178 @@ traceExteriorBoundaries(FloatTreeT& tree)
...
+        NEGATIVE = 2,
+        NOT_ASSIGNED = 3
+    };
+
+    const auto DIM = leafNode.DIM;
+    const auto SIZE = leafNode.SIZE;
+
+    std::vector<VoxelState> voxelState(SIZE, NOT_VISITED);
+
+    std::vector<std::pair<Index, VoxelState>> offsetStack;
```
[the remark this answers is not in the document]
jmlait: That's a good point, std::array could be used.  I'm a bit nervous about making that change at this point so put it in the future enhancement category and try to include it in future workflows.

PR #1529, opened 2022-11-28
Support for IlmBase versions < 3.1 is deprecated and will be removed

malaterre:


malaterre:
@apradhana any comment ?

PR #1530, opened 2022-11-28
Remove the explicit default assignment operator.

jmlait:
Remove the explicit default assignment operator - if this is specified, so should the copy constructor and move operators, as per rule of five.


kmuseth (changes requested):
In that case line 166 should also be removed

kmuseth (approved):
looks consistent now

PR #1534, opened 2022-11-30
Add missing separate_arguments cmake call

malaterre:
This is documented at:

```
include_directories(${LLVM_INCLUDE_DIRS})
separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
add_definitions(${LLVM_DEFINITIONS_LIST})
```

* https://github.com/llvm/llvm-project/blob/main/llvm/docs/CMake.rst#embedding-llvm-in-your-project

See also discussion at:

* https://discourse.cmake.org/t/how-to-use-target-compile-definitions/6982


Idclip:
Thanks @malaterre - so should we be using the `NATIVE_COMMAND` invocation of `separate_arguments`? What was the issue you experienced with the current implementation?

malaterre:
@Idclip I am not sure I understand your comment. Instead could you copy/paste the sentence that is not clear from my comment above:

* https://github.com/AcademySoftwareFoundation/openvdb/pull/1534#issue-1469377059

thanks

PR #1545, opened 2022-12-11
Fixed forked repo action runners

Idclip:

### DIFF
--- a/.github/workflows/ax.yml
+++ b/.github/workflows/ax.yml
@@ -61,11 +61,11 @@ jobs:
       matrix:
         config:
           # Unified
-          - { image: '2021-clang10', cxx: 'clang++', build: 'Release', j: '8', components: 'core,python,bin,axcore,axbin,axtest' }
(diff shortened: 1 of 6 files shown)


danrbailey (approved):
This is great, thanks for looking into this Nick. Should we perhaps bump the weekly CI builds down to use the free runners unless they're run manually? The regular ones are not as performance-critical, but if there is memory reasons for continuing to want to use the bigger machines, that obviously trumps the core counts.

Idclip:
> This is great, thanks for looking into this Nick. Should we perhaps bump the weekly CI builds down to use the free runners unless they're run manually? The regular ones are not as performance-critical, but if there is memory reasons for continuing to want to use the bigger machines, that obviously trumps the core counts.

Yeah there are still memory reasons - even with this change allowing the jobs to run on forks, some of the windows/debug/ax jobs still intermittently fail due to disk/memory limits. Still, it's better than nothing, and allows us to at least test most of the build on forks (and also indirectly keep tabs on our compilation memory consumption!)

FWIW, here's where I found the solution: https://github.com/actions/runner/issues/409

PR #1550, opened 2022-12-15
Fix NanoVDB build errors with HIP on Linux

brechtvl:
For reference:
```
openvdb/include/nanovdb/NanoVDB.h:1880:12: error: call to '__ffsll' is ambiguous
    return __ffsll(v);
           ^~~~~~~
/opt/rocm-5.3.2/include/hip/amd_detail/amd_device_functions.h:69:39: note: candidate function
__device__ static inline unsigned int __ffsll(unsigned long long int input) {
                                      ^
/opt/rocm-5.3.2/include/hip/amd_detail/amd_device_functions.h:77:39: note: candidate function
__device__ static inline unsigned int __ffsll(long long int input) {
                                      ^
openvdb/include/nanovdb/NanoVDB.h:5567:61: warning: returning reference to local temporary object [-Wreturn-stack-address]
    __hostdev__ const uint64_t& valueCount() const { return mGrid.valueCount(); }
                                                            ^~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/nanovdb/nanovdb/NanoVDB.h
+++ b/nanovdb/nanovdb/NanoVDB.h
@@ -1877,7 +1877,7 @@ __hostdev__ static inline uint32_t FindLowestOn(uint64_t v)
 {
     NANOVDB_ASSERT(v);
 #if (defined(__CUDA_ARCH__) || defined(__HIP__)) && defined(NANOVDB_USE_INTRINSICS)
-    return __ffsll(v);
+    return __ffsll(static_cast<unsigned long long int>(v));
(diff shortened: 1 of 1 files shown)


kmuseth (changes requested):
the ffsll fix looks good but the solution for the valueCount is to change `Grid::valueCount()` to return `const uint64_t&` rather than `ChannelAccessor::valueCount()`  to return `uint64_t`. Please make this minor change and I'll approve

kmuseth (approved):
great - thanks

PR #1559, opened 2023-01-04
Fix Segfault in Projection Mode of VDB Advect Points SOP

danrbailey:
When I made the changes to switch from `GEOvdbProcessTypedGrid` to `GEOvdbApply`, I accidentally introduced this bug in the projection mode of the VDB Advect Points SOP:

https://github.com/AcademySoftwareFoundation/openvdb/pull/591/files#diff-3da92c28c566b6736ba31e46a2c8a84a00c8beada13f7c3cda90de1782014874R1022

This change fixes the bug so that this mode no longer crashes.

@jmlait


richhones (approved):
Looks good to me

jmlait:
I had never merged the change from GEOvdbProcessTypedGrid over to the Houdini side (as I think we have some other deltas so since it was a no-op I just skipped it).  Thus Houdini escaped this yyp error.

jmlait (approved):

PR #1563, opened 2023-01-10
Fix all the int-in-bool-context warnings with GCC9

danrbailey:
The int-in-bool-context warnings are everywhere with GCC9, this PR uses lots of constexpr to try and fix them all.

One thing to note is that this also disables Bool Grids from being used in the VDB LOD SOP - I'm unsure if that's the correct thing to do, but seems to me that the prolong/restrict operations don't mean much in a bool context?


jmlait (approved):
I've had to keep limiting LOD to the non-bool grids, so this is a good change.

Issue #1124, opened 2021-07-14
[REQUEST] Replace Boost.Python with PyBind11


Idclip:
### Looking for help with Boost.Python -> PyBind11 migration

These points represent the work required to modify python bindings, located in `openvdb/python/`, to build with `PyBind11` and remove the `boost-python` dependency. Any work on the below should be done off of the `feature/pybind` branch. This branch will be synced with `master` on releases and allows for any work to be contributed incrementally:

 - Completely replace the `boost-python` dependency for the pyopenvdb module with `PyBind11`[*]. This includes working with NumPy support.
 - Maintain API and feature parity with the existing `boost-python` bindings. The Python unit tests should ideally require no modification and pass as they are after this work has been completed.
 - Maintain the existing folder structure/file layout where possible
 - Use the existing CMake build system and work with OpenVDBs minimum version requirements, including CMake, GCC/Clang, CXX standard, Python versions >= 2.7 and `PyBind11` versions >= 2.6 [**].
 - Conform to the OpenVDB Coding standards
 - Have commits signed-off (DCO checks) and contributors should be prepared to sign a CLA.
 - Bonus: Modify the CI, where necessary, to build and test with PyBind11. See `.github/workflows/build.yml`
------------------
[*] Note that, similar to the existing build system, `PyBind11` should be expected to exist in a users environment and should *not* be incorporated into the OpenVDB repository (there are outstanding licensing questions with this approach so this should *not* be done at this stage).

[**] `PyBind11` 2.6 was released in October 2020. The latest version is 2.6.2. I think ensuring it works with 2.6.X is enough.

chrisflesher:
Refer to issue #1047 for info on how to get started.

chrisflesher:
I started on an attempt for this, am pretty inexperienced with pretty much all aspects though... so any collaboration is welcome.
https://github.com/chrisflesher/openvdb/tree/replace-boost-python-with-pybind11

Am using the scikit-build system and trying to get it to compile using the following commands:
```
cd openvdb/openvdb/openvdb/myopenvdb
pip3 install . 2> stderr.txt
```

nachovizzo:
As a reference, I have been working with OpenVDB and pybind11 for 1 year now and I have already some open-source bindings:

- [vdbfusion[(https://github.com/PRBonn/vdbfusion)
- [vdb_to_numpy](https://github.com/PRBonn/vdb_to_numpy)

[...]

PR #1571, opened 2023-01-27
Feature/pybind11

kmuseth:
This MR replaces Boost.Python with pybind11 for the OpenVDB Python bindings and addresses https://github.com/AcademySoftwareFoundation/openvdb/issues/1124.

In addition to maintaining parity with the existing Boost.Python bindings, this MR also adds:

Stricter type safety overall thanks to the py::tuple, py::list, py::int_, etc. constructs that pybind11 provides over the generic boost::python::object. This also eliminates the need for several helper methods such as extractArg.
Additional type safety and flexibility when using NumPy courtesy of the pybind11 py::array_t<T> type. This reduces the need for runtime verification code.
More consistent usage of native C++ types as opposed to Python types, i.e. std::tuple<T1, T2> as opposed to py::tuple. This provides a greater degree of compile-time type safety and significantly less code in certain functions.
Removal of the boost::python::numeric legacy code path. boost::python::numeric was removed in Boost 1.65.
(Non-extensible) bindings for GridBase. Previously, conversions to GridBase were loosely handled using an implicit conversion of the holder types GridBase::Ptr and GridT::Ptr. As a result, many functions in GridBase were actually bound at the GridT level, which caused the same function to be instantiated and bound for each grid type. Exposing the proper GridBase -> GridT inheritance hierarchy in Python removes these duplicate instantiations and bindings and removes the need for implicit type conversion.
Potential future changes (for the sake of keeping this MR as minimal as possible aside from the port)

Remove outdated Python 2 code
Refactor pyGrid into a separate translation unit to improve incremental build speed
Refactor NumPy dependent functionality into separate files


kmuseth:
Note, this PR replaces PR1515 by Matthew Cong

Idclip (approved):
I made a small CI fix to this branch which was blatted over by the rebase - looks good

Idclip:
I'll be merging this once the CI passes. Note that we still need to do a few things post:
 - Update doxygen dependency/build docs
 - Decide on renaming import from pyopenvdb to openvdb
 - Progress #1377

Issue #175, opened 2017-10-02
Redundant call to render function in vdb viewer


Aakash1312:
https://github.com/dreamworksanimation/openvdb/blob/ffb02d5527379a08c07449e39d3324ac2e5eff28/openvdb/viewer/Viewer.cc#L1159

The `mUpdates` variable is reset at every key callback. However, I believe `mUpdates` should be reset only when key is pressed. Presently this variable is reset when the key is pressed as well as when it released. Due to this, the `render` function is called four times instead of two.

Nivig1310:
Hello I would like to work on this. I am a little confused. What lines of code are you referring to and what is the key?

ruchiparmar7:
Please assign this issue to me.
I have already created PR for this issue.

ganesh1828:
Can you please assign me this issue.

PR #1572, opened 2023-01-29
#175 Issue: Update Viewer.cc

ruchiparmar7:
setNeedsDisplay() moved to Line 970 Inside IF condition only for keyPress Event

### DIFF
--- a/openvdb_cmd/vdb_view/Viewer.cc
+++ b/openvdb_cmd/vdb_view/Viewer.cc
@@ -967,6 +967,7 @@ ViewerImpl::keyCallback(int key, int action)
             glfwSetWindowShouldClose(mWindow, true);
             break;
         }
+        setNeedsDisplay();
     }
 
     switch (key) {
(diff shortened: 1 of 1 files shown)


ruchiparmar7:
what should i do for the authorization under a signed CLA ?

kmuseth:
Please take a look at the CLI and DOC requirements described here:

https://www.openvdb.org/license/
https://easycla.lfx.linuxfoundation.org

Let me know if for any reason you can't (or don't want to) sign the CLA. In that case your changes are small enough that we might add them to another PR that we sign off. The best if however if you do this yourself.

ruchiparmar7:
I was previously confused how to signed CLA. but I figured out, Now you can review code and merge it

kmuseth (approved):

kmuseth (approved):

kmuseth:
@ruchiparmar7  can you please add description outlining both the problem you're addressing and how your code change fixes it (looking at the code changes neither are clear to me)

najibna (commented):
Moving `setNeedsDisplay()` into the key-press `if` means releases no longer dirty the viewer. That is incomplete: `mClipBox->shiftIsDown()` / `ctrlIsDown()` are still assigned after this block, so a modifier release can change clip-box state with no redraw.

Please either gate those updates on press as well, or call `setNeedsDisplay()` whenever clip/modifier state actually changes.

PR #1573, opened 2023-01-29
Fixed symbol visibility issues with explicit instantiation

Idclip:


e4lam:
Have we tested using this on Windows? I'm missing how an OpenVDB user of the instantiations on Windows if they're no longer declared as dllimport.

Idclip:
@danrbailey bumping this

danrbailey (approved):

PR #1575, opened 2023-01-29
Fixed a bug with LeafNodeBool TopologyCopy constructor

### LINKED ISSUE #1520: Bool LeafNode initialization behavior with designated on and off values
Hi!

I was trying to do a topology copy of a bool tree, with both active and inactive value initialized to `false` for the new tree. But the result has active voxels initialized to `true`, and active tiles set to `false`.

```openvdb::BoolTree* NewTree = new openvdb::BoolTree(OldGrid->tree(), /*inactiveValue*/ false, /*activeValue*/ false, openvdb::TopologyCopy());```

I traced to this line:
https://github.com/AcademySoftwareFoundation/openvdb/blob/be0e7a78861d2b7d9643f7a0cab04f3ab5951686/openvdb/openvdb/tree/LeafNodeBool.h#L847

Seems like the copy behavior for bool leaf node is to simply use the value mask when `offValue` is set to `false`? Is this intended behavior? Any suggested route I should take instead?

Thanks!

### PR DESCRIPTION
**Pull request by Idclip**

Closes #1520

### DIFF
--- a/openvdb/openvdb/tree/LeafNodeBool.h
+++ b/openvdb/openvdb/tree/LeafNodeBool.h
@@ -79,13 +79,35 @@ public:
     template<typename OtherValueType>
     explicit LeafNode(const LeafNode<OtherValueType, Log2Dim>& other);
 
-    /// Topology copy constructor
+    /// @brief Deprecated topology copy constructor
+    /// @note  This constructor initialises the bool buffer to the ValueMask
+    ///   states (i.e. value will be true if the active state is on and
+    ///   vice-versa). This is not really a "TopologyCopy" and is therefor
+    ///   deprecated. Use the explicit mask/buffer constructor instead:
+    /// @code
+    ///    // build new leaf node with the mask of 'a', but with the mask of
+    ///    // 'b' as the the new value buffer.
+    ///    const LeafNode a = ... ;
+    ///    const LeafNode b = ... ;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: openvdb/openvdb/tree/LeafNodeBool.h [hunk 1]
```diff
@@ -79,13 +79,35 @@ class LeafNode<bool, Log2Dim>
...
+    ///    // build new leaf node with the mask of 'a', but with the mask of
+    ///    // 'b' as the the new value buffer.
+    ///    const LeafNode a = ... ;
+    ///    const LeafNode b = ... ;
+    ///    const LeafNode copy(a.origin(), /*mask=*/a.getValueMask(),
+    ///      /*buff=*/b.getValueMask());
+    /// @endcode
     template<typename ValueType>
+    OPENVDB_DEPRECATED_MESSAGE("Use LeafNodeBool component constructor.")
     LeafNode(const LeafNode<ValueType, Log2Dim>& other, TopologyCopy);
```
Idclip, line 97: As part of this, I've decided to deprecate this constructor too as it's not a TopologyCopy. The non-specialized cases (i.e. float, int, etc) don't have this constructor. LeafNodeMasks do, but it makes sense there. I've introduced a component constructor to provide this behaviour more explicitly

danrbailey (approved):

PR #1578, opened 2023-01-30
Fixed a crash in AX when target does not have f16c support

Idclip:


danrbailey (approved):
Tested and this fixes the issue I was seeing, thanks!

PR #1600, opened 2023-03-13
add next hit HDDA to PNanoVDB

gaida-exe:
Signed-off-by: Sebastian Gaida <sebastian-gaida@gmx.de>


gaida-exe:
This PR adds a new variant of the HDDA algorithm to PNanoVDB.
Before, there was only a zero crossing function, and this would add a variant that gets the next hit above 0.
Reason: The background value is often 0 and therefore the other HDDA zero crossing would never report a hit, because it checks for a sign change from start of the ray to tested voxel.

gaida-exe:
Hello @kmuseth I checked NanoVDB and if there is a function that matches the current proposed function and there is.
The [TreeMatcher](https://github.com/sideeffects/openvdb_dev/blob/sendupstream_remove_boost_uuid/nanovdb/nanovdb/util/HDDA.h#L367), like the proposed function, also marches with the HDDA to find the first active value and return true as well as the coordinates ijk. The difference is that the [PNanoVDB function](https://github.com/AcademySoftwareFoundation/openvdb/pull/1600/files#diff-a7090645320b9c338f39bf6318545a53705772ed2f3a7bc8326a317cd4d5e43dR2602) only returns if the value is greater than a given minimal value.
This is essentially the same check that you are doing in NanoVDB for the [Generic Grids(this is a link)](https://github.com/sideeffects/openvdb_dev/blob/sendupstream_remove_boost_uuid/nanovdb/nanovdb/util/HDDA.h#L458)
In PNanoVDB we can't check for nullptr, but we can check for a value greater than 0 and also add the functionality to check for a value greater than the desired ([link](https://github.com/AcademySoftwareFoundation/openvdb/pull/1600/files#diff-a7090645320b9c338f39bf6318545a53705772ed2f3a7bc8326a317cd4d5e43dR2651)).

nanovdb/nanovdb/PNanoVDB.h
```diff
@@ -2575,16 +2577,78 @@ PNANOVDB_FORCE_INLINE pnanovdb_bool_t pnanovdb_hdda_zero_crossing(
...
+
+PNANOVDB_FORCE_INLINE pnanovdb_bool_t pnanovdb_hdda_next_hit(
+    pnanovdb_grid_type_t grid_type,
+    pnanovdb_buf_t buf,
+    PNANOVDB_INOUT(pnanovdb_readaccessor_t) acc,
+    PNANOVDB_IN(pnanovdb_vec3_t) origin, float tmin,
+    PNANOVDB_IN(pnanovdb_vec3_t) direction, float tmax,
+    PNANOVDB_INOUT(float) thit,
+    PNANOVDB_INOUT(float) valueAtHit,
+    PNANOVDB_INOUT(float) minimalValue
```
kmuseth, line 2610: That probably only needs to be PNANOVDB_IN(float) minimalValue

gaida-exe: Hello Ken, I agree with you and I removed the minimalValue. I also renamed the function to be the same of NanoVDB.

gaida-exe:
Sorry for the CLA noise.

PR #1603, opened 2023-03-29
Make find_package(TBB) work with TBBConfig.cmake

lordjaxom:
Building oneTBB generates export configuration files (e.g. TBBConfig.cmake) which can be imported with find_package in CONFIG mode. So it should be possible to build openvdb using the configuration provided by TBB by setting TBB_ROOT and CMAKE_FIND_PACKAGE_PREFER_CONFIG. However, the (standard) configs generated by TBB set TBB_VERSION instead of Tbb_VERSION, so configuring openvdb fails.

This PR makes it possible to use either FindTBB.cmake (find_package in module mode) and TBBConfig.cmake (find_package in config mode).


najibna (commented):
Renaming `Tbb_VERSION` / `Tbb_VERSION_MAJOR` / `Tbb_VERSION_MINOR` in `FindTBB.cmake` and switching the deprecation check to `${TBB_VERSION}` matches TBBConfig.cmake.

It is still a silent break for any module-mode consumer that reads `Tbb_VERSION`, and the module still mixes `Tbb_FOUND` / `Tbb_INCLUDE_DIRS` with `TBB_VERSION`. Consider exporting both names.

PR #1609, opened 2023-04-26
Fixed a crash on Android in the StreamState static instance constructor

jnkvca:
I built OpenVDB for Android as a shared library and it seemed to crash every time when it was loaded. Further investigation showed that it crashes in the static initialization of the `sStreamState` [instance](https://github.com/AcademySoftwareFoundation/openvdb/blob/5cb821242f833d78fbd9506c9d5e20e4c7280f32/openvdb/openvdb/io/Archive.cc#L79-L103). The class has a non-trivial constructor which makes calls to `xalloc()`, `iword()` and `pword()` as can be seen [here](https://github.com/AcademySoftwareFoundation/openvdb/blob/5cb821242f833d78fbd9506c9d5e20e4c7280f32/openvdb/openvdb/io/Archive.cc#L112-L118). These functions crash on Android when called from static initializers when the library is being loaded.

The fix is to delay initialization of `sStreamState` object and do it from the `openvdb::initialize()` function.

### DIFF
--- a/openvdb/openvdb/io/Archive.cc
+++ b/openvdb/openvdb/io/Archive.cc
@@ -80,8 +80,14 @@ struct StreamState
 {
     static const long MAGIC_NUMBER;
 
-    StreamState();
-    ~StreamState();
+    inline StreamState() : magicNumber(-1) {}
+    inline ~StreamState()
+    {
+        uninitialize();
+    }
+
+    void initialize();
+    void uninitialize();
 
     // Important:  The size and order of these member variables must *only* change when
     //             OpenVDB ABI changes to avoid potential segfaults when performing I/O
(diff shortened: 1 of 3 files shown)


Idclip:
Hey @jnkvca thanks a lot for highlighting this. Indeed we have a number of non-trivially constructable objects declared at file scope and thus suffer from SIOF (i.e. non-deterministic/compiler dependent construction order). I've incorporated your changes in a slightly different way and additionally fixed all remaining occurrences of such behavior (as well as enabling Clang's `-Wglobal-constructors` warning):

https://github.com/AcademySoftwareFoundation/openvdb/pull/1620

Let me know if this also looks like it would solve your issue

Idclip:
This should now be fixed as part of https://github.com/AcademySoftwareFoundation/openvdb/pull/1620 - thanks again for highlighting this!

PR #1621, opened 2023-05-26
Remove another std::call_once to avoid initialization crashes

jmlait:
We ran into this with GCC 11.  Similar fix as before, removing the call_once for a static initializer.


Idclip:
@jmlait I've already fixed this in the PR I put up yesterday, see #1620

PR #1626, opened 2023-06-13
Fix prime number in voxel hashing

benemer:
Hi!

I found that one of the constants used for hashing is not a prime number as claimed in the paper (19349663=41*471943). Note that this is also the case in the referred publication of [Teschner et al. [2003]](https://matthias-research.github.io/pages/publications/tetraederCollision.pdf).

I replaced it with the next closest prime number which is 19349669, which has been done [here](https://niessnerlab.org/papers/2013/4hashing/niessner2013hashing.pdf) as well.

Best
Benedikt

### DIFF
--- a/nanovdb/nanovdb/NanoVDB.h
+++ b/nanovdb/nanovdb/NanoVDB.h
@@ -1134,8 +1134,10 @@ public:
 
     /// @brief Return a hash key derived from the existing coordinates.
     /// @details For details on this hash function please see the VDB paper.
+    /// The prime numbers are modified based on the ACM Transactions on Graphics paper:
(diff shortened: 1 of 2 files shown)


Idclip (approved):
Thanks for this! I'd consider adding a comment akin to this PR description in the code (seeing as it now differs from the paper) but otherwise looks good

benemer:
> Thanks for this! I'd consider adding a comment akin to this PR description in the code (seeing as it now differs from the paper) but otherwise looks good

Good point, I just added two comments.

kmuseth (approved):
Good catch

PR #1629, opened 2023-06-23
Use fabsf() for single-precision absolute value

c0d1f1ed:
Overloads of `fabs()` for both `double` and `float` are only available from C++'s <cmath> header, not the C <math.h> header.

Every other math function already correctly used the non-overloaded function name; this was the only outlier. While NanoVDB is written in C++, this change helps avoid requiring C++ headers to be included, which significantly affect compilation time, especially for projects performing just-in-time compilation at runtime.


kmuseth (approved):
Looks great! Since the DCO-test filaed (the this change is so small) I'll first adopt this change in NVIDIA's repo for NanoVDB and then submit a larger PR that includes this change.

PR #1633, opened 2023-07-04
Fix mDelta initialization for NanoVDB DDA

Depersonalizc:
Hi! It seems that the way `mDelta` is currently initialized for `nanovdb::DDA` doesn't take the step size `Dim` into account. This would cause a mismatch between `mStep` and `mDelta` and thus incorrect `step` updates, whenever `Dim` is strictly greater than 1.

I changed the computation of `mDelta` to align with that in `openvdb::math::DDA`.


najibna (commented):
Setting `mDelta[axis] = mStep[axis] * inv[axis]` is the correct analogue of OpenVDB's DDA: when `Dim > 1`, `mStep` is already `±Dim` but `mNext` was still advanced by a one-voxel delta. For `Dim == 1` the new values match the old ones.

A small test with `Dim > 1` would make this much safer to land.

PR #1634, opened 2023-07-04
Introduced Anisotropic Surfacing Kernels and PCA methods

Idclip:
New:
        - Introduced anisotropic surfacing tools in points/PointRasterizeSDF.h
        and principal component analysis (PCA) methods in points/PrincipalComponentAnalysis.h.
        The latter tool analyses point neighborhoods and computes affine
        transformations representing elliptical distributions. The prior tool
        takes rotational and stretch point attributes (optionally computed from
        the prior PCA methods) and generates anisotropic surfaces.

    API Changes:
        - Re-worked the API in PointRasterizeSDF.h with the addition of anisotropic
        surfacing kernels. The new API exposes more concise free functions which
        take structs of parameters. The old API remains supported but is
        deprecated.

    Improvements:
        - Improved the performance of PointRasterizeSDF.h tools with larger radii
        with better bound computation approximations.

    Fixes:
        - Fixed a bug in points::rasterizeSpheres and points::rasterizeSmoothSpheres
        which could cause sections of the generated surface to incorrectly be marked
        as negative interior values.
        - Fixed a bug in points::rasterizeSmoothSpheres which could result in
        incomplete surface reconstruction with significantly larger search distances.
        - Fixed some precision issues in various Matrix methods

### DIFF
--- a/openvdb/openvdb/math/Mat.h
+++ b/openvdb/openvdb/math/Mat.h
@@ -213,8 +213,8 @@ MatType
 rotation(Axis axis, typename MatType::value_type angle)
 {
     using T = typename MatType::value_type;
-    T c = static_cast<T>(cos(angle));
-    T s = static_cast<T>(sin(angle));
+    T c = static_cast<T>(std::cos(angle));
+    T s = static_cast<T>(std::sin(angle));
 
     MatType result;
     result.setIdentity();
@@ -256,9 +256,9 @@ rotation(const Vec3<typename MatType::value_type> &_axis, typename MatType::valu
     Vec3<T> axis(_axis.unit());
 
     // compute trig properties of angle:
-    T c(cos(double(angle)));
-    T s(sin(double(angle)));
-    T t(1 - c);
+    const T c = static_cast<T>(std::cos(double(angle)));
+    const T s = static_cast<T>(std::sin(double(angle)));
+    const T t = T(1) - c;
 
     MatType result;
     // handle diagonal elements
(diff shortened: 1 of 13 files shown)


openvdb/openvdb/points/PrincipleComponentAnalysis.h
```diff
@@ -0,0 +1,148 @@
+// Copyright Contributors to the OpenVDB Project
```
apradhana, line 1: It's actually PrincipalComponentAnalysis, instead of Principle.

openvdb/openvdb/points/impl/PrincipleComponentAnalysisImpl.h
```diff
@@ -0,0 +1,768 @@
+// Copyright Contributors to the OpenVDB Project
```
apradhana, line 1: Same as above, it should be PrincipalComponentAnalysisImpl.h

openvdb/openvdb/points/PrincipleComponentAnalysis.h
Idclip: Oof thanks @apradhana. I'm not a big fan of this header name as there's obviously lots of different types of PCA, this is just one application of it in a very specific way, but I renamed it from something I didn't like either... so I'll fix this, but want to have a better think about the naming here

jmlait (approved):

danrbailey (changes requested):
As mentioned, please remove templated interrupter from all of these API methods.

Idclip:
Re-creating this here: https://github.com/AcademySoftwareFoundation/openvdb/pull/2067

PR #1641, opened 2023-07-11
Add missing log4cplus/version.h include

vimproved:
Currently, the logging.h header checks the LOG4CPLUS_VERSION macro to determine whether to use std::auto_ptr or std::unique_ptr. However, the LOG4CPLUS_VERSION macro is defined in log4cplus/version.h, which is not explicitlyincluded from logging.h. Explicitly include log4cplus/version.h to ensure that the LOG4CPLUS_VERSION macro is defined.

### DIFF
--- a/openvdb/openvdb/util/logging.h
+++ b/openvdb/openvdb/util/logging.h
@@ -14,6 +14,7 @@
 #include <log4cplus/layout.h>
 #include <log4cplus/logger.h>
 #include <log4cplus/spi/loggingevent.h>
+#include <log4cplus/version.h>
 #include <algorithm> // for std::remove()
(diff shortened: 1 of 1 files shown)


Idclip (approved):
Hey thanks, lgtm - we should probably bump our log4cplus min version to 2.X anyway (which would remove the need to use this macro)

najibna (commented):
Adding `#include <log4cplus/version.h>` next to the other log4cplus headers is the right fix so `LOG4CPLUS_VERSION` is defined before the `auto_ptr` / `unique_ptr` branch.

Please confirm `version.h` exists at the project's minimum log4cplus version, or wrap it with the same guard used for the rest of the log4cplus includes.

PR #1643, opened 2023-07-18
Make installing CMake module files depend on whether each specific option is turn on

jasonliu--:
Greetings, I'm one of the people who is in charge of [packaging OpenVDB](https://ports.macports.org/port/openvdb/details/) for the [MacPorts package management system](https://www.macports.org/). In the course of updating our package, I noticed that your project is installing all of the CMake module files, even when the CMake option for that module have been set to `OFF`. So here is a PR that makes the CMake modules depend on whether the CMake option has been enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,19 +285,44 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
 if(OPENVDB_INSTALL_CMAKE_MODULES)
   set(OPENVDB_CMAKE_MODULES
-    cmake/FindBlosc.cmake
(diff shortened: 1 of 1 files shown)


najibna (commented):
Gating `FindBlosc.cmake` / `OpenVDBHoudiniSetup.cmake` on the matching `USE_*` options is reasonable for packaging, but two conditions look wrong for *installed* modules.

`OpenVDBUtils.cmake` is appended only when `OPENVDB_BUILD_BINARIES` is on, yet FindOpenVDB typically needs it even if CLI tools were not built. `FindIlmBase.cmake` is gated on `USE_IMATH_HALF`, which is a half-float typedef option, not whether IlmBase/Imath is required.

PR #1650, opened 2023-07-26
Add missing braces

quagla:
This solves a compile error when using -Wmissing-braces.


jmlait (approved):
Specifically, mData is a union, so the { foo } will initialize the first member with the foo.  Since that is an array, it pedantically should have foo = { a, b, c, d }; hence the double braces.

kmuseth (approved):
Looks good - thanks! I've also run into this issue before and as Jeff said it's caused by the fact that I'm using a union for mData.

PR #1651, opened 2023-08-01
Massive improvements to NanoVDB

kmuseth:
Massive dump of improvements to NanoVDB:

- Minor version changed from 4 to 6 (major version is unchanged since the ABI is preserved)
- Transition from C++11 to C++17 in NanoVDB.h and its tools
- New (backwards compatible ) file format that allows serialized grids to with streamed directly to file without headers
- Several new ways to construct and modify NanoVDB grids on the GPU
- New device function to convert points into a compact grid: nanovdb::cudaPointsToGrid
- Improved and renamed device function that converts voxels into a grid: nanovdb::cudaVoxelsToGrid
- Introduced a new extendable API for acceleration of custom random-access methods, e.g. getValue(ijk)
- Index grids in 4 flavors (Index, OnIndex, IndexMask, and OnIndexMask)
- Introduced new (dummy) build-type nanovdb::Points and nanovdb::GridType::PointIndex
- Introduced new types nanovdb::GridType::Vec3u16 and nanovdb::GridType::Vec3u8 used for compressed representations of point coordinates as blind data
- CreateNanoGrid.h is replacing GridBuilder.h, IndexGridBuilder.h and OpenToNanoVDB.h
- Moved CudaDeviceBuffer.h to cuda/CudaDeviceBuffer.h
- Added cuda/CudaUtils.h and cuda/GpuTimer.h with cuda utility functions
- Added cuda/CudaPointToGrids.cuh that constructs device grids from lists of points or voxels
- Added cuda/CudaIndexToGrid.cuh that converts IndexGrids and values into regular Grids
- Added cuda/CudaSignedFloodFill.cuh that performs signed-flood filing on SDF on the GPU
- Added cuda/CudaAddBlindData.cuh that adds bind data to an existing grid on the GPU
- Added cuda/CudaGridChecksum.cuh that computes CRC32 checksums of grids on the GPU
- Added cuda/CudaGridHandle.cuh that handles grids on the GPU
- Added cuda/CudaNodeManager.cuh that constructs a NodeManager on the GPU
- Added cuda/CudaGridStats.cuh that computes grid statistics on the GPU
- The move constructor in GridHandle now requires the GridBuffer to actually contain a valid grid
- Added new types: Ve4f, Ve4d, ValueIndex, ValueOnIndex, ValueIndexMask, and ValueOnIndexMask
- Major improvements to GridBuilder.h, which allows user to construct grids with random access on the host
- Numerous improvements in NanoVDB.h: e.g. Customizable get/set methods on ValueAccessor, BitFlags, transform(Map), expandAtomic(BBox), expandAtomic(Coord), intersectAtomic(BBox), pi<T>(), BuildTraits, more documentation, Mask:: DenseIterator, Mask:: setOnAtomic,Mask:: setOffAtomic, Map constructors, DataType are now public vs private in all node types, GridMetaData can now be copied
- PNanoVDB.h is now in sync with NanoVDB.h
- Added PrefixSum.h for concurrent computation of prefix sum on the host
- Primitives.h can now create grids on the CPU with SDF, FOG and point of torus


kmuseth:
I'm getting weird compiler errors in tbb when building  linux-nanovdb:cxx:clang++-Debug/Release. Any ideas what's causing this issue? Clearly it has nothing to do with my code :)

Idclip (changes requested):
Lets make sure we squash all these commits before we merge this!

kmuseth:
> Lets make sure we squash all these commits before we merge this!

don't know how to do that after the push. Also don't understand why this is a blocking request

Idclip:
> > Lets make sure we squash all these commits before we merge this!
> 
> don't know how to do that after the push. Also don't understand why this is a blocking request

Because the VDB commit history should be as clean as possible, not full of "wip fix" commits and not retroactively fixed once pushed to master

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -529,92 +742,93 @@ __hostdev__ inline static const T* alignPtr(const T* p)
...
-    } mData;
-public:
-    static const int SIZE = 4;
-    using ValueType = uint8_t;
+    return gridType == GridType::Float ||
+           gridType == GridType::Double ||
+           gridType == GridType::Fp4 ||
+           gridType == GridType::Fp8 ||
+           gridType == GridType::Fp16 ||
+           gridType == GridType::FpN;
```
apradhana, line 798: What about `GridType::Half`?

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -529,92 +742,93 @@ __hostdev__ inline static const T* alignPtr(const T* p)
...
-    return gridType == GridType::Float  ||
-           gridType == GridType::Double ||
-           gridType == GridType::Fp4    ||
-           gridType == GridType::Fp8    ||
-           gridType == GridType::Fp16   ||
-           gridType == GridType::FpN;
+    return gridType == GridType::Int16 ||
+           gridType == GridType::Int32 ||
+           gridType == GridType::Int64 ||
+           gridType == GridType::UInt32;
```
apradhana, line 820: I'm not sure if `GridType::Index`, `GridType::OnIndex`, `GridType::IndexMask`, `GridType::OnIndexMask` should also be included here. I think they belong to another 'semantic'.

nanovdb/nanovdb/NanoVDB.h
kmuseth: Good catch - I'll add it. However, GridType::Half a little strange since in practice it is not (yet) used anywhere. The reason is I   introduce GridType::Half for IEEE 16 half floating points (as implemented in OpenEXR), but in reality NanoVDB is only using GridType::Fp16 which is my own (quantized) implementation of 16-bit floating points. As soon as OpenVDB supports Half in memory grids (we currently only support it out-of-core) then NanoVDB should also support it, in which case nanovdb::isFloatingPoint(GridType::Half) should also work. However, there is no harm in fixing this now :) So, consider this fixed ...

nanovdb/nanovdb/NanoVDB.h
kmuseth: The types you mention are handled by isIndex defined below. Since these index grids don't actually store anything per voxel (not even an integer) they are treated differently than for instance GridType::Int32 that actually encodes a single int per voxel. In other words, these methods reflect both the implementation as well as the return type. So, long story short it's correct as is but I  do agree that the documentation is confusing so I'll fix that instead :)

kmuseth: pushed up improved documentation

nanovdb/nanovdb/Readme.md
```diff
@@ -3,10 +3,11 @@
...
 
-* [Build instructions for make and cmake](../../doc/nanovdb/HowToBuild.md)
-* [Frequently asked questions](../../doc/nanovdb/FAQ.md)
-* [Source tree](../../doc/nanovdb/SourceTree.md)
-* [Examples](../../doc/nanovdb/HelloWorld.md)
+* [Build instructions for make and cmake](docs/HowToBuild.md)
+* [Frequently asked questions](docs/FAQ.md)
+* [Grid cells vs grid nodes](docs/GridCells_vs_GridNodes/Main.pdf)
+* [Source tree](docs/SourceTree.md)
+* [Examples](docs/HelloWorld.md)
```
apradhana, line 10: Please revert this to the previous version:
```
* [Build instructions for make and cmake](../../doc/nanovdb/HowToBuild.md)
* [Frequently asked questions](../../doc/nanovdb/FAQ.md)
* [Source tree](../../doc/nanovdb/SourceTree.md)
* [Examples](../../doc/nanovdb/HelloWorld.md)
```
We are using the `../../doc/` directory because we are consolidating the notes from the OpenVDB and OpenVDB AX under that same directory.

Secondly, you may want to upload `Main.pdf` to this repository if you want to include that in this repository.

nanovdb/nanovdb/unittest/TestNanoVDB.cu
```diff
@@ -0,0 +1,2528 @@
+// Copyright Contributors to the OpenVDB Project
```
apradhana, line 1: I see that `TestNanoVDB.cu` has been added, but I do not see any change in the `unittest/CMakeLists.txt`. Would you mind adding, the following lines at the end of `unittest/CMakeLists.txt`:
```
if(NANOVDB_USE_CUDA)
  add_executable(nanovdb_test_cuda "TestNanoVDB.cu")
  target_link_libraries(nanovdb_test_cuda PRIVATE nanovdb GTest::GTest GTest::Main)
  add_test(nanovdb_cuda_unit_test nanovdb_test_cuda)
endif()
```

nanovdb/nanovdb/Readme.md
kmuseth: reverted Readme.md as requested

nanovdb/nanovdb/unittest/TestNanoVDB.cu
kmuseth: done

apradhana (approved):

PR #1652, opened 2023-08-02
Add an OpenVDB example on how to write a fluid solver

apradhana:
This is supposed to be an example on how to build a fluid simulation using the tools that OpenVDB provides.


apradhana:
Wrong EasyCLA.

PR #1656, opened 2023-08-04
Add VDB Blend and SOP_OpenVDB_Fillet.

sanikache:


danrbailey (commented):
Would be to fix up the DSO and CLA so we can consider getting this merged in.

There's something I'm not understanding about the parallel reduction. It seems to me that if the resolutions of the two grids match (if the SOP resamples before doing the blend), then it doesn't need a merge/reduction step at all?

PR #1663, opened 2023-08-29
Basic Nix package and flake support

kdmarrett:
Pros:
- Nix is the largest and most up to date package manager [repology.org]. Nix is multi-language and works on any Unix system, it wraps existing CMake infrastructure cleanly and is strong complement to C++'s weak packaging ecosystem
- Nix is state of the art in deterministic builds (nix makes docker look like pip in terms of reproducibility). For example you can pin a nix package and all of its dependencies and compiler to specific git commits
- Allows installing openvdb and a subset of its command line tools (vdb_view and vdb_print enabled so far) via a single command line invocation
- Allows adding openvdb (and all of its dependencies) as a dependency to downstream C++ libraries via a single line
- Its strict reproducibililty standards make it ideal for CI tasks (like github actions). I have done github action builds with openvdb and Nix and can help with that if desired.

Cons:
- Strictly for Linux (or windows subsystem for linux) installation at the moment, private development is underway to support MacOS (mesa dependency is temporarily broken). Only works on Unix, it will probably never support raw windows installation
- I set parallel builds to ON, for a 32 thread system this consumes 43 GB of memory, so some high thread low memory systems will fail to build


kdmarrett:
Not ready for merge yet, need to remove unnecessary files that were previously used for backwards compatibility and fix the MacOS mesa nix package

Issue #1590, opened 2023-03-02
C++20 compatibility


jdumas:
The following line:
https://github.com/AcademySoftwareFoundation/openvdb/blob/7edd8cd86f105a01a41ad7b2bf59a81034cd79fb/openvdb/openvdb/points/AttributeSet.cc#L283
is not compiling for me on C++20 with MSVC 19.28.29916.0, due to [std::shared_ptr<>::unique()](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique) being deprecated and removed in C++20.

PR #1628, opened 2023-06-21
Fix C++20 compatibility by removing deprecated shared_ptr::unique()

jdumas:
Fix #1590.

PR #1664, opened 2023-09-01
Fixed C++20 compatibility of OpenVDB and TBB

maletsden:
## Change List:
- Fixed compatibility of OpenVDB and TBB by fixing invocations of `tbb::parallel_reduce`, which from C++20 have `__TBB_requires(tbb_range<Range> && parallel_reduce_body<Body, Range>)`. In that requires expression `parallel_reduce_body<Body, Range>` was failing because Body didn't have an `operator()` that takes a constant reference to `Range`:
```cpp
template <typename Body, typename Range>
concept parallel_reduce_body = splittable<Body> &&
                               requires( Body& body, const Range& range, Body& rhs ) {
                                   body(range);
                                   body.join(rhs);
                               };
```
- Replaced usage of `std::shared_ptr<T>::unique` (which was removed in C++20) by `std::shared_ptr<T>::use_count` (similiar to what @jdumas did in #1628 )

Fixes issue: #1590

### DIFF
--- a/openvdb/openvdb/points/AttributeSet.cc
+++ b/openvdb/openvdb/points/AttributeSet.cc
@@ -280,7 +280,8 @@ AttributeSet::isShared(size_t pos) const
 {
     assert(pos != INVALID_POS);
     assert(pos < mAttrs.size());
-    return !mAttrs[pos].unique();
+    // Warning: In multithreaded environment, the value returned by use_count is approximate.
+    return mAttrs[pos].use_count() != 1;
 }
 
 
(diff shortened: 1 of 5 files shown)


Idclip (approved):
Thanks a lot @maletsden, I've verified this too and bumped our CI (https://github.com/AcademySoftwareFoundation/openvdb/pull/1674) to build with newer compilers and C++20 to catch these going forward (this issue currently only triggers with TBB 2020+, GCC10+ and C++20)

@danrbailey

richhones (approved):

PR #1666, opened 2023-09-04
feat: allow maskSdf to optionally resurface the level set to the reacha…

kdmarrett:
…ble voxels (connected component) in mask


kdmarrett:
This approach is fundamentally flawed and inefficient. The better approach is to simply inactivate (prune) voxels that are active but that have values of FLT_MAX, FLT_MIN

PR #1679, opened 2023-10-10
Fix nullptr dereferences found by gcc 11.4

e4lam:
I'm not sure if we can actually hit these code paths in practice but this at least shuts up the warnings.


jmlait (approved):
Might become more important when we have adaptive grids...

danrbailey (approved):

PR #1686, opened 2023-10-13
Removed Redundant Render Function Call

ganesh1828:


diiigle:
Rings some alarm bells for me: 

 1. effectively a 1 line change, that would be "easy to approve"
 2. no description provided
 3. Why is there a binary blob committed, and then removed in a second commit? Just to place it in the history?!

smells fishy

ganesh1828:
@diiigle Binary was pushed by mistake, that's why I removed it. and description is provided now, you can check it.

PR #744, opened 2020-06-18
Removed boost/any in favour of openvdb::MetaMap usage

Idclip:

PR #1687, opened 2023-10-13
Replace boost::any with std::any

bpmckinnon:
Replace BOOST_STATIC_ASSERT with static_assert

### DIFF
--- a/openvdb/openvdb/io/Archive.cc
+++ b/openvdb/openvdb/io/Archive.cc
@@ -330,10 +330,10 @@ namespace {
 
 template<typename T>
 inline bool
-writeAsType(std::ostream& os, const boost::any& val)
+writeAsType(std::ostream& os, const std::any& val)
 {
     if (val.type() == typeid(T)) {
-        os << boost::any_cast<T>(val);
+        os << std::any_cast<T>(val);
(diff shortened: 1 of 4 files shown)


Idclip (approved):

Idclip:
Thanks a lot @bpmckinnon! I actually tried to fix this years ago with our metadata implementation (linking just for future reference https://github.com/AcademySoftwareFoundation/openvdb/pull/744) but happy to accept this change instead. 

Do you think you can sign off your commits and fill out the CLA?

I think you're right in that all that's left is the conversion_traits header (and it would be nice to have our own iostream/interprocess impl), we can look to solve that too.

bpmckinnon:
I'm working on the CLA.  I was doing this as part of work, so I'm looking for a corporate manager for the CLA.

Idclip:
> I'm working on the CLA. I was doing this as part of work, so I'm looking for a corporate manager for the CLA.

No worries, keep us posted - we can contribute these changes on your behalf but would be nicer if you're able to get it signed

bpmckinnon:
Alright, this should be ready to go

danrbailey (approved):

PR #1690, opened 2023-10-18
Fix Houdini CI Caching

Idclip:
# We bumped from the 2021 CI image to 2023 to fix some OpenSSL issues
    # with the Houdini download script. In so doing we broke some of the caching
    # between this job and the jobs in houdini.yml which _don't_ use the 2023
    # image yet. The issue is that the cache action will use zstd if it's
    # available to zip the cache and this causes it to be inserted with a unique
    # hash which images without zstd (i.e. the 2021/2022 images don't have
    # access to). For now, uninstall zstd here instead of installing it
    # everywhere and ask the LF to add zstd to the older base images.


Idclip:
This should now pass once in master

apradhana (approved):

PR #1693, opened 2023-10-24
Cleanup several aspects of the Python bindings

matthewdcong:
This MR cleans up several aspects of the pybind11 Python bindings. It's intended to have no change in outward user-facing functionality, but provides future-proofing for the bindings as pybind11 evolves.

1. The redundant explicit bindings for the Math types have been removed. These are not used (and should not be exposed) since we have typecasters implemented for the functionality which take precedence.
2. Implemented type casters for TypedMetadata. This means that Vec3SMetadata values in the MetaMap will call the Vec3SMetadata type caster which will subsequently call the Vec3f typecaster resulting in re-usable functionality. This also enables us to pass TypedMetadata values between C++ and Python more explicitly.
3. Removed Python 2 support
4. Switched from `obj.cast<T>()` to the equivalent but preferred `py::cast<T>(obj)`.


kmuseth (approved):
looks good to me

PR #1698, opened 2023-10-31
Update change notes

danrbailey:


Idclip:
Did we end up deciding if the nanovdb notes can be consolidated?

danrbailey:
> Did we end up deciding if the nanovdb notes can be consolidated?

No. I've done a first pass at updating the docs and am going to merge this PR now and re-publish them to make sure the process is all working and everything is ready to go, but I won't publish the 11.0.0 draft yet. I think it would be better if the NanoVDB notes are re-written, so can be a topic for the TSC tomorrow.

PR #1701, opened 2023-10-31
Fix for potential crash when reading invalid .vdb files.

jmlait:
Original fix thanks to Matthias Ueberheide.

### DIFF
--- a/openvdb/openvdb/io/Compression.cc
+++ b/openvdb/openvdb/io/Compression.cc
@@ -123,19 +123,22 @@ unzipFromStream(std::istream& is, char* data, size_t numBytes)
 {
     // Read the size of the compressed data.
     // A negative size indicates uncompressed data.
-    Int64 numZippedBytes;
+    Int64 numZippedBytes{0};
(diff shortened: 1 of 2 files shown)


kmuseth (approved):
Looks good to me (just initialization of variable and moving an error check forward)

kmuseth:
weird, one of the nanovdb unit-tests failed on Windows because it timed out! Never seen that error before ... @jmlait  can you re-run that test?

Idclip:
Looks like it segfaulted, it didn't time out

kmuseth:
re-ran the test and it passed so it might have been a resource issue, like insufficient memory

Idclip:
As far as I know your `nanovdb_openvdb_unit_test` target doens't use any VDB IO anyway? So that crash was probably not related to this change. 

I'd keep an eye out if it happens again in case it's a non-deterministic issue with the latest nanovdb changes

jmlait:
Built 10.1 with this version and ran sop-based reg tests on it with no change in behaviour from before the change.

PR #1716, opened 2023-12-04
Improve support for GCC dual ABI with LLVM/AX/Houdini builds

Idclip:


Idclip:
Closing, this is more invovled and at the moment it's easier to rebuild LLVM with the old CXX11 abi

Issue #1718, opened 2023-12-04
[BUILD] Windows - Visual Studio 2019 : fatal error LNK1248


nyue:
### Environment
**Operating System:** (e.g. Windows 2019, macOS-10.15, Linux CentOS 7.4)

Windows 10

**Version / Commit SHA:** (e.g VDB 7.2, SHA: 1a2b3c4d5e...)

OpenVDB 11.0.0

**CMake Version:** (e.g. 3.18)

3.25

**Compiler:** (e.g. gcc 10, msvc 2019)

VS 2019

### Describe the problem

Linker error

```
openvdb\openvdb\libopenvdb.lib : fatal error LNK1248: image size (1017E3216) exceeds maximum allowable size (FFFFFFFF)
[155/156] Building CXX object openvdb_cmd\vdb_print\CMakeFiles\vdb_print.dir\main.cc.obj
ninja: build stopped: subcommand failed.
```

vdb_print is not critical to me at this point so I can remove it from the build but reporting this anyway.

Idclip:
Hey @nyue this is usually related to explicit template instatiation. We should probably turn this off on Windows. Try configuring with `-D USE_EXPLICIT_INSTANTIATION=OFF`

nyue:
That works for me.

I have a new issue with the Windows build

The Ninja build encountered multiple build rules but only if I build the AX stuff.

![image](https://github.com/AcademySoftwareFoundation/openvdb/assets/486172/f5546fcb-c99a-4c03-b445-bcb25f5e433f)

Idclip:
Ah this looks to be a bug in our CMake, both the shared and static builds of AX are producing the same library name. 

You can either disable the static/shared build with `OPENVDB_AX_SHARED=ON|OFF` or `OPENVDB_AX_STATIC=ON|OFF` or add:

```cmake
  if(WIN32)
    set_target_properties(openvdb_ax_static PROPERTIES PREFIX "lib")
  endif()
```

to somewhere in openvdb_ax/CMakeLists.txt:

https://github.com/AcademySoftwareFoundation/openvdb/blob/41bb1279c2c96d0d09614bc57830da0184002c51/openvdb_ax/openvdb_ax/CMakeLists.txt#L304-L308


I'll make the latter change to master

Issue #1624, opened 2023-06-08
libopenvdb.lib : fatal error LNK1248:


ActiveXSturcture:
PS D:\openvdb\build> cmake -DCMAKE_TOOLCHAIN_FILE=D:\vcpkg\scripts\buildsystems\vcpkg.cmake -DOPENVDB_BUILD_NANOVDB=ON  ..
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.19044.
-- The CXX compiler identification is MSVC 19.36.32532.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.36.32532/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Suppressing some noisy MSVC CXX warnings, set OPENVDB_CXX_STRICT=ON to re-enable them.
-- Configuring for OpenVDB Version 10.0.2
-- Configuring for OpenVDB ABI Version 10
-- ----------------------------------------------------
-- ------------- Configuring OpenVDBCore --------------
-- ----------------------------------------------------
CMake Warning at openvdb/openvdb/CMakeLists.txt:64 (message):
  Detected VCPKG toolchain is using a mismatching triplet for OpenVDB build
  artifacts.  x64-windows is dynamic, but OPENVDB_CORE_STATIC is ON.  The
  static build of OpenVDB may not build correctly.  It it recommended to only
  configure either the static or shared library on Windows in a single run of
  CMake.

CMake Warning at C:/Program Files/CMake/share/cmake-3.22/Modules/FindBoost.cmake:1384 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.22/Modules/FindBoost.cmake:1507 (_Boost_COMPONENT_DEPENDENCIES)
  C:/Program Files/CMake/share/cmake-3.22/Modules/FindBoost.cmake:2117 (_Boost_MISSING_DEPENDENCIES)
  D:/vcpkg/installed/x64-windows/share/boost/vcpkg-cmake-wrapper.cmake:11 (_find_package)
  D:/vcpkg/scripts/buildsystems/vcpkg.cmake:807 (include)
[...]

PR #1721, opened 2023-12-06
Turn off USE_EXPLICIT_INSTANTIATION on Windows by default

Idclip:
https://github.com/AcademySoftwareFoundation/openvdb/issues/1718
https://github.com/AcademySoftwareFoundation/openvdb/issues/1624


danrbailey (commented):
Maybe a first step is to just disable it when building static libs on Windows?

danrbailey:
> Maybe a first step is to just disable it when building static libs on Windows?

Never mind, we can come back and make that additional change later.

danrbailey (approved):

Issue #1700, opened 2023-10-31
[BUG] Vec3H Grid Registration no longer builds


matthewdcong:
To reproduce, add the following registration to vdb_print/main.cc:
`openvdb::Grid<openvdb::tree::Tree4<openvdb::Vec3H, 4, 3, 3>::Type>::registerGrid();`

This will fail to build with a `no match for operator+` error on GCC 7.3.1 but building with Clang 16 reveals more information:
```
openvdb/openvdb/math/Vec3.h:540:43: note: candidate template ignored: requirement 'std::is_arithmetic_v<openvdb::v11_0::math::internal::half>' was not satisfied [with S = float, T = openvdb::v11_0::math::internal::half]
inline Vec3<typename promote<S, T>::type> operator+(const Vec3<T> &v, S scalar)
```
This appears to be due to #1688 because of the `enable_if` because `std::is_arithmetic` is true for a hardcoded list of system types that does not include half.

I've attempted to come up with a fix, but I've ran into two issues so far. Specializing `std::is_arithmetic` is undefined behavior, so we'll need to define an alternative template check. Secondly, this would create a circular dependency of sorts between `Math.h` and `Types.h`, since the latter is where the `half` type, either internal to OpenVDB or externally imported from Imath, is defined.

PR #1723, opened 2023-12-06
Replace boost conversion_traits with std::common_type

Idclip:
Re-creating this so that we don't lose visibility on it. It currently cannot be merged due to https://github.com/AcademySoftwareFoundation/openvdb/issues/1700


bpmckinnon:
One possible option given that std::numeric_limits is overridden is to try using std::numeric_limits<T>::is_specialized instead.
https://en.cppreference.com/w/cpp/types/numeric_limits/is_specialized

bpmckinnon:
I've verified that this is at least a potential option on my windows build... but I can't guarantee it will work in the build environment that is failing.

apradhana:
As for the problem that Matthew raised in [Issue 1700 ](https://github.com/AcademySoftwareFoundation/openvdb/issues/1700), I have a fix that I included in the [draft PR for supporting half grid](https://github.com/AcademySoftwareFoundation/openvdb/pull/1730). What's needed is adding a `cwiseAdd` specialization for `Vec3<math::half>` as shown [here](https://github.com/apradhana/openvdb/blob/half_grid_support/openvdb/openvdb/Types.h#L715-L730).

bpmckinnon:
Has there been any update on the cwiseAdd function?  Otherwise I can add the is_specialized change I proposed.

bpmckinnon:
Before this change can be resubmitted std::is_arithmetic_v should be replaced with std::numeric_limits::is_specialized to prevent the initial issue that was flagged

kmuseth (approved):

danrbailey:
Merging to a feature branch to bring this PR up-to-date before merging to master.

PR #1729, opened 2023-12-12
Various improvements

Idclip:
Houdini:
    - When OPENVDB_ENABLE_RPATH is ON, the location of libopenvdb_houdini is now added to the rpath of all Houdini dsos.

Build:
    - Jemalloc is now the preferred allocator of choice on all platforms when CONCURRENT_MALLOC is set to Auto.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -315,10 +315,7 @@ jobs:
     - name: llvm
       run: ./ci/install_llvm_windows.sh ${{ matrix.config.crt }}
     - name: install
-      run: |
-        vcpkg update
-        vcpkg install zlib tbb cppunit blosc python3 \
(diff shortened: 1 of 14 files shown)


danrbailey (commented):
RPath change looks good to me.

> On Windows, we follow SideFX's example in using Tbbmalloc due to the challenges of injecting into the Windows runtime to replace the system allocator.

Has this situation changed now?

Idclip:
> RPath change looks good to me.
> 
> > On Windows, we follow SideFX's example in using Tbbmalloc due to the challenges of injecting into the Windows runtime to replace the system allocator.
> 
> Has this situation changed now?

Support for Jemalloc on windows is significantly better now and vcpkg has a native build port. This also removes the platform dependent branching. Unless @jmlait has any concerns I think this is now a better default choice (CI is passing too)

danrbailey (approved):

PR #1733, opened 2023-12-16
Bump internal copy of half to fix issues on older x86_64 CPUs on Windows

brechtvl:
On some older CPUs without support for the lzcnt instruction, the conversion from half to float was wrong.

See AcademySoftwareFoundation/Imath#358 for more details.

Thanks to Ray Molenkamp for tracking down this bug.

### DIFF
--- a/openvdb/openvdb/math/Half.h
+++ b/openvdb/openvdb/math/Half.h
@@ -342,8 +342,14 @@ imath_half_to_float (imath_half_bits_t h)
         // other compilers may provide count-leading-zeros primitives,
         // but we need the community to inform us of the variants
         uint32_t lc;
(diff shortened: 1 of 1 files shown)


brechtvl:
For reference, we hit this in Blender when upgrading to the VFX platform 2024, from OpenVDB 10 to 11.

CPU's like Sandy Bridge and Ivy Bridge that are affected by this are relatively old, but still support the minimum requirement of SSE 4.2 used by Blender, Maya and Houdini.

image

Idclip (approved):
@brechtvl and co, nice catch. Looks good to me

danrbailey (approved):

PR #1734, opened 2023-12-18
Preserve Directory Structure for NanoVDB util/ and cuda/ Headers on Installation

eclipse0922:
This PR addresses an issue where the directory structure of the util/ and util/cuda/ headers in NanoVDB is not preserved when installed through CMake. The current installation process flattens the directory structure, leading to incorrect include paths when NanoVDB is used from the installed directories.

For example,
https://github.com/AcademySoftwareFoundation/openvdb/blob/master/nanovdb/nanovdb/util/GridHandle.h
uses cuda utils like this, but there is no 'cuda' directory in installed path.
```
#if defined(__CUDACC__)
#include <nanovdb/util/cuda/CudaGridHandle.cuh>
#endif// defined(__CUDACC__)

#endif // NANOVDB_GRID_HANDLE_H_HAS_BEEN_INCLUDED
```

Changes
Modified the CMakeLists.txt to use install(DIRECTORY ...) instead of install(FILES ...) for the util/ directory. This change ensures that the directory hierarchy (specifically the util/cuda/ subdirectory) is maintained during installation.


eclipse0922:
Another approach: Split NANOVDB_INCLUDE_UTILFILES into two parts, cuda and non-cuda, then install them separately.

kmuseth (approved):
Good catch - I've notice the same issue and is preparing a very large PR that introduces a major refactoring of the directory structure and namespaces for NanoVDB. However, in the meantime I'm happy to  adopt your PR

apradhana (approved):

PR #1750, opened 2024-01-09
Move the pure-reference files, GU_PrimVDB and GEO_PrimVDB, to a reference directory

jmlait:
Create a reference subdirectory to make it clear things are not for compiling directly, but are to provide an open source reference to otherwise properietary Houdini code.


openvdb_houdini/openvdb_houdini/CMakeLists.txt
```diff
@@ -174,14 +172,21 @@ file(COPY
...
 )
+file(MAKE_DIRECTORY ${OPENVDB_HOUDINI_LOCAL_DIR}/reference)
+file(COPY
+    reference/GEO_PrimVDB.h
+    reference/GU_PrimVDB.h
+    reference/GEO_PrimVDB.cc
+    reference/GU_PrimVDB.cc
+  DESTINATION
+    ${OPENVDB_HOUDINI_LOCAL_DIR}/reference
+)
```
Idclip, line 183: You can remove this - this copies the headers into a directory in the build folder for include path setup, but as these aren't used at all they don't need to be moved/copied

Idclip: That being said there is this comment in GU_PrimVDB:

> // Using the native OpenVDB Primitive shipped with Houdini is strongly recommended,
// as there is no guarantee that this code will be kept in sync with Houdini.
// However, for debugging it can be useful, so supply -DSESI_OPENVDB_PRIM to
// the compiler to build this custom primitive.

Is this something we still support?

jmlait: Good to know about the lack of need to copy into the build directory, I wasn't sure if that became a distribution directory that was needed.   (We've got non-built stuff that needs to copy out to HFS...)   However, since the original git repo is available, there probably is no need for any distribution to have reference code.

Regarding the comment in GU_PrimVDB...  I noticed we are not at all consistent before if we grabbed <GU/GU_PrimVDB.h> or <openvdb_houdini/GU_PrimVDB.h>, so I think we were grabbing from both so I'm not certain we actually supported that.  (There is also some hvdb:: that I think I still have to deal with where we'd intentionally pick one or the other).  I recall from previous discussions that we weren't really supporting that going forward, and instead focusing on this being a reference implementation so users can peek into the .C file.  Which can be useful as there are some openvdb relevant code there.  So I think this does become a line in the sand to say that we aren't supporting replacing GU_PrimVDB, (at least not by a simple compile option), going forward....

The main reason to do this change at all is in hopes of open sourcing more internal Houdini code that we can't expose in a manner that can be compiled, but we'd like the code available if people want to build their own versions.  This is part of that problem we have where if we use Houdini libraries it means the SOPs aren't really as open as they should be, if we can at least have reference implementations of what they call people could port the SOPs to different environments.   But we also don't want to spend the time to strip the files of all the Houdini dependencies and make part of core VDB.

Idclip: Thanks @jmlait, that all makes sense and sounds good to me

Idclip (approved):

PR #1760, opened 2024-01-30
Initial Adaptive Grid prototype

danrbailey:
Sharing an initial Adaptive Grid prototype (@kmuseth). In this case, it's not even a dense grid, it just contains a single background value and I've done the minimal work to make this compile with most of the required methods set to throw a not implemented error currently. No IO implemented yet. Some brief notes:

I'm using an `adaptive::` namespace, this sits within an `openvdb/adaptive` sub-directory and I'm calling this `AdaptiveTree`. It derives from `TreeBase` as we discussed and this is nested within a `Grid` as well and aliased to `AdaptiveGrid`.

I've changed the existing `GridTypes` alias to `SparseGridTypes` and re-introduced `GridTypes` to be `SparseGridTypes` + `AdaptiveGridTypes`. You can now do `grid.apply<SparseGridTypes>()` to get back the original behavior. There's a question as to whether we should leave `GridTypes` alone to maintain backwards-compatibility and to introduce something new like `AllGridTypes`, but not sure if this will make sense if we add more.

I introduced a new `TreeTraits` class to determine whether grids are adaptive or sparse and added a specialization to the relevant template classes. The main goal here was to not need to add new virtual methods or static member variables to the Tree class (or any other derived classes there might be in the wild). I've changed the `memUsage()` method which conveniently is used in `vdb_print` to see how this might affect the logic, it now does this:

```
Index64 memUsage(const TreeT& tree, bool threaded)
{
    if constexpr (TreeTraits<TreeT>::IsSparse) {
        // sparse implementation
    } else if constexpr (TreeTraits<TreeT>::IsAdaptive) {
        // adaptive implementation
    }
    OPENVDB_THROW(NotImplementedError, "");
}
```

This should be sufficient to get some conversation going as to how this kind of thing could work. It might make sense to consider using an `impl` sub-directory as we have done elsewhere, but where should the implementation for sparse trees and adaptive trees live? Should we just include both implementations in the body of this function as I've done here? Should we add a new sparse sub-directory and put the original implementation of this method there and then put the adaptive one in the adaptive sub-directory? How would this work for some of the much larger, more complicated methods? What about if we added more logic for handling dense trees? How is this going to work with explicit template instantiation?

### DIFF
--- a/openvdb/openvdb/CMakeLists.txt
+++ b/openvdb/openvdb/CMakeLists.txt
@@ -402,6 +402,10 @@ set(OPENVDB_LIBRARY_IO_INCLUDE_FILES
   io/TempFile.h
 )
 
+set(OPENVDB_LIBRARY_ADAPTIVE_INCLUDE_FILES
+  adaptive/AdaptiveGrid.h
+)
+
 set(OPENVDB_LIBRARY_MATH_INCLUDE_FILES
   math/BBox.h
   math/ConjGradient.h
--- a/openvdb/openvdb/Types.h
+++ b/openvdb/openvdb/Types.h
@@ -237,6 +237,17 @@ using make_index_sequence =
 ////////////////////////////////////////
 
 
+template<typename TreeT>
+struct TreeTraits
+{
+    static const bool IsSparse = false;
+    static const bool IsAdaptive = false;
+};
+
+
+////////////////////////////////////////
+
+
 template<typename T, bool = IsSpecializationOf<T, math::Vec2>::value ||
                             IsSpecializationOf<T, math::Vec3>::value ||
                             IsSpecializationOf<T, math::Vec4>::value>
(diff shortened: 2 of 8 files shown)


danrbailey:
This PR should pass all the CI checks when #1775 and #1776 are merged in.

@kmuseth - to follow up on one of your questions from the meeting yesterday. The ValueAccessor registry stores the `ValueAccessorBase` which means any variant of the ValueAccessor will be registered provided that it derives from this base class:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/Tree.h#L1016

Moving the accessor alias and ValueAccessor construction from the Grid to the Tree does not affect the registry in any way. Finally, this registry is stored on the tree and not the grid, so for an adaptive tree there is no registry needed as implemented.

Issue #1269, opened 2021-11-25
[BUILD] CMake RelWithDebInfo produces invalid TBB library path resulting in an error


joshbainbridge:
### Environment
**Operating System:** Linux CentOS 7.0 (Core)
**Version / Commit SHA:** VDB 9.0
**CMake Version:** 3.12
**Compiler:** clang++ 12.0.1

### Describe the problem
Recently we updated from OpenVDB 8.0.1 to 9.0.0 and found that our **RelWithDebInfo** CMake configurations were failing. This was due to the changes in `cmake/FindTBB.cmake` and output string that is ill configured. The error we get back from Clang is:

```
error: 'optimized;/path/to/install/lib64/libtbbmalloc.so.2;debug;/path/to/install/lib64/libtbbmalloc_debug.so.2', needed by 'src/some/library.so', missing and no known rule to make it.
```

It appears that instead of a specific library file being passed as IMPORTED_LOCATION when constructing the TBB CMake IMPORT target, we are instead passing a CMake string with both release and debug library locations. I believe these might be the key lines.

`cmake/FindTBB.cmake:288`:
```
set(Tbb_${COMPONENT}_LIBRARY optimized ${Tbb_${COMPONENT}_LIBRARY_RELEASE} debug ${Tbb_${COMPONENT}_LIBRARY_DEBUG})
```

`cmake/FindTBB.cmake:418`:
```
set_target_properties(TBB::${COMPONENT} PROPERTIES IMPORTED_LINK_INTERFACE_LANGUAGES "CXX" IMPORTED_LOCATION "${Tbb_${COMPONENT}_LIBRARY}")
```

My expectation when building a CMake **RelWithDebInfo** configuration is that my library will be linked against the debug build of TBB, rather than a release build. But release would also not be an issue. As long as one is selected and the build works.

### To Reproduce
1. Checkout repo '...'
2. Run CMake with `-D CMAKE_BUILD_TYPE=RelWithDebInfo`
3. Build binary

### Additional context
Our downstream project sets the CMAKE_MODULE_PATH to that of the modules included with OpenVDB. As a result we pick up the FindTBB.cmake module. We might be better off with a CMake config if one was installed along with the OpenVDB binary, avoiding this need. It looks like there is already a ticket for this https://github.com/AcademySoftwareFoundation/openvdb/issues/1160

Idclip:
[...]

PR #1763, opened 2024-02-08
Support relwithdebinfo and minsizerel build

zhangha182:
Support other release build by mapping relwithdebinfo and minsizerel to release.


danrbailey (approved):

danrbailey:
@Idclip - are you ok with this proposed change?

Idclip:
Yep this is fine. Usually you'd control this with `CMAKE_MAP_IMPORTED_CONFIG_<CONFIG>` (see https://github.com/AcademySoftwareFoundation/openvdb/issues/1269) but we can do it per package too. We should actually just switch to using `find_package(CONFIG)` for TBB and EXR.

Idclip (approved):

PR #1776, opened 2024-03-13
Workaround for python3 brew mac CI issue on GitHub runners

danrbailey:


danrbailey:
@Idclip - there seems to be a persistent issue with the Mac OS builds over the last few days when doing a `brew update`. I'm not too familiar with this CI workflow, but I tried a few different options and finally I found this work-around to delete the python3 binaries in `/usr/local/bin` before running the update and that appears to successfully resolve the issue. Does this seem ok to you?

Idclip:
@danrbailey looks like this might be a slightly better solution: https://github.com/actions/setup-python/issues/577#issuecomment-1431468702 `brew install --overwrite`. But happy with this too. 

Might want to add this issue as a comment to this change: https://github.com/actions/setup-python/issues/577

Idclip (approved):

danrbailey:
> @danrbailey looks like this might be a slightly better solution: [actions/setup-python#577 (comment)](https://github.com/actions/setup-python/issues/577#issuecomment-1431468702) `brew install --overwrite`. But happy with this too.
> 
> ~Might want to add this issue as a comment to this change: [actions/setup-python#577](https://github.com/actions/setup-python/issues/577)~ Woops, you've already done this!

Maybe I missed something, but I tried that overwrite method a bunch of different ways and couldn't seem to get it to work.

PR #1780, opened 2024-03-22
Support for instantiating Half-Typed-grid from stored float or double grids

konivo:
This is still WIP and it depends on #1730 .

The main motivation for this patch is to be able to load half-typed grids directly from files containing non-half-typed grids in order to keep the process peak memory low. When having large grids or large number of grids this can be important.

The PR does (atm) the conversion according to the following rules:
| stored grid value type | desired conversion | returned grid |
---------------|----------------------|-------------
float | half | half |
double | half | half |
vec2f | half | vec2h |
vec3f | half | vec3h |
vec2d | half | vec2h |
vec3d | half | vec3h |

The above rules can be extended if desirable, and in the future there could be also 'store' counterpart to do conversion into other grid types while storing. That could potentially replace the need for "saveFloatAsHalf" flag.

The core implementation has been already tested with various vdb files, some of them using saveFloatAsHalf, and the approach seems to be working without issues.

The implementation is straightforward and follows these steps:
- User specifies desired conversion when calling `File::open` or when instantiating `Stream`
- Desired conversion information then propagates into `StreamMetadata`
- `GridDescriptor` picks up the conversion from the metadata and from `io::ConvertingReaderFactory` it gets a reader and the resulting grid valuetype. It then instantiates a grid with this new valuetype instead of using the original value type. If there is no suitable reader for the desired conversion, descriptor instantiates a grid with the original valuetype.
- The reader returned from `io::ConvertingReaderFactory` is also stored in `StreamMetadata` and used during topology/buffers loading.

### DIFF
--- a/openvdb/openvdb/io/Archive.cc
+++ b/openvdb/openvdb/io/Archive.cc
@@ -212,6 +212,8 @@ struct StreamMetadata::Impl
     bool mDelayedLoadMeta = DelayedLoadMetadata::isRegisteredType();
     uint64_t mLeaf = 0;
     uint32_t mTest = 0; // for testing only
+    std::string mDesiredScalarType = "";
+    SharedPtr<ConvertingReaderBase> mConvertingReader;
 }; // struct StreamMetadata
 
 
@@ -278,6 +280,12 @@ uint64_t        StreamMetadata::leaf() const            { return mImpl->mLeaf; }
 MetaMap&        StreamMetadata::gridMetadata()          { return mImpl->mGridMetadata; }
 const MetaMap&  StreamMetadata::gridMetadata() const    { return mImpl->mGridMetadata; }
 uint32_t        StreamMetadata::__test() const          { return mImpl->mTest; }
+
+const std::string&    StreamMetadata::desiredScalarType() const { return mImpl->mDesiredScalarType; }
+void StreamMetadata::setDesiredScalarType(std::string t) { mImpl->mDesiredScalarType = t; }
+
+const ConvertingReaderBase* StreamMetadata::convertingReader() const {return mImpl->mConvertingReader.get();}
+void StreamMetadata::setConvertingReader(SharedPtr<ConvertingReaderBase> t) {mImpl->mConvertingReader = t;}
 
 StreamMetadata::AuxDataMap& StreamMetadata::auxData() { return mImpl->mAuxData; }
 const StreamMetadata::AuxDataMap& StreamMetadata::auxData() const { return mImpl->mAuxData; }
@@ -1186,6 +1194,17 @@ Archive::isDelayedLoadingEnabled()
 }
 
 
+Name Archive::conversionToString(Archive::ScalarConversion conv)
+{
+    switch (conv)
+    {
+        case Archive::ScalarConversion::None:
+            return "";
+        case Archive::ScalarConversion::Half:
+            return typeNameAsString<math::half>();
+    }
+}
+
 namespace {
 
 struct NoBBox {};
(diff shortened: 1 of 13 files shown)


apradhana:
Hi, @konivo , I was trying to test this PR on my end using this function:
```cpp
void testDirectConversionToHalfGrid() {
    using namespace openvdb;
    std::string fileName = "./dragon.vdb";

    io::File file(fileName);
    file.open(false /* delay load*/, io::MappedFile::Notifier(), io::Archive::ScalarConversion::Half);

    HalfGrid::registerGrid();

    HalfGrid::Ptr grid;
    // Loop over the names of all of the grids in the file.
    for (io::File::NameIterator nameIter = file.beginName();
        nameIter != file.endName(); ++nameIter)
    {
        std::string gridName = nameIter.gridName();
        grid = gridPtrCast<HalfGrid>(file.readGrid(gridName));
        std::cout << "gridName = " << gridName << "grid = " << grid << "\n";
    }
}
```
My expectation is that `grid` should be non-nullptr. However, I couldn't make this to work. Am I missing something here? Also, with the release of `VDB`-13, `HalfGrid` registration should already happen in `openvdb::initialize()`, so if you rebase on top of the master branch, `HalfGrid::registerGrid()` shouldn't be necessary.

@danrbailey and @Idclip for viz.

apradhana:
The test I proposed for converting float to half passes for [commit ID b1f157](https://github.com/AcademySoftwareFoundation/openvdb/pull/1780/commits/b1f157830180beff35b4d0a30216319a9ea65f51). However, there are 12 unit-tests that are not-passing, which can be due to ABI change introduced in this commit (i.e. adding new member variables in `StreamMetadata::Impl`). The uni-tests that are not passing are `TestLeafIOTest.testBufferInt:TestLeafIOTest.testBufferFloat:TestLeafIOTest.testBufferDouble:TestLeafIOTest.testBufferByte:TestLeafIOTest.testBufferVec3R:TestFile.testWriteGrid:TestFile.testWriteMultipleGrids:TestFile.testReadGridDescriptors:TestFile.testEmptyGridIO:TestFile.testDelayedLoadMetadata:TestGridDescriptor.testIO:TestTree.testHalf:TestTree.testIO`.

@konivo and I decided the following action item: create convertingReader in GridDescriptor.cc while we are creating grid, then pass it down to `Archive::doReadGrid` and later to `RootNode::readTopology` and `LeafNode::read`. @apradhana is to test this approach and present it in the next TSC meeting.

openvdb/openvdb/io/Archive.cc
```diff
@@ -212,6 +212,8 @@ struct StreamMetadata::Impl
     bool mDelayedLoadMeta = DelayedLoadMetadata::isRegisteredType();
     uint64_t mLeaf = 0;
     uint32_t mTest = 0; // for testing only
+    std::string mDesiredScalarType = "";
+    SharedPtr<ConvertingReaderBase> mConvertingReader;
```
apradhana, line 216: Modifying `StreamMetadata::Impl` changes the ABI and introduces segfaults when performing I/O. See comment at the beginning.

openvdb/openvdb/tree/RootNode.h
```diff
@@ -2388,9 +2388,11 @@ RootNode<ChildT>::readTopology(std::istream& is, bool fromHalf)
     // Delete the existing tree.
     this->clear();
 
-    // Read a RootNode that was stored in the current format.
+    // converting reader for reading grid values
+    auto& convertingReader = io::ConvertingReader<ValueType>::get(is);
```
apradhana, line 2392: What makes more sense is to move this to `readTopologyWithValueType` since this function is templated with `SourceValueT`. In fact, I suspect that for `readTopologyWithValueType`, we should be able to make the conversion from `Real -> Half` type to work without `ConvertingReader` by providing the correct type in `SourceValueT`.

PR #1786, opened 2024-04-04
Fix undefined aliasing issue when compiling with clang-cl

bgottamd:
Fixes the [inner ValueT](https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/RootNode.h#L1157) alias throw an undefined error when compiling with clang/clang-cl on Windows by converting the `convertValue` into a template function.


bgottamd:
Adding a quick bump. This small code change is essential so that Blender can support both clang - [merged PR here](https://projects.blender.org/blender/blender/pulls/120317) - and openvdb. Without it, blender fails to build due to the cited issue.

LazyDodo:
any progress here? We'd like to get this working out of the box for blender, and while i have no issue including selected upstream landed patches for our dependencies, upstream unreviewed patches are a bit of a different story.

bgottamd:
Thanks @jo-rn - this does fix the issue with the header file compilation in blender. However, there is some performance degradation when this flag is used:

|File|args|fastest|slowest|avg|
|----|------|--------------------|---|--|
|[classroom ](https://download.blender.org/demo/test/classroom.zip)3840x2160 256 samples|header template fix|1209|1217|1213|
||-fno-delayed-template-parsing|1218|1228|1223|

This equates to ~1% drop in performance just from this flag. Hoping this templating change or a modification of the alias can still be merged in.

bgottamd:
I isolated the flag to projects that complained about the header issue. This ended up being only `extern_mantaflow` and 'cycles_scene'

anthony-linaro:
For those keeping track of this issue affecting Blender, I have created a PR to work around the issue using the approach described in this thread: https://projects.blender.org/blender/blender/pulls/126236

PR #1787, opened 2024-04-09
Initial support for HalfGrid and ComputeType

ghurstunither:
Initial support for `HalfGrid` and `ComputeType` in OpenVDB, where the latter enables intermediate `float` computations for `HalfGrid` to increase accuracy and speed.

* Uses `ValueToComputeMap` in `Types.h` to set `ComputeType` at the leaf level and then to propagate all the way up to the grid level.
  * For `math::half`, `ComputeType` is `float` and for any other type `T`, `ComputeType` is `T`.
* Defines and registers `HalfGrid` (half vectors not in this PR).
* Moved the inclusion/declaration of `math::half` from `Types.h` to its own `HalfDecl.h`
  * Added some arithmetic support for `math::half` in `Math.h` (to avoid casting in some places) so there were now circular dependancies.
* Refactors many tools and utilities to perform intermediate computations with `ComputeType` instead of `ValueType`.

Note, I decided not to define `openvdb::is_floating_point` and friends. Instead I opted to use `std::is_floating_point<ComputeType>()::value` instead.

________

Still to come in this PR

*Uunit tests for half and to verify no unit tests have broken.
* Quantification of the level of support in tools.
  * e.g. some tools didn't need refactoring, and so this PR doesn't reflect that they fully support half grids
* More examples within this description.


apradhana:
@ghurstunither. I'm wondering if you have ever compiled this PR with `EXPLICIT_TEMPLATE_INSTANTIATION = ON`.

danrbailey:
Should we close this PR and re-open using the feature branch as the source?

ghurstunither:
> Should we close this PR and re-open using the feature branch as the source?

Yes, I’ve gone ahead and closed this one.

PR #1794, opened 2024-04-16
Prevent integer overflow in NodeManager and LeafManager

ghurstunither:
For large enough grids, `leafCounts` in `initLeafArray()` and `nodeCounts` in `initNodeChildren()` can overflow as `std::vector<Index32>`.


ghurstunither:
> Needs a change for the log

Is that something I add in this PR?

> would be ideal if we could add a simple unit test that exceeds 32-bit integer.

I'm not sure how to do that without making a gigantic tree.

danrbailey:
I think there's actually more to talk about here. The Tree and RootNode also make assumptions that the number of leaf nodes cannot exceed 2^32:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/Tree.h#L116

The tree ones in particular are actually virtual functions, so if we want to bump this arbitrary limit, we should also change those methods as an ABI change.

ghurstunither:
> I think there's actually more to talk about here. The Tree and RootNode also make assumptions that the number of leaf nodes cannot exceed 2^32:
> 
> https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/tree/Tree.h#L116
> 
> The tree ones in particular are actually virtual functions, so if we want to bump this arbitrary limit, we should also change those methods as an ABI change.

Thanks for pointing this out. I think the following should return `Index64`:
* `unallocatedLeafCount`
* `leafCount`
* `nodeCount`

The other methods seem fine to return `Index32` if we assume we'll never exceed 2^32 for other nodes, which I think is reasonable.

openvdb/openvdb/tree/Tree.h
```diff
@@ -113,11 +113,11 @@ class OPENVDB_API TreeBase
     /// A tree with only a root node and leaf nodes has depth 2, for example.
     virtual Index treeDepth() const = 0;
     /// Return the number of leaf nodes.
-    virtual Index32 leafCount() const = 0;
```
danrbailey, line 116: Thanks for making all of those changes, I agree with the idea of moving to 64-bit unsigned integers across the whole API so that we can support larger grid sizes. One important thing to note, I believe that these types of change will affect ABI, because it's changing the return type of a virtual function which can (theoretically) lead to a restructuring of the vtable which can cause incompatibilities between different versions of OpenVDB. As a result, I believe it is required that we do an ABI check here and conditionally change this for ABI=12 only. It is good timing to discuss this because we're about the lock the ABI for next year!
```
#if OPENVDB_ABI_VERSION_NUMBER >= 12
virtual Index64 leafCount() const = 0;
#else
virtual Index32 leafCount() const = 0;
#endif
```

Would be good to get thoughts from @kmuseth and/or @jmlait on this one, and I recommend we wait until next week's TSC to finally merge this one.

ghurstunither: > I agree with the idea of moving to 64-bit unsigned integers across the whole API so that we can support larger grid sizes.

Do you mean the three leafCount-type methods I changed to 64-bit, or do you mean internal node count methods too?

> I recommend we wait until next week's TSC to finally merge this one.

Definitely agree.

danrbailey: I believe it's only required for the virtual methods on the Tree classes, but changing non-virtual methods may need logic elsewhere to be adjusted if it could be a mix of Index32 and Index64 for older ABIs.

openvdb/openvdb/tree/Tree.h
```diff
@@ -111,15 +115,31 @@ class OPENVDB_API TreeBase
     /// @brief Return the depth of this tree.
     ///
     /// A tree with only a root node and leaf nodes has depth 2, for example.
-    virtual Index treeDepth() const = 0;
+#if OPENVDB_ABI_VERSION_NUMBER >= 12
```
danrbailey, line 118: I don't think we need to change `treeDepth()`, the standard tree depth is only 4 levels and it is never going to exceed the 32-bit limit.

openvdb/openvdb/tree/InternalNode.h
```diff
@@ -274,9 +274,18 @@ class InternalNode
     /// Set the transient data value.
     void setTransientData(Index32 transientData) { mTransientData = transientData; }
 
+#if OPENVDB_ABI_VERSION_NUMBER >= 12
```
danrbailey, line 277: I think it might be cleaner if we can avoid the ABI logic in RootNode/InternalNode/LeafNode by only providing the Index64 version (regardless of ABI) and then doing a `static_cast<Index32>` in the virtual Tree methods that use the older ABI?

openvdb/openvdb/tree/InternalNode.h
```diff
@@ -274,9 +274,18 @@ class InternalNode
...
+    Index64 leafCount() const;
+    Index64 nonLeafCount() const;
+#else
     Index32 leafCount() const;
-    void nodeCount(std::vector<Index32> &vec) const;
     Index32 nonLeafCount() const;
+#endif
+    void nodeCount(std::vector<Index64> &vec) const;
+#if OPENVDB_ABI_VERSION_NUMBER < 12
```
danrbailey, line 285: Ideally, we should be aiming to offer both `nodeCount(std::vector<Index32>&)` and nodeCount(std::vector<Index64>&)` for this release except that the 32-bit one is marked as deprecated. If we make the 32-bit one disappear in 12.0 ABI=12, this will be a breaking API change without a chance for people to see the deprecation message.

openvdb/openvdb/unittest/TestTree.cc
```diff
@@ -2755,3 +2756,204 @@ TEST_F(TestTree, testNodeCount)
     EXPECT_EQ(tree.leafCount(), nodeCount2.front());// leaf nodes
     for (size_t i=0; i<nodeCount2.size(); ++i) EXPECT_EQ( nodeCount1[i], nodeCount2[i]);
 }
+
```
danrbailey, line 2759: I think this is a diffing error - I've moved these methods to TestRootNode and TestInternalNode now.

danrbailey (commented):
Just one small diffing error to resolve, otherwise this all looks good to me, thanks for making all those changes!

openvdb/openvdb/unittest/TestTree.cc
ghurstunither: Thanks for finding this! Should be fixed now.

danrbailey (approved):

PR #1799, opened 2024-04-25
Add NANOVDB_USE_SYNC_CUDA_MALLOC define to force sync CUDA malloc

w0utert:
In virtualized environments that slice up the GPU and share it between instances as vGPU's, GPU unified memory is usually disabled out of security considerations. Asynchronous CUDA malloc/free depends on GPU unified memory, so before, it was not possible to deploy and run NanoVDB code in such environments.

This commit adds macros CUDA_MALLOC and CUDA_FREE and replaces all CUDA alloc/free calls with these macros. CUDA_MALLOC and CUDA_FREE expand to asynchronous CUDA malloc & free if the following two conditions are met:

  - CUDA version needs to be >= 11.2 as this is the first version that supports cudaMallocAsync/cudaMallocFree
  - NANOVDB_USE_SYNC_CUDA_MALLOC needs to undefined

In all other cases, CUDA_MALLOC and CUDA_FREE expand to synchronous cudaMalloc/cudaFree.

Since NanoVDB is distributed as header-only, setting the NANOVDB_USE_SYNC_CUDA_MALLOC flag should be handled by the project's build system itself.

### DIFF
--- a/nanovdb/nanovdb/util/cuda/CudaAddBlindData.cuh
+++ b/nanovdb/nanovdb/util/cuda/CudaAddBlindData.cuh
@@ -62,7 +62,7 @@ cudaAddBlindData(const NanoGrid<BuildT> *d_grid,
     // extract byte sizes of the grid, blind meta data and blind data
     enum {GRID=0, META=1, DATA=2, CHECKSUM=3};
     uint64_t tmp[4], *d_tmp;
-    cudaCheck(cudaMallocAsync((void**)&d_tmp, 4*sizeof(uint64_t), stream));
+    cudaCheck(CUDA_MALLOC((void**)&d_tmp, 4*sizeof(uint64_t), stream));
     cudaLambdaKernel<<<1, 1, 0, stream>>>(1, [=] __device__(size_t) {
         if (auto count = d_grid->blindDataCount()) {
             d_tmp[GRID] = PtrDiff(&d_grid->blindMetaData(0), d_grid);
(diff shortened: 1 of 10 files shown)


kmuseth (commented):
great contribution ! However, before I approve it let me try out your fix in the private development branch of NanoVDB. I will sync that repo up with the this (public) repo in the coming week - it includes several changes and improvements :)

kmuseth (commented):
at closer inspection, why simply replace this existing line:

#if CUDART_VERSION < 11020

with

#if (CUDART_VERSION < 11020) || defined(NANOVDB_USE_SYNC_CUDA_MALLOC) // 11.2 introduced cudaMallocAsync and cudaFreeAsync

This avoid the need to introduce the new macro and also works with existing client code of NanoVDB that may already be using cudaMallocAsync and cudaFreeAsync

w0utert:
@kmuseth that was the first solution I tried, but it didn’t work, because the cudaMalloc and cudaFree would still be resolved to the CUDA ones and not the redefined ones from the NanoVDB header. Without any modification to our build I still got CUDA ‘not supported’ errors on allocations.

Maybe this can be worked around with some linker directives but that seems brittle and could require possible annoying changes to the build system of project that includes the NanoVDB headers.

w0utert:
This was on Linux by the way, so it’s interesting it did work on your side, I assume there could be some link differences between our builds. I could double check next week to verify again to be sure and to find out why.

kmuseth:
@w0utert I think you're right so I changed my implementation by simply placing the definitions of a namespace (nanovdb). So in the end my solution looks very much like yours, except I define functions vs macros. Let me create a PR that includes this fix plus a ton of other (long overdue) improvements to NanoVDB. I'll point you the the relevant changes so you can validate that it does indeed work for you.

A warning, this new PR introduces new namespace in NanoVDB tools so your client code might need to be tweaked a bit :)

w0utert:
@kmuseth sounds good, that’s actually nicer than using macro’s! I will try your PR early next week and report back, but I’m pretty sure it will work.

w0utert:
@kmuseth 
I tested the `feature/nanovdb_v32.7` branch from your fork and it works on Linux as well, thanks!

-Wouter

kmuseth:
@w0utert  excellent - so are you okay if we close this PR?

w0utert:
@kmuseth yes, you can close this PR!

kmuseth:
not sure why I left this PR open for so long (since we agreed to close it:)

PR #1815, opened 2024-05-14
Fix NanoVDB CI

apradhana:
This is an attempt to fix NanoVDB CI. For Windows, this boils down to downgrading the CMake to `cmake 3.28`. I follow an example given on[ this page ](https://github.com/marketplace/actions/actions-setup-cmake) to downgrade the CMake version. Perhaps I should also update the documentation here: https://www.openvdb.org/documentation/doxygen/dependencies.html#depKnownIssues.

I tried the fix that @Idclip suggested for MacOS, but the CI is still failing when the homebrew installs the dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -151,6 +151,12 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v3
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+      with:
+        cmake-version: '3.28.x'
+    - name: Print cmake version
+      run: cmake --version
     - name: path
       shell: pwsh
       run: |
--- a/.github/workflows/nanovdb.yml
+++ b/.github/workflows/nanovdb.yml
@@ -161,8 +161,8 @@ jobs:
     strategy:
       matrix:
         config:
-          - { runner: 'macos-11', cxx: 'clang++', build: 'Release' }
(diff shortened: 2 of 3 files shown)


kmuseth (approved):
looks good to me

.github/workflows/build.yml
```diff
@@ -151,6 +151,12 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v3
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+      with:
+        cmake-version: '3.28.x'
```
Idclip, line 157: I'd rather not use random people's actions here

Idclip:
@apradhana thanks for trying to look into this - but are you saying we only support CMake version up to 3.28 with this change? That doesn't sounds appropriate to me, can you link me to any files issues with the newer versions of CMake and zlib through vcpkg please

Idclip:
@apradhana the issue you linked is talking about a regression in CMake 3.27. Your PR seems to be back-porting CMake from 3.29 to 3.28. I'm not sure why your change is related to this.

It also looks like more recent Windows builds have been succeeding so I believe this has been fixed on vcpkg mainline.

Currently all our CI runners are offline (I have no idea why). When they are back up I will investigate this

Idclip:
All of the ASWF CI runners are still offline

apradhana:
Closing this PR. Will add a new PR on-top of [PR-1837](https://github.com/AcademySoftwareFoundation/openvdb/pull/1837).

PR #1818, opened 2024-05-16
Overhauled OpenVDB Clip SOP camera frustum padding + misc QOL changes

kjyager:
The Clip SOPs handling of frustum clipping was odd in several ways, and this caused confusion among users. Padding in particular was opaque and unintuitive.

This change implements more consistent and explainable padding for the camera clipping mode. Padding is specified and applied in normalized device coordinates. Rather than reading the same float3 padding parm used by the other modes, this new implementation adds float2 parms 'padwinx' and 'padwiny', specifying the amount of padding to add on the left/right and top/bottom sides of the camera frustum. These are enabled only in camera mode. Padding is applied symmetrically by default, similar to the old method, but can be made asymmetric by unlinking the parameters. There is no Z-axis padding value, as the near/far plane override parameters make this functionality redundant.

A `legacycamclip` toggle was added to optionally revert back to the original implementation for compatibility reasons.

To implement this change, it was necessary to make minor modifications to the clipping implementation in the OpenVDB library. A padding arg was added for the camera frustum signature of `clip`. A similar change was made to the `hvdb::drawFrustum()` function. All other changes were made just to the SOP itself.

**Additional Changes:**
+ Tooltips and documentation for the node's parameters have been updated and expanded.
+ Some parameters that were previously non-animatable can now be animated. There was no documented reason given for them being fixed, and I've observed no issues with their being animated. It is the padding (old and new) and near/far clipping plane parameters which have been changed.
+ The `xVoxelCount` argument of `hvdb::frustumTransformFromCamera()` is now set to the cameras x-resolution, avoiding the rounding error that previously made the resulting frustum slightly inaccurate.
+ Mask clipping mode now disables the y and z components of the 'padding' parm, to indicate that only uniform padding is supported in that mode.
+ When clipping in mask mode, the warning regarding nonuniform not being supported previously triggered whenever the three 'padding' values were not equal. This caused it to be raised for valid non-zero values as well. I've fixed it to trigger only when the y or z components are non-zero.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Clip.cc
+++ b/openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Clip.cc
@@ -41,12 +41,14 @@ public:
     {
     public:
         openvdb::math::Transform::Ptr frustum() const { return mFrustum; }
+        openvdb::math::Vec4d padding() const { return mPadding; }
     protected:
         OP_ERROR cookVDBSop(OP_Context&) override;
     private:
         void getFrustum(OP_Context&);
 
         openvdb::math::Transform::Ptr mFrustum;
+        openvdb::math::Vec4d mPadding = openvdb::Vec4d::zero();
     }; // class Cache
 
 protected:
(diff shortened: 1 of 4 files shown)


openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Clip.cc
```diff
@@ -82,23 +84,36 @@ newSopOperator(OP_OperatorTable* table)
...
+            "    Use the active voxels of a VDB volume from the second input as a clipping mask.\n"));
+
+    parms.add(hutil::ParmFactory(PRM_TOGGLE_E, "legacycamclip", "Legacy")
+        .setDefault(PRMzeroDefaults)
+        .setInvisible()
+        .setTooltip("Enables legacy camera clipping implementation and controls")
+        .setDocumentation(
+            "Enables legacy camera clipping implementation and controls.\n\n"
+            
+            "Implementation used prior to (OpenVDB version - / Houdini -). Overestimates "
```
kjyager, line 110: I've bookmarked the doc-string here so that version numbers can be added when they are known later.

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Clip.cc
```diff
@@ -454,24 +517,36 @@ SOP_OpenVDB_Clip::Cache::getFrustum(OP_Context& context)
     }
 
     const bool pad = (0 != evalInt("setpadding", 0, time));
-    const auto padding = pad ? evalVec3f("padding", time) : openvdb::Vec3f{0};
+    const bool useLegacyPadding = evalInt("legacycamclip", 0, 0.0f);
```
kjyager, line 520: Legacy clipping toggle is non-animatable, so evaluating at `t = 0` is correct.

kjyager:
Suggesting that Jeff Lait also review

PR #1821, opened 2024-05-19
Added feature to read points into vdb_tool using PDAL

tommatterson:
PDAL being Point Data Abstraction Library  https://pdal.io/
* Uses any available pdal reader plugin to read input file
* Writes RGB into Cd channel of points (also extended to PTS implementation)
* Adds a requirement on PDAL, so hidden behind a feature VDB_TOOL_USE_PDAL - off by default


kmuseth:
@tommatterson  at first glance this looks like a great contribution. Please allow me a little more time to do a proper review

openvdb_cmd/vdb_tool/include/Geometry.h
```diff
@@ -393,8 +407,16 @@ void Geometry::read(const std::string &fileName)
         this->readGEO(fileName);
         break;
     default:
-      throw std::invalid_argument("Geometry::read: File \""+fileName+"\" has an invalid extension");
-      break;
+#if VDB_TOOL_USE_PDAL
```
kmuseth, line 410: this looks strange to me. Why is this file type handles in the default case and not according to a file extension like the other file types?

kmuseth (changes requested):
please see question. Also, a unit-test is missing which is especially important since this is a pretty uncommon format (I've never come across a pdal file before:)

openvdb_cmd/vdb_tool/include/Geometry.h
tommatterson: Indeed, you've never come across a pdal file because pdal is not a file type itself, it is the "point data abstraction library"  pdal.io

PDAL supports many different file types through its own plugins - such as LAS, E57, PLY and so forth.  
https://pdal.io/en/2.8.1/stages/readers.html

As it is pluggable, I couldn't simply code in a supported file extension or list of extensions, and used `pdal::StageFactory::inferReaderDriver(fileName)` to determine if the file can be opened.  This was as neat as I could be without interrogating the library to build up a list of supported file extensions - which it doesn't directly let me access, they are referred to as `drivers` and don't need to match the extension.  

We could add a test for a .las or .e57 or even .npy file as they are supported by default, but it might be overkill to test every supported type.

swahtz: Hi @kmuseth, per our discussion I added a unittest which will use PDAL to read an ASCII formatted file that it supports.

kmuseth: now that I understand that pdal can import multiple file types I do agree that it makes sense to place this is the "default" case. So, please ignore my comment above. The only remaining thing a minimum unit-test to this code at get's compiled and run during the CI

kmuseth (approved):

swahtz:
Hi @danrbailey, we could include this in vdb12.0 now if we wanted to, we resolved the discussions and added a unit test.

openvdb_cmd/vdb_tool/include/Geometry.h
swahtz: Great, I've added that unit test which uses PDAL functionality to read some points similar to other similar format readers' unit tests.

PR #1822, opened 2024-05-21
Digital Domain 3.0, Inc. approved relicense from MPL-2.0 to Apache-2.0

plattfot:
Our approval is marked by the addition of our name to the RE-LICENSE_NOTE.txt.

### DIFF
--- a/RE-LICENSE_NOTE.txt
+++ b/RE-LICENSE_NOTE.txt
@@ -8,3 +8,4 @@ DreamWorks Animation
 Side Effects Software Inc.
 Blender Foundation
 NVIDIA Corporation
+Digital Domain 3.0, Inc.


jmlait:
The DCO test is failing, which is a bit confusing to me as the commit appears to have been signed off properly?

plattfot:
Could it be an issue with this is not being signed by the commit author (i.e. me)?

plattfot:
Yeah, that was the issue. I just pushed a commit with me signing off the commit and not Deke.  I missed that the DCO must be signed off by the author.

jmlait (approved):

jmlait:
Excellent, thank you, we can commit when we get a second TSC signoff.

apradhana (approved):

kmuseth (approved):

PR #1870, opened 2024-08-13
More CI Fixes

apradhana:
Fix NanoVDB CI: `NanoVDB` lite and on Windows: upgrade to CUDA 12.4 for Visual Studio 2022, allow use of unsecure node version, use node 16 in ax, build, docs, houdini, and weekly workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,6 +86,9 @@ jobs:
           - { cxx: g++,     image: '2022-clang11',   abi: '9',  build: 'Release', cmake: '-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON' }
       fail-fast: false
     steps:
+    - name: Enable Node 16
+      run: |
+        echo "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true" >> $GITHUB_ENV
     - uses: actions/checkout@v3
     - name: pybind11
       #if: contains(matrix.config.image, '2023') == false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,8 @@ jobs:
       fail-fast: false
     steps:
     - name: Enable Node 16
+      # Solution taken from https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default
+      if: contains(matrix.config.image, '2022')
       run: |
         echo "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true" >> $GITHUB_ENV
     - uses: actions/checkout@v3
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -55,6 +55,8 @@ jobs:
     #   run: pip install epydoc
     - name: install_latex
       run: |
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -86,6 +86,9 @@ jobs:
           - { cxx: g++,     image: '2022-clang11',   abi: '9',  build: 'Release', cmake: '-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON' }
       fail-fast: false
     steps:
+    - name: Enable Node 16
+      run: |
+        echo "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true" >> $GITHUB_ENV
```
danrbailey, line 91: Thanks @apradhana, this seems to fix the nodejs-related CI issues which appears to be causing wide-spread chaos with GHA, not just for VDB. However, I think this is only an issue for the old vfx 2022 images though, it would be good if we can gate this temporary fix to just those versions so that when we remove the 2022 images, it's more obvious we can also remove this temporary fix. I think we only have to support older glibc for a few more months (until VDB 12.0 is released).

I haven't tried this, but something like `if: contains(matrix.config.image, '2022') == true` if that works?

.github/workflows/docs.yml
```diff
@@ -51,7 +54,10 @@ jobs:
     # - name: install_epydoc
     #   run: pip install epydoc
     - name: install_latex
-      run: yum -y install texlive-latex-bin texlive-dvips texlive-collection-fontsrecommended texlive-collection-latexrecommended
+      run: |
+        sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
```
danrbailey, line 58: Could you add a comment documenting what these patches are doing?

.github/workflows/nanovdb.yml
```diff
@@ -153,21 +158,21 @@ jobs:
...
+          - { runner: 'macos-12', cxx: 'clang++', build: 'Debug'   }
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
       - name: install_deps
         run: ./ci/install_macos.sh
       - name: build
         run: >
-          ./ci/build.sh -v
+          sudo ./ci/build.sh -v
```
danrbailey, line 170: Why do we need sudo here?

swahtz: The error that I believe this is trying to fix is this error which happens very early in the macos-build job:
```
getopt: illegal option -- l
getopt: illegal option -- g
```

This was happening to me on my Mac when trying to test out this action.  This is probably happening because the bash script is picking up the system's getopt and not the brew-installed `gnu-getopt` package installed binary.  The brew-installed `getopt` is attempting to be path'd here in the `.ci/install_macos.sh`

https://github.com/AcademySoftwareFoundation/openvdb/blob/6719224567c11211df7b239c13fa0025ae2a062c/ci/install_macos.sh#L34

But that doesn't seem to be working if we're getting this getopt error.

swahtz: I believe that this `getopt` issue is also the source of the problems with the `macos-houdini` CI jobs because the line crashing the build is also a `getopt` command

danrbailey (approved):

apradhana:
Thank you for approving this, @danrbailey!

PR #1875, opened 2024-08-17
fixed precision issue in unit-test on OSX M3

kmuseth:
fixed precision issue in unit-test on OSX M3


apradhana (approved):
Looks good to me!

PR #1894, opened 2024-09-10
Fixing build issue caused by adding cwiseAdd in `math/Vec3.h`

apradhana:
Template specialization of `cwiseAdd` for `Vec3` half-type should be in be in `Types.h` instead of in `math/Vec3.h`.


danrbailey (approved):

apradhana:
Thank you for approving this, @danrbailey!

PR #1897, opened 2024-09-13
Fix Windows Static Build

danrbailey:
Earlier versions of Blosc required explicit library prefix/suffix modification. With more recent versions, this is no longer required and resolves a long-standing issue on Windows static builds where a "lib" prefix was being incorrectly applied and failing to find Blosc dependencies like LZ4 as a result. This change was informed by the patches being applied by package maintainers to build OpenVDB static libraries.


apradhana (approved):
Thanks for finding out what's wrong with Windows static build @danrbailey! I was able to narrow it down to a problem in `FindBlosc.cmake` and you found the problem. This looks good.

PR #1837, opened 2024-06-01
Fixed Various CI

Idclip:

PR #1898, opened 2024-09-13
Fix MacOS Houdini Build

danrbailey:
There was a missing backslash in the changes made in this PR (https://github.com/AcademySoftwareFoundation/openvdb/pull/1837) which I believe addresses the recent failures in the macos-houdini build. For good measure, I also used the updated version of bash from homebrew in the build script.


danrbailey:
Also needed to pick the arm64 version of Houdini, use Houdini tbb and update the homebrew llvm path to the Apple Silicon location. This will need the Houdini weekly builds to run to cache the Houdini arm64 downloads before this one will succeed.

jmlait (approved):

PR #1899, opened 2024-09-13
Fix NanoVDB Build

danrbailey:
This fixes an issue building the NanoVDB unit tests on Windows due to these errors:
```
unittest\TestNanoVDB.cu(2555): error : namespace "std" has no member "setw" 
unittest\TestNanoVDB.cu(2555): error : namespace "std" has no member "setfill" 
```

(@kmuseth @apradhana)


apradhana:
Thanks for catching this, @danrbailey.

PR #1902, opened 2024-09-17
WIP: Setting up fVDB GHA tests

swahtz:


danrbailey:
This PR seems to still be triggering lots of CI runs it shouldn't. I think the `.github` directory is ignored by default, so my best guess is that it's perhaps because the changes to exclude `fvdb/**` are not part of the master branch (ie the trigger is based on the result of merging this branch to master). Perhaps it makes sense to create a new PR just to add those changes and then we merge that? Also, if you merge `origin/master` to this branch, that will at least remove a bunch of the more expensive Windows builds.

swahtz:
> This PR seems to still be triggering lots of CI runs it shouldn't. I think the `.github` directory is ignored by default, so my best guess is that it's perhaps because the changes to exclude `fvdb/**` are not part of the master branch (ie the trigger is based on the result of merging this branch to master). Perhaps it makes sense to create a new PR just to add those changes and then we merge that? Also, if you merge `origin/master` to this branch, that will at least remove a bunch of the more expensive Windows builds.


Yes perhaps I'll do that for the moment.  I've been manually cancelling all the tests I don't need thinking I've solved these runner issues and the fix is 'just around the corner' but the corner keeps getting farther away.

PR #1903, opened 2024-09-17
Move all debug and static windows builds to the weekly CI

danrbailey:
As we recently discussed, this moves all debug and static windows builds to the weekly CI instead of running on every PR. This leaves just the CI for the shared library release build for OpenVDB and for NanoVDB. A potential next step is to consolidate these two builds under a single CI job as we are still seeing long times for installing all of the Windows dependencies (when we are not making use of the wider machines). It is always possible to manually kick off a weekly CI run when making substantial changes or build system changes.

This is intended to significantly reduce our use of ASWF CI budget as Windows is the most expensive CI and is our longest running.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,29 +130,16 @@ jobs:
       run: ccache --evict-older-than 1d
 
   windows:
-    # Windows CI. Tests static and dynamic builds with MT and MD respectively.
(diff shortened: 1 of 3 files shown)


jmlait (approved):

apradhana (approved):
Thanks for moving the Windows Static test for Open and Nano to the weekly CI, @danrbailey. It looks good to me.

Issue #1917, opened 2024-09-26
[BUILD] Migrate AX Unit Tests from CppUnit to GTest


danrbailey:
A few years ago, we migrated OpenVDB unit tests from CppUnit to GTest:

https://github.com/AcademySoftwareFoundation/openvdb/pull/875

However, we never got around to migrating the OpenVDB AX unit tests across. The core unit tests were relatively straight-forward to migrate and CppUnit sometimes causes issues because it's very old and is no longer actively maintained (the last release was 2017). We'd like to finally eliminate this dependency by porting the OpenVDB AX unit tests.

https://freedesktop.org/wiki/Software/cppunit

To build these unit tests, make sure that you are building AX and that you enable the AX unit tests (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/CMakeLists.txt#L77).

PR #1919, opened 2024-09-27
Replace CppUnit with GoogleTest in AX tests

tstraubinger:
This is a work-in-progress pull request. I've begun by adding GoogleTest to the AX unit test runner and am currently running both the existing CppUnit test suite as well as any added GoogleTest tests. The intention is to migrate the test cases one-by-one and then remove CppUnit completely afterwards.

Resolves https://github.com/AcademySoftwareFoundation/openvdb/issues/1917


tstraubinger:
These changes are ready for review. I've completely removed CppUnit from the AX tests but it remains a dependency at the top level. I think I'll put in a second pull request to remove it there as well, separately from this one.
@Idclip @richhones @danrbailey

openvdb_ax/openvdb_ax/test/integration/TestStandardFunctions.cc
```diff
@@ -173,7 +48,7 @@ inline void testFunctionOptions(unittest_util::AXTestHarness& harness,
     const ast::Tree::Ptr syntaxTree = ast::parse(code.c_str());
     timer.stop();
 
-    CPPUNIT_ASSERT_MESSAGE(syntaxTree, "Invalid AX passed to testFunctionOptions.");
+    ASSERT_TRUE(syntaxTree) << "Invalid AX passed to testFunctionOptions.";
```
tstraubinger, line 51: The arguments to CPPUNIT_ASSERT_MESSAGE were reversed here. If I understand correctly then, this was ensuring that the string literal evaluates to true (which as a non-null pointer, it does) and would otherwise have printed  out the pointer to the AST (perhaps via shared_ptr's operator<<(ostream&))

openvdb_ax/openvdb_ax/test/backend/TestFunctionTypes.cc
```diff
@@ -1699,29 +1662,43 @@ TestFunctionTypes::testIRFunctions()
...
+        // NOTE: GTest's ASSERT_* macros create a 'return;' statement
+        // which errors here. Exceptions are being used instead to
+        // exit early.
+
         // test the builder is pointing to the correct location
-        CPPUNIT_ASSERT(&B != &_B);
+        if (&B == &_B)
+        {
+            throw std::runtime_error("Builder should be different");
+        }
```
tstraubinger, line 1673: Some opinionated changes: GTest `ASSERT_*` macros can't be used in a non-void function because they expand to a void return statement (I suppose Google wanted to support code bases with C++ exceptions disabled). I've replaced the assertions with equivalent if-statements and explicit C++ exceptions. I've confirmed that GTest reports a failure as expected when these are thrown here.

openvdb_ax/openvdb_ax/compiler/Logger.h
```diff
@@ -197,7 +199,9 @@ class OPENVDB_AX_API Logger
 
 private:
 
-    friend class ::TestLogger;
+    FRIEND_TEST(TestLogger, testParseSetsTree);
+    FRIEND_TEST(TestLogger, testAddError);
+    FRIEND_TEST(TestLogger, testAddWarning);
```
tstraubinger, line 204: It might have been possible to avoid these changes to a header file in order to test private members, but this is the GTest- recommended approach. See https://google.github.io/googletest/advanced.html#testing-private-code

tstraubinger:
I've created a secondary branch off of this one that further removes all mentions of CppUnit from the build dependencies and instructions. Shall I create a merge request into this branch, wait until this is merged, or just push the changes to this branch?
https://github.com/tstraubinger/openvdb/tree/feature/remove_cppunit

danrbailey:
> I've created a secondary branch off of this one that further removes all mentions of CppUnit from the build dependencies and instructions. Shall I create a merge request into this branch, wait until this is merged, or just push the changes to this branch? https://github.com/tstraubinger/openvdb/tree/feature/remove_cppunit

Thanks, I'm fine to have these changes in this branch as they're all closely related. I've just approved the CI running, but note there are some conflicts, so this branch likely needs to be updated and the conflicts resolved which may affect the result of the CI.

tstraubinger:
Changes have been rebased onto the current master branch and I've included removing CppUnit as a dependency. Whitespace that was causing a linting check to fail has also been removed. Unfortunately I don't know where to see other failed checks now that newer changes have been pushed.

openvdb_ax/openvdb_ax/test/main.cc
```diff
@@ -37,218 +15,6 @@ int sGenerateAX = false;
...
-            getTestNames(allTests, tests);
-            delete tests;
-            for (const auto& name: allTests) { std::cout << name << "\n"; }
-            return EXIT_SUCCESS;
-        } else if (arg == "-shuffle") {
-            shuffle = true;
-        } else if (arg == "-v") {
-            verbose = true;
-        } else if (arg == "-g") {
-            sGenerateAX = true;
```
Idclip, line 145: This used to be callable from the test binary - in practice we don't use it very much, so in the rare cases where the tests need to be re-generated, I think it's fine for this to need a re-compile. So just commenting for future visibility, nothing else to do

openvdb_ax/openvdb_ax/test/integration/TestStandardFunctions.cc
Idclip: Good catch!

Idclip (commented):
@tstraubinger thank you very much for these changes - on the whole these look great, there's a few files I need to look at a bit closer (mainly TestFunctionTypes due to the logic switch within the non-void lambdas), but my only comment is to avoid including gtest in the core ax lib Logger/Executables files (re-work the friend setup)

Idclip:
I've taken these changes and resolved the conflicts here: https://github.com/AcademySoftwareFoundation/openvdb/pull/2068

Closing this, will aim to get 2068 in as soon as possible, thanks again @tstraubinger!

PR #1925, opened 2024-10-04
Add 2024 CI Images and Drop 2021 CI Images

danrbailey:
This drops CI jobs for the VFX Reference Platform 2021 images which we are about to drop support for and introduces 2024 images which we haven't used until now. I've also removed the clang-specific images in favor of the base image (where possible, AX does not yet work on LLVM>15). There was an outstanding issue where the version of glfw was incorrectly installed in the 2023 image so this is now installed from source to resolve this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,13 +79,13 @@ jobs:
           # @note  we specifically use clang15.0 (not clang15) here as the newest
           #  versions of the clang15.X containers have some issues with the GLFW
           #  installation
-          - { cxx: clang++, image: '2023-clang15.0', abi: '11', build: 'Release', cmake: '' }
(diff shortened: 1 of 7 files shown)


.github/workflows/weekly.yml
```diff
@@ -157,7 +157,7 @@ jobs:
       # @note  we specifically use clang15.0 (not clang15) here as the newest
       #  versions of the clang15.X containers have some issues with the GLFW
       #  installation
```
Idclip, line 159: remove this comment

.github/workflows/build.yml
```diff
@@ -79,13 +79,13 @@ jobs:
           # @note  we specifically use clang15.0 (not clang15) here as the newest
           #  versions of the clang15.X containers have some issues with the GLFW
           #  installation
```
Idclip, line 81: Remove this comment

Idclip (approved):

PR #1926, opened 2024-10-09
Update NanoVDB to 2024 OS Image

danrbailey:
NanoVDB for Linux is still built on the old 2022 EL7 images and the version of Cuda being used here (11.6) is over two years old. I think it would be a good idea for this to be updated before we release VDB 12.0. I've included the initial change on a feature branch which is currently failing and needs work to update it. Hopefully the changes required are trivial.

(Tagging all the NVidia people - @apradhana @kmuseth @matthewdcong @swahtz)


swahtz:
Hi @danrbailey, looked into this a bit and fixed up the nanovdb action for RHEL8.  It looks like this 2024 image comes with the cuda 12.6 toolkit, so I've opted to use that pre-installed `nvcc` compiler

I'm just having issues with the clang++ builds erroring that we need to compile with `-fPIE`.  It looks like perhaps the default `-fPIE` behaviour changed in clang 14:

https://github.com/llvm/llvm-project/commit/1042de90589c74c6998c127d1ee1b33ff3efb8cc

danrbailey:
> Details

Thanks for taking a look at this. From what I can tell, almost all of our Linux environments where we use VDB should already be PIE/PIC as it's the default in RHEL7+ and GCC6+/Clang14+. It's a bit surprising to me that we aren't picking up this default. If we set `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` as a cmake argument, does this resolve this issue?

swahtz:
> > Details
> 
> Thanks for taking a look at this. From what I can tell, almost all of our Linux environments where we use VDB should already be PIE/PIC as it's the default in RHEL7+ and GCC6+/Clang14+. It's a bit surprising to me that we aren't picking up this default. If we set `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` as a cmake argument, does this resolve this issue?

Yes I was a bit confused as well...  but this is the error we're seeing when building in this image with clang 14 and I have the same issue when I've replicated it with this image+compiler:

https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/11432517892/job/31803124110?pr=1926#step:6:365

danrbailey:
> Yes I was a bit confused as well... but this is the error we're seeing when building in this image with clang 14 and I have the same issue when I've replicated it with this image+compiler:
> 
> https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/11432517892/job/31803124110?pr=1926#step:6:365

It sounds like we might be hitting an incompatibility where clang 17 under this image has PIC/PIE disabled by default, but the Cuda libraries we are linking against do? I don't think that's completely surprising because I believe we compile clang 17 ourselves (as part of the OS image), but I'm not aware we have intentionally configured it like this. Can we confirm that the clang we are using in the image has PIC/PIE disabled? Then I think we should ask the CI WG about this on slack.

swahtz:
I can indeed confirm that building with `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` compiles correctly.  I'll ask the CI WG about this, I can't seem to see definitively if the `CLANG_DEFAULT_PIE_ON_LINUX` is set in our clang build.

swahtz:
Per CI WG slack chat, I made a ticket about this clang behaviour issue and have just added the `CMAKE_POSITION_INDEPENDENT_CODE` to the test job for the time being.

https://github.com/AcademySoftwareFoundation/aswf-docker/issues/228

apradhana (approved):

PR #1927, opened 2024-10-09
Drop ABI=9 support

danrbailey:
This drops ABI=9 support in preparation for VDB 12.0, but keeps the version as 11.x for the time being.

This also edits the deprecation document to accurately reflect the change in policy around when to remove support for obsolete ABIs.


danrbailey:
(Note that CI is expected to fail until https://github.com/AcademySoftwareFoundation/openvdb/pull/1925 is merged, I shall update this PR when those changes go in)

PR #1928, opened 2024-10-09
Update License to Apache 2.0

danrbailey:
To merge the update-to-date `feature/apache2` branch into master. No code changes needed beyond #1813.

@dgovil

PR #1931, opened 2024-10-09
RootNode API Changes

danrbailey:
This implements a number of changes to the RootNode. These methods are now part of the public API:

- RootNode::tileCount() - renamed from getTileCount
- RootNode::activeTileCount() - renamed from getActiveTileCount
- RootNode::inactiveTileCount() - renamed from getInactiveTileCount
- RootNode::hasKey()
- RootNode::coordToKey()

This new method is added - RootNode::deleteChildOrTile()

This introduces the ability to delete a child or tile from a RootNode. This was previously not possible, except for an unintuitive workaround by adding a background tile (which implicitly replaced the child or tile), then to erase all background tiles.

A bug was fixed in RootNode::setOrigin() where the origin was updated _before_ the error was thrown allowing the user to catch the error and potentially leaving the root in an invalid and unsupported state.

Finally, I added a new RootNode unit test that added additional validation that wasn't previously being done and to confirm the changes in behaviour mentioned above. (Note this unit test is added using SPDX Apache-2.0 so #1928 needs to be merged first).


kmuseth (approved):
looks good to me

PR #1933, opened 2024-10-11
Fix Windows CI

danrbailey:
The Windows vcpkg CI is failing again, I ported the bash install script to power shell and this appears to have fixed the issue.


swahtz (approved):
Looks great to me, thanks @danrbailey

PR #1941, opened 2024-10-22
Add support for pip install

matthewdcong:
This PR adds preliminary support for pip install. Assuming that the dependencies required for building OpenVDB and NanoVDB are installed, then `pip install <openvdb_source_root>` will

1. Build OpenVDB, NanoVDB, and their corresponding Python bindings.
2. Install these Python bindings and dependent libraries in the `site-packages` directory
3. Enable the user to `import openvdb` and `import nanovdb` when launching Python.

This is a first-step towards making installation of OpenVDB easier for new users, and will help users dive into the library via the Python interface.


pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+description= "Python bindings for OpenVDB: sparse volume data structure and tools."
+dependencies = [
+    "numpy",
+]
+authors = [
+  { name = "OpenVDB Developer Team", email = "openvdb-dev@lists.aswf.io" },
+]
+requires-python = ">=3.8"
+classifiers = [
+  "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)",
```
swahtz, line 17: Apache 2 now, yes?

matthewdcong: Good catch, thanks!

matthewdcong:
Why is the CLA check failing?

openvdb/openvdb/CMakeLists.txt
```diff
@@ -718,29 +718,58 @@ endif()
...
 # Installation
 
-if(OPENVDB_CORE_STATIC)
-  install(TARGETS openvdb_static
-    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-  )
-endif()
+if(SKBUILD)
```
danrbailey, line 721: Wondering if there's a cleaner way to handle this kind of case without all of this conditional branching. What about using an INSTALL_PREFIX and including the forward slash in the prefix (although potentially needs thought for multi-platforms):
```
if (SKBUILD)
  set(INSTALL_PREFIX "openvdb/")
else
  set(INSTALL_PREFIX "")
endif()

install(TARGETS openvdb_static
  RUNTIME_DESTINATION ${INSTALL_PREFIX}${CMAKE_INSTALL_BINDIR}
)
```
Ultimately, it would be nice to use less `if (SKBUILD)` logic if possible?

matthewdcong: Yeah, I'm not a huge fan of all the `SKBUILD` logic either. I'm happy to take another look.

In an ideal world, it'd be possible to separate them entirely, but in practice Python packaging has very limited support for _two_ modules in a single package. Thus, we need to delegate a lot of the pip install location logic to CMake since Python just doesn't provide us with the expressiveness to do so.

matthewdcong: Thanks for the suggestions @danrbailey ! I've consolidated the conditional branching significantly. Please take another look and let me know what you think.

nanovdb/nanovdb/CMakeLists.txt
```diff
@@ -322,16 +322,16 @@ if(TARGET Threads::Threads)
   target_link_libraries(nanovdb INTERFACE Threads::Threads)
 endif()
 
-set(NANOVDB_INSTALL_INCLUDE_DIR ${CMAKE_INSTALL_INCLUDEDIR}/nanovdb)
-set(NANOVDB_INSTALL_CUDA_DIR ${NANOVDB_INSTALL_INCLUDE_DIR}/cuda)
-set(NANOVDB_INSTALL_IO_DIR ${NANOVDB_INSTALL_INCLUDE_DIR}/io)
-set(NANOVDB_INSTALL_MATH_DIR ${NANOVDB_INSTALL_INCLUDE_DIR}/math)
-set(NANOVDB_INSTALL_TOOLS_DIR ${NANOVDB_INSTALL_INCLUDE_DIR}/tools)
+set(NANOVDB_INSTALL_ROOT_DIR ${NANOVDB_INSTALL_INCLUDEDIR}/nanovdb)
+set(NANOVDB_INSTALL_CUDA_DIR ${NANOVDB_INSTALL_ROOT_DIR}/cuda)
```
danrbailey, line 326: There are a few of these variables that are swapping INCLUDE paths for ROOT paths?

danrbailey (commented):
This is much better, thanks Matt! Just one small comment which I think is probably a find-replace typo...

nanovdb/nanovdb/CMakeLists.txt
matthewdcong: Not a typo I think.

Prior to this change, `NANOVDB_INSTALL_INCLUDE_DIR` (note the underscore between include and dir) was already defined. I've renamed this variable to `NANOVDB_INSTALL_INCLUDE_DIR` to `NANOVDB_INSTALL_ROOT_DIR` in order to avoid confusion with the newly introduced `NANOVDB_INSTALL_INCLUDEDIR` (that defaults to `CMAKE_INSTALL_INCLUDEDIR`).

danrbailey: Ah, that makes sense, thanks for explaining.

danrbailey (approved):

apradhana (approved):
Tested on Linux with the latest changes. LGTM.

danrbailey (approved):

PR #1956, opened 2024-10-30
Fix a memory-leak in RootNode::deleteChildOrTile()

danrbailey:
The address sanitizer CI picked up that there was a memory leak in the `RootNode::deleteChildOrTile()` method and I realize that the NodeStruct doesn't delete the child on destruction so I've fixed this by explicitly deleting it.

I also added a warning that this method will invalidate existing iterators as previously requested.

(@kmuseth)


jmlait (approved):
I see.... the RootNode::NodeStruct::set method will delete the existing child, which means the clip methods:
            setTile(this->findCoord(xyz), bgTile); // delete any existing child node first
            mTable.erase(xyz);
would be implicitly deleting the child in the setTile method.

PR #1958, opened 2024-10-31
VDB 12.0.0 Release

danrbailey:
The release branch for VDB 12.0.0. This bumps the version up, updates the change log and makes a small fix to the CI docs workflow.

This also bumps up a number of minimum versions for our dependencies. We now support the VFX Reference Platforms 2023, 2024, 2025. These are the most notable version bumps:

* CMake 3.18 -> 3.20
* GCC 9.3.1 -> 11.2.1 (C++ ABI=0 no longer supported)
* MSVC 2019 -> MSVC 2022
* Python 3.9 -> 3.10
* Boost 1.73 -> 1.80
* LLVM 10.0.0 -> 13.0.0

I have also updated the CI accordingly. Finally, the address sanitizer picked up a bug in the RootNode, which I have also fixed (using emplace for an existing element in one case).


cmake/config/OpenVDBVersions.cmake
```diff
@@ -34,24 +34,23 @@ if(NOT DISABLE_DEPENDENCY_VERSION_CHECKS)
...
   set(MINIMUM_TBB_VERSION 2020.3)
-  set(MINIMUM_LLVM_VERSION 10.0.0)
+  set(MINIMUM_LLVM_VERSION 13.0.0)
   set(MINIMUM_BLOSC_VERSION 1.17.0)
   set(MINIMUM_GLFW_VERSION 3.1)
 
-  set(MINIMUM_PYTHON_VERSION 3.9.1)
-  set(MINIMUM_NUMPY_VERSION 1.20.0)
+  set(MINIMUM_PYTHON_VERSION 3.10)
+  set(MINIMUM_NUMPY_VERSION 1.23.0)
```
apradhana, line 54: Do we need to add `MINIMUM_NANOBIND_VERSION` here?

openvdb/openvdb/openvdb.cc
```diff
@@ -26,14 +26,20 @@
...
+#endif
 
 // If using a future OPENVDB_ABI_VERSION_NUMBER, issue an error directive.
 // This can be optionally suppressed by defining:
 //   OPENVDB_USE_FUTURE_ABI_<VERSION>=ON.
-#ifndef OPENVDB_USE_FUTURE_ABI_12
-    #if OPENVDB_ABI_VERSION_NUMBER == 12
+#ifndef OPENVDB_USE_FUTURE_ABI_13
+    #if OPENVDB_ABI_VERSION_NUMBER == 13
         #error ABI = 12 is still in active development and has not been finalized, \
```
apradhana, line 41: I think the error should be ABI = 13 is still in active development.

cmake/config/OpenVDBVersions.cmake
danrbailey: I just tried that, but it seems that nanobind does not store the version number in `2.0.0`, so I'll revert that change.

This issue was reported here (https://github.com/wjakob/nanobind/pull/618) and fixed in `2.1.0`, so I'll document the future minimum version as that and we can bump it next year.

apradhana: Thanks for trying this. You're right, seems that adding a requirement for NanoBind version 2.0.0 breaks a few of the CIs.

CMakeLists.txt
```diff
@@ -479,8 +479,8 @@ endif()
 if(OPENVDB_BUILD_PYTHON_MODULE OR (OPENVDB_BUILD_NANOVDB AND NANOVDB_BUILD_PYTHON_MODULE))
 
   # Call find_package(Python ...)
-  find_package(Python 3.8 REQUIRED COMPONENTS Development Interpreter)
-  find_package(nanobind REQUIRED)
+  find_package(Python ${MINIMUM_PYTHON_VERSION} REQUIRED COMPONENTS Development Interpreter)
+  find_package(nanobind ${MINIMUM_NANOBIND_VERSION} REQUIRED)
```
apradhana, line 483: Because nanobind does not store version number in `2.0.0` to be ingested by CMake, should we take the ${MINIMUM NANOBIND_VERSION` here?

apradhana (approved):

apradhana (approved):
Thanks for addressing my comments. Other than my comment on setting the minimum NanoBind version, everything else looks good. Once all the CIs passed, this looks good.

Idclip:
> GCC 9.3.1 -> 11.2.1 (C++ ABI=0 no longer supported)

Can you still build with ABI=11 and the old C++ ABI?

danrbailey:
> > GCC 9.3.1 -> 11.2.1 (C++ ABI=0 no longer supported)
> 
> Can you still build with ABI=11 and the old C++ ABI?

Yes, it still works. In fact the only thing we do regarding ABI currently is trying to align with Houdini's ABI (https://github.com/AcademySoftwareFoundation/openvdb/blob/master/cmake/OpenVDBHoudiniSetup.cmake#L349) and I didn't remove that logic in this release. Otherwise, you get whatever the compiler/compiler flags request.

However, it's not guaranteed to continue working. I removed all the CI jobs that test with GCC 9 / ABI=0 and if you build with GCC 9, you'll get a warning that it is unsupported which you have to suppress.

.github/workflows/ax.yml
```diff
@@ -126,10 +120,9 @@ jobs:
     strategy:
       matrix:
         config:
-          #@note llvm10 never got its own brew formula...
+          #@note llvm15 never got its own brew formula...
```
Idclip, line 123: remove this comment

.github/workflows/ax.yml
```diff
@@ -126,10 +120,9 @@ jobs:
...
           # Last macos runner befor M1 (macos-14)
-          - { runner: 'macos-13', cxx: 'clang++', build: 'Release', llvm: '12' }
-          - { runner: 'macos-13', cxx: 'clang++', build: 'Release', llvm: '13' }
+          - { runner: 'macos-13', cxx: 'clang++', build: 'Release', llvm: '15' }
```
Idclip, line 125: You remove llvm14 from the weekly workflow - add it back here? Also add 13 back if our minimum is 13

Idclip (approved):
minor comments about the llvm ci, otherwise lgtm, thanks a lot @danrbailey

.github/workflows/ax.yml
danrbailey: Done, thanks

.github/workflows/ax.yml
danrbailey: Yep, I just added 13 back in (we do also test 14 via Windows currently).

PR #1971, opened 2024-11-20
added read/write support of OFF files to vdb_tool

kmuseth:


openvdb_cmd/vdb_tool/include/Geometry.h
```diff
@@ -519,6 +544,52 @@ void Geometry::readPDAL(const std::string &fileName)
...
+    std::getline(is, line);
+    if (line!="OFF") throw std::invalid_argument("Geometry::readOFF: expected header \"OFF\" but read \"" + line + "\"");
+    int vtxCount=0, faceCount=0, edgeCount=0;
+    while (vtxCount==0 && std::getline(is, line)) {
+        if (line[0]=='#') continue;
+        std::istringstream iss(line);
+        iss >> vtxCount >> faceCount >> edgeCount;
+    }
+    Vec3f p;
+    for (int i=0; i<vtxCount && std::getline(is, line); ++i) {
```
swahtz, line 570: It seems like it's possible in this format to have blank lines and comment lines in between the header, vertices or faces sections.  It even seems you can have line breaks inside of these sections.  Is it worth being robust to those?

https://www.software3d.com/StellaManual.php?prod=stella4D#import
http://www.geomview.org/docs/html/OFF.html
https://people.sc.fsu.edu/~jburkardt/data/off/off.html
https://people.sc.fsu.edu/~jburkardt/data/off/br2.off

kmuseth: good point - I've fixed this and tested it with some of the files in the links

swahtz (approved):
Looks great, thanks!

Issue #1976, opened 2024-12-02
[BUG]Nanovdb: Invalid code in GridBuilder fails to compile with clang19


mont29:
### Environment
**Operating System:** Linux Debian Testing, x86_64
**Version / Commit SHA:** VDB 11.0.0
**Other:** Clang-19

### Describe the bug
Trying to build a project using NanoVDB with Clang19 fails with following error:

```
[...]/lib/linux_x64/openvdb/include/nanovdb/util/GridBuilder.h:1161:72: error: no member named 'isActive' in 'LeafNode<BuildT>'
 1161 |         bool isActive() const { NANOVDB_ASSERT(*this); return mParent->isActive(mPos);}
      |                                                               ~~~~~~~  ^
```

Issue discovered when trying to build Blender with Clang 19.

This is related to [this change](https://releases.llvm.org/19.1.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics:~:text=Clang%20now%20looks%20up%20members%20of%20the%20current%20instantiation%20in%20the%20template%20definition%20context%20if%20the%20current%20instantiation%20has%20no%20dependent%20base%20classes.) in the compiler (see also [the PR](https://github.com/llvm/llvm-project/pull/84050)).

dgovil:
Bastien, I just put up a PR at https://github.com/AcademySoftwareFoundation/openvdb/pull/1977 that should resolve this for you.

PR #1977, opened 2024-12-03
Fix Clang Template errors

dgovil:
Upcoming versions of Clang/LLVM change some behaviour around templating that causes OpenVDB to fail to compile.

This commit addresses two issues and allows VDB to compile again.

1. There were three instances of an unnecessary template keyword in NodeManager.h that were not followed by a template call, and therefore are illegal to the compiler.

2. GridBuilder.h had a template call to a non-existent method. This was not previously validated, but is now validated.  Switching the call from `isActive` to `isOn` works (as suggested by Ken).

See the [changelog](https://releases.llvm.org/19.1.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics:~:text=Clang%20now%20looks%20up%20members%20of%20the%20current%20instantiation%20in%20the%20template%20definition%20context%20if%20the%20current%20instantiation%20has%20no%20dependent%20base%20classes.) for Clang and the associated [PR for the second point above](https://github.com/llvm/llvm-project/pull/84050)

This fixes https://github.com/AcademySoftwareFoundation/openvdb/issues/1976

Built on macOS with the following build args

```
-DBOOST_ROOT=/usr/local/apps/boost/1.82.0-a0
-DTBB_ROOT=/usr/local/apps/tbb/2020_U3-a1/platform-osx
-DBlosc_ROOT=/usr/local/apps/blosc/1.21.5-a0/apple_universal-2
-DOPENVDB_BUILD_NANOVDB=ON
-DNANOVDB_USE_OPENVDB=ON
```

### DIFF
--- a/nanovdb/nanovdb/tools/GridBuilder.h
+++ b/nanovdb/nanovdb/tools/GridBuilder.h
@@ -1158,7 +1158,7 @@ struct LeafNode
         ValueIterator& operator=(const ValueIterator&) = default;
         ValueType operator*() const { NANOVDB_ASSERT(*this); return mParent->mValues[mPos];}
         Coord getCoord() const { NANOVDB_ASSERT(*this); return mParent->offsetToGlobalCoord(mPos);}
-        bool isActive() const { NANOVDB_ASSERT(*this); return mParent->isActive(mPos);}
+        bool isActive() const { NANOVDB_ASSERT(*this); return mParent->mValueMask.isOn(mPos);}
         operator bool() const {return mPos < SIZE;}
         ValueIterator& operator++() {++mPos; return *this;}
         ValueIterator operator++(int) {
--- a/openvdb/openvdb/tree/NodeManager.h
+++ b/openvdb/openvdb/tree/NodeManager.h
@@ -328,7 +328,7 @@ private:
         void operator()(const NodeRange& range) const
         {
             for (typename NodeRange::Iterator it = range.begin(); it; ++it) {
-                OpT::template eval(mNodeOp, it);
+                OpT::eval(mNodeOp, it);
             }
         }
         const NodeOp mNodeOp;
@@ -348,7 +348,7 @@ private:
         void operator()(const NodeRange& range) const
         {
             for (typename NodeRange::Iterator it = range.begin(); it; ++it) {
-                OpT::template eval(mNodeOp, it);
+                OpT::eval(mNodeOp, it);
             }
         }
         const NodeOp& mNodeOp;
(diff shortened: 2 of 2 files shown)


nanovdb/nanovdb/tools/GridBuilder.h
```diff
@@ -1158,7 +1158,6 @@ struct LeafNode
         ValueIterator& operator=(const ValueIterator&) = default;
         ValueType operator*() const { NANOVDB_ASSERT(*this); return mParent->mValues[mPos];}
         Coord getCoord() const { NANOVDB_ASSERT(*this); return mParent->offsetToGlobalCoord(mPos);}
-        bool isActive() const { NANOVDB_ASSERT(*this); return mParent->isActive(mPos);}
```
kmuseth, line 1161: can you change the implementation to this:

`bool isActive() const { NANOVDB_ASSERT(*this); return mParent->mValueMask.isOn(mPos);}`

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -328,7 +328,7 @@ class NodeList
         void operator()(const NodeRange& range) const
         {
             for (typename NodeRange::Iterator it = range.begin(); it; ++it) {
-                OpT::template eval(mNodeOp, it);
+                OpT::eval(mNodeOp, it);
```
kmuseth, line 331: good catch

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -348,7 +348,7 @@ class NodeList
...
```
kmuseth, line 351: good catch

openvdb/openvdb/tree/NodeManager.h
```diff
@@ -373,7 +373,7 @@ class NodeList
         void operator()(const NodeRange& range)
         {
             for (typename NodeRange::Iterator it = range.begin(); it; ++it) {
-                OpT::template eval(*mNodeOp, it);
+                OpT::eval(*mNodeOp, it);
```
kmuseth, line 376: good catch

nanovdb/nanovdb/tools/GridBuilder.h
kmuseth: if that also fails then I agree that we can simply remove this method

kmuseth (changes requested):
please see request for minor change

nanovdb/nanovdb/tools/GridBuilder.h
dgovil: Ah yeah, that works. I updated the PR with the change.

mont29:
Can confirm that this PR fixes the issue I have when building Blender with clang19 on linux (reported as #1976).

kmuseth (approved):

meshula:
@dgovil @kmuseth Thanks for identifying, patching, and landing this one so quickly.

We're tracking this one for our next release of OpenUSD in May. I was wondering if there'll be a tagged release or patch we can point to in order to pick up this fix? 

cc/ @jesschimein

PR #1986, opened 2024-12-17
Added SPDX license identifiers to files missing them

swahtz:


nanovdb/nanovdb/python/test/TestNanoVDB.py
```diff
@@ -1,4 +1,6 @@
-# /usr/bin/env python
+#!/usr/bin/env python
```
apradhana, line 1: @swahtz , @matthewdcong : is the first line necessary for this python script? Can we get rid of it?

dgovil: It's necessary if someone wants to run it directly `./TestnanoVDB.py`.
Also minor nit, it might be good to use `python3` instead of just `python`

dgovil (approved):

apradhana (approved):

nanovdb/nanovdb/python/test/TestNanoVDB.py
apradhana: Got it. Thanks for the clarification @dgovil.

PR #1988, opened 2024-12-17
SPDX Identifier Check Action

swahtz:
Added a new action to check for correct SPDX identifiers in a new workflow for code formatting

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,41 @@
+name: Code Formatting
+
+on:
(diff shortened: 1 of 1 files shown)


swahtz:
Depends on #1986  for checks to pass

swahtz:
TODO:  Switch to the upstream action if this PR is accepted:

https://github.com/enarx/spdx/pull/29

swahtz:
This should be good to go.

PR #1995, opened 2025-01-15
Fix typos

rex4539:


swahtz:
Thanks for this!  Looks like a few of the changes apply to variable/function names so might get some more eyes on the changes.

openvdb/openvdb/tools/MeshToVolume.h
```diff
@@ -2860,14 +2860,14 @@ struct TransformValues
 };
 
 
-// Inactivate values outside the (exBandWidth, inBandWidth) range.
+// Deactivate values outside the (exBandWidth, inBandWidth) range.
 template<typename TreeType>
-struct InactivateValues
+struct DeactivateValues
```
danrbailey, line 2865: This one shouldn't be changed, we should keep the original name.

danrbailey (changes requested):
I could find just one case where a struct was being changed that should not. Otherwise, I'm happy for this to be merged in.

openvdb/openvdb/TypeList.h
```diff
@@ -418,27 +418,27 @@ struct TSRemoveIndicesImpl<TypeList<T, Ts...>, First, Last, Idx>
...
 /// @tparam  OpT  User struct to convert each type
 /// @tparam Ts  Remaining types in the @c TypeList
-template<template <typename> class OpT, typename... Ts> struct TSTranformImpl;
+template<template <typename> class OpT, typename... Ts> struct TSTransformImpl;
 
 /// @brief  Partial specialization for an empty @c TypeList
 /// @tparam  OpT  User struct to convert each type
 template<template <typename> class OpT>
-struct TSTranformImpl<OpT> {
+struct TSTransformImpl<OpT> {
```
danrbailey, line 426: Could we also revert this change? We'd like to review and merge only the changes related to comments in this pass if possible.

rex4539:
> I could find just one case where a struct was being changed that should not. Otherwise, I'm happy for this to be merged in.

Is it ok now or you still need some changes?

rex4539:
Fixed some more typos, from the new code that has been added recently.

swahtz (approved):
I have done a read through of these changes and can confirm there are now no changes to signatures or class names; all the changes are to comments, documentation, error messages and the like minus 2 changes to local variable naming for typos.

Thanks for this contribution!  I think a next step might be for us to instrument an action that checks for these kinds of typos so they don't creep back in as we develop.  Perhaps an action like https://github.com/marketplace/actions/typos-action

openvdb/openvdb/unittest/TestFilter.cc
```diff
@@ -89,7 +89,7 @@ TEST_F(TestFilter, testMedian)
     const int width = 2;
     openvdb::math::DenseStencil<openvdb::FloatGrid> stencil(*referenceGrid, width);
     openvdb::tools::Filter<openvdb::FloatGrid> filter(*filteredGrid);
-    filter.median(width, /*interations=*/1);
+    filter.median(width, /*interactions=*/1);
```
jmlait, line 92: should be iterations.

openvdb/openvdb/unittest/TestFilter.cc
```diff
@@ -132,7 +132,7 @@ TEST_F(TestFilter, testMean)
     const int width = 2;
     openvdb::math::DenseStencil<openvdb::FloatGrid> stencil(*referenceGrid, width);
     openvdb::tools::Filter<openvdb::FloatGrid> filter(*filteredGrid);
-    filter.mean(width,  /*interations=*/1);
+    filter.mean(width,  /*interactions=*/1);
```
jmlait, line 135: should be iterations

openvdb/openvdb/unittest/TestTools.cc
```diff
@@ -1028,13 +1028,13 @@ TEST_F(TestTools, testPointAdvect)
             std::vector<openvdb::Vec3d>, true> constrainedAdvectionTool(*velocityGrid, *cptGrid, 0);
         constrainedAdvectionTool.setThreaded(false);
 
-        // change the number of constraint interation from default 0 to 5
+        // change the number of constraint interaction from default 0 to 5
```
jmlait, line 1031: iteration

openvdb/openvdb/unittest/TestTools.cc
```diff
@@ -1028,13 +1028,13 @@ TEST_F(TestTools, testPointAdvect)
...
         constrainedAdvectionTool.setConstraintIterations(5);
 
         // test the pure-projection mode (order = 0)
         constrainedAdvectionTool.setIntegrationOrder(0);
 
-        // change the number of constraint interation (from 0 to 5)
+        // change the number of constraint interaction (from 0 to 5)
```
jmlait, line 1037: iteration

jmlait (approved):
Please change interation -> iteration rather than interaction :>

There are two places where real-code is changed, but I've validated both are safe.

rex4539:
Closing due to no interest.

PR #2008, opened 2025-02-27
Slowdown fix - Avoid unnecessary getLastWriteTime calls for memory mapped files on windows

jjtParadox:
We are reading very large openvdb volumes (8gb and larger) and discovered that, on our Windows systems, using `delayLoad=true` for volumes causes 10x slower file reads compared to `delayLoad=false`.

Profiling pointed us to this behavior in `Archive::getLastWriteTime()`, which makes three system calls (open file, get file time, close file): https://github.com/AcademySoftwareFoundation/openvdb/blob/5ff0edc2cc333a70923c0122d6a6deec3bc80cc4/openvdb/openvdb/io/Archive.cc#L506-L512

The above is called once per node when reading the node from the file due to this check in `MappedFile::createBuffer()`: https://github.com/AcademySoftwareFoundation/openvdb/blob/5ff0edc2cc333a70923c0122d6a6deec3bc80cc4/openvdb/openvdb/io/Archive.cc#L557-L565

This check and warning is unnecessary on Windows as memory mapped files are automatically write-locked and cannot be modified while the memory mapping is open. The only time it *might* be necessary is reading a volume over a network.

This PR guards the last write time check in `MappedFile::createBuffer()` with `#ifndef _WIN32`.

### DIFF
--- a/openvdb/openvdb/io/Archive.cc
+++ b/openvdb/openvdb/io/Archive.cc
@@ -554,6 +554,7 @@ MappedFile::filename() const
 SharedPtr<std::streambuf>
 MappedFile::createBuffer() const
 {
+#ifndef _WIN32
     if (!mImpl->mAutoDelete && mImpl->mLastWriteTime > 0) {
         // Warn if the file has been modified since it was opened
         // (but don't bother checking if it is a private, temporary file).
@@ -563,6 +564,7 @@ MappedFile::createBuffer() const
             mImpl->mLastWriteTime = 0; // suppress further warnings
         }
     }
+#endif
 
     return SharedPtr<std::streambuf>{
         new boost::iostreams::stream_buffer<boost::iostreams::array_source>{


Idclip:
Hey @jjtParadox thanks for this! Discussed this briefly today, we're considering removing this check across all platforms (maped files on most linux distros are not write locked, but this check does nothing in the way of solving that issue anyway). I'm not sure of the performance implication of calling `stat` in this pattern either, but I wouldn't be surprised if it had similar performance problems.

Now that this has been "fixed", do you have any comparison numbers to hand with delay loading on/off? Curious to see what the overhead looks like

jjtParadox:
Here's some numbers using a 7.82GB file of dense grids.

Before the fix:
| `delayLoad` | Time |
| ---- | ---- |
| true | 208s |
| false | 13.2s |

After the fix:
| `delayLoad` | Time |
| ---- | ---- |
| true | 11.9s |
| false | 13.1s |

Code used for testing:
```c++
openvdb::initialize();

openvdb::io::File file("Bench-Large.vdb");
file.open(true); // Change this to `false` to disable delayLoad

openvdb::GridPtrVecPtr grids = std::make_shared<openvdb::GridPtrVec>();

for (auto name_iter = file.beginName(), name_end = file.endName(); name_iter != name_end; ++name_iter)
{
	TestGridType::Ptr gridPtr = openvdb::gridPtrCast<TestGridType>(file.readGrid(*name_iter));

	// Force all voxel data to be read from the file into CPU memory
	openvdb::tools::foreach(gridPtr->tree().cbeginLeaf(), [](const TestGridType::TreeType::LeafCIter leaf)
		{
			leaf->buffer().data();
		});

	grids->push_back(gridPtr);
}
```

Idclip:
Yikes, it took me a moment to realise that was 208 seconds, not ms! 

Thanks a lot for these - and to your point about things like antivirus, I think we should just remove this check entirely. 

Happy for this to go in as is too

Idclip (approved):

jmlait (approved):

Idclip:
Merging this as is. Will add release notes and remove this code entirely in a subsequent PR

PR #2013, opened 2025-03-14
Fixed child offset computation in nanovdb::tools::cuda::indexToGrid

heandreas:
I ran into a problem when using `nanovdb::tools::cuda::indexToGrid<nanovdb::Vec3f>` to build a grid of `Vec3fs` from an index grid. After debugging for a while I managed to find a fix that works for me. I believe the computation for `dstTile.child` in `processRootTilesKernel` is not correct for all types. It seems to work for float and double grids, but not for Vec3f and probably many (or all?) other types that do not align with the size of the index type. My fix is inspired by the code in `processNodesKernel` that also computes similar byte offsets.

I extended the unit test `CudaIndexGridToGrid_ValueOnIndex` to show the problem. It only passes with my fix and not the original code. Feedback is highly appreciated!

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/IndexToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/IndexToGrid.cuh
@@ -183,7 +183,17 @@ __global__ void processRootTilesKernel(typename IndexToGrid<SrcBuildT>::NodeAcce
     auto &dstTile = *nodeAcc->template dstRoot<DstBuildT>().tile(tid);
     dstTile.key   = srcTile.key;
     if (srcTile.child) {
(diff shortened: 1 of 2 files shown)


kmuseth:
thanks @heandreas ! Looking at your changes now :)

kmuseth:
yes this does indeed look like a valid fix - great catch!

since you didn't sign the EasyCLA I'm taking the liberty to re-write your PR slightly with more documentation explaining how this fix works

heandreas:
Sounds great, thanks for looking into it!

kmuseth:
@heandreas  - see PR #2021 - I added a credit to you - let me know if you prefer to be anonymous.

kmuseth:
@heandreas  - take a look [here](https://github.com/AcademySoftwareFoundation/openvdb/pull/2021)

PR #2016, opened 2025-03-24
Transpose on NanoToOpenVDB allowing round tripping

jmlait:
The openvdb::Mat3::transpose() method returns a transpose of itself, it doesn't do an in-place transpose.

Thus the line

    mat.transpose(); // the 3x3 in nanovdb is transposed relative to openvdb's 3x3

is a no-op, and not actually transposing the matrix.  Since we DO transpose when we do an OpenToNano; it is important we transpose here or we can't round-trip.

Unfortunately, it has been this way since the initial commit, so it is quite likely tooling has already worked around this....


Idclip:
@jmlait can we update the vdb api with a [[nodiscard]]?

jmlait:
[[nodiscard]] is a good idea, I was wondering what the right approach should be for this.  I think I'd want to do that as separate PR to avoid having to also find out what else is broken here :<

Idclip (approved):

kmuseth (commented):
looks good to me

PR #2065, opened 2025-07-27
AX backend restructure for LLVM opaque ptr support (LLVM 16+)

Idclip:


jmlait (approved):
Impressive to have to maintain the various flavours of LLVM in one build, I do not envy the changes required.

Idclip:
> Impressive to have to maintain the various flavours of LLVM in one build, I do not envy the changes required.

This was extremely painful, but it's somewhat on me for relying too heavily on LLVM's IR and not having our own - I've made steps in that direction in this MR. The opaque ptr changes were the only significant issue, primarily in the backend function framework which was handling resolution - without being able to introspect pointer types this all fell apart.  Once this was solved it was minor API discrepancies between LLVM 16->20 (with a symbol issue in LLVM 17, hence the new future minimum of LLVM 18 and the new absolutely minimum of LLVM 15).

PR #2067, opened 2025-08-03
 Introduced Anisotropic Surfacing Kernels and PCA methods

Idclip:
Re-created from https://github.com/AcademySoftwareFoundation/openvdb/pull/1634 with additional improvements
```
OpenVDB:
    New:
        - Introduced anisotropic surfacing tools in points/PointRasterizeSDF.h
        and principle component analysis (PCA) methods in points/PrincipleComponentAnalysis.h.
        The latter tool analyses point neighborhoods and computes affine
        transformations representing elliptical distributions. The prior tool
        takes rotational and stretch point attributes (optionally computed from
        the prior PCA methods) and generates anisotropic surfaces.
        - Added OPENVDB_HAS_INVOKABLE_MEMBER_FUNCTION and OPENVDB_HAS_MEMBER_FUNCTION
        macros to help determine the presence and compatibility of interface
        member functions at compile time.

    API Changes:
        - Re-worked the API in PointRasterizeSDF.h with the addition of anisotropic
        surfacing kernels. The new API exposes more concise free functions which
        take structs of parameters. The old API remains supported but is
        deprecated.
        - Deprecated providing a filter or interrupter as arguments to
        points::rasterize. These can instead be implemented on user transfer
        schemes.
        - Transfer schemes provided to points::rasterize can now optionally
        implement a rasterizePoints callback that takes a range of points per
        voxel.

    Improvements:
        - Improved the performance of PointRasterizeSDF.h tools with larger radii
        with better bound computation approximations.

    Fixes:
        - Fixed a bug in points::rasterizeSpheres and points::rasterizeSmoothSpheres
        which could cause sections of the generated surface to incorrectly be marked
        as negative interior values.
        - Fixed a bug in points::rasterizeSmoothSpheres which could result in
        incomplete surface reconstruction with significantly larger search distances.
        - Fixed some precision issues in various Matrix methods
```

### DIFF
--- a/openvdb/openvdb/math/Mat3.h
+++ b/openvdb/openvdb/math/Mat3.h
@@ -716,8 +716,7 @@ pivot(int i, int j, Mat3<T>& S, Vec3<T>& D, Mat3<T>& Q)
         S(i,k) = T(cosin_of_theta * temp - sin_of_theta * S(j,k));
         S(j,k) = T(sin_of_theta * temp + cosin_of_theta * S(j,k));
     }
-    for (int k = 0; k < n; ++k)
-    {
+    for (int k = 0; k < n; ++k) {
         temp = Q(k,i);
         Q(k,i) = T(cosin_of_theta * temp - sin_of_theta * Q(k,j));
         Q(k,j) = T(sin_of_theta * temp + cosin_of_theta * Q(k,j));
--- a/openvdb/openvdb/math/Mat3.h
+++ b/openvdb/openvdb/math/Mat3.h
@@ -680,7 +680,7 @@ pivot(int i, int j, Mat3<T>& S, Vec3<T>& D, Mat3<T>& Q)
 
     double Sjj_minus_Sii = D[j] - D[i];
 
-    if (fabs(Sjj_minus_Sii) * (10*math::Tolerance<T>::value()) > fabs(Sij)) {
+    if (std::abs(Sjj_minus_Sii) * (10*math::Tolerance<T>::value()) > std::abs(Sij)) {
         tan_of_theta = Sij / Sjj_minus_Sii;
     } else {
         /// pivot on Sij
@@ -688,40 +688,40 @@ pivot(int i, int j, Mat3<T>& S, Vec3<T>& D, Mat3<T>& Q)
 
         if (cotan_of_2_theta < 0.) {
             tan_of_theta =
-                -1./(sqrt(1. + cotan_of_2_theta*cotan_of_2_theta) - cotan_of_2_theta);
+                -1./(std::sqrt(1. + cotan_of_2_theta*cotan_of_2_theta) - cotan_of_2_theta);
         } else {
             tan_of_theta =
-                1./(sqrt(1. + cotan_of_2_theta*cotan_of_2_theta) + cotan_of_2_theta);
+                1./(std::sqrt(1. + cotan_of_2_theta*cotan_of_2_theta) + cotan_of_2_theta);
         }
     }
 
-    cosin_of_theta = 1./sqrt( 1. + tan_of_theta * tan_of_theta);
+    cosin_of_theta = 1./std::sqrt( 1. + tan_of_theta * tan_of_theta);
     sin_of_theta = cosin_of_theta * tan_of_theta;
     z = tan_of_theta * Sij;
     S(i,j) = 0;
-    D[i] -= z;
-    D[j] += z;
+    D[i] -= T(z);
+    D[j] += T(z);
     for (int k = 0; k < i; ++k) {
         temp = S(k,i);
-        S(k,i) = cosin_of_theta * temp - sin_of_theta * S(k,j);
-        S(k,j)= sin_of_theta * temp + cosin_of_theta * S(k,j);
+        S(k,i) = T(cosin_of_theta * temp - sin_of_theta * S(k,j));
+        S(k,j) = T(sin_of_theta * temp + cosin_of_theta * S(k,j));
     }
     for (int k = i+1; k < j; ++k) {
         temp = S(i,k);
-        S(i,k) = cosin_of_theta * temp - sin_of_theta * S(k,j);
-        S(k,j) = sin_of_theta * temp + cosin_of_theta * S(k,j);
+        S(i,k) = T(cosin_of_theta * temp - sin_of_theta * S(k,j));
+        S(k,j) = T(sin_of_theta * temp + cosin_of_theta * S(k,j));
     }
     for (int k = j+1; k < n; ++k) {
         temp = S(i,k);
-        S(i,k) = cosin_of_theta * temp - sin_of_theta * S(j,k);
-        S(j,k) = sin_of_theta * temp + cosin_of_theta * S(j,k);
+        S(i,k) = T(cosin_of_theta * temp - sin_of_theta * S(j,k));
+        S(j,k) = T(sin_of_theta * temp + cosin_of_theta * S(j,k));
     }
     for (int k = 0; k < n; ++k)
-        {
-            temp = Q(k,i);
-            Q(k,i) = cosin_of_theta * temp - sin_of_theta*Q(k,j);
-            Q(k,j) = sin_of_theta * temp + cosin_of_theta*Q(k,j);
-        }
+    {
+        temp = Q(k,i);
+        Q(k,i) = T(cosin_of_theta * temp - sin_of_theta * Q(k,j));
+        Q(k,j) = T(sin_of_theta * temp + cosin_of_theta * Q(k,j));
+    }
 }
 
 } // namespace mat3_internal
@@ -758,7 +758,7 @@ diagonalizeSymmetricMatrix(const Mat3<T>& input, Mat3<T>& Q, Vec3<T>& D,
         double er = 0;
         for (int i = 0; i < n; ++i) {
             for (int j = i+1; j < n; ++j) {
-                er += fabs(S(i,j));
+                er += std::abs(S(i,j));
             }
         }
         if (std::abs(er) < math::Tolerance<T>::value()) {
@@ -773,12 +773,12 @@ diagonalizeSymmetricMatrix(const Mat3<T>& input, Mat3<T>& Q, Vec3<T>& D,
         for (int i = 0; i < n; ++i) {
             for (int j = i+1; j < n; ++j){
 
-                if ( fabs(D[i]) * (10*math::Tolerance<T>::value()) > fabs(S(i,j))) {
+                if ( std::abs(D[i]) * (10*math::Tolerance<T>::value()) > std::abs(S(i,j))) {
                     /// value too small to pivot on
                     S(i,j) = 0;
                 }
-                if (fabs(S(i,j)) > max_element) {
-                    max_element = fabs(S(i,j));
+                if (std::abs(S(i,j)) > max_element) {
+                    max_element = std::abs(S(i,j));
                     ip = i;
                     jp = j;
                 }
--- a/openvdb/openvdb/points/PointRasterizeSDF.h
+++ b/openvdb/openvdb/points/PointRasterizeSDF.h
@@ -5,37 +5,62 @@
 ///
 /// @file PointRasterizeSDF.h
 ///
-/// @brief Transfer schemes for rasterizing point positional and radius data to
-///  signed distance fields with optional closest point attribute transfers.
-///  All methods support arbitrary target linear transformations, fixed or
-///  varying point radius, filtering of point data and arbitrary types for
-///  attribute transferring.
+/// @brief Various transfer schemes for rasterizing point positions and radius
+///  data to signed distance fields with optional closest point attribute
+///  transfers. All methods support arbitrary target linear transformations,
+///  fixed or varying point radius, filtering of point data and arbitrary types
+///  for attribute transferring.
 ///
-/// @details There are two main transfer implementations; rasterizeSpheres and
-///  rasterizeSmoothSpheres. The prior performs trivial narrow band stamping
-///  of spheres for each point, where as the latter calculates an averaged
-///  position of influence per voxel as described in:
-///    [Animating Sand as a Fluid - Zhu Bridson 05].
+/// @details There are currently three main transfer implementations:
 ///
-///  rasterizeSpheres() is an extremely fast and efficient way to produce both a
-///  valid symmetrical narrow band level set and transfer attributes using
-///  closest point lookups.
+/// - Rasterize Spheres
 ///
-///  rasterizeSmoothSpheres() produces smoother, more blended connections
-///  between points which is ideal for generating a more artistically pleasant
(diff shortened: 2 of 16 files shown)


openvdb/openvdb/math/Mat3.h
```diff
@@ -680,48 +680,48 @@ pivot(int i, int j, Mat3<T>& S, Vec3<T>& D, Mat3<T>& Q)
...
+        S(i,k) = T(cosin_of_theta * temp - sin_of_theta * S(j,k));
+        S(j,k) = T(sin_of_theta * temp + cosin_of_theta * S(j,k));
     }
     for (int k = 0; k < n; ++k)
-        {
-            temp = Q(k,i);
-            Q(k,i) = cosin_of_theta * temp - sin_of_theta*Q(k,j);
-            Q(k,j) = sin_of_theta * temp + cosin_of_theta*Q(k,j);
-        }
+    {
```
jmlait: I, for one, enjoy the brace being in the one-true-location.   But I think pedantically the original code was also wrong, and it needs to be dragged back to the previous line?

openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -5,37 +5,62 @@
...
+///
+///     Calculates an averaged position of influence per voxel as described in:
+///       [Animating Sand as a Fluid - Zhu Bridson 2005].
+///
+///     This technique produces smoother, more blended connections between
+///     points which is ideal for generating a more artistically pleasant
+///     surface directly from point distributions. It aims to avoid typical
+///     post filtering operations used to smooth surface volumes. Note however
+///     that this method may not necessarily produce a *symmetrical* narrow
+///     band level set; the exterior band may be smaller than desired depending
```
jmlait: Maybe we should mention the temporal instability which is the usual problem with this method?  Connections can "snap" on and off as they leave the threshold?

openvdb/openvdb/points/PointTransfer.h
```diff
@@ -475,6 +574,16 @@ struct RasterizePoints
...
                     localFilter.reset(*pointLeaf);
 
+                    if (this->interrupted()) return;
+
+                    // It's actually faster to go through the ValueIter API than
+                    // the leaf API as the value iterators cache the value buffer
+                    // ptrs (the leaf buffer API has to check the ptr on every
+                    // access).
+                    // @todo  Once we've improved the leaf buffer impl this
+                    //   should be removed
```
jmlait, line 584: Haven't we improved that recently?

openvdb/openvdb/points/PrincipalComponentAnalysis.h
```diff
@@ -0,0 +1,194 @@
...
+    ///   attribute which represents the scaling components of each points
+    ///   ellipse or (1.0,1.0,1.0) for isolated points.
+    using StretchT = math::Vec3<float>;
+    std::string stretch = "stretch";
+
+    /// @brief  Settings for the "rotation" attribute, a floating point matrix
+    ///   attribute which represents the rotation of each points ellipse or
+    ///   the identity matrix for isolated points.
+    using RotationT = math::Mat3<float>;
+    std::string rotation = "rotation";
```
jmlait, line 167: Since our "rotation" is a 3x3, technically I could have any skew or stretch baked in that matrix?  Does the rasterization support skews?   Is the stretch used to determine overlap, so a scale factor in this rotation will result in missing voxels?   Should this just be a normalized quaternion to bypass this argument?  Or documented that MUST be a rotation matrix?

openvdb/openvdb/points/impl/PointRasterizeEllipsoidsSDFImpl.h
```diff
@@ -0,0 +1,718 @@
...
+        else {
+            for (Index i = 0; i < Index(rad.size()); ++i) {
+                maxr = math::maxComponent(maxr, rad.eval(i).get());
+            }
+        }
+
+        // The max stretch coefficient. We can't analyze each xyz component
+        // individually as we don't take into account the ellips rotation, so
+        // have to expand the worst case uniformly
+        const Real maxRadius = std::max(maxr.x(), std::max(maxr.y(), maxr.z()));
```
jmlait: This suggests the answer to the earlier question was that the rotation matrix has to be orthonormal.  We are also relying on it preserving distances when we transform into the ellipse space to do the distance computation earlier.

openvdb/openvdb/points/impl/PointRasterizeEllipsoidsSDFImpl.h
```diff
@@ -0,0 +1,718 @@
...
+                Coord::round(this->mSurfaceTransform.worldToIndex(maxPos)));
+            surfaceBounds.min() -= Coord(dist);
+            surfaceBounds.max() += Coord(dist);
+            this->activate(surfaceBounds);
+            /// @todo deactivate min
+            this->updateMaxLookup(minPos, maxPos, dist, leaf);
+        }
+    }
+
+    /// @brief  Fill activity by analyzing the axis aligned ellips bounding
```
jmlait: ellipse not ellips

openvdb/openvdb/points/impl/PointRasterizeEllipsoidsSDFImpl.h
```diff
@@ -0,0 +1,718 @@
...
+    ///   boxes on points in this leaf. Slightly slower than just looking at
+    ///   ellips stretches but produces a more accurate/tighter activation
```
jmlait: maybe s/ellips/ellipse/g  :>

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
```diff
@@ -1045,315 +1284,493 @@ rasterizeSurface(const PointDataGridT& points,
...
-    const Real minBandRadius = rad.min();
-    const Real maxBandRadius = rad.max();
-    const size_t width = static_cast<size_t>(math::RoundUp(maxBandRadius));
+    typename SdfT::Ptr surface;
+    GridPtrVec grids;
 
-    typename SdfT::Ptr surface =
-        initFixedSdf<SdfT, InterrupterT>
+    if (settings.radius.empty())
+    {
```
jmlait: Brace on previous line.

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
```diff
@@ -1045,315 +1284,493 @@ rasterizeSurface(const PointDataGridT& points,
...
+    // activate topology further outside the bounds of the point data grid).
+    // This bounds is expanded by the halfband + max radii
+    CoordBBox bounds;
+    points.tree().evalLeafBoundingBox(bounds);
+
+    typename SdfT::Ptr surface;
+    GridPtrVec grids;
+
+    if (settings.radius.empty())
+    {
```
jmlait: Collapse brace

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
```diff
@@ -1045,315 +1284,493 @@ rasterizeSurface(const PointDataGridT& points,
...
+
+        const FixedRadius<Real> rad(settings.radiusScale / vs);
+        if (interrupter) interrupter->start("Rasterizing particles to level set using constant Zhu-Bridson");
+
+        grids = doRasterizeSurface<SdfT, AveragePositionTransfer, AttributeTypes, FilterT>
+            (points, attributes, *surface,
+                width, rad, indexSpaceSearch, points.transform(), filter, interrupter, *surface); // args
+    }
+    else
+    {
```
jmlait: This looks so much better than } else {, but our standard differs...

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
```diff
@@ -0,0 +1,789 @@
...
+
+        OPENVDB_ASSERT(std::isfinite(searchRadiusInv));
+        OPENVDB_ASSERT(std::isfinite(searchRadius2));
+        OPENVDB_ASSERT(std::isfinite(searchRadiusIS2));
+
+        const Coord& a(bounds.min());
+        const Coord& b(bounds.max());
+
+        for (Index srcid = start; srcid < end; srcid += step)
+        {
```
jmlait: Collapse brace

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
```diff
@@ -0,0 +1,789 @@
...
```
jmlait: Collapse brace

openvdb/openvdb/points/PointTransfer.h
Idclip: No, I improved the `ValueIterators` API, not the `LeaNode` API - the latter still goes through the `LeafBuffer`. So we have

ValueIterators:

```
iter.getValue(offset) = ValueType* mData[offset];
```

vs LeafNode:

```
node.getValue(offset) = LeafNode::mBuffer[offset] = LeafBuffer::at(offset)

// impl of LeafBuffer::at()

static const ValueType sZero = zeroVal<T>();
OPENVDB_ASSERT(i < SIZE);
this->loadValues(); // if (this->isOutOfCore()) this->doLoad();
if (mData) return mData[i]; else return sZero; // this still involves a check 
```

`isOutOfCore`, `doLoad()` and the check on `mData` all cause mis-optimizatons. It's very annoying, it means that going through the iterators is currently more efficient that the LeafNode API. But there's no real way to solve this, the `LeafNode` API is the thing involved in deciding whether to load delay loaded data off disk. This logic _has_ to be performed here.

Idclip: Alternatively, instead of calling
```
const auto valiter = pointLeaf->cbeginValueAll(); // outer loop
auto val = valiter.getItem(offset); // inner loop
```

I can instead just cache the data pointer locally:
```
const auto* buffer = pointLeaf->buffer().data();  // outer loop
auto val = buffer[offset]; // inner loop
```

Both now achieve the same thing and are more efficient than:
```
auto val = pointLeaf->getValue(offset); // inner loop
```

openvdb/openvdb/math/Mat3.h
Idclip: fixed

openvdb/openvdb/points/impl/PointRasterizeEllipsoidsSDFImpl.h
Idclip: fixed

openvdb/openvdb/points/impl/PointRasterizeEllipsoidsSDFImpl.h
Idclip: fixed all, including type names

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
Idclip: fixed

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
Idclip: fixed

openvdb/openvdb/points/impl/PointRasterizeSDFImpl.h
Idclip: fixed

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
Idclip: fixed

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
Idclip: fixed, and fixed other cases that I can see in this file

openvdb/openvdb/points/PointRasterizeSDF.h
Idclip: Yeah good point, added a note

openvdb/openvdb/points/PrincipalComponentAnalysis.h
Idclip: Yeah as you've pointed out below, input to the rasterizer is expected to be an orthogonal rotational matrix. I have extra work in this space regarding outputting quats such as computing anisotropy based on grid topology, storing and slerping as quaternions etc - for now I'd like to get this in as is, but the API is designed to support both output types in the future, for example:

```c++
struct PcaAttributes {
    // ...
    using RotationT = math::Mat3<float>;    
    using QuatT = math::Quat<float>;
    enum class RotationOutput { Matrix, Quaternion };
    std::string rotation = "rotation";
    RotationOutput rotmode = RotationOutput::Matrix;
    // ...
};
```

or:

```c++
struct PcaAttributes {
    // @brief  rotation as ortho 3x3 matrix, clear string to avoid outputting attribute
    using RotationT = math::Mat3<float>;    
    std::string rotation = "rotation";   
    // @brief  rotation as unit quaternion, clear string to avoid outputting attribute
    using QuatT = math::Quat<float>;
    std::string uquat = "uquat";
};
```

Though the covariance has to be computed on a mat, so the conversion to quat could technically be performed after the call to `points::pca`. On the EllipsoidSettings, it simply takes `std::string rot = "rotation"`, and we can update the internal algorithm to handle both 3x3 mats and quats

Idclip: FWIW, dug out my old code, here's what I have:

```c++
    enum class AttributeOutput
    {
        // When selected, the "stretch" output attribute won't be created.
        // Instead, only a 3x3 affine unitary "xform" attribute will be created
        // with the combined stretch and rotational transformations
        COMBINED_TRANSFORM,
        // When selected, both the "stretch" vector and "xform" output attributes
        // will be created. The xform will take the form of a quaternion
        // representing the rotation.
        STRETCH_AND_QUATERNION,
        // When selected, both the "stretch" vector and "xform" output attributes
        // will be created. The xform will take the form of a orthogonal 3x3 matrix
        // representing the rotation.
        STRETCH_AND_ROTATION_MATRIX
    };

    AttributeOutput format = AttributeOutput::COMBINED_TRANSFORM;
```

openvdb/openvdb/points/PointTransfer.h
jmlait: Thank you for the explanation, that makes sense.

openvdb/openvdb/points/PrincipalComponentAnalysis.h
jmlait: Thank you, that all makes sense.  And it is something we can deal with in future iterations, I'm not suggesting we delay this to have quaternions.

jmlait (approved):

PR #2069, opened 2025-08-05
Update LevelSetTubesImpl.h

ghurstunither:
Preprocess level set tapered capsule data to handle negative radii inputs by truncating the line segment to where radius equals 0.

This stops the assert failure in debug mode here: https://github.com/Idclip/openvdb/actions/runs/16716413607


Idclip:
Can you add a note in `pendingchanges/`, otherwise lgtm!

PR #2081, opened 2025-09-11
Bump Houdini CI builds to 20.5/21

Idclip:


Idclip:
@danrbailey good catch, will investigate that thanks

PR #2085, opened 2025-09-21
Fixed a compilation issue with the OpenVDB Activate SOP with Houdini 21

Idclip:


danrbailey (approved):

Idclip:
Looks like we're hitting the 32gb memory limit on the aswf runner for the g++ images... and we have symbol issues with the 2025 VFX builds with TBB. I will merge this as is as hopefully this is all that is needed to get up and running with H21, though the symbol issues are concerning

PR #2087, opened 2025-09-24
Remove Old IO Modes

danrbailey:
This removes all code for reading VDB files earlier than 222 and generates an error instead. This was first introduced prior to VDB 1.0 (2013) and is primarily there to support Dreamworks adoption of earlier versions of the file format.

This also removes LegacyFrustum and RootNodeMask - both classes which only exist for earlier file formats.


openvdb/openvdb/io/Archive.cc
```diff
@@ -786,11 +786,14 @@ Archive::setGridCompression(std::ostream& os, const GridBase& grid) const
 void
 Archive::readGridCompression(std::istream& is)
 {
-    if (getFormatVersion(is) >= OPENVDB_FILE_VERSION_NODE_MASK_COMPRESSION) {
-        uint32_t c = COMPRESS_NONE;
-        is.read(reinterpret_cast<char*>(&c), sizeof(uint32_t));
-        io::setDataCompression(is, c);
+    if (getFormatVersion(is) < OPENVDB_FILE_VERSION_GRID_INSTANCING) {
```
Idclip: Did you mean to use `OPENVDB_FILE_VERSION_GRID_INSTANCING` here? Previous clause was >= `OPENVDB_FILE_VERSION_NODE_MASK_COMPRESSION` and the comment you added reflects that

kmuseth (commented):
overall this looks good (removed the legacy frustum transform and the run-time sized NodeMask), but it's still unclear to me how files older than 221 are handled. That is, what kind of error message will a user get when trying to read these super old files? Hope something more useful than a generic "file format not supported" :) I'd prefer we give the user some hints on how they can convert old files (using version 12 or older) ...

kmuseth (approved):

PR #2088, opened 2025-09-25
Fix Houdini CI

danrbailey:
I looked into this issue and we are using RPATH, however the problem is the RPATH points at `/usr/local/lib64` which is the same location as other dependencies are installed using the ASWF docker containers. (blosc, etc). I don't see an easy solution here. I tried a few fixes and settled on this one - using the VFX 2026 containers (as suggested by @swahtz) which has TBB 2022 installed. Although it is not technically correct, this seems to work fine for our builds of H21.0. (@jmlait)

As part of this change, I also adjusted this so that we are building the debug build correctly (and not evaluating the unit tests) and dropped the number of threads for GCC down from 8 to 6 which sufficiently reduces the memory to keep within the available RAM on this GHA hosts.

The only issue is that H21 doesn't evaluate the AX unit tests correctly for the GCC builds. I have disabled them for Houdini builds now (they still build and run fine in the ax workflow):
```
test 5
    Start 5: vdb_ax_unit_test

5: Test command: /__w/openvdb/openvdb/build/openvdb_ax/openvdb_ax/test/vdb_ax_test "-v"
5: Working Directory: /__w/openvdb/openvdb/openvdb_ax/openvdb_ax/test/../
5: Test timeout computed to be: 9999795
5: : CommandLine Error: Option 'print-summary-global-ids' registered more than once!
5: LLVM ERROR: inconsistency in registered CommandLine options
5/5 Test #5: vdb_ax_unit_test .................Subprocess aborted***Exception:   0.32 sec
```
@Idclip - I don't know if you have any ideas about this one? (Could be related to the use of the VFX 2026 container which brings in GCC 14). Either way, I think we should merge this as it fixes all the failing Houdini CI issues and then we can switch them back on if we can figure out this issue.

### DIFF
--- a/.github/workflows/houdini.yml
+++ b/.github/workflows/houdini.yml
@@ -129,6 +129,7 @@ jobs:
             -DDISABLE_DEPENDENCY_VERSION_CHECKS=ON \
             \"
     - name: test
+      if: matrix.config.build == 'Release'
       run: cd build && ctest -V
     # Keep ccache light by stripping out any caches not accessed in the last day
     - name: ccache_clean


Idclip:
That error usually comes from mixing LLVM versions, I'll have to investigate whether this is an issue with the sesi one or multiple on the system

Idclip:
> > That error usually comes from mixing LLVM versions, I'll have to investigate whether this is an issue with the sesi one or multiple on the system
> 
> Do you mind if we merge this in the time being?

Let me have a look this weekend, will merge Sun/Monday latest and do the release regardless

danrbailey:
> > > That error usually comes from mixing LLVM versions, I'll have to investigate whether this is an issue with the sesi one or multiple on the system
> > 
> > 
> > Do you mind if we merge this in the time being?
> 
> Let me have a look this weekend, will merge Sun/Monday latest and do the release regardless

Ok, thank you!

Idclip:
The error is due to linking LLVM statically and AX not having a well defined visible symbol table (i.e. no `-fvisibility=hidden`). The unittests need LLVM symbols, but these are being embedded and exported as part of `openvdbax.so` - the `vdb_ax` binary doesn't need the LLVM symbols, so the tests which test the `vdb_ax` command line binary are fine.

Three possible solutions:
 1) Build AX with `visibility=hidden` and export the appropriate API - we should be doing this regardless, but the FunctionRegistry is complicated and will need to properly hide all LLVM impl. Not really possible at the moment
 2) Link against `LLVM.so` - doesn't seem to be installed on these images, only the static libs are available
 3) Don't link the unit tests against `openvdbax.so`, either embed the obj's into `vdb_ax_test` or link statically.

Not a huge fan of (3) considering (1) is what we should be doing - but it's not hard to do so might give it a shot. Regardless we should be able to re-enable the `vdb_ax` tests, just doesn't seem to be a switch just for them at the moment.

Idclip:
@danrbailey I've updated the `h21_ci` branch in this repo with a few commits to add the ax tests back:
https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/18034405143/job/51317984168

If this passes, can you try updating your branch with these changes so we can test all workflows?

danrbailey:
> @danrbailey I've updated the `h21_ci` branch in this repo with a few commits to add the ax tests back: https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/18034405143/job/51317984168
> 
> If this passes, can you try updating your branch with these changes so we can test all workflows?

I have applied your changes with a few small edits and everything passes, thanks! My changes were that I added a build suffix to the job name to distinguish release vs debug and I've disabled actually evaluating the debug unit tests as they are very slow to run. If we want to do that, I think doing it through the weekly CI is more appropriate.

Happy to merge this?

Idclip (approved):

Issue #2090, opened 2025-09-25
[BUG] CheckNormGrad() will square FLT_MAX causing a FP overflow.


stolk:
### Environment
Debian Trixie

### Describe the bug
Running the unit tests with FP exceptions enabled showed that CheckNormGrad() will square FLT_MAX, causing a FP overflow.

### To Reproduce
Enable FP exceptions (MR with toggle for the test will follow.)

### Expected behavior
No FP overflow

### Additional context

I have a fix, and will be creating an MR to address this issue.

Issue #2092, opened 2025-09-25
[BUG] ConvexVoxelizer will create NaN values as it takes square roots of negative values.


stolk:
### Environment
Debian Trixie

### Describe the bug
With FP exceptions enabled, the unit test will catch ConvexVoxelizer taking the square root of negative values, leading to NaN values with unpredictable results.

### To Reproduce
Enable FP exceptions (MR that adds toggle for this will follow.)

### Expected behavior
No FP errors.

### Additional context
I have a fix, MR follows.

stolk:
Dropping this commit from the MR, that would have fixed this:
```
commit c679e3be71dd0c24987e5f06c3145f15b9626586 (HEAD -> fpe, origin/fpe)
Author: Bram Stolk <b.stolk@gmail.com>
Date:   Wed Sep 24 14:11:23 2025 -0700

    Do not take square root of negative values.
    FP exceptions were triggered in unit tests.
    
    FIXES #2092
    
    Signed-off-by: Bram Stolk <b.stolk@gmail.com>

diff --git a/openvdb/openvdb/math/Stencils.h b/openvdb/openvdb/math/Stencils.h
index 0c4ed2c0..5d145815 100644
--- a/openvdb/openvdb/math/Stencils.h
+++ b/openvdb/openvdb/math/Stencils.h
@@ -1627,7 +1627,9 @@ public:
         Real alphaM, alphaG, normGrad;
         if (this->curvatures(alphaM, alphaG, normGrad)) {
           const Real mean = alphaM*mInv2Dx/math::Pow3(normGrad);
-          const Real tmp = std::sqrt(mean*mean - alphaG*mInvDx2/math::Pow4(normGrad));
+          const Real base = mean*mean - alphaG*mInvDx2/math::Pow4(normGrad);
+          const Real clampVal = 0;
+          const Real tmp = std::sqrt(std::max(base, clampVal));
           pair.first  = ValueType(mean - tmp);
           pair.second = ValueType(mean + tmp);
         }
diff --git a/openvdb/openvdb/tools/impl/ConvexVoxelizer.h b/openvdb/openvdb/tools/impl/ConvexVoxelizer.h
index d018e362..c3fb6839 100644
--- a/openvdb/openvdb/tools/impl/ConvexVoxelizer.h
+++ b/openvdb/openvdb/tools/impl/ConvexVoxelizer.h
@@ -389,7 +389,10 @@ protected:
[...]

Issue #2091, opened 2025-09-25
[BUG] RandomLeafFilter constructor will divide by zero during unit test.


stolk:
### Environment
Debian Trixie

### Describe the bug
With FP exceptions enabled, the unit test will throw a FP exception for the RandomLeafFilter code, as it will divide by zero.

### To Reproduce
Enable FP exceptions (MR that adds toggle for this will follow.)

### Expected behavior
No FP exception thrown.

### Additional context
I have a fix, will create MR.

PR #2093, opened 2025-09-25
Fixes div by zero. Adds FPE to tests.

stolk:
This MR adds a toggle to run the vbe_test program with FP exceptions enabled.

~~This MR also fixes issues #2090 #2091 #2092 that were found with this facility:~~

This MR also fixes a division by zero.

### DIFF
--- a/openvdb/openvdb/points/IndexFilter.h
+++ b/openvdb/openvdb/points/IndexFilter.h
@@ -240,7 +240,8 @@ public:
             currentPoints += iter->pointCount();
         }
 
-        const float factor = targetPoints > currentPoints ? 1.0f : float(targetPoints) / float(currentPoints);
+        const float denom = currentPoints > 0 ? float(currentPoints) : 1.0f; // Do not divide by zero.
+        const float factor = targetPoints > currentPoints ? 1.0f : float(targetPoints) / denom;
 
         std::mt19937 generator(seed);
         std::uniform_int_distribution<unsigned int> dist(0, std::numeric_limits<unsigned int>::max() - 1);


openvdb/openvdb/tools/Diagnostics.h
```diff
@@ -441,7 +442,7 @@ struct CheckNormGrad
         : acc(grid.getConstAccessor())
         , invdx2(ValueType(1.0/math::Pow2(grid.voxelSize()[0])))
         , minVal2(_min*_min)
-        , maxVal2(_max*_max)
+        , maxVal2(_max < FLT_MAX ? _max*_max : _max)
```
danrbailey: Although this tool asserts that it can only be used with floating-point, it can be used with other floating-point types, such as double or potentially half, so we can't use FLT_MAX here.

I would also prefer that we slightly re-engineer this to use `std::numeric_limits<ValueType>::max()` instead of using the C-style macro.

openvdb/openvdb/unittest/main.cc
```diff
@@ -12,11 +12,24 @@
...
+
 #include <gtest/gtest.h>
 
 int
 main(int argc, char *argv[])
 {
     ::testing::InitGoogleTest(&argc, argv);
+
+#if defined(__linux)
+    const char* s = getenv("VDB_TEST_FPE");
```
danrbailey: This looks good to me in general, thanks for fixing these issues. However, instead of this being enabled with an environment variable, I would prefer that this be a CMake option such as `OPENVDB_TESTS_FPE` that we could more easily enable in our CI, somewhere like this:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/unittest/CMakeLists.txt#L15

I wouldn't expect that it is useful to be able to enable/disable FPEs at runtime.

openvdb/openvdb/tools/Diagnostics.h
stolk: Excellent point, will change per suggestion.

openvdb/openvdb/unittest/main.cc
stolk: Now toggled with:
```
cmake -D OPENVDB_TESTS_FPE=ON ...
```

openvdb/openvdb/unittest/CMakeLists.txt
```diff
@@ -196,6 +196,10 @@ else()
   )
 endif()
 
+if (OPENVDB_TESTS_FPE)
```
danrbailey, line 201: Looks good, but this still needs to be added as a CMake option at the head of this file with a default value of OFF. It should also be noted in the description of this option that this only has an effect on Linux too.

openvdb/openvdb/unittest/main.cc
```diff
@@ -12,11 +12,20 @@
...
+
 #include <gtest/gtest.h>
 
 int
 main(int argc, char *argv[])
 {
     ::testing::InitGoogleTest(&argc, argv);
+
+#if defined(__linux__) && defined(OPENVDB_TESTS_FPE)
+    feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
```
Idclip: I wonder if we really want `FE_OVERFLOW` on by default. It's perfectly valid if it's handled and my understanding is it will fire regardless. For example, the change in Diagnostics.h:

```
maxVal2(_max < std::numeric_limits<ValueType>::max() ? _max*_max : _max)
```

can still overflow if _max is large. A better solution would be:

```
maxVal2([&](){ 
    auto m2 = _max*_max; 
    return std::isfinite(m2) ? m2 : std::numeric_limits<T>::max();
}())
```

But this will always trigger an overflow, even though the program is well formed. I guess it might still be useful to identify these cases - but as it causes an immediate abort it's probably not super useful.

Issue hit with `FE_DIVBYZERO` and `FE_INVALID` should (usually) always be avoided, so those definitely make sense.

stolk: First off: FLT_MAX is a special value to denote a lack of bounds, so will frequently be encountered. Values smaller than that, but larger than the sqrt(FLT_MAX) will not occur in practice.

They are either small, or they are FLT_MAX, nothing in between.

Hence the solution used here makes sense: special treatment of FLT_MAX, because it **is** a special case, used to signal something.

Second, I disagree on your comment on FE_OVERFLOW. It is important to catch these in unit tests.

NOTE: This change does not enable exceptions for production code, nor for applications, not for `vdb_view`, `vdb_render`, etc. 

This change only (optionally) affects `vdb_test`, which does need to catch these cases.

Idclip: If you want to make the case that FLT_MAX is special that's fine, but that is not explicitly stated for many of our tools - and saying values are either small or FLT_MAX is a huge generalization.

Yes I'm aware this only impacts the tests. There are lots of places where we check for finite and branch accordingly. So having it fire in these situations will be annoying (assuming our tests actually hit these paths).

danrbailey: I think adding just FE_DIVBYZERO and FE_INVALID would be a good start. That's better than what we have right now. By yeah, making it default to OFF for the unit tests would be my vote.

I agree that there are cases where we should allow overflow/underflow to happen. That's something we can guard against so that we can still globally set these flags:
```
fenv_t env;
feholdexcept(&env);
// operation that can generate overflow
fesetenv(&env);
```
I'm not sure it's worth maintaining that level of strictness though.

Idclip: To be clear, I'm fine leaving this in so long as all our tests currently pass with it enabled. I'm not a fan of arbitrarily treating FLT_MAX as a special value in places where input bounds aren't well defined - but if your change is simply to achieve the prior then for this PR it makes sense

stolk: The tests will pass because VDB_TESTS_FPE is off by default. 

This is a (very useful) tool for developers that want it. It finds real bugs.

Also, on breaking tests **if** enabled... I really can't see that as a negative: Breaking unit tests is good. It finds bugs. Having tests pass is not the goal: having correct code is the goal.

It forces a developer to tighten up the use of floating point values. It finds invalid assumptions made by the developer.

Code that produces FP overflows is suspect, and should be flagged: 3.402823466e+38F is a really big number. If your code exceeds it, you probably did something wrong. Just like the case it found, where an 'unbounded' special value was squared, e.g.

stolk: I have removed `FE_OVERFLOW` so that it no longer fires by default. It will only conditionally fire if you toggle FP exceptions on via cmake, and then set the env var `OPENVDB_TEST_OVERFLOW` at runtime. If not set, it will only fire on `FP_INVALID` and `FP_DIVBYZERO` thereby allowing code to overflow float values.

openvdb/openvdb/tools/Diagnostics.h
```diff
@@ -441,7 +442,7 @@ struct CheckNormGrad
         : acc(grid.getConstAccessor())
         , invdx2(ValueType(1.0/math::Pow2(grid.voxelSize()[0])))
         , minVal2(_min*_min)
-        , maxVal2(_max*_max)
+        , maxVal2(_max < std::numeric_limits<ValueType>::max() ? _max*_max : _max)
```
danrbailey: I think we should focus this PR on only divbyzero and invalid FPEs. This change doesn't appear to prevent overflows anyway? a max value of "numeric limits max - 1" will still overflow here.

stolk: I will remove it, per your suggestion, although I do disagree: It does not catch all overflows, but does catch the most prevalent one where `FLT_MAX` is used as un-initialized value. Not only is it common, it is actually occurring in the existing test suite!

openvdb/openvdb/unittest/CMakeLists.txt
stolk: Would this work?
```
+option(OPENVDB_TESTS_FPE "Enable FP exceptions in unit tests (linux only)" FALSE)
```

openvdb/openvdb/tools/Diagnostics.h
stolk: Also, taking it out, means that #2090 bug is not fixed.

stolk: @danrbailey I added the option documentation as you suggested, to the cmake config.
And I removed the FP overflow fix, as you suggested.

Can you please approve the MR? Or at the very least, approve the work-flow:

image

openvdb/openvdb/unittest/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ set(OPENVDB_TESTS "" CACHE STRING [=[
   \"Activate;NodeManager\" would build just the TestActivate.cc and TestNodeManager.cc
   unit tests.]=])
 
+option(OPENVDB_TESTS_FPE "Enable FP exceptions in unit tests (linux only)" FALSE)
```
danrbailey: Can you use `OFF` not `FALSE` please? This is to be consistent with all the other options in VDB.

openvdb/openvdb/unittest/main.cc
```diff
@@ -12,11 +12,25 @@
...
 #include <gtest/gtest.h>
 
 int
 main(int argc, char *argv[])
 {
     ::testing::InitGoogleTest(&argc, argv);
+
+#if defined(__linux__) && defined(OPENVDB_TESTS_FPE)
+    int excepts = FE_DIVBYZERO | FE_INVALID;
+    if (getenv("OPENVDB_TEST_OVERFLOW")) {
```
danrbailey: I'd like to get rid of this overflow runtime check. The current set of unit tests do not pass with overflow checks enabled anyway and I'd rather not mix compile-time and run-time variables here. Can we make this simply:
```
feenableexcept(FE_DIVBYZERO | FE_INVALID)
```
If a user wishes to add overflow checks, they will have to modify this line to add more checks, but it's not an option I think we should support at present.

danrbailey: Actually, we did discuss this in one of our TSC meetings. The ConvexVoxelizer changes you are making break the expected behavior currently. As it stands, the implementation relies on passing values as NaN and detecting them. We believe that the correct thing to do is for us to refactor this implementation so that we can safely enable these NaN checks. However, we don't want to hold up this PR any further as adding in the necessary framework to enable these checks is already a benefit.

What we propose is:

* Only set FE_DIVBYZERO and make the change to IndexFilter.h
* Drop all changes to Stencils.h and ConvexVoxelizer.h

It can be a subsequent effort to tackle adding FE_INVALID and the relevant code changes to unlock this.

openvdb/openvdb/unittest/CMakeLists.txt
stolk: Changed per suggestion.

openvdb/openvdb/unittest/main.cc
stolk: I do not understand. How can the implementation rely on taking the square root of negative numbers?

The ConvexVoxelizer change does not change the NaN passing. It changes so that it does not take the square root of -0.00001 e.g.

danrbailey: The algorithm does `y = std::sqrt(x)` so if x is negative then y is NaN. Subsequently the method receiving the result does `std::isfinite(y)` and based on true/false value does some other logic. By clamping `x`, that ensures that no NaNs are ever generated and thus the subsequent conditional always evaluates to true. This breaks the existing algorithm.

To be clear, this approach is not generally considered to be good practice and we think that the implementation should be adjusted so that it doesn't rely on NaN values, but at present that is not the case.

stolk: Changed per suggestion.

NOTE: I think it made no sense to actually remove the flags, as the unit tests run just fine, before and after this patch, as FPE is off by default. 

Passing tests is not a goal. It is good to have tests fail, as it means you have found something. And only a developer who turned FPE on will see those fails. Your CI builds will just pass fine as is.

In any case, those flags have been removed, as was the commit with the sqrt() fixed in them.

PR #2101, opened 2025-10-04
Half Grid Support PR

apradhana:
**What are the most important things that I should take-away from this PR at the code-level?**
We want to make progress on making `HalfGrid` native in `OpenVDB`. This means:
- We have `HalfDecl.h` that brings in **either** an `OpenEXR half` type or the `internal half` type that we have in `openvdb`. It's good to note at this point that this type is included in `math.h` and `Vec3.h`.
- We have a few type-traits functions and class to handle "specialization" with Half.
- We have an `ExtendedRealGridTypes` in `Types.h`. This is used to build `NumericGridTypes`, which is used to build `GridTypes`. The function `openvdb::initialize` makes sure that `HalfGrid` is registered (by calling `GridTypes::foreach<RegisterGrid>()`).
- These tools work with `HalfGrid`:

**What `type_traits` functions/idioms do you add?**
- Adds `openvdb::is_floating_point` which inherits from `std::is_floating_point`. Itd allows specialization for `Half` type.
- Adds `openvdb::is_signed` which inherits from `std::is_signed`. It allows specialization for `Half` type.
- Adds a trait class called `ComputeTypeFor` to convert `Half -> float`, `Vec2H -> Vec2s`, `Vec3H -> Vec3s`, Vec4H -> Vec4s`.

**What medium tests did you do?**
    - authoring `sphere level set`
    - authoring `platonic solid level set`
    - change level set background
    - level set measure
    - mesh to volume
    - volume to mesh
    - rendering half grid using `vdb_render` (to test interpolation)

### DIFF
--- a/openvdb/openvdb/Metadata.h
+++ b/openvdb/openvdb/Metadata.h
@@ -359,6 +359,7 @@ operator<<(std::ostream& ostr, const Metadata& metadata)
 using BoolMetadata   = TypedMetadata<bool>;
 using DoubleMetadata = TypedMetadata<double>;
 using FloatMetadata  = TypedMetadata<float>;
+using HalfMetadata   = TypedMetadata<math::half>;
 using Int32Metadata  = TypedMetadata<int32_t>;
 using Int64Metadata  = TypedMetadata<int64_t>;
 using StringMetadata = TypedMetadata<std::string>;
--- a/openvdb/openvdb/Types.h
+++ b/openvdb/openvdb/Types.h
@@ -8,27 +8,7 @@
 #include "Platform.h"
 #include "TypeList.h" // backwards compat
 
-#ifdef OPENVDB_USE_IMATH_HALF
-#ifdef OPENVDB_IMATH_VERSION
-#include <Imath/half.h>
-#else
-#include <OpenEXR/half.h>
-#endif
-namespace openvdb {
-OPENVDB_USE_VERSION_NAMESPACE
-namespace OPENVDB_VERSION_NAME {
-namespace math {
-using half = half;
-}}}
-#else
-#include <openvdb/math/Half.h>
-namespace openvdb {
-OPENVDB_USE_VERSION_NAMESPACE
-namespace OPENVDB_VERSION_NAME {
-namespace math {
-using half = internal::half;
-}}}
-#endif
+#include <openvdb/math/HalfDecl.h>
 
 #include <openvdb/math/Math.h>
 #include <openvdb/math/BBox.h>
@@ -58,12 +38,13 @@ using Int64   = int64_t;
 using Int     = Int32;
 using Byte    = unsigned char;
 using Real    = double;
+using Half    = math::half;
 
 // Two-dimensional vector types
 using Vec2R = math::Vec2<Real>;
 using Vec2I = math::Vec2<Index32>;
 using Vec2f = math::Vec2<float>;
-using Vec2H = math::Vec2<math::half>;
+using Vec2H = math::Vec2<Half>;
 using math::Vec2i;
 using math::Vec2s;
 using math::Vec2d;
@@ -72,7 +53,7 @@ using math::Vec2d;
 using Vec3R = math::Vec3<Real>;
 using Vec3I = math::Vec3<Index32>;
 using Vec3f = math::Vec3<float>;
-using Vec3H = math::Vec3<math::half>;
+using Vec3H = math::Vec3<Half>;
 using Vec3U8 = math::Vec3<uint8_t>;
 using Vec3U16 = math::Vec3<uint16_t>;
 using math::Vec3i;
@@ -87,7 +68,7 @@ using BBoxd = math::BBox<Vec3d>;
 using Vec4R = math::Vec4<Real>;
 using Vec4I = math::Vec4<Index32>;
 using Vec4f = math::Vec4<float>;
-using Vec4H = math::Vec4<math::half>;
+using Vec4H = math::Vec4<Half>;
 using math::Vec4i;
 using math::Vec4s;
 using math::Vec4d;
@@ -499,6 +480,54 @@ struct CanConvertType<ValueMask, T> { enum {value = CanConvertType<bool, T>::val
 ////////////////////////////////////////
 
 
+/// @brief Maps low precision floating point types to a floating-point type suitable for computation.
+///
+/// Selectively promotes low-precision types (e.g., `math::half`) to a higher-precision
+/// type (e.g., `float`) to balance numerical accuracy and performance in
+/// compute-intensive code.
+///
+/// Types already suitable for computation, float precision and higher, are left unchanged.
+///
+/// This trait is useful when intermediate compute precision must differ from storage
+/// precision to prevent precision loss or instability.
+///
+/// @tparam T The input type to be mapped to a compute-appropriate floating-point type.
+template <typename T>
+struct ComputeTypeFor
+{
+    using type = T;
+};
+
+// Specialization for half -> float
+template <>
+struct ComputeTypeFor<Half>
+{
+    using type = float;
+};
+
+// Specialization for Vec<half> -> Vec<float>
+template <>
+struct ComputeTypeFor<Vec2H>
+{
+    using type = Vec2s;
+};
+
+template <>
+struct ComputeTypeFor<Vec3H>
+{
+    using type = Vec3s;
+};
+
+template <>
+struct ComputeTypeFor<Vec4H>
+{
+    using type = Vec4s;
+};
+
+
+////////////////////////////////////////
+
+
 /// @brief CopyConstness<T1, T2>::Type is either <tt>const T2</tt>
 /// or @c T2 with no @c const qualifier, depending on whether @c T1 is @c const.
 /// @details For example,
@@ -515,6 +544,21 @@ template<typename FromType, typename ToType> struct CopyConstness<const FromType
     using Type = const ToType;
 };
 /// @endcond
+
+
+////////////////////////////////////////
+template<class T>
+struct is_floating_point : std::is_floating_point<T> { };
+
+template<>
+struct is_floating_point<Half> : std::is_floating_point<float> { };
+
+
+template<class T>
+struct is_signed : std::is_signed<T> { };
+
+template<>
+struct is_signed<Half> : std::is_signed<float> { };
 
 
 ////////////////////////////////////////
@@ -762,11 +806,11 @@ class PartialCreate {};
 // For half compilation
 namespace math {
 template<>
-inline auto cwiseAdd(const math::Vec3<math::half>& v, const float s)
+inline auto cwiseAdd(const Vec3H& v, const float s)
 {
-    math::Vec3<math::half> out;
-    const math::half* ip = v.asPointer();
-    math::half* op = out.asPointer();
+    Vec3H out;
+    const Half* ip = v.asPointer();
+    Half* op = out.asPointer();
     for (unsigned i = 0; i < 3; ++i, ++op, ++ip) {
         OPENVDB_NO_TYPE_CONVERSION_WARNING_BEGIN
         *op = *ip + s;
--- a/openvdb/openvdb/math/FiniteDifference.h
+++ b/openvdb/openvdb/math/FiniteDifference.h
@@ -299,10 +299,14 @@ temporalIntegrationSchemeToMenuName(TemporalIntegrationScheme tis)
 /// ( f(x+dx/2) - f(x-dx/2) ) / dx  = df/dx (x) + error,
 /// where the error is fifth-order in smooth regions: O(dx) <= error <=O(dx^5)
 template<typename ValueType>
-inline ValueType
+inline typename ComputeTypeFor<ValueType>::type
 WENO5(const ValueType& v1, const ValueType& v2, const ValueType& v3,
     const ValueType& v4, const ValueType& v5, float scale2 = 0.01f)
 {
+    using ComputeType = typename ComputeTypeFor<ValueType>::type;
+
+    const ComputeType f1 = v1, f2 = v2, f3 = v3, f4 = v4, f5 = v5;
+
     const double C = 13.0 / 12.0;
     // WENO is formulated for non-dimensional equations, here the optional scale2
     // is a reference value (squared) for the function being interpolated.  For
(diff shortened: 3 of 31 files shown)


openvdb/openvdb/Metadata.h
```diff
@@ -359,6 +359,7 @@ operator<<(std::ostream& ostr, const Metadata& metadata)
 using BoolMetadata   = TypedMetadata<bool>;
 using DoubleMetadata = TypedMetadata<double>;
 using FloatMetadata  = TypedMetadata<float>;
+using HalfMetadata   = TypedMetadata<math::half>;
```
Idclip, line 362: Use your Half alias in Types?

openvdb/openvdb/Types.h
```diff
@@ -8,27 +8,7 @@
...
-#else
-#include <openvdb/math/Half.h>
-namespace openvdb {
-OPENVDB_USE_VERSION_NAMESPACE
-namespace OPENVDB_VERSION_NAME {
-namespace math {
-using half = internal::half;
-}}}
-#endif
+#include <openvdb/math/HalfDecl.h>
```
Idclip, line 11: Was there any reason for this change?

openvdb/openvdb/Types.h
```diff
@@ -87,7 +68,7 @@ using BBoxd = math::BBox<Vec3d>;
 using Vec4R = math::Vec4<Real>;
 using Vec4I = math::Vec4<Index32>;
 using Vec4f = math::Vec4<float>;
-using Vec4H = math::Vec4<math::half>;
+using Vec4H = math::Vec4<Half>;
```
Idclip, line 71: Did you consider adding alias for half matrices? Or rather, was there a specific reason we only add aliases for vec2/3/4s?

openvdb/openvdb/math/FiniteDifference.h
```diff
@@ -299,55 +299,66 @@ temporalIntegrationSchemeToMenuName(TemporalIntegrationScheme tis)
...
-        A2*(5.0*v3 -     v2 +  2.0*v4) +
-        A3*(2.0*v3 + 5.0*v4 -      v5))/(6.0*(A1+A2+A3)));
+    const double A1=0.1/math::Pow2(C*math::Pow2(f1-2*f2+f3)+0.25*math::Pow2(f1-4*f2+3.0*f3)+eps),
+                 A2=0.6/math::Pow2(C*math::Pow2(f2-2*f3+f4)+0.25*math::Pow2(f2-f4)+eps),
+                 A3=0.3/math::Pow2(C*math::Pow2(f3-2*f4+f5)+0.25*math::Pow2(3.0*f3-4*f4+f5)+eps);
+
+    return static_cast<ComputeType>(static_cast<ComputeType>(
+        A1*(2.0*f1 - 7.0*f2 + 11.0*f3) +
+        A2*(5.0*f3 -     f2 +  2.0*f4) +
+        A3*(2.0*f3 + 5.0*f4 -      f5))/(6.0*(A1+A2+A3)));
```
Idclip, line 324: Instead of doing this, why not just recurse?

```c++
template<typename ValueType>
inline typename ComputeTypeFor<ValueType>::type
WENO5(const ValueType& v1, const ValueType& v2, const ValueType& v3,
    const ValueType& v4, const ValueType& v5, float scale2 = 0.01f)
{
	using CT = typename ComputeTypeFor<ValueType>::type;
	if constexpr (!std::is_same_v<CT, ValueType>) {
		return WENO5<CT>(CT(v1), CT(v2), CT(v3), CT(v4), CT(v5), scale);
	}

	... // actual impl
}
```

This would make the diff much smaller (potentially reducing errors), constrains the conversion logic to the invocation of the function and avoid redundant copies when the types are the same. You can apply this pattern to other places too.

openvdb/openvdb/math/FiniteDifference.h
```diff
@@ -299,55 +299,66 @@ temporalIntegrationSchemeToMenuName(TemporalIntegrationScheme tis)
 /// ( f(x+dx/2) - f(x-dx/2) ) / dx  = df/dx (x) + error,
 /// where the error is fifth-order in smooth regions: O(dx) <= error <=O(dx^5)
 template<typename ValueType>
-inline ValueType
+inline typename ComputeTypeFor<ValueType>::type
```
Idclip, line 302: Why return the ComputeType? When/how do we draw the line between the CT being an internal arithmetic type vs propagating it between function which may be involved in many fused arithmetic ops?

openvdb/openvdb/Types.h
```diff
@@ -499,6 +480,54 @@ struct CanConvertType<ValueMask, T> { enum {value = CanConvertType<bool, T>::val
...
+struct ComputeTypeFor<Vec3H>
+{
+    using type = Vec3s;
+};
+
+template <>
+struct ComputeTypeFor<Vec4H>
+{
+    using type = Vec4s;
+};
```
Idclip, line 525: Have you considered using `ValueTraits` or like types to support all VDB types which can be templated on scalars to promote half element types?

openvdb/openvdb/Types.h
```diff
@@ -499,6 +480,54 @@ struct CanConvertType<ValueMask, T> { enum {value = CanConvertType<bool, T>::val
...
+struct ComputeTypeFor
+{
+    using type = T;
+};
+
+// Specialization for half -> float
+template <>
+struct ComputeTypeFor<Half>
+{
+    using type = float;
```
Idclip, line 505: Why not use `PromoteType` here and below

openvdb/openvdb/Types.h
```diff
@@ -517,6 +546,21 @@ template<typename FromType, typename ToType> struct CopyConstness<const FromType
 /// @endcond
 
 
+////////////////////////////////////////
+template<class T>
+struct is_floating_point : std::is_floating_point<T> { };
+
+template<>
+struct is_floating_point<Half> : std::is_floating_point<float> { };
```
Idclip, line 553: Would be better to derive from std::true_type?

openvdb/openvdb/math/Math.h
```diff
@@ -73,6 +74,16 @@ template<> inline std::string zeroVal<std::string>() { return ""; }
...
 
+
+/// @note Extends the implementation of std::is_arithmetic to support math::half
+template<typename T>
+struct is_arithmetic : std::is_arithmetic<T> {};
+template<>
+struct is_arithmetic<math::half> : std::true_type {};
+// Helper variable template (equivalent to std::is_arithmetic_v)
+template<typename T>
+inline constexpr bool is_arithmetic_v = is_arithmetic<T>::value;
```
Idclip, line 85: Put these in Types, alongside the other traits

openvdb/openvdb/math/Math.h
```diff
@@ -145,16 +156,18 @@ template<> inline std::string negative(const std::string& val) { return val; }
 
 //@{
 /// Tolerance for floating-point comparison
-template<typename T> struct Tolerance { static T value() { return zeroVal<T>(); } };
-template<> struct Tolerance<float>    { static float value() { return 1e-8f; } };
-template<> struct Tolerance<double>   { static double value() { return 1e-15; } };
+template<typename T> struct Tolerance   { static T value() { return zeroVal<T>(); } };
+template<> struct Tolerance<math::half> { static math::half value() { return math::half(0.00097656f); } };
```
Idclip, line 174: A comment for the choice

openvdb/openvdb/math/Math.h
```diff
@@ -361,6 +374,10 @@ isApproxZero(const Type& x, const Type& tolerance)
 }
 
 
+/// Return @c true if @a x is less than zero.
+inline bool
+isNegative(math::half& x) { return x.isNegative(); }
```
Idclip, line 393: Why not override the below specialization instead of this non-const overload?

openvdb/openvdb/math/Math.h
```diff
@@ -589,44 +627,61 @@ Pow(double b, double e)
...
+{
+    return a > b ? a : b;
+}
+
+template<typename Type>
+inline const Type&
+max_impl(const Type& a, const Type& b)
+{
+    return std::max(a,b);
+}
```
Idclip, line 657: You can probably just delete these internal functions and change `Max()` just do `a < b ? b : a` instead of calling `std::max` - note that you need to use the less than op, your code currently isn't equivalent (max uses lt and returns first arg when equal)

openvdb/openvdb/math/Math.h
```diff
@@ -644,44 +699,64 @@ inline const Type&
...
+{
+    return a < b ? a : b;
+}
+
+template<typename Type>
+inline const Type&
+min_impl(const Type& a, const Type& b)
+{
+    return std::min(a,b);
+}
```
Idclip, line 735: Same point as Max

openvdb/openvdb/math/Stencils.h
```diff
@@ -35,9 +35,11 @@ template<typename DerivedType, typename GridT, bool IsSafe>
 class BaseStencil
 {
 public:
-    typedef GridT                                       GridType;
-    typedef typename GridT::TreeType                    TreeType;
-    typedef typename GridT::ValueType                   ValueType;
+    typedef GridT                    GridType;
+    typedef typename GridT::TreeType TreeType;
+
+    typedef typename ComputeTypeFor<typename GridT::ValueType>::type ValueType;
```
Idclip, line 41: People rely on the ValueType typename representing the actual ValueType - it's a public typename. Obviously doesn't change existing code, but removes the meaning of the typename for anything which overrides a ComputeType - so this should be changed. Applies elsewhere

openvdb/openvdb/openvdb.h
```diff
@@ -89,12 +91,14 @@ using VectorGrid   = Vec3fGrid;
 /// @{
 /// The floating point Grid types which OpenVDB will register by default.
 using RealGridTypes   = TypeList<FloatGrid, DoubleGrid>;
+/// The extended floating point Grid types which OpenVDB will register by default.
+using ExtendedRealGridTypes = RealGridTypes::Append<HalfGrid>;
```
Idclip, line 95: Unsure about this, I think it should ideally live in the above list. Would like other people to comment

openvdb/openvdb/tools/Interpolation.h
```diff
@@ -766,17 +781,18 @@ inline typename TreeT::ValueType
 BoxSampler::sample(const TreeT& inTree, const Vec3R& inCoord)
```
Idclip, line 781: Here we continue to return ValueType, not ComputeT?

openvdb/openvdb/math/Vec3.h
```diff
@@ -663,11 +663,13 @@ using Vec3i = Vec3<int32_t>;
 using Vec3ui = Vec3<uint32_t>;
 using Vec3s = Vec3<float>;
 using Vec3d = Vec3<double>;
+using Vec3h = Vec3<math::half>;
```
Idclip, line 666: So none of Vec3<Half> methods (length/cross/etc) will use ComputeType? I've run into some code such as:
```c++
ComputeType(Vec3H().length());
```

Where the intended use should probably be:
```c++
ComputeType(Vec3H()).length();
```
This ties in to the somewhat arbitrary line that has been drawn with where the CT stops being propagated. It should probably apply to the underlying math of the types when running on grid operations, but this is not an easy distinction to make - without it it is easy to break.

I'd like others to comment if they have thoughts.

openvdb/openvdb/tools/LevelSetAdvect.h
```diff
@@ -448,19 +450,19 @@ sampleField(ValueType time0, ValueType time1)
     this->cook("Sampling advection field");
 
     // Find the extrema of the magnitude of the velocities
-    ValueType maxAbsV = 0;
+    ComputeType maxAbsV = 0;
     VectorType* v = mVelocity;
     for (size_t i = 0; i < voxelCount; ++i, ++v) {
-        maxAbsV = math::Max(maxAbsV, ValueType(v->lengthSqr()));
+        maxAbsV = math::Max(maxAbsV, ComputeType(v->lengthSqr()));
```
Idclip, line 456: I believe you're computing the lengthsqr on the half vec here?

openvdb/openvdb/tools/LevelSetAdvect.h
```diff
@@ -564,9 +566,9 @@ euler(const LeafRange& range, ValueType dt, Index phiBuffer, Index resultBuffer)
         for (VoxelIterT voxelIter = leafIter->cbeginValueOn(); voxelIter; ++voxelIter, ++vel) {
             const Index i = voxelIter.pos();
             stencil.moveTo(voxelIter);
-            const ValueType a =
-                stencil.getValue() - dt * vel->dot(GradT::result(map, stencil, *vel));
-            result[i] = Nominator ? Alpha * phi[i] + Beta * a : a;
+            const ComputeType a =
+                stencil.getValue() - dt * ComputeType(vel->dot(GradT::result(map, stencil, *vel)));
```
Idclip, line 570: vel->dot can run at half precision?

openvdb/openvdb/tools/LevelSetMorph.h
```diff
@@ -512,12 +515,12 @@ sampleXformedSpeed(const LeafRange& range, Index speedBuffer)
             bool isZero = true;
             for (VoxelIterT voxelIter = leafIter->cbeginValueOn(); voxelIter; ++voxelIter) {
                 const Vec3R xyz = map.applyMap(voxelIter.getCoord().asVec3d());//world space
-                const ValueType a = math::SmoothUnitStep((mask.wsSample(xyz)-min)*invNorm);
+                const ComputeType a = math::SmoothUnitStep((mask.wsSample(xyz)-min)*invNorm);
                 ValueType& s = speed[voxelIter.pos()];
                 s -= target.wsSample(xyz);
-                s *= invMask ? 1 - a : a;
+                s *= invMask ? ValueType(1 - a) : ValueType(a);
```
Idclip, line 521: Arithmetic at ValueType precision

openvdb/openvdb/math/FiniteDifference.h
ghurstunither: Yes, this is a good question. For utilities like finite differences, stencil, and operators, when they returned `ValueType`, I noticed the tools that called them always immediately casted them to `ComputeType` for further computations.

The problem then was the finite difference was calculated with `ComputeType`, casted to `ValueType` upon return, only to immediately be recast back to `ComputeType`. This of course introduces rounding errors for `ValueType == half`.

Perhaps I was a bit too presumptuous, but then I couldn't think of a reason these should ever return half precision, especially since well behaved functions can easily have numerically unstable derivatives.

My justification then was that for all other types `ComputeType` is `ValueType`, and so no backward compatibility is broken.

[I realize this still doesn't answer your question of where the line in the sand is...]

openvdb/openvdb/math/FiniteDifference.h
ghurstunither: This is a great idea. I'll look into doing this.

openvdb/openvdb/math/Stencils.h
ghurstunither: Yes, sloppy coding on my part!

openvdb/openvdb/math/Math.h
ghurstunither: It's equal to $2^{-10}$, which is essentially the machine epsilon for `math::half`.

openvdb/openvdb/math/Math.h
ghurstunither: The missing `const` seems to be an oversight. Thank you for going through all of these changes with a fine tooth comb to find this.

For all methods the half class implemented, I tried to call from the Math.h analogues.

PR #2103, opened 2025-10-08
Add const mask access to bool LeafBuffer

danrbailey:
This extends the bool partial specialization of LeafBuffer to allow direct const access to the mask outside of the Tree hierarchy classes.

### DIFF
--- a/openvdb/openvdb/tree/LeafBuffer.h
+++ b/openvdb/openvdb/tree/LeafBuffer.h
@@ -514,6 +514,9 @@ public:
     /// @warning This method should only be used by experts seeking low-level optimizations.
     const WordType* data() const { return const_cast<LeafBuffer*>(this)->data(); }
 
+    /// @brief Return raw LeafBuffer data
+    const NodeMaskType& storage() const { return mData; }
+
 private:
     // Allow the parent LeafNode to access this buffer's data.
     template<typename, Index> friend class LeafNode;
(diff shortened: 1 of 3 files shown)


openvdb/openvdb/tree/LeafBuffer.h
```diff
@@ -514,6 +514,9 @@ class LeafBuffer<bool, Log2Dim>
     /// @warning This method should only be used by experts seeking low-level optimizations.
     const WordType* data() const { return const_cast<LeafBuffer*>(this)->data(); }
 
+    /// @brief Return LeafBuffer value mask
+    const NodeMaskType& mask() const { return mData; }
```
Idclip: This isn't strictly the mask though right? For LeafBuffer's, this is the data - which of course is stored via a node-mask, but I wonder if calling the API method "mask" is confusing - even the comment referring to it as a "value mask" is a bit odd. Unfortunately can't be `data`... `storage()` ? Otherwise lgtm

kmuseth: yes I agree  - this is confusing naming since it only applies to the bool template specialization. It's that payload of the leaf node, not its mask

pendingchanges/bool_mask_access.txt
```diff
@@ -0,0 +1,3 @@
+API changes:
+- Add LeafBuffer<bool, Log2Dim>::mask() for direct const access to the
```
kmuseth: confusing naming

openvdb/openvdb/tree/LeafBuffer.h
danrbailey: Renamed to `storage()` now, thanks!

PR #2105, opened 2025-10-09
Removed unnecessary USE_NUMPY build option

Idclip:


.github/workflows/weekly.yml
```diff
@@ -160,7 +160,7 @@ jobs:
           - { name: 'half',  build: 'Release', components: 'core,python,bin,view,render,test',               cmake: '-DUSE_BLOSC=OFF -DUSE_IMATH_HALF=ON' }
           - { name: 'sse',   build: 'Release', components: 'core,python,bin,view,render,test',               cmake: '-DOPENVDB_SIMD=SSE42' }
           - { name: 'avx',   build: 'Release', components: 'core,python,bin,view,render,test',               cmake: '-DOPENVDB_SIMD=AVX' }
-          - { name: 'numpy', build: 'Release', components: 'core,python,bin,view,render,test',               cmake: '-DUSE_NUMPY=ON -DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON' }
+          - { name: 'pygrd', build: 'Release', components: 'core,python,bin,view,render,test',               cmake: '-DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON' }
```
matthewdcong: A bit pedantic, but can we just change this to `pygrid`? `pygrd` seems unnecessarily concise.

openvdb/openvdb/python/test/TestOpenVDB.py
```diff
@@ -728,12 +718,7 @@ def createArrays():
 
     def testMeshConversion(self):
         import time
-
-        # Skip this test if NumPy is not available.
-        try:
-            import numpy as np
-        except ImportError:
-            return
+        import numpy as np
```
matthewdcong: If we assume that numpy is available, we can move these imports to the top of the file.

matthewdcong:
Two small comments, but otherwise lgtm.

PR #2106, opened 2025-10-13
Minimum Viable Product for Half Grid Support

apradhana:
This PR is trying to offer a Minimum Viable Product for Half Grid Support in OpenVDB 13.

Things covered in the PR are:
- Native HalfGrid registration in the OpenVDB library.

### DIFF
--- a/openvdb/openvdb/unittest/TestMath.cc
+++ b/openvdb/openvdb/unittest/TestMath.cc
@@ -392,8 +392,8 @@ TEST_F(TestMath, testMinMaxImpl)
 {
     using namespace openvdb;
     { // Test equality
-        const float a = 3.14;
-        const float b = 3.14;
+        const float a = 3.14f;
+        const float b = 3.14f;
         const Half c = static_cast<Half>(a);
         const Half d = static_cast<Half>(b);
         EXPECT_EQ(&std::max(a, b), &math::internal::max_impl(a, b));


openvdb/openvdb/math/HalfDecl.h
```diff
@@ -0,0 +1,30 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: MPL-2.0
```
danrbailey: Note this license needs updating.

openvdb/openvdb/openvdb.h
```diff
@@ -89,12 +91,14 @@ using VectorGrid   = Vec3fGrid;
 /// @{
 /// The floating point Grid types which OpenVDB will register by default.
```
danrbailey, line 92: I don't think these typedefs are used for anything except deciding which grids to register (and a few unit tests). We should however, update the instantiation implementation to add half support to align with this implementation:

https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/CMakeLists.txt#L304

All the generic topology algorithms like clipping or activate should be explicitly instantiated for half grids.

danrbailey (commented):
This looks good in general. However, if we do go this route and split this into two separate PRs, this first PR should disable anything for which we need a compute type algorithm, for example `BoxSampler::sample()`. There should be a static assert ensuring that it is not a half tree, otherwise this will (currently) do trilinear interpolation at half type which will ultimately change behavior when we subsequently add compute type.

openvdb/openvdb/math/HalfDecl.h
apradhana: Thanks for catching this, Dan!

openvdb/openvdb/unittest/TestGridIO.cc
```diff
@@ -190,7 +191,92 @@ TestGridIO::readAllTest()
     ::remove("something.vdb2");
 }
 
+void TestGridIO::testCreateWriteReadHalf() {
```
apradhana, line 194: @danrbailey and @Idclip: is this the serialization test that you guys are thinking of for HalfGrid?

apradhana: @ghurstunither for viz.

openvdb/openvdb/CMakeLists.txt
```diff
@@ -324,6 +324,7 @@ if(USE_EXPLICIT_INSTANTIATION)
 
   set(OPENVDB_LIBRARY_INSTANTIATION_ALL_TREES
     MaskTree
+    HalfTree
```
apradhana, line 333: Is this what you are thinking about, @danrbailey? @ghurstunither for viz.

openvdb/openvdb/unittest/TestGridIO.cc
danrbailey: Yep, this looks good to me, thank you!

openvdb/openvdb/CMakeLists.txt
danrbailey: Yep, thanks!

openvdb/openvdb/math/Math.h
```diff
@@ -361,6 +374,10 @@ isApproxZero(const Type& x, const Type& tolerance)
 }
 
 
+/// Return @c true if @a x is less than zero.
+inline bool
+isNegative(math::half& x) { return x.isNegative(); }
+
```
Idclip: Why is this necessary, default version below should suffice so can just remove?

openvdb/openvdb/math/Math.h
```diff
@@ -589,44 +627,61 @@ Pow(double b, double e)
 
 // ==========> Max <==================
 
+namespace internal {
+
+inline const math::half&
+max_impl(const math::half& a, const math::half& b)
+{
+    return a > b ? a : b;
+}
```
Idclip, line 632: Please use < (match the standard so that you are returning the same value at it when a== b). Maybe even add a unit test that `&math::max_impl(a,b) == &std::max(a,b)` where a==b

apradhana: Thanks for catching this, @Idclip . Added unit tests to address this issue and caught a different behavior with `math::Min` too.

openvdb/openvdb/unittest/TestGridIO.cc
apradhana: Thank you, @danrbailey !

openvdb/openvdb/openvdb.h
apradhana: Thanks for catching this, @danrbailey . Based on TSC conversation a week ago, I registered `HalfGrid` under `NumericGridTypes`.

openvdb/openvdb/math/Math.h
apradhana: Thanks for catching this, @Idclip . Removed the template specialization according to the comment.

danrbailey (approved):
Thanks Andre, looks good! Squash before merge please. :)

PR #2108, opened 2025-10-14
Improvements to NanoVDB

kmuseth:
- Major rewrite of nanovdb::NanoToOpenVDB, which now allows nanovdb::IndexGrids and blind (or side-car) data to be converted into an openvdb grid.
- Added stand-alone function nanovd::writeUncompressedGrid that allows nanovdb::IndexGrids and blind (or side-car) data to be written directly to a stream.
- Extended nanovdb::GridBlindDataSemantic with nanovdb::GridClass types to support conversion between IndexGrids and regular grids.
- nanovdb_print now (optionally) prints the blind data count per grid.
- Removed nanovdb::ValueIndexMask and nanovdb::ValueOnIndexMask since they are no longer used by fVDB (or elsewhere).
- NanoVDB version is now 32.9.0, so the ABI of the grid is preserved but not the API. However, the API changes are very minor!

### DIFF
--- a/pendingchanges/nanovdb.txt
+++ b/pendingchanges/nanovdb.txt
@@ -16,6 +16,7 @@ New features:
 - Added nanovdb/util/cuda/Injection.cuh, which facilitates sidecar data copying - on the GPU - in the overlap of two grids with different topology.
 - Added Mat classes and other math types to NanoVDB.
 - Major rewrite of nanovdb::NanoToOpenVDB, which now allows nanovdb::IndexGrids and blind (or sidecar) data to be converted into an openvdb grid.
+- Added support for LevelSet, FogVolume, and Staggered grid semantics. Added toGridClass and toSemantic to map Grid Blind data semantic to Grid class and vice versa.
 
 Improvements:
 - Improvements to mGPU grid building and introduction of a separate test executable for mGPU tests


pendingchanges/nanovdb.txt
```diff
@@ -0,0 +1,18 @@
+NanoVDB:
+
+Improvements:
+- Added nanovdb/tools/cuda/DilateGrid.cuh which performs topology dilation on the GPU.
+- Added nanovdb/tools/cuda/MergeGrids.cuh which merges the topology of two grids on the GPU.
+- Added Mat classes and other math types to NanoVDB
+- Improvements to mGPU grid building and introduction of a separate test executable for mGPU tests
+- Major rewrite of nanovdb::NanoToOpenVDB, which now allows nanovdb::IndexGrids and blind (or side-car) data to be converted into an openvdb grid.
+- Added stand-alone function nanovd::writeUncompressedGrid that allows nanovdb::IndexGrids and blind (or side-car) data to be written directly to a stream.
```
matthewdcong: nanovdb

nanovdb/nanovdb/tools/NanoToOpenVDB.h
```diff
@@ -213,129 +350,171 @@ NanoToOpenVDB<T>::processNode(const SrcNodeT *srcNode)
...
-        *dst++ = srcNode->getValue(i);
-        *dst++ = srcNode->getValue(i+1);
-        *dst++ = srcNode->getValue(i+2);
-        *dst++ = srcNode->getValue(i+3);
+    dstLeaf->setOrigin(trait::mapCoord(srcLeaf->origin()));
+    dstLeaf->setValueMask(trait::mapMask(srcLeaf->valueMask()));
+
+    if constexpr(!BuildTraits<NanoBuildT>::is_special) {
+        const auto* src = trait::mapPtr(srcLeaf->data()->mValues);// doesn't work for compressed data, bool or ValueMask
+        for (auto *dst = dstLeaf->buffer().data(), *end = dst + 512; dst != end; dst += 4, src += 4) {
```
matthewdcong, line 414: Is this loop needed? It seems like this could be a memcpy or std::copy.

nanovdb/nanovdb/tools/NanoToOpenVDB.h
```diff
@@ -213,129 +350,171 @@ NanoToOpenVDB<T>::processNode(const SrcNodeT *srcNode)
...
-    OpenNode0* dstNode = new OpenNode0(); // un-initialized for fast construction
-    dstNode->setOrigin(srcNode->origin());
-    dstNode->setValueMask(srcNode->valueMask());
+    dstLeaf->setOrigin(trait::mapCoord(srcLeaf->origin()));
+    dstLeaf->setValueMask(trait::mapMask(srcLeaf->valueMask()));
+
+    if constexpr(!BuildTraits<NanoValueT>::is_special) {
+        if constexpr(util::is_same_v<NanoIndexT, ValueIndex>) {
+            const auto* src = trait::mapPtr(sideCar + srcLeaf->data()->mOffset);
+            for (auto *dst = dstLeaf->buffer().data(), *end = dst + 512; dst != end; dst += 4, src += 4) {
```
matthewdcong, line 451: same as above

nanovdb/nanovdb/tools/NanoToOpenVDB.h
kmuseth: good point

kmuseth: looking at the documentation for std::copy it looks like it performs a pretty naive for-loop without loop-un-rolling. So, I'm convinced it will actually outperform my code ...

kmuseth: switched to std::copy but kept the old implementation just in case it was in fact faster

matthewdcong (approved):
Looks good, thanks for the changes.

pendingchanges/nanovdb.txt
```diff
@@ -0,0 +1,25 @@
...
+
+Highlights:
+- NanoVDB should no longer be considered a static grid, since we keep adding more tools that support dynamic topology.
+  In other words, the feature gap between OpenVDB and NanoVDB is closing, and the main distinguishing characteristic is
+  the fact that only NanoVDB runs on the GPU. However, OpenVDB of course continues to have a much larger toolbox.
+
+Improvements:
+- Added nanovdb/tools/cuda/DilateGrid.cuh, which performs topology dilation on the GPU.
+- Added nanovdb/tools/cuda/MergeGrids.cuh, which merges the topology of two grids on the GPU.
+- Added nanovdb/tools/cuda/CoarsenGrid.cuh, which performs topological coarsening of grids on the GPU.
```
sifakis, line 13: (Optional) Additional features to mention, if useful:
- Added nanovdb/tools/cuda/PruneGrid.cuh which prunes the topology of a grid using a sidecar mask on the GPU.
- Added nanovdb/util/cuda/Injection.cuh which facilitates sidecar data copying - on the GPU - in the overlap of two grids with different topology.

nanovdb/nanovdb/tools/NanoToOpenVDB.h
```diff
@@ -213,129 +353,179 @@ NanoToOpenVDB<T>::processNode(const SrcNodeT *srcNode)
...
+                dst[0] = src[0];
+                dst[1] = src[1];
+                dst[2] = src[2];
+                dst[3] = src[3];
+            }
+#endif
+        } else {
+            const auto* src = trait::mapPtr(sideCar);
+            auto *dst = dstLeaf->buffer().data();
+            for (int i = 0; i < 512; i += 4) {
```
sifakis, line 461: Somewhat related  to @matthewdcong 's comment above ... a possible optimization might be
```
            auto offset = srcLeaf->firstOffset();
            for (int w = 0; w < 8; ++w) {
                auto word = srcLeaf->valueMask().words()[w];
                for (int bit = 0; bit < 64; ++bit, word >>= 1) *dst++ = src[(word&0x1UL)?offset++:0];
            }
```
in order to avoid calls to popCnt; although those might vectorize on AVX512.
At any rate, this is not essential as leaf node allocation is likely the key contributor to runtime in this tool

sifakis (approved):
Looks good to me!

pendingchanges/nanovdb.txt
```diff
@@ -0,0 +1,33 @@
...
+- Added nanovdb/tools/cuda/RefineGrid.cuh, which performs topological refinement of grids on the GPU.
+- Added nanovdb/tools/cuda/PruneGrid.cuh, which prunes the topology of a grid using a sidecar mask on the GPU.
+- Added nanovdb/util/cuda/Injection.cuh, which facilitates sidecar data copying - on the GPU - in the overlap of two grids with different topology.
+- Added Mat classes and other math types to NanoVDB.
+- Major rewrite of nanovdb::NanoToOpenVDB, which now allows nanovdb::IndexGrids and blind (or sidecar) data to be converted into an openvdb grid.
+
+Improvements:
+- Improvements to mGPU grid building and introduction of a separate test executable for mGPU tests
```
apradhana, line 22: (Optional). Consider mentioning:
- Added support for `LevelSet`, `FogVolume`, and `Staggered` grid semantic. Added `toGridClass` and `toSemantic` to map Grid Blind data semantic to Grid class and vice versa.

apradhana (approved):
One nitpick comment regarding the Change Notes. Looks good to me!

PR #2109, opened 2025-10-14
Feature improvements to points::pca and points::rasterizeSdf

Idclip:
```
OpenVDB:
  New features:
    - Added customization of the output transformation attribute for points::pca(). 
      Can be output as a quaternion or as a combined affine transform with the 
      scale/stretch ommitted.
    - Added support for input quaternions attributes and affine transformations
      (that include scale) in points::rasterizeSdf() for ellipsoidal rasterizations.
```

### DIFF
--- a/openvdb/openvdb/points/PrincipalComponentAnalysis.h
+++ b/openvdb/openvdb/points/PrincipalComponentAnalysis.h
@@ -176,16 +176,16 @@ struct PcaAttributes
         // attributes will be created. The xform will take the form of an
         // orthogonal 3x3 matrix representing the rotation OR a pure
         // reflection.
-        STRETCH_AND_ROTATION_MATRIX
+        STRETCH_AND_UNITARY_MATRIX
     };
 
-    XformOutput xformOutput = XformOutput::STRETCH_AND_ROTATION_MATRIX;
+    XformOutput xformOutput = XformOutput::STRETCH_AND_UNITARY_MATRIX;
 
     /// @brief  Settings for the "stretch" attribute, a floating point vector
     ///   attribute which represents the scaling components of each points
     ///   ellipse or (1.0,1.0,1.0) for isolated points.
     /// @note   This will only be created if the XformOutput is set to
-    ///   STRETCH_AND_QUATERNION or STRETCH_AND_ROTATION_MATRIX.
+    ///   STRETCH_AND_QUATERNION or STRETCH_AND_UNITARY_MATRIX.
     using StretchT = math::Vec3<float>;
     std::string stretch = "stretch";
 
--- a/openvdb/openvdb/unittest/TestPCA.cc
+++ b/openvdb/openvdb/unittest/TestPCA.cc
@@ -516,7 +516,7 @@ TEST_F(TestPCA, testPCAxforms)
     s.nonAnisotropicStretch = 1.0f;
 
     points::PcaAttributes a;
-    a.xformOutput = points::PcaAttributes::XformOutput::STRETCH_AND_ROTATION_MATRIX;
+    a.xformOutput = points::PcaAttributes::XformOutput::STRETCH_AND_UNITARY_MATRIX;
     points::pca(*points1, s, a);
 
     a.xformOutput = points::PcaAttributes::XformOutput::COMBINED_TRANSFORM;


openvdb/openvdb/points/PointRasterizeSDF.h
```diff
@@ -301,17 +302,34 @@ struct EllipsoidSettings
...
 
     /// @param pws  An optional attribute which represents the world space
     ///   position of a point.
     /// @details  This can be useful to override the position of a point in
     ///   index space. If it exists, it must be a Vec3d type.
     std::string pws = "";
+
+    /// Old style "xform" attribute which has since been renamed and ONLY
+    /// supported rotation. If set, will be prioritised and infered to only
+    /// hold a rotation. This will be removed and should not be used.
```
richhones: This comment is a bit confusing, Maybe 'Before 'xform' support was added, only rotation attributes were supported. This has been superceded and should no longer be used. If set, will be prioritised and newer 'xform' will be ignored for backwards compatibility.'

openvdb/openvdb/points/PrincipalComponentAnalysis.h
```diff
@@ -154,17 +158,41 @@ struct PcaSettings
...
+    ///   STRETCH_AND_QUATERNION or STRETCH_AND_ROTATION_MATRIX.
     using StretchT = math::Vec3<float>;
     std::string stretch = "stretch";
 
-    /// @brief  Settings for the "rotation" attribute, a floating point matrix
-    ///   attribute which represents the orthogonal rotation of each points
-    ///   ellipse or the identity matrix for isolated points.
+    /// @brief  Settings for the "rotation" attribute, either a floating point
+    ///   matrix or quaternion attribute which represents the orthogonal rotation
+    ///   of each points ellipse or the identity matrix for isolated points.
```
richhones: this bit of the comment needs updating to 'xform' and to consider the combined transform version.

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
```diff
@@ -583,21 +586,53 @@ pca(PointDataGridT& points,
...
+        if (attrs.format == PcaAttributes::AttributeOutput::STRETCH_AND_QUATERNION) {
+            temps.emplace_back(descriptor.uniqueName("_covariance"));
+            return temps.back();
+        }
+        return attrs.xform;
+    }();
+
+    const std::string stretchAttribName = [&]() {
+        if (attrs.format == PcaAttributes::AttributeOutput::COMBINED_TRANSFORM)  {
+            temps.emplace_back(descriptor.uniqueName("_stetch"));
```
richhones: typo: stretch

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
```diff
@@ -583,21 +586,53 @@ pca(PointDataGridT& points,
...
     const auto& descriptor = leaf->attributeSet().descriptor();
-    const std::vector<std::string> temps {
+    std::vector<std::string> temps {
         descriptor.uniqueName("_weightedpositionsums"),
         descriptor.uniqueName("_inv_weightssum")
     };
 
+    // If we're storing the rotational component as a quaternion, we also remove
+    // the temporary covariance output. If we're storing a combined transform,
+    // we remove the individual stretch component
```
richhones: this comment is a bit confusing. Perhaps 'Separate rotation and stretch attributes are required in the below calculations, if storing outputting rotation as a quarternion, create a temporary covariance matrix, else reuse output xform matrix attribute. Same for stretch.'

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
```diff
@@ -729,22 +768,100 @@ pca(PointDataGridT& points,
...
+    if (attrs.format == PcaAttributes::AttributeOutput::STRETCH_AND_QUATERNION)
+    {
+        manager.foreach([&](LeafNodeT& leafnode, size_t) {
+            AttributeWriteHandle<Mat3T, NullCodec> rotHandle(leafnode.attributeArray(rotIdx));
+            AttributeWriteHandle<Vec3T, NullCodec> stretchHandle(leafnode.attributeArray(strIdx));
+            PcaAttributes::QuatT* Q = initPcaArrayAttribute<PcaAttributes::QuatT>(leafnode, qutIdx, /*fill=*/false);
+            for (Index idx = 0; idx < rotHandle.size(); ++idx) {
+                const auto& rot = rotHandle.get(idx);
+                // If we're a pure reflection (partial or global, det of -1),
+                // we need to convert to a rotation for quat support. Could just
```
richhones: If we have a negative determinant, I think we can just multiply one of the columns by {-1, -1, -1} to fix this? Because we dont actually care about the actual way the vector is pointed, but instead the principal axis this defines which is the same after multiplication with -1. We dont need to update the stretch values (which should always be > 0 anyway iirc).

Idclip: Nice catch on both accounts, fixed

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
Idclip: Fixed

openvdb/openvdb/points/impl/PrincipalComponentAnalysisImpl.h
Idclip: Fixed

openvdb/openvdb/points/PrincipalComponentAnalysis.h
Idclip: Changed to simply "read above docs on output info"

Idclip:
> Happy with the API change, one question about how we fix the reflections.

I've made the suggested changes. I've also renamed `AttributeOutput` -> `XformOutput` and `format` to `xformOutput`

richhones (approved):
Looks good to me!

PR #2112, opened 2025-10-21
Fix minor spelling issues in Convex tools

danrbailey:
This is pulled from https://github.com/AcademySoftwareFoundation/openvdb/pull/2108 and I believe because there are changes to the openvdb/ codebase, this PR is causing the NanoVDB PR to rebuild lots of additional GHA workflows (core library, ax, houdini, etc) that is contributing to the large amount of build and workflows and caches we are consuming at the moment. Once this PR is merged, we can merge master into that PR2108 to eliminate these openvdb/ changes from that PR and reduce the GHA workflows needed for that PR.

@kmuseth @ghurstunither


Idclip:
I mean sure, but this is going to happen again (and it's silly we can so quickly blow through the cache). Can approve, but at this point we should handle the case where houdini has been evicted and just download it again

PR #2115, opened 2025-10-22
Dependency bumps for VDB 13

Idclip:


.github/workflows/ax.yml
```diff
@@ -68,14 +68,17 @@ jobs:
         # Only test unified builds (core+ax). weekly CI should test standalone
         # @note  LLVM 17 builds are incompatible with the vdb python plugin on linux
         config:
+          - { image: '2026',           cxx: 'clang++', build: 'Release', cmake: '' }
```
danrbailey, line 70: Do we want to use a specific clang version here? `2026-clang20.0` ?

This is now a lot of AX builds, is it possible that we could move a few of the in-between versions to the weekly build? If it builds on the earliest and latest supported, we should have reasonable confidence that it will build on the versions in between.

.github/workflows/build.yml
```diff
@@ -70,6 +70,9 @@ jobs:
     strategy:
       matrix:
         config:
+          - { cxx: clang++, image: '2026',           abi: '13', build: 'Release', cmake: '' }
+          - { cxx: g++,     image: '2026',           abi: '13', build: 'Release', cmake: '' }
+          - { cxx: clang++, image: '2026',           abi: '12', build: 'Debug',   cmake: '' }
           - { cxx: clang++, image: '2025-clang19.1', abi: '12', build: 'Release', cmake: '' }
           - { cxx: g++,     image: '2025-clang19.1', abi: '12', build: 'Release', cmake: '' }
           - { cxx: clang++, image: '2025-clang19.1', abi: '12', build: 'Debug',   cmake: '' }
```
danrbailey: I think we should drop the g++ and debug builds for 2025.

.github/workflows/weekly.yml
```diff
@@ -503,10 +501,10 @@ jobs:
     runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-22.04-8c-32g-300h') || 'ubuntu-latest' }}
     name: linux-blosc:${{ matrix.blosc }}
     container:
-      image: aswf/ci-openvdb:2025-clang19.1
+      image: aswf/ci-openvdb:2026
     strategy:
       matrix:
-        blosc: ['1.18.0','1.19.0','1.20.0','1.21.0']
+        blosc: ['1.18.0','1.19.0','1.20.0','1.21.0', 'latest']
```
danrbailey: Our blosc minimum is 1.17.0, but we're not actually testing this? I also think we should remove some of the in-between versions here or move them to weekly.

Idclip: Removed all but 1.21.0 and latest (as discussed, these are already running on weekly.yml)

.github/workflows/build.yml
Idclip: Done!

.github/workflows/ax.yml
Idclip: As discussed, only reason we use the clang suffix is to work around issues in the alias. Here are my open issues on the docker repo:

https://github.com/AcademySoftwareFoundation/aswf-docker/issues/293
https://github.com/AcademySoftwareFoundation/aswf-docker/issues/295

I removed a bunch of AX builds (and moved others to weekly)

PR #2126, opened 2025-10-31
Fix unittest build issue in 

apradhana:
Based on report [54033589094](https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/18926180730/job/54033589094) from @danrbailey:

- the global scope has no "aligned_alloc"
- issue with typename inside of lambda with MSVC


apradhana:
I'm checking this by running the CI here: https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/18959393041/job/54143399653.

PR #2127, opened 2025-10-31
Fix NumPy issue on Windows CI

danrbailey:
This removes the numpy installation and the python component from the Windows CI builds. It also evaluates the debug unit tests on workflow dispatch for easier testing.

(This PR is still relevant for addressing the NanoVDB issue on Windows - https://github.com/AcademySoftwareFoundation/openvdb/pull/2126)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,6 +142,9 @@ jobs:
     - name: install
       shell: powershell
       run: .\ci\install_windows.ps1
+    - name: install_numpy
+      shell: powershell
+      run: .\ci\install_windows_numpy.ps1
     - name: build
       run: >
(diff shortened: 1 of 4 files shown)


danrbailey:
> > * Build / windows (push)
> 
> The issue is that NumPy installation fails on Windows, so I have removed the attempted installation. That means the python unit tests cannot run on any of the Windows CI builds without getting this error:
> 
> ```
> ModuleNotFoundError: No module named 'numpy'
> ```

I just confirmed this - https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/18980501600/job/54211525483

We need to fix numpy installation to be able to evaluate python unit tests on Windows (or otherwise disable portions of the tests that use numpy).

PR #2131, opened 2025-11-02
13.0.0 Release

danrbailey:
All changes notes, version bumps and ABI macros updated.


kmuseth (approved):
Looks good! My only nitpick is the fact that you didn't add highlights at the very top of the change files. It's useful for PR and answering the simple question: "why should I care about v13?" :)

PR #2183, opened 2026-03-20
PointData I/O

danrbailey:
This makes a few related to changes to PointData I/O.

1) The overloads for `readCompressedValues()` and `writeCompressedValues()` are moved to a new header file (`PointDataIO.h`). This is mainly to aid migration, it is now possible to include those overloads without needing to include the rest of `PointDataLeafNode.h`. (https://github.com/AcademySoftwareFoundation/openvdb/commit/45ba507304a6cfc6144e67453b36b54c71b3e7f7)

2) The PointDataGrid is the only grid type that takes advantage of grid-based multi-pass I/O which is where you navigate over the grid multiple times to read attribute-by-attribute. This change adapts this I/O code so that only the PointDataGrid can now use this mechanism and it will eventually be removed entirely as the new codec framework natively supports multi-pass I/O (https://github.com/AcademySoftwareFoundation/openvdb/commit/9b307aa4406e9167979539e334ac0f2223f0094d)

3) Remove the storing and loading of the AttributeArray out-of-core atomic member variable. Now that delayed loading has been removed, this can be eliminated. Also, scheduled to remove this entirely in ABI=14. This also eliminates the locking of the mutex during read as there is now no possibility of a race condition. (https://github.com/AcademySoftwareFoundation/openvdb/commit/aeedf18a0d01b352456fcb006d5f4f1d14363ded)

@Idclip


danrbailey:
Closing out this PR, all these changes have now been include in https://github.com/AcademySoftwareFoundation/openvdb/pull/2197

PR #2186, opened 2026-03-27
[WIP: Do not review] NanoVDB: VoxelBlockManager CPU port of decodeInverseMaps

sifakis:
Draft for CI and diff review. WIP.


danrbailey:
I know that this is marked as not for review yet, but I just wanted to comment and say that specifically there is one aspect of this PR which is something I've wanted in OpenVDB for a long time. I brought this up in our TSC meetings again recently. A lot of the time when we use the ValueAccessor, it is not true random access, but stencil access patterns and the fact that the ValueAccessor is only able to store one leaf node means you end up thrashing the cache in the ValueAccessor when accessing different neighboring leaf nodes along leaf boundaries.

I have done a little experimentation in this area and the lazy leaf neighbor pattern that you use here is what I was envisioning, though I was contemplating whether a templated stencil argument (ie 6-neighbor vs 27-neighbor) may provide improved performance depending on the use case. I also hadn't settled on a name - NeighborAccessor, StencilAccessor, etc.

I would be very interested in comparing performance of a similar structure for OpenVDB that doesn't include the SIMD logic to compare with performance of the ValueAccessor (that also does not use SIMD). If it outperforms the ValueAccessor, I think we should look to switch some of our neighbor use cases across. 

The other big issue with the ValueAccessor IMO is the tight coupling with the Grid and Tree (`getValueAndCache()` etc). This hurts impacts compile times and a new acceleration structure that extends the ValueAccessor in a new direction might be a good opportunity to revisit this and the attach/release accessor mechanism. I know NanoVDB took a slightly different approach on these trade offs.

(@kmuseth - a good topic for a future TSC meeting I think)

PR #2187, opened 2026-03-27
CNanoVDB sync with Houdini

jmlait:
Improvements:
  - CNanoVDB supports integer (I) grids
  - CNanoVDB has setValue methods to write to voxels
  - CNanoVDB has getLeaf methods to acquire raw leaf pointers

Bug Fixes:
  - static used for local kernel methods in CNanoVDB
  - Remove auto-computed reserved sizes in favour of hard coded sizes in CNanoVDB.  While less maintainble, C99 forbids zero sized arrays so some compilers crashed when encountering this.
  - Initialize key value when constructing an accessor, some compilers entered bad states with this undefined.

### DIFF
--- a/nanovdb/nanovdb/CNanoVDB.h
+++ b/nanovdb/nanovdb/CNanoVDB.h
@@ -5,14 +5,15 @@
 // Simple C-wrapper for the nanovdb structure
 // Meant for systems where you lack a C++ compiler.
 //
-
-/// @warning This header file will soon be deprecated in favor of PNanoVDB.h
-
 #ifndef __CNANOVDB__
 #define __CNANOVDB__
 
-#define CNANOVDB_DATA_ALIGNMENT 32
-#define CNANOVDB_ALIGNMENT_PADDING(x, n) (-(x) & ((n)-1))
+
+// We compute alignment, but if it evaluates to 0
+// it will be invalid C99, allowing pedantic compilers to
+// crash and claim they remain standards conforming.
+//#define CNANOVDB_DATA_ALIGNMENT 32
+//#define CNANOVDB_ALIGNMENT_PADDING(x, n) (-(x) & ((n)-1))
 
 #define USE_SINGLE_ROOT_KEY
 
@@ -53,6 +54,21 @@ enum cnanovdb_GridType
     cnanovdb_GridType_End     = 10
 };
 
+// #define versions of the grid type to allow pre-processor
+// tests.
+#define CNANOVDB_GRIDTYPE_UNKNOWN 0
+#define CNANOVDB_GRIDTYPE_FLOAT   1
+#define CNANOVDB_GRIDTYPE_DOUBLE  2
+#define CNANOVDB_GRIDTYPE_INT16   3
+#define CNANOVDB_GRIDTYPE_INT32   4
+#define CNANOVDB_GRIDTYPE_INT64   5
+#define CNANOVDB_GRIDTYPE_VEC3F   6
+#define CNANOVDB_GRIDTYPE_VEC3D   7
+#define CNANOVDB_GRIDTYPE_MASK    8
+#define CNANOVDB_GRIDTYPE_FP16    9
+#define CNANOVDB_GRIDTYPE_END     10
+
+
 #define ROOT_LEVEL 3
 
 #define DEFINEMASK_int(LOG2DIM, SIZE) \
@@ -88,7 +104,7 @@ typedef struct
     double mInvMatD[9]; // r,c = 3*r + c
     double mVecD[3];
     double mTaperD;
-} cnanovdb_map;
+} cnanovdb_map;         // Size 264
 
 typedef struct
 {
@@ -209,7 +225,8 @@ typedef struct
     uint32_t    mDataClass;    // 4 bytes
     uint32_t    mDataType;     // 4 bytes
     char        mName[256];
-    uint8_t     _reserved[CNANOVDB_ALIGNMENT_PADDING(sizeof(int64_t)+sizeof(uint64_t)+2*sizeof(uint32_t)+2*sizeof(uint32_t)+256*sizeof(char), CNANOVDB_DATA_ALIGNMENT)];
+    // 2*8+4*4+256 : aligned, so don't include.
+    // uint8_t     _reserved[CNANOVDB_ALIGNMENT_PADDING(sizeof(int64_t)+sizeof(uint64_t)+2*sizeof(uint32_t)+2*sizeof(uint32_t)+256*sizeof(char), CNANOVDB_DATA_ALIGNMENT)];
 } cnanovdb_gridblindmetadata;
 
 typedef struct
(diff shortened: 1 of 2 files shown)


nanovdb/nanovdb/CNanoVDB.h
apradhana: Do we need `static inline void` here?

nanovdb/nanovdb/CNanoVDB.h
```diff
@@ -340,6 +373,23 @@ cnanovdb_node##LEVEL##SUFFIX##_getValueAndCache(const CNANOVDB_GLOBAL cnanovdb_n
 } \
 \
 static bool \
+cnanovdb_node##LEVEL##SUFFIX##_setValue(CNANOVDB_GLOBAL cnanovdb_node##LEVEL##SUFFIX *RESTRICT node, const cnanovdb_coord *RESTRICT ijk, VALUETYPE value) \
+{ \
+    uint32_t n = cnanovdb_node##LEVEL##SUFFIX##_CoordToOffset(ijk); \
+    node->mVoxels[n] = value;  \
```
apradhana, line 379: Please consider also updating the mask.

pendingchanges/cnanonsize.txt
```diff
@@ -0,0 +1,14 @@
+NanoVDB:
+  Improvements:
+  - CNanoVDB supports integer (I) grids
+  - CNanoVDB has setValue methods to write to voxels
+  - CNanoVDB has getLeaf methods to acquire raw leaf pointers
+
+  Bug Fixes:
+  - static used for local kernel methods in CNanoVDB
+  - Remove auto-computed reserved sizes in favour of hard coded sizes
+    in CNanoVDB.  While less maintainble, C99 forbids zero sized
```
apradhana: Minor typo: maintainble -> maintainable

nanovdb/nanovdb/CNanoVDB.h
apradhana: Just pointing out that the corresponding member in `NanoVDB.h` is `mValueSize` instead of `mFlags` (cf. https://github.com/AcademySoftwareFoundation/openvdb/blob/master/nanovdb/nanovdb/NanoVDB.h#L1575).

nanovdb/nanovdb/CNanoVDB.h
```diff
@@ -229,7 +246,10 @@ typedef struct
     uint32_t         mGridType; // 4B.
     uint64_t         mBlindMetadataOffset; // 8B. offset of GridBlindMetaData structures.
     int32_t          mBlindMetadataCount; // 4B. count of GridBlindMetaData structures.
```
apradhana: Nitpick: Just pointing out that `mBlindMetadataCount` in `NanoVDB.h` is `uint32_t` (cf. https://github.com/AcademySoftwareFoundation/openvdb/blob/master/nanovdb/nanovdb/NanoVDB.h#L1940)

nanovdb/nanovdb/CNanoVDB.h
```diff
@@ -229,7 +246,10 @@ typedef struct
     uint32_t         mGridType; // 4B.
     uint64_t         mBlindMetadataOffset; // 8B. offset of GridBlindMetaData structures.
```
apradhana: Nitpick: Just pointing out that `mBlindMetadataOffset` in `NanoVDB.h` is `int64_t` (cf. https://github.com/AcademySoftwareFoundation/openvdb/blob/master/nanovdb/nanovdb/NanoVDB.h#L1939).

nanovdb/nanovdb/CNanoVDB.h
jmlait: I believe static and inline have opposite meanings in this context as they refer to linking options?

nanovdb/nanovdb/CNanoVDB.h
jmlait: Hmm.... This is intentionally NOT updating the mask.  The intent is to keep updates as speed-of-light as possible.  (Which is also why there is the getLeaf() exposed).  In practice this is being used with leaf-activated grids in Pyro so the mask is all active and we don't want to spend any time activating.

nanovdb/nanovdb/CNanoVDB.h
jmlait: Not really a fan of uint for this as you'll need a -1 escape code eventually anyways?  But I can synchronize.

nanovdb/nanovdb/CNanoVDB.h
apradhana: Thanks for clarifying. Makes sense for pyro purposes.

nanovdb/nanovdb/CNanoVDB.h
apradhana: Thank you!!

nanovdb/nanovdb/CNanoVDB.h
apradhana: I was basically looking at the signature of `cnanovdb_readaccessor_computeDirty` that adds the keyword `static` before `inline`. To the best of my understanding having `inline` by itself can fail to link in strict C99 and adding `static` makes sure that each translation unit gets its own internal copy, so there is no missing symbol of multiple definition problem. 

I was able to produce a sample of where linking can fail without `static` with O0 flag. I have `helper.c`:
```
#include <stdint.h>
#include <stdbool.h>
#include "../../CNanoVDB.h"

void helper_function(const void *rootdata)
{
    cnanovdb_readaccessor acc;
    cnanovdb_readaccessor_init(&acc, rootdata);
    (void)acc;
}
```
and `main.c`:
```
#include <stdbool.h>
#include "../../CNanoVDB.h"

extern void helper_function(const void *rootdata);

int main(void)
{
    cnanovdb_readaccessor acc;
    cnanovdb_readaccessor_init(&acc, (const void *)0);
    printf("acc key: %d %d %d\n", acc.mKey.mVec[0], acc.mKey.mVec[1], acc.mKey.mVec[2]);

    helper_function((const void *)0);
    return 0;
}
```
Then, I compile it with:
```
gcc -std=c99 -O0 -c main.c -o main.o
gcc -std=c99 -O0 -c helper.c -o helper.o
gcc main.o helper.o -o test
```
The linker error I get is:
```
/usr/bin/ld: main.o: in function `main':
main.c:(.text+0x3b4e): undefined reference to `cnanovdb_readaccessor_init'
/usr/bin/ld: helper.o: in function `helper_function':
helper.c:(.text+0x3b54): undefined reference to `cnanovdb_readaccessor_init'
collect2: error: ld returned 1 exit status`
```
However, it's not a problem if either I compile it with `O1` flag or adding a `static` keyword just like `cnanovdb_readaccessor_computeDirty`.

Since it's not a problem with O1 flag, I'm approving the PR as is.

apradhana (approved):

apradhana:
Thank you for addressing all the comments. Approved.

nanovdb/nanovdb/CNanoVDB.h
jmlait: Oh, I see, I added the static to the files below!

My understanding was inline would suppress symbol creation so stop it from linking as there won't be a symbol; but it shouldn't matter as all callers in the same translation unit should have inlined it...

I'm not sure why I made it static inline elsewhere, to be honest.  I've checked the revision locally and sadly the change didn't comment why static was added to the inline methods.  But they should be consistent in any rate, and your example shows why "static inline" is a meaningful setup.  So I'll add static here.

apradhana: Thank you for adding this.

apradhana:
LGTM!

apradhana (approved):

PR #2197, opened 2026-04-22
Add PointIndexCodec, PointDataCodec and other Point changes

danrbailey:
This adds the two codecs for handling point data - PointIndexCodec and PointDataCodec.

I have also addressed the feedback that was offered previously to rename MultiPass tag class to PointDataGridMultiPass and this now locks down the old multi-pass mechanism specific to only be usable by VDB Points.

Some other related changes include removing the out-of-core atomic in AttributeArray for `ABI=14` and some minor re-organisation to be able to access the `readCompressedValues()`/`writeCompressedValues()` specializations from the new codec.


openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+/// Look up an existing paged stream or create a new one for the given attribute index
+template <typename PagedStreamPtrT>
+inline typename PagedStreamPtrT::element_type* getOrCreatePagedStream(
+    std::map<Index, PagedStreamPtrT>& pagedStreams, Index attributeIndex)
+{
+    auto it = pagedStreams.find(attributeIndex);
+    if (it != pagedStreams.end()) return it->second.get();
+    using PagedStreamT = typename PagedStreamPtrT::element_type;
+    pagedStreams[attributeIndex] = std::make_shared<PagedStreamT>();
+    return pagedStreams[attributeIndex].get();
```
Idclip: Assuming PagedStreamPtrT is a shared_ptr<>, why not just `return  pagedStreams[attributeIndex].get()`? Avoid the double lookup

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+            array->readPagedBuffers(*pagedStream);
+        }
+    }
+}
+
+template <typename LeafT>
+inline void readPointDataVoxelData(const std::vector<LeafT*>& leaves,
+    std::istream& is, bool saveFloatAsHalf,
+    const typename LeafT::ValueType& background,
+    const std::map<Coord, uint16_t>& voxelBufferSizes)
```
Idclip: maybe make the unordered_map?

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+{
+    (void) voxelBufferSizes;
```
Idclip: remove

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+template <typename LeafT>
+inline void readPointDataAttributeData(const std::vector<LeafT*>& leaves,
+    std::istream& is, Index attributeIndex,
+    std::map<Index, compression::PagedInputStream::Ptr>& pagedStreams)
+{
+    for (auto* leaf : leaves) {
+        points::AttributeArray* array = attributeIndex < leaf->attributeSet().size() ?
+            &leaf->attributeArray(attributeIndex) : nullptr;
+        if (array) {
+            auto* pagedStream = getOrCreatePagedStream(pagedStreams, attributeIndex);
```
Idclip, line 133: You call this a lot in all these methods but it looks cache-able outside of the leaf loops?

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+template <typename LeafT>
+inline Index countPointDataPasses(const std::vector<const LeafT*>& leaves)
+{
+    Index maxRequiredPasses = 0;
+    for (const auto* leaf : leaves) {
+        const Index requiredPasses = leaf->buffers();
+        if (requiredPasses > maxRequiredPasses) {
+            maxRequiredPasses = requiredPasses;
+        }
+    }
```
Idclip, line 187: parallel reduce?

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+/// Per-grid-type codec-specific options for PointDataCodec
+/// Contains point attribute filtering options
+struct OPENVDB_API PointDataCodecTypeData : public io::ReadTypedOptions
+{
+    // Point Attribute Options - which attributes to read
+    std::vector<std::string> pointAttributeNames;
+}; // struct PointDataCodecTypeData
+
+template <typename GridT>
+struct PointDataCodec : public TopologyCodec<GridT>
```
Idclip: If it can't be final, docs as to expected derived usage

openvdb/openvdb/codecs/PointDataCodec.h
```diff
@@ -0,0 +1,449 @@
...
+{
+    using Ptr = std::unique_ptr<PointDataCodec<GridT>>;
+
+    ~PointDataCodec() noexcept = default;
+
+    static inline std::string name() { return GridT::gridType(); }
+
+    void readBuffers(std::istream& is, io::CodecData& data, const io::ReadOptions& options, io::ReadDiagnostics&) final
+    {
+        GridT& grid = static_cast<GridT&>(*data.grid);
```
Idclip, line 317: OPENVDB_ASSERT(dynamic_cast<GridT*>(data.grid.get()))l

openvdb/openvdb/codecs/PointIndexCodec.h
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    std::ostream& os;
+    const bool saveFloatAsHalf;
+}; // struct WritePointIndexBuffersOp
+
+} // namespace internal
+
+template <typename GridT>
+struct PointIndexCodec : public TopologyCodec<GridT>
```
Idclip: final else docs

openvdb/openvdb/codecs/PointIndexCodec.h
```diff
@@ -0,0 +1,157 @@
...
+{
+    using Ptr = std::unique_ptr<PointIndexCodec<GridT>>;
+
+    ~PointIndexCodec() noexcept = default;
+
+    static inline std::string name() { return GridT::gridType(); }
+
+    void readBuffers(std::istream& is, io::CodecData& data, const io::ReadOptions& options, io::ReadDiagnostics& diagnostics) final
+    {
+        GridT& grid = static_cast<GridT&>(*data.grid);
```
Idclip, line 121: dynamic cast assert

openvdb/openvdb/codecs/TopologyCodec.h
```diff
@@ -191,13 +193,27 @@ struct ReadTopologyOp
...
+            if (oldVersion) {
+                Index n = 0;
+                if constexpr (std::is_same_v<ValueT, StorageValueT>) {
+                    for (auto iter = node.beginValueAll(); iter; ++iter) {
+                        node.setValueOnlyUnsafe(iter.pos(), values[n++]);
+                    }
+                } else {
+                    for (auto iter = node.beginValueAll(); iter; ++iter) {
+                        node.setValueOnlyUnsafe(iter.pos(), static_cast<ValueT>(values[n++]));
+                    }
```
Idclip: similar comment about just using the cast branch

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: Thanks, fixed

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: Thanks, fixed

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: Yes, it is. I'm not going to try and change this, because this existing pattern matches how things work in the old multi-pass tree I/O. I will focus more heavily on performance improvements for the new codecs.

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: Again, going to keep this as-is to match the existing tree I/O implementation. Optimization effort can focus on new codecs.

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: Fixed, thanks

danrbailey:
@Idclip - I believe I've addressed all your feedback now. The primary goal here is to migrate the existing code off the tree/points API and into these codecs. As mentioned earlier, I'll try and further optimize performance in the new codecs.

openvdb/openvdb/codecs/PointDataCodec.h
Idclip: I don't quite follow if you mean that this code won't be used much going forward or that you'll fix it in a subsequent MR? Happy to defer to whatever you think is best, but I think think change would be good to action at some point

Idclip (approved):

openvdb/openvdb/codecs/PointDataCodec.h
danrbailey: I'm not sure if you're primarily suggesting this to improve the readability or maintenance of the code or because you see opportunities to improve performance.

When we introduce the new codecs, these older codecs will be used less and less as I expect that all new read and writes will go through the newer, more efficient codecs. I don't think we'll be able to completely eliminate these codecs for a long time (if ever), but I would prefer to prioritize spending time on improving the new codecs. And I will happily accept all performance-related feedback when I share the new codecs. :)

PR #2198, opened 2026-04-24
nanovdb/python: fix writeGrids dropping all but the last grid

swahtz:
The Python binding for `nanovdb.io.writeGrids` iterated the handle list and called `io::writeGrid(fileName, ...)` per element. That file-level overload opens the path with `std::ios::trunc`, so every iteration wiped the previously written grid and the resulting .nvdb only ever contained the last handle in the list (same story for deviceWriteGrids on the CUDA side).

Mirror `io::writeGrids(fileName, handles, codec, verbose)` from nanovdb/io/IO.h directly in the wrapper: open the output stream once with trunc, then call the ostream overload of writeGrid for each handle cast from the nb::list. We can't simply forward to `io::writeGrids` because `GridHandle<BufferT>` is move-only and the nb::list holds references we must not invalidate.

Tests (nanovdb/python/test/TestNanoVDB.py):
- TestGridHandleExchange.test_list_to_vector now reads the written file back via readGridMetaData and readGrids and asserts the full handle count is preserved.
- Added TestDeviceReadWriteGrids.test_device_write_grids_multi as a CUDA-guarded regression for deviceWriteGrids with multiple handles.

### DIFF
--- a/nanovdb/nanovdb/python/PyIO.cc
+++ b/nanovdb/nanovdb/python/PyIO.cc
@@ -69,8 +69,15 @@ template<typename BufferT> nb::list readGrids(const std::string& fileName, int v
 
 template<typename BufferT> void writeGrids(const std::string& fileName, nb::list handles, io::Codec codec, int verbose)
 {
+    std::ofstream os(fileName, std::ios::out | std::ios::binary | std::ios::trunc);
+    if (!os.is_open()) {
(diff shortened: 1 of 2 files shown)


kmuseth (approved):
looks good to me .. I see the issue and agree with the fix

apradhana (approved):
Good catch!

PR #2206, opened 2026-05-19
Add Python typing stubs

scottdmilner:
This PR uses nanobind's `nanobind_add_stub()` to add typing stubs to the Python module. The recommended nanobind version is raised to 2.9.2 to add support for the `RECURSIVE` flag.

Can be toggled by cmake flags `OPENVDB_BUILD_PYTHON_STUBS` / `NANOVDB_BUILD_PYTHON_STUBS`


nanovdb/nanovdb/python/CMakeLists.txt
```diff
@@ -37,6 +39,19 @@ else()
...
+  nanobind_add_stub(
+    nanovdb_stub
+    INSTALL_TIME
+    RECURSIVE
+    MODULE nanovdb
+    OUTPUT_PATH nanovdb
+    PYTHON_PATH "."
+    MARKER_FILE py.typed
+  )
+endif()
```
Idclip, line 58: What happens if someone pulls and is building with < 2.9.2, does the RECURSIVE flag silently do nothing? Would it be worth a nested branch here for `if(nanobind_VERSION VERSION_LESS 2.9.2)` which outputs a `message(WARNING)` that no python type info will be generated (as it requires 2.9.2)?

Same in the OpenVDB CMakeLists.

Idclip (approved):
Thanks for this @scottdmilner, I haven't tested but looks good. Minor comment and the failing test can be ignored

nanovdb/nanovdb/python/CMakeLists.txt
scottdmilner: Good catch, the current behavior is to fail build configuration. I added some checks and warnings.

nanovdb/nanovdb/python/CMakeLists.txt
```diff
@@ -37,6 +39,24 @@ else()
   install(TARGETS nanovdb_python DESTINATION ${VDB_PYTHON_INSTALL_DIRECTORY})
 endif()
 
+# typing stubs
+if(NANOVDB_BUILD_PYTHON_STUBS)
+  if(nanobind_VERSION VERSION_LESS ${FUTURE_MINIMUM_NANOBIND_VERSION})
```
Idclip: I'm fine with bumping `FUTURE_MINIMUM_NANOBIND_VERSION`, but I think here this should be hard coded to `if(nanobind_VERSION VERSION_LESS 2.9.2)` - that's the version which added support for the `RECURSIVE` flag and is technically unrelated to `FUTURE_MINIMUM_NANOBIND_VERSION`. Hard coding it will stop the branch silently moving if  `FUTURE_MINIMUM_NANOBIND_VERSION` is updated again.

openvdb/openvdb/python/CMakeLists.txt
```diff
@@ -61,7 +63,7 @@ endif()
 ##########################################################################
 
 set(NANOBIND_MODULE_ARGS "")
-if(${nanobind_VERSION} GREATER_EQUAL 2.5.0)
+if(${nanobind_VERSION} GREATER_EQUAL ${FUTURE_MINIMUM_NANOBIND_VERSION})
```
Idclip: In similar fashion, pretty sure `NB_SUPPRESS_WARNINGS` was introduced in `2.5.0` and is not directly related to the value of `FUTURE_MINIMUM_NANOBIND_VERSION`, so I'd revert this

Idclip (approved):

PR #2207, opened 2026-05-20
NanoVDB Math: refactor onto `MatBase`/`VecBase`, fix correctness bugs, add tests

swahtz:
Consolidates `nanovdb/math/Math.h` by lifting Mat / Vec arithmetic onto two shared base classes (`MatBase<T, R, C>` and a new `VecBase<T, N>`). The five Mat and three Vec derived classes become thin delegating wrappers. The pass also fixes several long-standing correctness bugs and fills inconsistent / partial APIs.

**Files changed**: `Math.h` (the refactor), `NanoVDB.h` (`Map` API now constexpr — depends on the `matMult` rework), `tools/cuda/PointsToGrid.cuh` (three call-site casts from the `explicit` ctor tightening), and three test files (additive: 11 new `TEST_F` blocks plus mechanical cast updates), and one python-binding kernel.

## Bug fixes

- `math::Min` / `Max` for `int32_t` / `uint32_t` no longer route through `fminf` / `fmaxf` — old form lost precision above 2^24 and could be UB near `INT_MAX`. Heavily used by `HDDA.h`.
- Removed dimensionally-invalid `operator*(Mat3, Mat2x3) -> Mat2x3`. No callers in the tree.
- `Mat2::inverse()` now returns an explicit zero on singular input (was uninitialised) and uses `T(1)/det` for `T=double`.
- `Vec*::operator/(T)` and `Mat*::operator/(T)` now per-element-divide. The old `(T(1)/s) * (*this)` did integer division for integer `T`, so `Vec3i(6,8,10) / 2` returned `(0,0,0)`.
- `Vec2`'s `operator+`/`-`/`+=`/`-=` against `Coord` now take `Coord2`. Old signature silently dropped `z`.
- `math::Round(Vec3<float>)` and `Round(Vec3<double>)` both use `floor(x + 0.5)` now. The `float` overload previously used `rintf` (round-half-to-even), so the two disagreed at half-integers.
- Corrected `floor()` / `ceil()` doc swap on `Vec2` / `Vec3`. Implementations were always right.

## Refactor

- **`MatBase<T, R, C>`**: shared element-wise / scalar / equality / transpose / mat*mat / mat*vec implementations, all `__hostdev__ constexpr`. `rows()` / `cols()` / `size()` promoted public.
- **`VecBase<T, N>`** (new): shared element-wise / scalar / equality / reductions / mutating min/max / integer rounding. `floorAs<Result>` / `ceilAs<Result>` / `roundAs<Result>` unify `Round` semantics across `float` / `double` / `long double`.
- All five Mat and three Vec classes become thin wrappers: ctors + per-type extras (`Mat2::inverse`, `Vec3::cross`/`outer`) + one-line operators delegating to the base.
- `Mat3` / `Mat3x2` / `Mat4` scalar ctors switched from `template<typename Source>` to `T`, matching `Mat2` / `Mat2x3`. Mixed-literal construction (`Mat3<double>(1, 2.0, 3, ...)`) now works uniformly. Array ctors keep `template<typename Source>` for cross-type interop.

## API completeness

- Every Mat type has the same operator surface: unary `-`, binary `+` / `-`, `+=` / `-=`, `*(T)` / `/(T)`, `*=(T)` / `/=(T)`, `==` / `!=`. Previously scattered.
- New mat*mat: `Mat3 * Mat3x2` standalone, `Mat4 * Mat4` member.
- New mat*vec members: `Mat2 * Vec2`, `Mat2x3 * Vec3 -> Vec2`, `Mat3x2 * Vec2 -> Vec3`. `Mat4 * Vec4` promoted to member.
- New `scalar * MatN` standalones for `Mat2x3` / `Mat3x2` / `Mat3` / `Mat4`.
- `Vec4` gains parity with `Vec3`: `min`, `max`, `asPointer`, `floor`, `ceil`, `round` (returns `Vec4<int32_t>` — no `Coord4`).

## `constexpr` sweep

Every function in `Math.h` and the `Map` API in `NanoVDB.h` that C++17 allows is now `constexpr`. Carve-outs: `length` / `normalize` (`std::sqrt`, C++26), the `float` / `double` overloads of `Min` / `Max` / `Floor` / `Ceil` / `Round` / `Abs` (C++23), and CUDA-device-only `*Atomic` helpers.

Rewriting `matMult` / `matMultT` from `fma(a, b, c)` to `a * b + c` was required (fma isn't constexpr until C++23, P1383R2). **GPU codegen unchanged** — nvcc `-fmad=true` contracts back to hardware FMA. **Host codegen unchanged** under default `-ffp-contract=on`. Only compile-time `constexpr` evaluation observes the double-rounded result. Worst-case ≤1 ulp drift, well below NanoVDB's geometric precision.

## Other quality passes

- **Bounds checks** on every `operator[]` in `Math.h` via `NANOVDB_ASSERT` (twelve sites). No-op under `NDEBUG`; const-index accesses fold at compile time even in debug.
- **Doxygen sweep**: every previously-undocumented public method gains an `@brief`; several `// @brief` typos (missing leading slash — silently invisible to Doxygen) fixed.
- **Hidden friends**: `scalar*Vec` / `scalar/Vec` / `scalar*Mat` converted from namespace-scope free templates to in-class `friend` definitions — ADL-found only, no longer pollute the enclosing namespace.

## Tests

Eleven additive `TEST_F` blocks in `TestNanoVDB.cc`:

| Test | Covers |
|---|---|
| `IntegerMinMax` | `2^24+1` precision; `INT_MAX` / `UINT_MAX` overflow safety. |
| `Round` | Half-integer agreement between `float` and `double` (especially `-1.5`). |
| `Vec2` / `Vec3Ops` / `Vec4Ops` | Full operator coverage; new `Vec4` methods; integer-T division. |
| `Mat2` / `Mat3` / `Mat4` / `Mat2x3_Mat3x2` | Constructors, every operator, transpose, mat*mat, mat*vec; `Mat2::inverse` non-singular roundtrip + singular-returns-zero; mixed-literal construction. |
| `MatMul` | Every dimensionally-valid mat*mat plus every member mat*vec across the five matrix types. |
| `MatVecIntrospection` | `static_assert`s for `rows`/`cols`/`size`/`SIZE`/`ValueType`; `data()`/`asPointer()` ordering. |

## Risk and compatibility

- **Behaviour preserved** for every previously-correct code path.
- **Behaviour changes** are exactly the bug fixes above — each moves an incorrect / undefined output to a defined, correct one.
- **`matMult` rounding**: ≤1 ulp drift only in pure abstract-machine evaluation. GPU and host codegen identical to before.
- **API additions are pure**. The only removed overload is the broken `Mat3 * Mat2x3` (no callers).
- **One signature tightening**: `Vec2`'s `Coord` overloads now require `Coord2`. Code passing a 3D `Coord` to `Vec2` will fail to compile — but it was silently dropping `z`.
- **Cross-template ctor tightened to `explicit`**: `Vec*(const Vec*T<T2>&)` now matches the same-class ctor's convention. Audit found 18 affected sites across the nanovdb + openvdb trees, all updated here (3 in `PointsToGrid.cuh`, 3 in `TestNanoVDB.cu`, 12 in `TestOpenVDB.cc`, 1 in `PySampleFromVoxels.cu`). All mechanical and value-preserving. **Downstream consumers that relied on the implicit conversion will need the same one-line cast** — typically `nanovdb::Vec3f(yourVec)` or an explicit template argument to a returning helper.

### DIFF
--- a/nanovdb/nanovdb/math/Math.h
+++ b/nanovdb/nanovdb/math/Math.h
@@ -1039,8 +1039,13 @@ public:
 // ----------------------------> Vec2 <--------------------------------------
 
 /// @brief A simple vector class with two components, similar to openvdb::math::Vec2
+///
+/// Aligned to 2*alignof(T) so the whole class fits in one SIMD-friendly
+/// chunk (e.g. 8 bytes for Vec2<float>, 16 bytes for Vec2<double>),
+/// without any tail padding because the byte size is already a power-of-2
+/// multiple of alignof(T).
 template<typename T>
-class Vec2 : public VecBase<T, 2>
+class alignas(alignof(T) * 2) Vec2 : public VecBase<T, 2>
 {
     using Base = VecBase<T, 2>;
 
@@ -1321,8 +1326,11 @@ template<typename T> class Vec4;
 
 
 
+// Aligned to 4*alignof(T) — Mat2 stores 4 elements (2x2), which is already
+// a power-of-2 multiple of alignof(T), so this is free in size and gives
+// SIMD-friendly placement (16 bytes for Mat2<float>, 32 bytes for double).
 template <typename T>
-class Mat2 : public MatBase<T, 2, 2> {
+class alignas(alignof(T) * 4) Mat2 : public MatBase<T, 2, 2> {
     using Base = MatBase<T, 2, 2>;
 public:
     Mat2() = default;
@@ -1376,6 +1384,10 @@ public:
     }
 };
 
+// Mat2x3 is intentionally NOT alignas-elevated: its byte size
+// (6*sizeof(T)) is not a power-of-2 multiple of alignof(T), so any
+// alignas(N > alignof(T)) would force tail padding and break
+// packed-array layout plus on-disk format compatibility.
 template <typename T>
 class Mat2x3 : public MatBase<T, 2, 3> {
     using Base = MatBase<T, 2, 3>;
@@ -1420,6 +1432,10 @@ public:
     __hostdev__ constexpr Mat3x2<T> transpose() const { return this->template transposeAs<Mat3x2<T>>(); }
 };
 
+// Mat3x2 is intentionally NOT alignas-elevated: its byte size
+// (6*sizeof(T)) is not a power-of-2 multiple of alignof(T), so any
+// alignas(N > alignof(T)) would force tail padding and break
+// packed-array layout plus on-disk format compatibility.
 template <typename T>
 class Mat3x2: public MatBase<T, 3, 2> {
     using Base = MatBase<T, 3, 2>;
@@ -1468,6 +1484,10 @@ public:
 
 };
 
+// Mat3 is intentionally NOT alignas-elevated: its byte size
+// (9*sizeof(T)) is not a power-of-2 multiple of alignof(T), so any
+// alignas(N > alignof(T)) would force tail padding and break
+// packed-array layout plus on-disk format compatibility.
 template <typename T>
 class Mat3 : public MatBase<T, 3, 3> {
     using Base = MatBase<T, 3, 3>;
@@ -1519,8 +1539,16 @@ public:
     __hostdev__ constexpr Mat3 transpose() const { return this->template transposeAs<Mat3>(); }
 };
 
+// Aligned to 16*alignof(T) — Mat4 stores 16 elements (4x4), which is
+// already a power-of-2 multiple of alignof(T), so the alignment is free
+// in size. Whole-matrix alignment (64 bytes for Mat4<float>, 128 bytes
+// for Mat4<double>) is heavy compared to row-alignment (4*alignof(T))
+// but matches the per-class "align to full size" rule used for Vec2 /
+// Vec4 / Mat2 above and lets a Mat4<float> load with a single AVX-512
+// instruction. Drop to alignas(alignof(T) * 4) here if the over-aligned
+// constraint causes friction with downstream allocators.
 template <typename T>
-class Mat4 : public MatBase<T, 4, 4> {
+class alignas(alignof(T) * 16) Mat4 : public MatBase<T, 4, 4> {
     using Base = MatBase<T, 4, 4>;
 public:
     Mat4() = default;
@@ -1621,6 +1649,13 @@ __hostdev__ constexpr Mat3x2<T> operator*(const Mat3<T>& lhs, const Mat3x2<T>& r
 // ----------------------------> Vec3 <--------------------------------------
 
 /// @brief A simple vector class with three components, similar to openvdb::math::Vec3
+///
+/// Vec3 is intentionally NOT alignas-elevated: its byte size
+/// (3*sizeof(T)) is not a power-of-2 multiple of alignof(T), so any
+/// alignas(N > alignof(T)) would force tail padding and break
+/// packed-array layout plus on-disk format compatibility. An opt-in
+/// over-aligned Vec3 wrapper (e.g. for CUDA-coalesced loads) is best
+/// added as a separate type rather than changing Vec3 itself.
 template<typename T>
 class Vec3 : public VecBase<T, 3>
 {
@@ -1754,8 +1789,13 @@ __hostdev__ inline constexpr Vec3<double> Coord::asVec3d() const
 // ----------------------------> Vec4 <--------------------------------------
 
 /// @brief A simple vector class with four components, similar to openvdb::math::Vec4
+///
+/// Aligned to 4*alignof(T) so the whole class fits in one SIMD register
+/// (16 bytes for Vec4<float>, 32 bytes for Vec4<double>), without any
+/// tail padding because the byte size is already a power-of-2 multiple
+/// of alignof(T).
 template<typename T>
-class Vec4 : public VecBase<T, 4>
+class alignas(alignof(T) * 4) Vec4 : public VecBase<T, 4>
 {
     using Base = VecBase<T, 4>;
 
--- a/nanovdb/nanovdb/math/Math.h
+++ b/nanovdb/nanovdb/math/Math.h
@@ -136,11 +136,11 @@ __hostdev__ inline Type Min(Type a, Type b)
 }
 __hostdev__ inline int32_t Min(int32_t a, int32_t b)
 {
-    return int32_t(fminf(float(a), float(b)));
+    return a < b ? a : b;
 }
 __hostdev__ inline uint32_t Min(uint32_t a, uint32_t b)
 {
-    return uint32_t(fminf(float(a), float(b)));
+    return a < b ? a : b;
 }
 __hostdev__ inline float Min(float a, float b)
 {
@@ -158,11 +158,11 @@ __hostdev__ inline Type Max(Type a, Type b)
 
 __hostdev__ inline int32_t Max(int32_t a, int32_t b)
 {
-    return int32_t(fmaxf(float(a), float(b)));
+    return a > b ? a : b;
 }
 __hostdev__ inline uint32_t Max(uint32_t a, uint32_t b)
 {
-    return uint32_t(fmaxf(float(a), float(b)));
+    return a > b ? a : b;
 }
 __hostdev__ inline float Max(float a, float b)
 {
@@ -252,18 +252,23 @@ __hostdev__ inline int Abs(int x)
 template<typename CoordT, typename RealT, template<typename> class Vec3T>
 __hostdev__ inline CoordT Round(const Vec3T<RealT>& xyz);
 
+/// @brief Round each component to its closest integer (round-half-toward-+inf)
+/// using @c floor(x+0.5). Same rule is applied to both single and double
+/// precision so float and double inputs yield the same integer coords.
 template<typename CoordT, template<typename> class Vec3T>
 __hostdev__ inline CoordT Round(const Vec3T<float>& xyz)
 {
-    return CoordT(int32_t(rintf(xyz[0])), int32_t(rintf(xyz[1])), int32_t(rintf(xyz[2])));
-    //return CoordT(int32_t(roundf(xyz[0])), int32_t(roundf(xyz[1])), int32_t(roundf(xyz[2])) );
-    //return CoordT(int32_t(floorf(xyz[0] + 0.5f)), int32_t(floorf(xyz[1] + 0.5f)), int32_t(floorf(xyz[2] + 0.5f)));
+    return CoordT(int32_t(floorf(xyz[0] + 0.5f)),
+                  int32_t(floorf(xyz[1] + 0.5f)),
+                  int32_t(floorf(xyz[2] + 0.5f)));
 }
 
 template<typename CoordT, template<typename> class Vec3T>
 __hostdev__ inline CoordT Round(const Vec3T<double>& xyz)
 {
-    return CoordT(int32_t(floor(xyz[0] + 0.5)), int32_t(floor(xyz[1] + 0.5)), int32_t(floor(xyz[2] + 0.5)));
+    return CoordT(int32_t(floor(xyz[0] + 0.5)),
+                  int32_t(floor(xyz[1] + 0.5)),
+                  int32_t(floor(xyz[2] + 0.5)));
 }
 
 template<typename CoordT, typename RealT, template<typename> class Vec3T>
@@ -802,154 +807,279 @@ public:
     __hostdev__ inline Coord2 round() const { return *this; }
 }; // Coord2 class
 
-// ----------------------------> Vec2 <--------------------------------------
+// ----------------------------> VecBase <-----------------------------------
 
-/// @brief A simple vector class with three components, similar to openvdb::math::Vec3
-template<typename T>
-class Vec2
+/// @brief Base class for fixed-size vectors. Provides shared element-wise
+/// arithmetic, scalar arithmetic, equality, length/dot reductions,
+/// component min/max, and integer-coord rounding. Derived classes provide
+/// constructors and any dimension-specific extras (e.g. cross/outer on Vec3).
+template<typename T, int N>
+class VecBase
 {
-    T mVec[2];
+protected:
+    T mVec[N];
 
 public:
-    static const int SIZE = 2;
-    static const int size = 2; // in openvdb::math::Tuple
     using ValueType = T;
+    static constexpr int SIZE = N;
+
+    VecBase() = default;
+
+    /// @brief Indexed element access (no bounds check; assumes 0 <= i < N)
+    __hostdev__ const T& operator[](int i) const { return mVec[i]; }
+    __hostdev__ T&       operator[](int i)       { return mVec[i]; }
+
+    /// @brief raw pointer to the underlying N-element storage
+    __hostdev__ T*       asPointer()       { return mVec; }
+    __hostdev__ const T* asPointer() const { return mVec; }
+
+    // ---- generic element-wise helpers (taking/returning Derived) ----
+
+    template<typename Derived>
+    __hostdev__ Derived plus(const Derived& rhs) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] + rhs[i];
+        return out;
+    }
+    template<typename Derived>
+    __hostdev__ Derived minus(const Derived& rhs) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] - rhs[i];
+        return out;
+    }
+    template<typename Derived>
+    __hostdev__ Derived mul(const Derived& rhs) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] * rhs[i];
+        return out;
+    }
+    template<typename Derived>
+    __hostdev__ Derived div(const Derived& rhs) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] / rhs[i];
+        return out;
+    }
+    template<typename Derived>
+    __hostdev__ Derived negate() const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = -mVec[i];
+        return out;
+    }
+    template<typename Derived>
+    __hostdev__ Derived scale(const T& s) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] * s;
+        return out;
+    }
+    /// @brief return (*this) / @a s element-wise as a @c Derived. Uses per-element
+    /// division (correct for integer @c T, unlike multiplying by 1/s).
+    template<typename Derived>
+    __hostdev__ Derived divideBy(const T& s) const {
+        Derived out;
+        for (int i = 0; i < N; ++i) out[i] = mVec[i] / s;
+        return out;
+    }
+
+    // ---- in-place compound assignment helpers ----
+
+    __hostdev__ VecBase& addAssign(const VecBase& rhs) {
+        for (int i = 0; i < N; ++i) mVec[i] += rhs.mVec[i];
+        return *this;
+    }
+    __hostdev__ VecBase& subAssign(const VecBase& rhs) {
+        for (int i = 0; i < N; ++i) mVec[i] -= rhs.mVec[i];
+        return *this;
+    }
+    __hostdev__ VecBase& mulAssign(const VecBase& rhs) {
+        for (int i = 0; i < N; ++i) mVec[i] *= rhs.mVec[i];
+        return *this;
+    }
+    __hostdev__ VecBase& divAssign(const VecBase& rhs) {
+        for (int i = 0; i < N; ++i) mVec[i] /= rhs.mVec[i];
+        return *this;
+    }
+    __hostdev__ VecBase& scaleAssign(const T& s) {
+        for (int i = 0; i < N; ++i) mVec[i] *= s;
+        return *this;
+    }
+    __hostdev__ VecBase& divideAssignScalar(const T& s) {
+        for (int i = 0; i < N; ++i) mVec[i] /= s;
+        return *this;
+    }
+
+    __hostdev__ bool equals(const VecBase& rhs) const {
+        for (int i = 0; i < N; ++i) if (mVec[i] != rhs.mVec[i]) return false;
+        return true;
+    }
+
+    // ---- reductions ----
+
+    /// @brief dot product. @a V must have @c operator[] valid for 0..N-1.
+    template<typename V>
+    __hostdev__ T dot(const V& v) const {
+        T s = T(0);
+        for (int i = 0; i < N; ++i) s += mVec[i] * v[i];
+        return s;
+    }
+    __hostdev__ T lengthSqr() const {
+        T s = T(0);
+        for (int i = 0; i < N; ++i) s += mVec[i] * mVec[i];
+        return s;
+    }
+    __hostdev__ T length() const { return Sqrt(this->lengthSqr()); }
+
+    /// @brief return the smallest of the N components
+    __hostdev__ T smallestComponent() const {
+        T m = mVec[0];
+        for (int i = 1; i < N; ++i) if (mVec[i] < m) m = mVec[i];
+        return m;
+    }
+    /// @brief return the largest of the N components
+    __hostdev__ T largestComponent() const {
+        T m = mVec[0];
+        for (int i = 1; i < N; ++i) if (mVec[i] > m) m = mVec[i];
+        return m;
+    }
+
+    // ---- component-wise (mutating) min/max ----
+
+    template<typename V>
+    __hostdev__ VecBase& mergeMin(const V& other) {
+        for (int i = 0; i < N; ++i) if (other[i] < mVec[i]) mVec[i] = other[i];
+        return *this;
+    }
+    template<typename V>
+    __hostdev__ VecBase& mergeMax(const V& other) {
+        for (int i = 0; i < N; ++i) if (other[i] > mVec[i]) mVec[i] = other[i];
+        return *this;
+    }
+
+    // ---- integer rounding (toward -inf / +inf / nearest) ----
+
+    /// @brief floor-rounded components into the @c Result type (whose @c [i]
+    /// must accept int32_t). For integer @c T the value is passed through.
+    template<typename Result>
+    __hostdev__ Result floorAs() const {
+        Result r;
+        if constexpr (util::is_floating_point<T>::value) {
+            for (int i = 0; i < N; ++i) r[i] = math::Floor(mVec[i]);
+        } else {
+            for (int i = 0; i < N; ++i) r[i] = static_cast<int32_t>(mVec[i]);
+        }
+        return r;
+    }
+    template<typename Result>
+    __hostdev__ Result ceilAs() const {
+        Result r;
+        if constexpr (util::is_floating_point<T>::value) {
+            for (int i = 0; i < N; ++i) r[i] = math::Ceil(mVec[i]);
+        } else {
+            for (int i = 0; i < N; ++i) r[i] = static_cast<int32_t>(mVec[i]);
+        }
+        return r;
+    }
+    /// @brief nearest-integer rounding using floor(x + 0.5) for floating @c T
+    /// (round-half-toward-positive-infinity). Unifies behaviour between float
+    /// and double; pass-through for integer @c T.
+    template<typename Result>
+    __hostdev__ Result roundAs() const {
+        Result r;
+        if constexpr (util::is_floating_point<T>::value) {
+            const T half = T(0.5);
+            for (int i = 0; i < N; ++i) r[i] = math::Floor(mVec[i] + half);
+        } else {
+            for (int i = 0; i < N; ++i) r[i] = static_cast<int32_t>(mVec[i]);
+        }
+        return r;
+    }
+}; // VecBase<T, N>
+
+// ----------------------------> Vec2 <--------------------------------------
+
+/// @brief A simple vector class with two components, similar to openvdb::math::Vec2
+template<typename T>
+class Vec2 : public VecBase<T, 2>
+{
+    using Base = VecBase<T, 2>;
+
+public:
+    using ValueType = T;
+    static const int size = 2; // openvdb::math::Tuple-compat alias of SIZE
+
     Vec2() = default;
-    __hostdev__ explicit Vec2(T x)
-        : mVec{x, x}
-    {
-    }
-    __hostdev__ Vec2(T x, T y)
-        : mVec{x, y}
-    {
-    }
+    __hostdev__ explicit Vec2(T x)            { this->mVec[0] = x;    this->mVec[1] = x;    }
+    __hostdev__ Vec2(T x, T y)                { this->mVec[0] = x;    this->mVec[1] = y;    }
+
     template<template<class> class Vec2T, class T2>
-    __hostdev__ Vec2(const Vec2T<T2>& v)
-        : mVec{T(v[0]), T(v[1])}
-    {
-        static_assert(Vec2T<T2>::size == size, "expected Vec2T::size==2!");
+    __hostdev__ Vec2(const Vec2T<T2>& v) {
+        static_assert(Vec2T<T2>::SIZE == 2, "expected Vec2T::SIZE==2!");
+        this->mVec[0] = T(v[0]); this->mVec[1] = T(v[1]);
     }
     template<typename T2>
-    __hostdev__ explicit Vec2(const Vec2<T2>& v)
-        : mVec{T(v[0]), T(v[1])}
-    {
+    __hostdev__ explicit Vec2(const Vec2<T2>& v) {
+        this->mVec[0] = T(v[0]); this->mVec[1] = T(v[1]);
     }
-    __hostdev__ explicit Vec2(const Coord2& ijk)
-        : mVec{T(ijk[0]), T(ijk[1])}
-    {
+    __hostdev__ explicit Vec2(const Coord2& ijk) {
+        this->mVec[0] = T(ijk[0]); this->mVec[1] = T(ijk[1]);
     }
-    __hostdev__ bool operator==(const Vec2& rhs) const { return mVec[0] == rhs[0] && mVec[1] == rhs[1]; }
-    __hostdev__ bool operator!=(const Vec2& rhs) const { return mVec[0] != rhs[0] || mVec[1] != rhs[1]; }
+
     template<template<class> class Vec2T, class T2>
-    __hostdev__ Vec2& operator=(const Vec2T<T2>& rhs)
-    {
-        static_assert(Vec2T<T2>::size == size, "expected Vec2T::size==2!");
-        mVec[0] = rhs[0];
-        mVec[1] = rhs[1];
-        return *this;
-    }
-    __hostdev__ const T& operator[](int i) const { return mVec[i]; }
-    __hostdev__ T&       operator[](int i) { return mVec[i]; }
-    template<typename Vec2T>
-    __hostdev__ T dot(const Vec2T& v) const { return mVec[0] * v[0] + mVec[1] * v[1]; }
-    __hostdev__ T lengthSqr() const
-    {
-        return mVec[0] * mVec[0] + mVec[1] * mVec[1]; // 3 flops
-    }
-    __hostdev__ T     length() const { return Sqrt(this->lengthSqr()); }
-    __hostdev__ Vec2  operator-() const { return Vec2(-mVec[0], -mVec[1]); }
-    __hostdev__ Vec2  operator*(const Vec2& v) const { return Vec2(mVec[0] * v[0], mVec[1] * v[1]); }
-    __hostdev__ Vec2  operator/(const Vec2& v) const { return Vec2(mVec[0] / v[0], mVec[1] / v[1]); }
-    __hostdev__ Vec2  operator+(const Vec2& v) const { return Vec2(mVec[0] + v[0], mVec[1] + v[1]); }
-    __hostdev__ Vec2  operator-(const Vec2& v) const { return Vec2(mVec[0] - v[0], mVec[1] - v[1]); }
-    __hostdev__ Vec2  operator+(const Coord& ijk) const { return Vec2(mVec[0] + ijk[0], mVec[1] + ijk[1]); }
-    __hostdev__ Vec2  operator-(const Coord& ijk) const { return Vec2(mVec[0] - ijk[0], mVec[1] - ijk[1]); }
-    __hostdev__ Vec2  operator*(const T& s) const { return Vec2(s * mVec[0], s * mVec[1]); }
-    __hostdev__ Vec2  operator/(const T& s) const { return (T(1) / s) * (*this); }
-    __hostdev__ Vec2& operator+=(const Vec2& v)
-    {
-        mVec[0] += v[0];
-        mVec[1] += v[1];
-        return *this;
-    }
-    __hostdev__ Vec2& operator+=(const Coord& ijk)
-    {
-        mVec[0] += T(ijk[0]);
-        mVec[1] += T(ijk[1]);
-        return *this;
-    }
-    __hostdev__ Vec2& operator-=(const Vec2& v)
-    {
-        mVec[0] -= v[0];
-        mVec[1] -= v[1];
-        return *this;
-    }
-    __hostdev__ Vec2& operator-=(const Coord& ijk)
-    {
-        mVec[0] -= T(ijk[0]);
-        mVec[1] -= T(ijk[1]);
-        return *this;
-    }
-    __hostdev__ Vec2& operator*=(const T& s)
-    {
-        mVec[0] *= s;
-        mVec[1] *= s;
-        return *this;
-    }
-    __hostdev__ Vec2& operator/=(const T& s) { return (*this) *= T(1) / s; }
-    __hostdev__ Vec2& normalize() { return (*this) /= this->length(); }
-    /// @brief Perform a component-wise minimum with the other Coord.
-    __hostdev__ Vec2& minComponent(const Vec2& other)
-    {
-        if (other[0] < mVec[0])
-            mVec[0] = other[0];
-        if (other[1] < mVec[1])
-            mVec[1] = other[1];
+    __hostdev__ Vec2& operator=(const Vec2T<T2>& rhs) {
+        static_assert(Vec2T<T2>::SIZE == 2, "expected Vec2T::SIZE==2!");
+        this->mVec[0] = rhs[0]; this->mVec[1] = rhs[1];
         return *this;
     }
 
-    /// @brief Perform a component-wise maximum with the other Coord.
-    __hostdev__ Vec2& maxComponent(const Vec2& other)
-    {
-        if (other[0] > mVec[0])
-            mVec[0] = other[0];
-        if (other[1] > mVec[1])
-            mVec[1] = other[1];
+    // ---- element-wise (Vec & Vec) ----
+    __hostdev__ Vec2  operator-() const            { return Base::template negate<Vec2>(); }
+    __hostdev__ Vec2  operator+(const Vec2& v) const { return Base::template plus<Vec2>(v); }
+    __hostdev__ Vec2  operator-(const Vec2& v) const { return Base::template minus<Vec2>(v); }
+    __hostdev__ Vec2  operator*(const Vec2& v) const { return Base::template mul<Vec2>(v); }
+    __hostdev__ Vec2  operator/(const Vec2& v) const { return Base::template div<Vec2>(v); }
+    __hostdev__ Vec2& operator+=(const Vec2& v)    { Base::addAssign(v); return *this; }
+    __hostdev__ Vec2& operator-=(const Vec2& v)    { Base::subAssign(v); return *this; }
+
+    // ---- mixed Vec2 / Coord2 ----
+    __hostdev__ Vec2  operator+(const Coord2& ijk) const { return Vec2(this->mVec[0] + ijk[0], this->mVec[1] + ijk[1]); }
+    __hostdev__ Vec2  operator-(const Coord2& ijk) const { return Vec2(this->mVec[0] - ijk[0], this->mVec[1] - ijk[1]); }
+    __hostdev__ Vec2& operator+=(const Coord2& ijk) {
+        this->mVec[0] += T(ijk[0]); this->mVec[1] += T(ijk[1]);
         return *this;
     }
+    __hostdev__ Vec2& operator-=(const Coord2& ijk) {
+        this->mVec[0] -= T(ijk[0]); this->mVec[1] -= T(ijk[1]);
+        return *this;
+    }
+
+    // ---- scalar ----
+    __hostdev__ Vec2  operator*(const T& s) const  { return Base::template scale<Vec2>(s); }
+    __hostdev__ Vec2  operator/(const T& s) const  { return Base::template divideBy<Vec2>(s); }
+    __hostdev__ Vec2& operator*=(const T& s)       { Base::scaleAssign(s); return *this; }
+    __hostdev__ Vec2& operator/=(const T& s)       { Base::divideAssignScalar(s); return *this; }
+    __hostdev__ Vec2& normalize()                  { return (*this) /= this->length(); }
+
+    // ---- equality ----
+    __hostdev__ bool operator==(const Vec2& rhs) const { return Base::equals(rhs); }
+    __hostdev__ bool operator!=(const Vec2& rhs) const { return !Base::equals(rhs); }
+
+    // ---- component-wise min/max ----
+    __hostdev__ Vec2& minComponent(const Vec2& other) { Base::mergeMin(other); return *this; }
+    __hostdev__ Vec2& maxComponent(const Vec2& other) { Base::mergeMax(other); return *this; }
+
     /// @brief Return the smallest vector component
-    __hostdev__ ValueType min() const
-    {
-        return mVec[0] < mVec[1] ? mVec[0] : mVec[1];
-    }
+    __hostdev__ ValueType min() const { return Base::smallestComponent(); }
     /// @brief Return the largest vector component
-    __hostdev__ ValueType max() const
-    {
-        return mVec[0] > mVec[1] ? mVec[0] : mVec[1];
-    }
-    /// @brief Round each component if this Vec<T> up to its integer value
-    /// @return Return an integer Coord
-    __hostdev__ Coord2 floor() const { return Coord2(Floor(mVec[0]), Floor(mVec[1])); }
-    /// @brief Round each component if this Vec<T> down to its integer value
-    /// @return Return an integer Coord
-    __hostdev__ Coord2 ceil() const { return Coord2(Ceil(mVec[0]), Ceil(mVec[1])); }
-    /// @brief Round each component if this Vec<T> to its closest integer value
-    /// @return Return an integer Coord
-    __hostdev__ Coord2 round() const
-    {
-        if constexpr(util::is_same<T, float>::value) {
-            return Coord2(Floor(mVec[0] + 0.5f), Floor(mVec[1] + 0.5f));
-        } else if constexpr(util::is_same<T, int>::value) {
-            return Coord2(mVec[0], mVec[1]);
-        } else {
-            return Coord2(Floor(mVec[0] + 0.5), Floor(mVec[1] + 0.5));
-        }
-    }
+    __hostdev__ ValueType max() const { return Base::largestComponent(); }
 
-    /// @brief return a non-const raw constant pointer to array of three vector components
-    __hostdev__ T* asPointer() { return mVec; }
-    /// @brief return a const raw constant pointer to array of three vector components
-    __hostdev__ const T* asPointer() const { return mVec; }
+    /// @brief Round each component down (toward negative infinity)
+    /// @return integer Coord2
+    __hostdev__ Coord2 floor() const { return Base::template floorAs<Coord2>(); }
+    /// @brief Round each component up (toward positive infinity)
+    /// @return integer Coord2
+    __hostdev__ Coord2 ceil()  const { return Base::template ceilAs<Coord2>(); }
+    /// @brief Round each component to its closest integer value
+    /// @return integer Coord2
+    __hostdev__ Coord2 round() const { return Base::template roundAs<Coord2>(); }
 }; // Vec2<T>
 
 template<typename T1, typename T2>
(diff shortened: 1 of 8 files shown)


nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +807,279 @@ class Coord2
...
+    /// @brief floor-rounded components into the @c Result type (whose @c [i]
+    /// must accept int32_t). For integer @c T the value is passed through.
+    template<typename Result>
+    __hostdev__ Result floorAs() const {
+        Result r;
+        if constexpr (util::is_floating_point<T>::value) {
+            for (int i = 0; i < N; ++i) r[i] = math::Floor(mVec[i]);
         } else {
-            return Coord2(Floor(mVec[0] + 0.5), Floor(mVec[1] + 0.5));
+            for (int i = 0; i < N; ++i) r[i] = static_cast<int32_t>(mVec[i]);
```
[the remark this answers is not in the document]
swahtz: Addressed in 1bef1d77. You're right that `util::is_floating_point<T>` doesn't include `long double` (its docstring claims to be a C++11 implementation of `std::is_floating_point` but it omits `long double`), so the refactor's constexpr branch was silently truncating `Vec<long double>` via `static_cast<int32_t>` instead of going through the Floor / Ceil rounding path.

Switched the three predicates in `VecBase::floorAs` / `ceilAs` / `roundAs` to `std::is_floating_point<T>`, added an explicit `#include <type_traits>` for it, and added a comment explaining the choice.

One nuance worth surfacing: the underlying `math::Floor` / `math::Ceil` only have `float` and `double` overloads (no `long double`), so the master version of `Vec3<long double>::round()` was ALSO a compile error (via ambiguous overload resolution on `Floor(long double)` — I confirmed this by compiling against the master Math.h). My fix therefore doesn't make `Vec<long double>` usable for floor / ceil / round — it restores the master behavior of failing with an ambiguous-overload diagnostic rather than silently truncating, which is strictly preferable. Anyone who wants `Vec<long double>` arithmetic going forward would need to add the corresponding `math::Floor(long double)` / `math::Ceil(long double)` overloads (or upgrade `util::is_floating_point` to match `std::is_floating_point` semantics, which is out of scope for this PR since it touches `util/Util.h`).

nanovdb/nanovdb/math/Math.h
```diff
@@ -136,11 +138,11 @@ __hostdev__ inline Type Min(Type a, Type b)
 }
 __hostdev__ inline int32_t Min(int32_t a, int32_t b)
 {
-    return int32_t(fminf(float(a), float(b)));
+    return a < b ? a : b;
```
harrism, line 171: **👏 praise:** ‏Nice. Wow, that previous code... 🙄

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
-        static_assert(Vec2T<T2>::size == size, "expected Vec2T::size==2!");
-        mVec[0] = rhs[0];
-        mVec[1] = rhs[1];
-        return *this;
-    }
+    static constexpr int SIZE = N;
+
+    VecBase() = default;
+
+    /// @brief Indexed element access (no bounds check; assumes 0 <= i < N)
```
harrism: **💡 suggestion:** ‏I understand why we don't bounds check, but perhaps an assert() to catch OOB access in debug builds?

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
     __hostdev__ const T& operator[](int i) const { return mVec[i]; }
```
harrism: **💡 suggestion:** ‏This method can and should be `constexpr`.

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
-    __hostdev__ Vec2  operator-(const Vec2& v) const { return Vec2(mVec[0] - v[0], mVec[1] - v[1]); }
-    __hostdev__ Vec2  operator+(const Coord& ijk) const { return Vec2(mVec[0] + ijk[0], mVec[1] + ijk[1]); }
-    __hostdev__ Vec2  operator-(const Coord& ijk) const { return Vec2(mVec[0] - ijk[0], mVec[1] - ijk[1]); }
-    __hostdev__ Vec2  operator*(const T& s) const { return Vec2(s * mVec[0], s * mVec[1]); }
-    __hostdev__ Vec2  operator/(const T& s) const { return (T(1) / s) * (*this); }
-    __hostdev__ Vec2& operator+=(const Vec2& v)
-    {
-        mVec[0] += v[0];
-        mVec[1] += v[1];
+    __hostdev__ T&       operator[](int i)       { return mVec[i]; }
```
harrism: **💡 suggestion:** ‏This method can and should be `constexpr`

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+
+    /// @brief raw pointer to the underlying N-element storage
+    __hostdev__ T*       asPointer()       { return mVec; }
+    __hostdev__ const T* asPointer() const { return mVec; }
```
harrism: **💡 suggestion:** ‏These methods can also be `constexpr`.

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+
+    // ---- generic element-wise helpers (taking/returning Derived) ----
+
+    template<typename Derived>
+    __hostdev__ Derived plus(const Derived& rhs) const {
+        Derived out;
```
harrism: **🤓 nitpick (non-blocking):** ‏Technically, I gather that pre-C++20 this is ill defined if this function is made constexpr due to the loop. Fix is to value initialize. The compiler will optimize it out anyway because every element is written.

```suggestion
        Derived out{};
```

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+    __hostdev__ T&       operator[](int i)       { return mVec[i]; }
+
+    /// @brief raw pointer to the underlying N-element storage
+    __hostdev__ T*       asPointer()       { return mVec; }
+    __hostdev__ const T* asPointer() const { return mVec; }
+
+    // ---- generic element-wise helpers (taking/returning Derived) ----
+
+    template<typename Derived>
+    __hostdev__ Derived plus(const Derived& rhs) const {
```
harrism: **💡 suggestion:** ‏All of the operators can be made `constexpr`.

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+    // and `double` whereas the former (correctly) also matches
+    // `long double`. Without the broader predicate, `Vec<long double>`
+    // would silently truncate via `static_cast<int32_t>` instead of using
+    // the floor/ceil/round rounding rules. The inner `math::Floor` /
+    // `math::Ceil` only have float/double overloads, so `long double`
+    // implicitly converts to `double` at the call site — matching the
+    // pre-refactor behaviour of `Vec*::round()` which routed `long
+    // double` through the same `Floor(x + 0.5)` path.
+
+    /// @brief floor-rounded components into the @c Result type (whose @c [i]
```
harrism, line 1112: **☑️ todo:** ‏You have an @brief here and then two methods run together under the same comment. Doxygen docs apply per method, function, variable, class. Each function should have its own doc string.

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+        T s = T(0);
+        for (int i = 0; i < N; ++i) s += mVec[i] * v[i];
+        return s;
+    }
+    __hostdev__ T lengthSqr() const {
+        T s = T(0);
+        for (int i = 0; i < N; ++i) s += mVec[i] * mVec[i];
+        return s;
+    }
+    __hostdev__ T length() const { return Sqrt(this->lengthSqr()); }
```
harrism: **🎵 note (non-blocking):** ‏This is basically the only method of this class that can't be compiled with `constexpr` (until C++26 -- due to newton-raphson iterations in std::sqrt).

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
 
-    /// @brief return a non-const raw constant pointer to array of three vector components
-    __hostdev__ T* asPointer() { return mVec; }
-    /// @brief return a const raw constant pointer to array of three vector components
-    __hostdev__ const T* asPointer() const { return mVec; }
+// ----------------------------> Vec2 <--------------------------------------
+
+/// @brief A simple vector class with two components, similar to openvdb::math::Vec2
+template<typename T>
+class Vec2 : public VecBase<T, 2>
```
harrism: **💡 suggestion:** ‏alignment.


```suggestion
class alignas(alignof(T) * 2) Vec2 : public VecBase<T, 2>
```

nanovdb/nanovdb/math/Math.h
harrism: **💡 suggestion:** ‏Add more legible x/y/z/w accessors, e.g. 

```
__hostdev__ constexpr const T& x() const { return mVec[0]; } 
__hostdev__ constexpr const T& y() const { static_assert(N > 1); return mVec[0]; } 
...
```

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+{
+    using Base = VecBase<T, 2>;
+
+public:
```
harrism, line 1170: **💡 suggestion:** ‏constexpr all methods.

nanovdb/nanovdb/math/Math.h
harrism: **💡 suggestion:** ‏Rename these `data()` to match `MatBase` and everything in STL...

nanovdb/nanovdb/math/Math.h
harrism, line 1316: **💡 suggestion:** ‏Everything in this class and its derived classes can be `constexpr`. One caveat below (`Inverse`). Start with the `operator[]` accessors. Those must be `constexpr` for all of the methods that call them to be `constexpr`.

nanovdb/nanovdb/math/Math.h
```diff
@@ -1033,40 +1304,39 @@ class Mat2 : public MatBase<T, 2, 2> {
...
     /// @brief returns transpose of this
-    __hostdev__ Mat2<T> transpose() const {
-        return Mat2<T>((*this)[0][0], (*this)[1][0], (*this)[0][1], (*this)[1][1]);
-    }
+    __hostdev__ Mat2 transpose() const { return this->template transposeAs<Mat2>(); }
 
     /// @brief returns inverse of this
     __hostdev__ Mat2<T> inverse() const {
         T det = (*this)[0][0] * (*this)[1][1] - (*this)[0][1] * (*this)[1][0];
         if (isApproxZero(det)) {
```
harrism, line 1559: **💡 suggestion:** ‏For this function to be `constexpr`, you also need to make `isApproxZero()` in math.h `constexpr` (it is constexprable), as well as ` Tolerance<T>::value()` for `float` and `double` (also in math.h).

nanovdb/nanovdb/math/Math.h
harrism, line 1515: **💡 suggestion:** ‏Alignment.


```suggestion
template <typename T>
class alignas(alignof(T) * 4) Mat2 : public MatBase<T, 2, 2> {
```

harrism: **❓ question:** ‏Why not name the square matrix types `Mat2x2, Mat3x3, Mat4x4` for consistency with `Mat2x3` and `Mat3x2`?

harrism: **🎵 note (non-blocking):** ‏Don't `alignas` this class, or any of the vector and matrix classes with one dimension of `3`.  Add a comment such as this..

```
// Mat2x3 / Mat3x2 / Mat3: intentionally NOT alignas. Their byte sizes
// (6*sizeof(T), 9*sizeof(T)) are not powers of 2, so alignas(N) for any
// N > alignof(T) would force tail padding and break packed-array layout
// and on-disk format compatibility.
```

We *could* add an opt-in aligned `Vec3Aligned` class that is aligned as `4 * sizeof(T)` for CUDA coalescing performance reasons. But due to the need to be compatible with byte layout on disk for files, best to make it opt-in.

nanovdb/nanovdb/math/Math.h
```diff
@@ -1252,281 +1511,181 @@ class Mat4 : public MatBase<T, 4, 4> {
...
+template<typename T>
+__hostdev__ Mat3x2<T> operator*(const Mat3<T>& lhs, const Mat3x2<T>& rhs) {
+    return lhs.template multiply<Mat3x2<T>, Mat3x2<T>>(rhs);
 }
 // ----------------------------> Vec3 <--------------------------------------
 
 /// @brief A simple vector class with three components, similar to openvdb::math::Vec3
 template<typename T>
-class Vec3
+class Vec3 : public VecBase<T, 3>
```
harrism: **💡 suggestion:** ‏Add a comment explaining why we don't `alignas` this class so somebody doesn't come along and "fix" it later.

nanovdb/nanovdb/math/Math.h
```diff
@@ -1033,40 +1304,39 @@ class Mat2 : public MatBase<T, 2, 2> {
...
+
+    // ---- equality ----
+    __hostdev__ bool operator==(const Mat2& m) const     { return Base::equals(m); }
+    __hostdev__ bool operator!=(const Mat2& m) const     { return !Base::equals(m); }
 
     /// @brief returns transpose of this
-    __hostdev__ Mat2<T> transpose() const {
-        return Mat2<T>((*this)[0][0], (*this)[1][0], (*this)[0][1], (*this)[1][1]);
-    }
+    __hostdev__ Mat2 transpose() const { return this->template transposeAs<Mat2>(); }
```
harrism: **☑️ todo:** ‏Add `[[nodiscard]]` on every value-returning query/method.  I'm commenting here on `transpose()` because forgetting to capture the result of `m.transpose()` thinking it mutated m in place is
the kind of bug `[[nodiscard]]` is designed to catch.

nanovdb/nanovdb/math/Math.h
harrism: **💡 suggestion:** ‏None of the methods throw exceptions, so they should be marked `noexcept`.

nanovdb/nanovdb/math/Math.h
```diff
@@ -802,154 +809,290 @@ class Coord2
...
+public:
+    using ValueType = T;
+    static constexpr int size = 2; // openvdb::math::Tuple-compat alias of SIZE
+
+    Vec2() = default;
+    __hostdev__ explicit Vec2(T x)            { this->mVec[0] = x;    this->mVec[1] = x;    }
+    __hostdev__ Vec2(T x, T y)                { this->mVec[0] = x;    this->mVec[1] = y;    }
+
+    template<template<class> class Vec2T, class T2>
+    __hostdev__ Vec2(const Vec2T<T2>& v) {
```
harrism: **☑️ todo:** ‏For consistency, I think this ctor should be `explicit` like the others.

nanovdb/nanovdb/math/Math.h
harrism: **💡 suggestion:** ‏Consider moving these to be hidden friends (`friend` inside the class. This restricts them to ADL on Vec2 rather than everywhere.

nanovdb/nanovdb/math/Math.h
harrism: **💡 suggestion:** ‏Consider marking all the derived classes `final` since they aren't intended to be base for derivation.

harrism (changes requested):
This is a really great cleanup and improvement.  I have comments in 4 areas.

1. constexpr -- might as well constexpr all the things
2. Alignas -- on the classes where it makes sense, let's guarantee alignment so the compiler can optimize / coalesce loads (this allows each row of a 4x4 matrix to be loaded with a single 128-bit global load, for example. Without it, the compiler will generate 4 32-bit loads).
3. Other modernization / best practices topics: `[[nodiscard]]`, `noexcept`, consistent `explicit` ctors, hidden friends.
4. Documentation -- didn't comment on this much inline, but make sure all methods have doxygen docs. Looks a bit inconsistent?

harrism:
I also asked an agent to review with modern C++ best practices in mind and then had it prune its list to not include what I already mentioned in my review.

This first one would be good to implement.

```
 Body-assignment in derived constructors. The new derived ctors do this->mVec[0] = x;
  this->mVec[1] = y; instead of the pre-PR : mVec{x, y} {}. Invisible for fundamental T, a
  regression for any future Vec<UserType>. Fix is a protected variadic constructor on VecBase:

  template<typename... Args>
  __hostdev__ explicit VecBase(Args... args) : mVec{T(args)...} {
      static_assert(sizeof...(Args) == N, "wrong number of args");
  }

  Derived ctors delegate: Vec2(T x, T y) : Base(x, y) {}.
```

The rest are all speculative / nice to have or depend on C++20 or later.

```
Defaulted operator== (C++20 only). bool operator==(const Vec3&) const noexcept = default;
  replaces the manual equals() chain; operator!= auto-synthesizes. Only worth raising if NanoVDB's
   C++ baseline allows.

  Tuple-like interface (tuple_size / tuple_element / get) for auto [x, y, z] = v; and std::apply.
  C++17.

  begin()/end() so Vec/Mat work with range-based for and <algorithm>. Six lines per base.

  [[deprecated]] on the lowercase size alias if the goal is to migrate callers to SIZE. Nudges
  without breaking.

  static_cast<T>(x) over functional-style T(x). Identical for arithmetic types but greppable and
  doesn't silently become a constructor call for non-arithmetic T. Minor.

  CTAD deduction guides so Vec3 v(1.0f, 2.0f, 3.0f); deduces Vec3<float>. Three lines per Vec /
  Mat class. C++17.

  C++20 requires clauses on cross-Vec templated functions (dot<V>, mergeMin<V>, the templated
  converting constructor). Replaces static_assert(V::SIZE == N) in the body with constraints at
  the declaration. Better diagnostics, SFINAE-friendly.

  std::array<T, N> instead of T[N] in the base. Free .data(), iterators, structured bindings,
  .size(), .fill(), ==. Layout-compatible with T[N] for fundamental T. Possibly tangles with CUDA
  portability (cuda::std::array exists). Out of scope for this PR; worth knowing.
```

nanovdb/nanovdb/math/Math.h
swahtz: Yes good idea.  In the spirit of trying to improve Math.h generally, I've applied this across the board so all `::operator[]`s of the various classes here have a `NANOVDB_ASSERT` for their appropriate ranges.

nanovdb/nanovdb/math/Math.h
swahtz: Addressed in 2f95f151 — `VecBase::operator[](int) const` is now `constexpr` (along with the non-const overload below, and the `asPointer()` pair on line 836).

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — non-const `operator[]` is now `constexpr` too.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — both `asPointer` overloads on `VecBase` are now `constexpr`.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — every helper in `VecBase` that builds a result element-by-element now uses `Derived out{};` (and `MatBase` helpers use `Result r{};`), so the constexpr rule about uninitialized locals is satisfied even on older standards. The compiler optimizes the value-init away because every element gets overwritten.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — every operator / helper on `VecBase` (plus, minus, mul, div, negate, scale, divideBy, addAssign, subAssign, mulAssign, divAssign, scaleAssign, divideAssignScalar, equals, dot, lengthSqr, smallestComponent, largestComponent, mergeMin, mergeMax, floorAs, ceilAs, roundAs) is now `constexpr`. Only `length()` remains non-constexpr — `std::sqrt` isn't constexpr until C++26, called out in your other comment on line 931.

nanovdb/nanovdb/math/Math.h
swahtz: Acknowledged — `length()` stays non-constexpr in 2f95f151 with a one-line comment noting "Not constexpr — std::sqrt isn't constexpr until C++26." so a future reader knows the blocker.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — every constructor, every operator, and every accessor on `Vec2` is now `constexpr`. Same for Vec3 and Vec4 (and the free `operator*(T1, Vec2)` / `operator/(T1, Vec2)` etc. overloads).

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — `MatBase::operator[]` (both overloads) and every other method on `MatBase` (`data`, `plus`, `minus`, `negate`, `scale`, `addAssign`, `subAssign`, `scaleAssign`, `divideBy`, `divideAssign`, `equals`, `transposeAs`, `multiply`, `multiplyVec`) is now `constexpr`. Same pass through every derived class — `Mat2` / `Mat2x3` / `Mat3x2` / `Mat3` / `Mat4` constructors, operators, and `transpose()` — plus the free `scalar*matrix` and mixed-shape `matrix*matrix` overloads.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 2f95f151 — `math::Tolerance<float>::value()` / `Tolerance<double>::value()` (both literal-returning one-liners) and `math::isApproxZero<T>` (one-line comparison using `Tolerance`) are now `constexpr`, which makes `Mat2::inverse()` `constexpr` too. Verified at compile time via a `static_assert` on the inverse of the identity in the standalone test TU.

nanovdb/nanovdb/math/Math.h
swahtz: Yes, we should do this and it would align with similar methods in OpenVDB's Math classes.  I've added these in this form to `VecBase`

nanovdb/nanovdb/math/Math.h
swahtz: This is probably worthy of asking @kmuseth for some guidance.  These functions in OpenVDB Math are called `asPointer`, so do we want to keep using `asPointer` to align with OpenVDB or use `data` to align with STL?

`MatBase::data()` is a new function in this PR and is actually a naming 'mistake' and should probably have been called `asPointer` to align with Vector and OpenVDB Math classes.  So the question is rather do we want to switch `nanovdb::math` to use `data` (STL) across the board rather than `asPointer` (OpenVDB).

nanovdb/nanovdb/math/Math.h
swahtz: So the square matrix types take their naming from OpenVDB Math (where the classes are `Mat4`, `Mat3`, etc.).  OpenVDB Math doesn't have any non-square matrix types.  

[In an old PR](https://github.com/NVIDIA-Omniverse/openvdb/pull/238) where we created `Mat3x2`, etc. we had the discussion about naming.  I was originally advocating we name them `Mat32` to follow the convention of `Imath` whose classes are named `Matrix33`, `Matrix44` (which is curious because they don't define any non-square matrices either and just use the square matrices for the affine transforms where you'd need a 2x3, 3x4, etc.).  In that PR, Ken notes that renaming the square Matrix types would be a large change we didn't want to undertake at the time.

nanovdb/nanovdb/math/Math.h
swahtz: Done in b2cf24c8. `Vec2` is now `class alignas(alignof(T) * 2) Vec2` — 8 bytes for `Vec2<float>`, 16 bytes for `Vec2<double>`, no tail padding because the byte size was already a power-of-2 multiple of `alignof(T)`. Same pattern applied to every other Vec/Mat class with a power-of-2 element count (Vec4, Mat2, Mat4); Vec3 / Mat3 / Mat2x3 / Mat3x2 deliberately stay at natural alignment and now carry an inline comment explaining the choice.

nanovdb/nanovdb/math/Math.h
swahtz: Done in b2cf24c8: `class alignas(alignof(T) * 4) Mat2`. 16 bytes for `Mat2<float>`, 32 bytes for `Mat2<double>`, no padding because the byte size was already a power-of-2 multiple of `alignof(T)`.

swahtz: Done in b2cf24c8. `Mat2x3`, `Mat3x2`, `Mat3`, and `Vec3` each now carry an inline comment along the lines you suggested — explaining that their byte sizes (6\*T, 9\*T, 3\*T) aren't powers of 2 so an `alignas(N > alignof(T))` would force tail padding and break packed-array layout plus on-disk format compatibility. Comments are placed immediately above the `template <typename T> class …` line so they catch anyone trying to add `alignas` later.

Took your suggestion on `Vec3Aligned` and left it as a future opt-in type rather than retrofitting `Vec3` — agree the on-disk compatibility is the right reason to keep `Vec3` itself unchanged.

nanovdb/nanovdb/math/Math.h
swahtz: Done in b2cf24c8 — `Vec3` now has an explanatory comment immediately above its template declaration noting it stays at natural alignment because `3*sizeof(T)` is not a power-of-2 multiple of `alignof(T)`, and any `alignas(N > alignof(T))` would force tail padding and break packed-array layout plus on-disk format compatibility. Same comment pattern applied to `Mat3` / `Mat2x3` / `Mat3x2` per the related thread.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 5890f0d4. 282 `[[nodiscard]]` annotations applied across `Math.h` on every value-returning query — `transpose()` (the trigger for your comment), `inverse()`, `cross`, `outer`, `dot`, `lengthSqr`, `floorAs/ceilAs/roundAs`, every `operator+/-/* /`, every comparison operator, the free math helpers (`Min/Max/Pow2/3/4/Abs/Sign/MinIndex/MaxIndex/AlignUp/Round/Floor/Ceil/Sqrt/Clamp/Fract/isApproxZero/matMult/matMultT` etc.), and so on.

Deliberately skipped to avoid noisy false-positive warnings on legitimate code:

- `operator[]`, `asPointer` / `data`, named accessors (`x/y/z/w/r/g/b/a`, `packed`) — pure accessors returning references / pointers to internal storage; the discard-the-result bug class doesn't apply.
- Compound-assignment operators (`+=`, `-=`, `*=`, `/=`, `&=`, `<<=`, `>>=`) and other chain-style methods that return `*this` by reference (`Vec::normalize`, `BBox::expand/intersect/translate`) — primary purpose is the mutation; `[[nodiscard]]` would warn on every `a += b;`.
- `operator++` (pre + post on the integer-bbox `Iterator`) — the post-increment's proxy return is conventionally discarded in `for (...; ++it; ...)` loops.
- The device-only `*Atomic` helpers — used for atomic side effects, not query reads.

Verified by recompiling `TestNanoVDB.cc` at `-Wall -Wextra -O2` — zero new `-Wunused-result` warnings, meaning the existing test code already consumes every return value it should. No latent bugs were caught (a slightly disappointing result, but it does confirm the surrounding code is already well-disciplined about capturing math results).

nanovdb/nanovdb/math/Math.h
swahtz: Done in 001dbe31 — confirmed via grep that `Math.h` has zero `throw` / `exception` usages, then added `noexcept` to every function in the file: 489 annotations across the free math helpers, every `Coord` / `Coord2` / `VecBase` / `Vec*` / `MatBase` / `Mat*` / `BBox` / `Iterator` / `Rgba8` method, and every standalone operator overload. Defaulted constructors are now `noexcept = default;`. Out-of-class definitions of `Coord::asVec3s`/`asVec3d` and `Coord2::asVec2s`/`asVec2d` got matching `noexcept` on both the in-class declarations and the out-of-class definitions so the exception specifications agree.

Recompiled `TestNanoVDB.cc` at `-Wall -Wextra -O2` — exit code 0, no new warnings. Constexpr sanity TU also still compiles, so the marks didn't break constant evaluation.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 001dbe31. The templated open `Vec2T<T2>` ctor (and its Vec3 / Vec4 siblings) was the odd one out — the more-specific typed `Vec2(const Vec2<T2>&)` and the `Vec2(const Coord2&)` ctors were already `explicit`. Made the open-template ctor `explicit` on all three classes for consistency. Same one-line change applied to `Vec3(const Vec3T<T2>&)` and `Vec4(const Vec4T<T2>&)`.

nanovdb/nanovdb/math/Math.h
swahtz: Done in 001dbe31. Added `final` to every leaf derived class: `Vec2`, `Vec3`, `Vec4`, `Mat2`, `Mat2x3`, `Mat3x2`, `Mat3`, `Mat4`. They each derive from `VecBase` or `MatBase` and aren't intended as further base classes, so `final` documents the design and prevents accidental further derivation. Placement is after the `alignas` (where present) and the class name, before the `: public …` inheritance, e.g. `class alignas(alignof(T) * 4) Vec4 final : public VecBase<T, 4>`.

nanovdb/nanovdb/math/Math.h
swahtz: Good idea, I didn't know about that pattern.  Implemented that for all the applicable Vec functions and where we could use this in the `scalar*Mat` cases too.

swahtz: Done in 993d4bda — converted the six `scalar*Vec` / `scalar/Vec` templates (Vec2/Vec3/Vec4) and the five `scalar*Mat` templates (Mat2/Mat2x3/Mat3x2/Mat3/Mat4) into hidden friends defined inline in each class body. They are now found only via ADL on the owning type and no longer participate in unrelated namespace-scope overload sets. Existing `scalar*Vec` / `scalar*Mat` tests in TestNanoVDB.cc still pass.

swahtz:
Implemented in a1072df7. Added a protected variadic ctor to both `VecBase` and `MatBase` and converted every derived Vec/Mat ctor (Vec2/3/4, Mat2/2x3/3x2/3/4) to delegate via initializer list — no more body-assignment, no more default-construct-then-overwrite of the base subobject. The `static_assert(sizeof...(Args) == N)` (and `== ROWS*COLS` for MatBase) catches arity mismatches at the call site.

Tracking the other suggestions for follow-up but not in this PR — happy to file separate issues for the C++17-feasible ones (tuple-like interface, begin/end, CTAD guides) if useful.

nanovdb/nanovdb/math/Math.h
swahtz: I separated out those comments.  I'm also going to do a docstring sweep to make sure doxygen docs are consistently defined.

harrism (approved):
Fantastic. Thanks for addressing all my suggestions.

nanovdb/nanovdb/math/Math.h
```diff
@@ -1236,463 +1754,380 @@ class Mat4 : public MatBase<T, 4, 4> {
...
 // ----------------------------> matMult <--------------------------------------
+//
+// All six matMult / matMultT overloads were originally written with
+// fma / fmaf for the single-rounding precision benefit. Those stdlib
+// functions are not constexpr in C++17, which transitively blocked
+// Map::applyMap and BBox<CoordT,false>::transform<Map> from being
+// constexpr. Switching to plain `a * b + c` form gives back the
+// constexpr-eligibility (worth ~1 ulp of rounding accuracy in the
+// worst case, well below NanoVDB's geometric precision) and the
+// device-side codegen is unchanged in practice — nvcc contracts
```
kmuseth, line 2127: The claim that "device-side codegen is unchanged in practice" (and the PR description's "Host codegen unchanged under default -ffp-contract=on" / "only compile-time constexpr evaluation observes the double-rounded result") isn't quite accurate, because this rewrite changed more than just FMA usage — it also reversed the accumulation order:

```
// before: right-to-left, FMA at each step
fmaf(x, m[0], fmaf(y, m[1], z * m[2]))
// after: left-to-right sum
x * m[0] + y * m[1] + z * m[2]
```

Contraction can re-fuse the multiplies, but it can't restore the original association, so results differ at runtime, not only under constexpr. Over 500k random inputs, -O2 with contraction enabled:

`215222/500000 differ (43.0%), worst absolute err 3.1e-05`

To be clear, I don't think this blocks the change — the conclusion holds. Typical drift really is ~1 ulp, and for the case that actually matters here, a uniform-scale Map, results are bit-identical (0 differing out of 500k). It's just the justification that's overstated. Suggest rewording to something like: "accumulation order changes from right-to-left to left-to-right, so results may differ by ~1 ulp on host and device; exact for uniform-scale maps."

nanovdb/nanovdb/unittest/TestNanoVDB.cu
```diff
@@ -2075,7 +2075,7 @@ TEST(TestNanoVDBCUDA, Sphere_CudaPointsToGrid_Voxel32)
                 EXPECT_LE(voxel[0],  0.5f);
                 EXPECT_LE(voxel[1],  0.5f);
                 EXPECT_LE(voxel[2],  0.5f);
-                test = (begin[i] - nanovdb::voxelToWorld(voxel, ijk, grid->map())).length() < 1e-9;
+                test = (begin[i] - nanovdb::voxelToWorld<Vec3T>(voxel, ijk, grid->map())).length() < 1e-9;
```
kmuseth, line 2078: This changes which code path is tested, not just the conversion. voxelToWorld is template<typename Vec3T = Vec3d> and dispatches internally:

```
if constexpr(util::is_same<Vec3T,Vec3d>::value) return map.applyMap(...);   // double
else                                            return map.applyMapF(...);  // float
```

Vec3T is nanovdb::Vec3f in this test, so adding the explicit argument flips all three sites from the double applyMap path to the float applyMapF path — while the tolerances stay at 1e-9 / 1e-6 / 1e-2.

The 1e-9 here is the concern: the sphere has radius 100 and float carries ~1e-7 relative precision, so this assertion is only satisfiable because createPointSphere(8, 100.0, Vec3d(0.0), 0.5) yields an exactly-representable uniform 0.5 scale. Coverage of the double path is silently dropped, and the test would fail for any non-power-of-two voxel size.

Since the goal was just to satisfy the now-explicit ctor, converting the result preserves the original behaviour:

`test = (begin[i] - Vec3T(nanovdb::voxelToWorld(voxel, ijk, grid->map()))).length() < 1e-9;`

Same applies at lines 2205 and 2334.

nanovdb/nanovdb/math/Math.h
```diff
@@ -1236,463 +1754,380 @@ class Mat4 : public MatBase<T, 4, 4> {
...
+    static constexpr int size = 3; // openvdb::math::Tuple-compat alias of SIZE
+
+    /// @brief Default-construct (components are left uninitialized for fundamental @c T).
+    Vec3() noexcept = default;
+    /// @brief Broadcast: set all three components to @a x.
+    __hostdev__ explicit constexpr Vec3(T x) noexcept            : Base(x, x, x) {}
+    /// @brief Component-wise construction.
+    __hostdev__ constexpr Vec3(T x, T y, T z) noexcept           : Base(x, y, z) {}
+
+    /// @brief Cross-template converting ctor (e.g. from @c openvdb::Vec3). Implicit
```
kmuseth, line 1852: The doc comment says "Implicit to preserve foreign-type interop" directly above a ctor declared explicit. Since making this ctor explicit is the PR's one source-breaking change, this is the comment most likely to be read by someone hitting the resulting compile error — worth getting right.

Same contradiction at lines 1181 (Vec2) and 2025 (Vec4).

nanovdb/nanovdb/unittest/TestNanoVDB.cu
```diff
@@ -2202,7 +2202,7 @@ TEST(TestNanoVDBCUDA, Sphere_CudaPointsToGrid_Voxel16)
             EXPECT_LE(count, maxPointsPerVoxel);
             bool test = false;
             for (uint64_t j=0; test == false && j<count; ++j) {
-                test = (begin[i] - nanovdb::voxelToWorld(start[j], ijk, grid->map())).length() < 1e-6;
+                test = (begin[i] - nanovdb::voxelToWorld<Vec3T>(start[j], ijk, grid->map())).length() < 1e-6;
```
kmuseth, line 2205: Drive-by, not caused by this PR: test is computed here but never asserted — there's no EXPECT_TRUE(test) after the loop, so this check is dead. Every sibling site pairs them (1581/1583, 1699/1704, 1818/1823, 1942/1947, 2069/2080, 2332/2336); only this Voxel16 one is missing, and it's already like that in master. Since you're touching this line anyway, might be worth fixing.

nanovdb/nanovdb/math/Math.h
```diff
@@ -759,250 +864,637 @@ class Coord2
...
+    /// @brief raw pointer to the underlying N-element storage
+    __hostdev__ constexpr T*       asPointer() noexcept       { return mVec; }
+    __hostdev__ constexpr const T* asPointer() const noexcept { return mVec; }
+
+    // ---- generic element-wise helpers (taking/returning Derived) ----
+
+    /// @brief Return @c *this + @a rhs as a @c Derived.
+    template<typename Derived>
+    __hostdev__ [[nodiscard]] constexpr Derived plus(const Derived& rhs) const noexcept {
+        Derived out{};
```
kmuseth, line 968: Minor: Derived out{}; value-initializes before the loop overwrites every element, so the zero-fill is a dead store on some of the hottest paths in the library. Compilers usually elide it at -O2, but it's avoidable. Applies to the whole family of helpers here and in MatBase (968, 975, 982, 989, 996, 1003, 1011, 1379, 1387, 1395).

kmuseth:
In addition to my (old) comments above I'd like to discuss the following (alignment and performance) issues identified by Astra:

# Review of OpenVDB PR #2207

**Recommendation: request changes.** The PR introduces a serialized-format break, a measured CPU normalization slowdown, and several arithmetic regressions. Its source compatibility impact is wider than the constructor tightening highlighted in the description.

Reviewed [PR #2207](https://github.com/AcademySoftwareFoundation/openvdb/pull/2207), head `f4beed62efca49f4f0a431fde70fdabd0954c8da`, against its merge base `7191b429e6bb0547d78d23041515db1b6771867d`. Review date: September 15, 2026. The comparison uses the PR's actual branch point, avoiding unrelated changes subsequently added to master.

## Findings

### 1. [P1] Preserve Vec4 alignment to avoid silently changing the file format

Location: [Math.h:2007](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L2007).

The new `alignas(alignof(T) * 4)` leaves standalone `sizeof(Vec4)` unchanged but changes padding **inside structures containing it**. Comparing compiled layouts gives:

| Property | Base | PR |
|---|---:|---:|
| `alignof(Vec4f)` | 4 | 16 |
| `alignof(Vec4d)` | 8 | 32 |
| Vec4f root background offset | 28 | 32 |
| Vec4f root tile value offset | 20 | 32 |
| Vec4d root tile value offset | 24 | 32 |
| Vec4d leaf minimum offset | 80 | 96 |
| Vec4d leaf voxel-array offset | 160 | 192 |
| Vec4d leaf size | 16544 | 16576 |

Existing Vec4 grids therefore acquire different interpretations. The major format version is unchanged. PNanoVDB's offset table also remains unchanged, so C++ and PNanoVDB in the same checkout already disagree: a byte buffer populated with background `(1,2,3,4)` at the PR's C++ offsets reads as `(0,1,2,3)` using PNanoVDB's offsets. The Vec4d leaf change also affects the stride between adjacent leaves, beyond just statistics.

**Requested change:** retain the existing alignment for serialized types; use separate aligned types or explicit alignment at allocation/use sites where useful. If a format change is intentional, it needs versioning and coordinated reader/writer changes. Add field-offset and leaf-stride tests for Vec4f/Vec4d, including PNanoVDB. Unchanged standalone sizes are insufficient.

The Vec2/Mat2/Mat4 alignment changes also alter downstream enclosing-structure layouts, although this review's concrete NanoVDB file-format failure is Vec4.

### 2. [P2] Keep the reciprocal path for floating-point normalization/division

Locations: [Math.h:1010](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L1010), [Math.h:1052](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L1052).

The integer division fix replaces the existing floating-point reciprocal-and-multiply implementation too. For Vec3 normalization, generated vectorized ARM64 code changes from one division plus three multiplies to three divisions. `normalize()` reaches this path through scalar `/=`. It is used by the repository's ray tracing and collision examples.

Local measurements, in nanoseconds per vector:

| Operation | Base | PR | Change |
|---|---:|---:|---:|
| Vec3f normalization | 0.852 | 1.064 | +25% |
| Vec3d normalization | 1.779 | 2.259 | +27% |
| Vec3d scalar division | 1.091 | 1.216 | +11% |

These are microbenchmarks on an ARM64 Mac with Clang 22.1.8, C++17, `-O3 -DNDEBUG -ffp-contract=on`, without fast math. Each process used 1,024 input vectors, varying inputs/divisors, 10,000 repetitions and seven timed samples. The table combines the two process medians per revision from alternating base/head runs, after compilation completed. Kernels were compiled in a separate translation unit without LTO. Normalization was slower in both comparisons.

Performance is operation- and platform-dependent: Vec4f scalar division became faster in this test, and matrix multiplication did not show a comparably large slowdown. These measurements establish a CPU normalization regression, not a blanket slowdown or a GPU result. No CUDA hardware benchmark was run.

**Requested change:** specialize the integer correction while preserving the existing floating-point path, then benchmark representative CUDA normalization and transform workloads. Also preserve the original reduction evaluation where practical; the new reductions introduce an initial addition to zero.

### 3. [P2] Preserve runtime FMA semantics and the order of accumulation

Location: [Math.h:2144](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L2144), with analogous changes in the other `matMult`/`matMultT` overloads.

The rewrite changes both fusion guarantees and the association of additions: the old nested FMA expression accumulates from the right, while the replacement expression accumulates from the left. Automatic FMA contraction does not restore that order.

Concrete runtime test: use matrix rows `(1,1,1)`, `(0,1,0)`, `(0,0,1)` and input `(1,1e8f,-1e8f)`. The base returns first component **1**, the PR returns **0**. This reproduces with `-O3 -ffp-contract=on`; it is not limited to constant evaluation. The matrix is invertible and all inputs are finite.

The description's claim that drift is at most one ulp and host/device code generation remains identical is therefore unsupported. These helpers feed Map transformations used for world/index coordinates and sampling.

**Requested change:** retain the original nesting and explicit FMA operations at runtime. Make constant evaluation a separate supported path, or defer constexpr on these operations. Add cancellation-sensitive tests rather than only small, well-conditioned inputs.

### 4. [P2] Avoid introducing a rounding step before rounding the coordinate

Location: [Math.h:317](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L317).

Adding `0.5f` can round the float before `floorf` sees it. Two examples:

| Input | Base float overload | PR float overload | PR double overload for the same value |
|---|---:|---:|---:|
| `8388609.f` (exact integer) | 8388609 | 8388610 | 8388609 |
| `0.4999999701976776123046875f` (next float below 0.5) | 0 | 1 | 0 |

Thus the change affects inputs beyond the intended half-integer tie cases and still disagrees across float/double. The order-zero nearest-neighbor sampler calls this function, so it can select an adjacent voxel for these inputs.

**Requested change:** implement the desired tie rule without this intermediate rounding error. Test exact large integers and floats immediately adjacent to half-integers. The related member rounding helpers should use the same corrected implementation where appropriate.

### 5. [P2] Snapshot a scalar divisor that aliases the vector

Location: [Math.h:1052–1053](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L1052).

For `Vec3f v(2,4,6); v /= v[0];`, the base returns `(1,2,3)` and the PR returns `(1,4,6)`. The scalar is a reference to the first component; the loop overwrites it with 1 before processing the remaining components. The old reciprocal was computed before any mutation.

**Requested change:** copy the divisor before the loop or accept it by value. Cover Vec2, Vec3 and Vec4. The newly added matrix `/=` helper has the same aliasing hazard.

### 6. [P2] Accumulate mixed-type dot products in the promoted type

Location: [Math.h:1067–1069](https://github.com/AcademySoftwareFoundation/openvdb/blob/f4beed62efca49f4f0a431fde70fdabd0954c8da/nanovdb/nanovdb/math/Math.h#L1067).

The old expression performed arithmetic in the promoted operand type and converted once on return. The new loop stores each intermediate sum in `T`. `Vec3i(1,1,1).dot(Vec3d(.4,.4,.4))` returns **1** in the base and **0** in the PR, because each partial sum is truncated. A float receiver with double operands likewise loses intermediate precision that it previously retained.

**Requested change:** use the promoted product/sum type for the accumulator and convert at return, seeding from the first product where appropriate.

## Source API compatibility inventory

All six small compile probes in the attached bundle compile against the base and fail against the PR. These are observable compatibility changes, regardless of whether the new API policy is preferred.

| Change | Previously valid usage affected | Suggested handling |
|---|---|---|
| Converting constructors become explicit | Passing an external vector or `Vec3d` to an argument of type `Vec3f` | Preserve implicit interop, or clearly release-note the change and provide migration examples. The PR acknowledges this change. |
| All eight Vec/Mat leaf classes become `final` | Deriving a small utility type from `Vec3f`, or extending a matrix type | Remove `final` for compatibility. These classes have no virtual dispatch to devirtualize. This break is absent from the PR's compatibility summary. |
| Scalar operators become hidden friends | Qualified `nanovdb::math::operator*(2.f, v)` calls and explicit function-template use | Preserve a namespace-visible compatibility surface if supported. Ordinary `2.f * v` expressions continue to work. |
| Mat4–Vec4 free multiplication becomes a member | Qualified `nanovdb::math::operator*(m, v)` calls | Assess or preserve the old callable interface; ordinary `m * v` still works. |
| Vec2–Coord arithmetic now takes Coord2 | Code intentionally using the first two components of a 3D coordinate | Prefer a deprecation/migration policy if source compatibility is required. The PR acknowledges this change. |

The dimensionally invalid Mat3–Mat2x3 overload is also removed. That removal is documented and dimensionally justified. It should still be distinguished from API additions.

## Verification and scope

- Compiled both revisions' headers and executed independent layout/arithmetic reproducers.
- Compared C++ offsets with PNanoVDB's own constants using a byte-buffer fixture; this was not a full historical `.nvdb` file round trip.
- Compiled six source-compatibility probes against both revisions.
- Examined optimized ARM64 assembly and benchmarked eight existing operations.
- Built the PR's `TestNanoVDB.cc` with GoogleTest v1.17.0. **21 selected tests passed**, covering the new math tests, vectors, matrices, Map, coordinates, vector sampling, leaves/roots and the existing CNanoVDB/PNanoVDB checks. These checks do not catch the demonstrated edge cases.
- Inspected the full changed-file list, surrounding math code, call sites, PR discussion and current checks. No full OpenVDB build, CUDA execution, end-to-end rendering benchmark, or downstream application build was performed.
- No review or comments were posted to GitHub.

## Reproducing the checks

Unzip `pr2207-reproducers.zip`. Obtain two OpenVDB source trees at the commits listed above. The included script takes repository roots, each containing the `nanovdb/` directory:

```sh
python3 reproduce.py --base /path/to/base-checkout --head /path/to/pr-checkout --out results --benchmark
```

It requires Python 3 and Clang with C++17 support. The standalone probes need no OpenVDB libraries, GoogleTest, CUDA or TBB. Omit `--benchmark` for just the layout, arithmetic and compilation checks. The bundle also contains the observed outputs, selected GoogleTest results and generated assembly.

## Suggested review decision

Resolve the Vec4 format break first. Keep the integer correctness fixes separate from floating-point algorithm changes, preserve existing public API unless a breaking release is intended, and add the failing regression cases above. Then require representative CUDA performance measurements before accepting claims of unchanged GPU performance.

kmuseth (changes requested):
Overall this looks good, but I have some concerns regarding alignment ant performance (see review above by Astra)

The MatBase/VecBase consolidation reads cleanly, and the bug fixes are genuine, pre-existing issues worth having caught: Min/Max for integer types routing through fminf/fmaxf, Vec/Mat integer operator/ returning zeros, Mat2::inverse() returning uninitialized data on singular input, and Vec2 arithmetic silently dropping z when given a 3D Coord.

I've left five inline comments. Two I'd like your take on before this merges — the matMult accumulation-order claim, and the voxelToWorld<Vec3T> change in the CUDA tests, which I think alters which code path is under test rather than just the conversion. The other three are minor (a stale doc comment, a pre-existing dead assertion, and a dead store).

Verification caveat: locally I built and ran nanovdb_test_nanovdb (163/163 pass) and confirmed sizeof/alignof parity for every Vec/Mat type against master, so the ABI-visible layout is unchanged. However, I have no nvcc on this machine, so I could not compile or run any of the CUDA translation units. My comments on TestNanoVDB.cu are from reading the code only — in particular the claim that voxelToWorld<Vec3T> flips those three assertions from the double applyMap path to the float applyMapF path is derived from the if constexpr dispatch in PointsToGrid.cuh, not from observed behaviour. Would appreciate a second pair of eyes there from someone who can build the CUDA targets.

nanovdb/nanovdb/math/Math.h
harrism: We can have our cake and eat it in this case.

Compilers honor parentheses as an association barrier (they can't reassociate across them without `-ffast-math`/`-fassociative-math`), so the robust, `constexpr`-eligible solution is to write each component as 

```
x * mat[0] + (y * mat[1] + z * mat[2])
```

This restores the right-to-left association, and with contraction enabled the compiler fuses it back into exactly the original `fmaf(x, m0, fmaf(y, m1, z*m2))`. Measured with `clang -O2`, 500k random inputs vs the base fmaf form:

```
┌────────────────────────┬─────────────────────┬────────────────────────────────┐
│     -ffp-contract      │ current PR (a+b+c)  │         parenthesized          │
├────────────────────────┼─────────────────────┼────────────────────────────────┤
│ on (clang/gcc default) │ 215042 differ (43%) │                       0 differ │
├────────────────────────┼─────────────────────┼────────────────────────────────┤
│ fast                   │       215042 differ │                       0 differ │
├────────────────────────┼─────────────────────┼────────────────────────────────┤
│ off                    │       223487 differ │ 178257 differ (~1 ulp, no FMA) │
└────────────────────────┴─────────────────────┴────────────────────────────────┘
```

Notes (from Fable 5.1):
- Astra's cancellation case mat=(1,1,1)·(1,1e8,-1e8): base=1, current PR=0, parenthesized=1 in all three modes — the catastrophic case is an association bug, so parentheses fix it regardless of FMA.
- Bit-identity to the old code relies on contraction being on. That's the default for nvcc (-fmad=true), GCC (-ffp-contract=fast), and clang (on), but MSVC /fp:precise and anyone building with -ffp-contract=off get a double-rounded result — ~1 ulp drift, same as the PR description already claims. That's an honest version of the "codegen unchanged" statement.
- Same change is needed in all six matMult/matMultT overloads (float/double, and the Mat3-typed variants).

harrism:
@kmuseth regarding your Astra point 1 (keep vec4 default alignment). This change was based on my original review comment. This is classic CUDA optimization. The way the code is written, with vec4<float> 4-byte-aligned, every load and store of a vec4 will result in 4x 32-bit load/store instructions. If the compiler can see it is 16-byte aligned, it will instead generate a single 128-bit load/store, which is what we want. We can't get close to peak GPU bandwidth with serialized vec4 loads. (Note for a Mat4x4<float>, 4 128-bit loads become 16 32-bit loads, and for all the other types we get some fraction of this vectorization (e.g. vec4d, vec2f, etc.))

My mistake in the May review was not thinking about format layout and data versioning. We can't break that, as you point out.

The good news is there is a cake-and-eat-it solution for this too.

1. At a major release we should consider adding `alignas()` to the base types as in b2cf24c8 *and* updating the PNanoVDB constants table.
2. For now (Claude helped with the following plan): 
  A. Leave `Vec4`/`Vec2`/`Mat4` without alignas (revert that part of b2cf24c8), with the same "this is on-disk layout" comment the `Vec3` family already has.
  B. Add an aligned load helper and use it in `LeafData::getValue`, `InternalData::getValue`/`tile` reads, and the value-array iterators — the paths that dominate kernel time:

```  
template<typename ValueT>
__hostdev__ inline ValueT loadAligned(const ValueT* p) {
    if constexpr (sizeof(ValueT) == 16 || sizeof(ValueT) == 32)
        return *static_cast<const ValueT*>(__builtin_assume_aligned(p, sizeof(ValueT)));
    else
        return *p;
}
```

`__builtin_assume_aligned` is supported by nvcc, clang, and gcc; for MSVC fall back to `__assume((uintptr_t)p % 16 == 0)` or plain load. With that, nvcc emits LD.E.128 for Vec4f and 2×128 for Vec4d. Alternatively an opt-in `Vec4fAligned` (`alignas(16)`, same storage) that the accessor reinterprets through — same effect, slightly more type noise (and `reinterpret_cast` smell).
3. Pin the assumption with `static_assert`s in `LeafData`/`InternalData`: `offsetof(..., mValues) % 16 == 0`, `sizeof(DataType) % 32 == 0` (already there), so nobody breaks the invariant silently.
4. For user-owned arrays (e.g. `PointsToGrid` inputs, blind data), document that Vec4 is naturally aligned and offer `Vec4Aligned` as an opt-in if they want vectorized loads — that's my `Vec3Aligned` idea from the May review generalized.

PR #2219, opened 2026-05-21
NanoVDB Python: restructure to mirror the C++ NanoVDB API

### LINKED ISSUE #2208: NanoVDB Python Bindings — C++ API Mirror
### Goal

**Mirror the meaningful C++ NanoVDB surface in Python.** Today's bindings expose roughly a quarter of the user-facing C++ API. This plan brings tree/node access, build grids, validation, stats, blind data, all 24 grid types, all primitives, and the type-erased introspection surface to Python — organized to match `nanovdb::*` namespaces one-for-one.

### What this enables

Even without expanding the GPU operation surface, mirroring the C++ API unlocks meaningful capabilities that aren't possible today:

- **Real-world `.nvdb` files load and inspect correctly.** Only 6 of the 24 `GridType` enum values are currently accessible from Python. After this plan, any grid type a user receives in a file — `Boolean`, `Vec3d`, `Half`, `OnIndex`, `PointIndex`, `Fp16`, `Mask`, etc. — can be read, introspected, and operated on without falling off the binding cliff.

- **Walk the tree from Python.** `Grid.tree().root()`, node iterators, `NodeManager`, and `VoxelBlockManager` let users build their own analytics, visualizations, exporters, and debuggers — counting active voxels by region, computing per-node histograms, dumping topology graphs, finding sparse hotspots, etc.

- **Construct grids programmatically in pure Python.** `nanovdb.tools.build.Grid<T>` exposes the voxel-by-voxel construction path that today is only reachable through a single bbox+callback factory. Procedural worlds, test fixtures, synthetic ML datasets, and programmatic editing all become feasible without dropping to C++.

- **Type-erased introspection.** `GridMetaData` lets a Python tool answer "what's in this buffer?" without knowing `BuildT` at compile time — essential for generic loaders, viewers, and converters.

- **Point grids become first-class.** `PointAccessor` (and surfacing the existing `PointGrid` properly via `handle.grid()`) makes the `PointData`/`PointIndex` grid classes actually usable from Python — gridPoints/leafPoints/voxelPoints iteration, zero-copy NumPy views of attribute arrays.

- **Blind data access.** `findBlindData(name)`, `findBlindDataForSemantic(...)`, and `getBlindData(n)` expose the sidecar channels that `OnIndexGrid` and `PointGrid` use to carry their actual values, colors, normals, radii, velocities, IDs, etc.

- **Integrity and statistics as first-class.** `validateGrid`, `checkGrid`, `isValid`, `evalChecksum`, `validateChecksum`, `updateGridStats`, and `getExtrema` are bindable in straightforward ways and turn quality-control workflows into one-liners.

- **Full primitive library.** Box, BBox (wireframe), octahedron, and the point-sampled variants — 8 missing primitives that exist in C++ today.

- **Quantized grids.** `Fp4`, `Fp8`, `Fp16`, `FpN` (with `AbsDiff`/`RelDiff` oracle classes) become creatable and readable from Python — important for memory-constrained workflows.

- **Scaling the binding surface.** A code-gen approach (one entry per `BuildT` in a master list) means adding a new grid type — say a new vector size or a new fixed-bit float — is one line instead of dozens of repeated `defineGrid<T>` blocks.

### What already exists

The current bindings live under [nanovdb/nanovdb/python/](nanovdb/nanovdb/python/) and ship a single nanobind extension built from `NanoVDBModule.cc` plus per-area `.cc/.cu` files.

- **Module structure**: `nanovdb` (root), `nanovdb.math`, `nanovdb.io`, `nanovdb.tools`, `nanovdb.tools.cuda`, and `nanovdb.math.cuda`. Built as `nanovdb_python` with `OUTPUT_NAME "nanovdb"`; `__init__.py` does `from .lib.nanovdb import *`.
- **Core data types**: 6 typed grids (`FloatGrid`, `DoubleGrid`, `Int32Grid`, `Vec3fGrid`, `RGBA8Grid`, `PointGrid`) with matching `*ReadAccessor` classes and (for scalars) `*NodeInfo` classes. `GridData` base class is exposed. `GridHandle` and `DeviceGridHandle` provide typed accessors (`floatGrid()`, `doubleGrid()`, `int32Grid()`, `vec3fGrid()`, `rgba8Grid()`, and `device*` equivalents).
- **Math primitives** (`nanovdb.math`): `Coord`, `Vec3f`/`Vec3d`, `Vec4f`/`Vec4d`, `Rgba8`, `BBox` family, `Map`, `Mask<3/4/5>` (LeafMask/LowerInternalNodeMask/UpperInternalNodeMask), plus samplers (`NearestNeighbor`/`Trilinear`/`Triquadratic`/`Tricubic`) for float/double/int32/Vec3f.
- **I/O** (`nanovdb.io`): `Codec` (NONE/ZIP/BLOSC), `FileMetaData`/`FileGridMetaData`, `readGrid(s)`/`writeGrid(s)`/`hasGrid`/`readGridMetaData` for host buffer; matching `deviceReadGrid(s)`/`deviceWriteGrid(s)` for device buffer.
- **Tools** (`nanovdb.tools`): 4 of 12+ primitives (`createLevelSetSphere`, `createLevelSetTorus`, `createFogVolumeSphere`, `createFogVolumeTorus` — float/double only), grid-from-callback factories `createFloatGrid`/`createDoubleGrid`/`createInt32Grid`/`createVec3fGrid`, `updateChecksum`, `validateGrids`, `StatsMode`/`CheckMode` enums. Optional `openToNanoVDB`/`nanoToOpenVDB`.
- **CUDA bindings** (out of scope for this plan, retained as-is): `nanovdb.tools.cuda.signedFloodFill` (float/double), `nanovdb.tools.cuda.pointsToRGBA8Grid` (Rgba8 only), `nanovdb.math.cuda.sampleFromVoxels` (float/double), and a stub `DeviceBuffer` class with no methods.
- **Tests**: PyTorch-based suite in [TestNanoVDB.py](nanovdb/nanovdb/python/test/TestNanoVDB.py).

### What is proposed (high-level summary)

| Area | Add / change |
|---|---|
| Module structure | Strict mirror of C++ namespaces; introduce `nanovdb.tools.build`; collapse the phantom `nanovdb.math.cuda` submodule into `nanovdb.tools.cuda` |
| Grid types | 18 additional `BuildT` — `Boolean`, `Half`, `Int16`, `Int64`, `UInt8`, `UInt32`, `Vec3d`, `Vec4f`, `Vec4d`, `Vec3u8`, `Vec3u16`, `Fp4`, `Fp8`, `Fp16`, `FpN`, `ValueIndex`, `ValueOnIndex`, `PointIndex`, `ValueMask` |
| Grid access | Polymorphic `handle.grid(n)` via `callNanoGrid` dispatch; `GridMetaData` for type-erased introspection |
| Blind data | `Grid.blindMetaData(n)`, `findBlindData(name)`, `findBlindDataForSemantic(...)`, `getBlindData(n)` returning zero-copy NumPy views; `GridBlindDataClass`/`Semantic` enums |
| Points | `PointAccessor` (`gridPoints`/`leafPoints`/`voxelPoints`) so `PointGrid` is actually usable |
| Tree / nodes | `Grid.tree()`, `NanoRoot`/`NanoUpper`/`NanoLower`/`NanoLeaf` with iterators, stats, and zero-copy NumPy leaf-value views |
| Node management | `NodeManager`/`NodeManagerHandle`/`createNodeManager` (host); `VoxelBlockManager` host build + decode for OnIndex grids |
| Grid construction | `nanovdb.tools.build.Grid<T>` for direct voxel-by-voxel construction; `ValueAccessor`/`WriteAccessor` |
| Statistics | `updateGridStats`, `getExtrema`, `Extrema<T>`, `Stats<T>` |
| Validation | `validateGrid` (single), `checkGrid`, `isValid`, `evalChecksum`, `validateChecksum` |
| Primitives | `createLevelSetBox`/`BBox`/`Octahedron`, `createFogVolumeBox`/`Octahedron`, `createPointSphere`/`Torus`/`Box`, `createPointScatter` (host flavor) |
| `createNanoGrid` | Quantized overloads (FpN with `AbsDiff`/`RelDiff` oracle, Fp4/Fp8/Fp16 with dither, Index/OnIndex with channels) |
| Packaging | Ship `nanovdb/` headers in the wheel under `nanovdb/include/`; add `nanovdb.get_include()` helper |
| Bug fixes | `GridHandle.__bool__` correctness, enum `__repr__` for `GridType`/`GridClass`/`Codec` |
| Quality | Code-gen-driven `BuildT` instantiation, auto-generated `.pyi` stubs |

### Departures from existing work

These are explicit breaks from the current Python API, chosen to mirror C++ and to reduce long-term API surface. All are pre-1.0 changes:

1. **Polymorphic `handle.grid(n)` replaces typed accessors** ([PyGridHandle.h:25-41](nanovdb/nanovdb/python/PyGridHandle.h), [PyDeviceGridHandle.cu:31-38](nanovdb/nanovdb/python/cuda/PyDeviceGridHandle.cu)). The methods `floatGrid()`, `doubleGrid()`, `int32Grid()`, `vec3fGrid()`, `rgba8Grid()`, `deviceFloatGrid()`, `deviceDoubleGrid()`, `deviceInt32Grid()`, `deviceVec3fGrid()`, and `deviceRGBA8Grid()` are removed. Migration: `handle.floatGrid(i)` → `handle.grid(i)` (host) and `handle.deviceFloatGrid(i)` → `handle.deviceGrid(i)`, each returning the appropriate typed `Grid` subclass selected by `gridType(n)`.

2. **`nanovdb.math.cuda` submodule is removed** ([PySampleFromVoxels.cu](nanovdb/nanovdb/python/cuda/PySampleFromVoxels.cu)). The batched `sampleFromVoxels` kernel moves to `nanovdb.tools.cuda.sampleFromVoxels`. This eliminates a Python-only namespace that has no C++ counterpart and aligns with `nanovdb::tools::cuda::*`.

3. **Hand-written `BuildT` template registration is replaced by code-gen** ([NanoVDBModule.cc:388-407](nanovdb/nanovdb/python/NanoVDBModule.cc), [PySampleFromVoxels.cc:68-86](nanovdb/nanovdb/python/PySampleFromVoxels.cc), [PyCreateNanoGrid.cc:48-51](nanovdb/nanovdb/python/PyCreateNanoGrid.cc)). The repeated per-type `defineGrid<T>`, `defineSampler<T>`, `defineCreateNanoGrid<T>` blocks become a single X-macro-driven loop (or a small Python build-time script) so adding/removing a `BuildT` is one line.

4. **Headers ship inside the Python wheel**. New `install(DIRECTORY ... DESTINATION nanovdb/include/nanovdb ...)` block in [CMakeLists.txt](nanovdb/nanovdb/python/CMakeLists.txt) — today the wheel contains only the binary extension and `__init__.py`. Existing system-level header install is unchanged.

5. **`__init__.py` gains Python-level helpers** ([__init__.py](nanovdb/nanovdb/python/__init__.py)). New `nanovdb.get_include()` joins the existing Windows DLL shim. The `from .lib.nanovdb import *` line stays.

6. **New `nanovdb.tools.build` submodule** that doesn't exist today, mirroring C++ `nanovdb::tools::build::*`. Imports the build grid types for in-Python construction.

7. **Pre-existing bug fixes that change observable behavior**:
   - `GridHandle.__bool__` will return `not handle.empty()` instead of `None` ([PyGridHandle.h:23-24](nanovdb/nanovdb/python/PyGridHandle.h)).
   - `repr(nanovdb.GridType.Float)` will return `"Float"` instead of nanobind's default `"GridType.Float"`-with-value (and likewise for `GridClass`, `Codec`).

## Target module layout — mirror the C++ namespaces

| C++ namespace | Python module | Contents |
|---|---|---|
| `nanovdb::*` (root) | `nanovdb` | `NanoGrid`, `GridHandle`, `HostBuffer`, `NodeManager`, `Mask`, `Version`, `GridType`/`GridClass`/`PointType`/`CheckMode`/`MagicType` enums, `Checksum` |
| `nanovdb::math::*` | `nanovdb.math` | `Coord`, `Vec3*`, `Vec4*`, `Rgba8`, `BBox`, `Map`, `SampleFromVoxels` (a single `__hostdev__` template — no separate cuda submodule in C++) |
| `nanovdb::io::*` | `nanovdb.io` | `Codec`, `FileHeader`, `FileMetaData`, `FileGridMetaData`, `readGrid(s)`, `writeGrid(s)`, `hasGrid`, `readGridMetaData` |
| `nanovdb::tools::*` | `nanovdb.tools` | `createNanoGrid`, primitives (`createLevelSetSphere`, …), `validateGrid(s)`, `updateGridStats`, `getExtrema`, `updateChecksum`, `StatsMode`, `Extrema`, `Stats`, `nanoToOpenVDB`, `openToNanoVDB`, `NodeAccessor`, `CreateNanoGrid`, `AbsDiff`, `RelDiff` |
| `nanovdb::tools::build::*` | `nanovdb.tools.build` | Mutable `Grid<T>`, `Tree<T>`, `RootNode`, `InternalNode`, `LeafNode`, `ValueAccessor`, `WriteAccessor` (CPU grid builder) |
| `nanovdb::tools::cuda::*` | `nanovdb.tools.cuda` | Retains current contents (`signedFloodFill`, `pointsToRGBA8Grid`); gains the relocated `sampleFromVoxels` kernel |
| `nanovdb::cuda::*` | `nanovdb.cuda` | Retains current `DeviceBuffer` stub |

```mermaid
graph TD
  root["nanovdb"]
  math["nanovdb.math"]
  io["nanovdb.io"]
  tools["nanovdb.tools"]
  build["nanovdb.tools.build"]
  toolsCuda["nanovdb.tools.cuda"]
  cuda["nanovdb.cuda"]

  root --> math
  root --> io
  root --> tools
  root --> cuda
  tools --> build
  tools --> toolsCuda
```

## Cross-cutting strategy

- **Code-gen for BuildT instantiations**: A master list of `(BuildT, suffix, traits)` drives an X-macro or a tiny build-time Python script that emits the `defineGrid`, `defineAccessor`, `defineSampler`, and `GridHandle.grid()` dispatch arms. Without this, broad type coverage is infeasible.
- **CUDA and OpenVDB stay strictly optional** behind `NANOVDB_USE_CUDA` and `NANOVDB_USE_OPENVDB`; CPU-only builds remain valid.
- **No new dependencies beyond what's already required.**

---

## Phase 0 — Foundation & Build Infrastructure

1. **Code-gen for type instantiations**. Replace hand-rolled blocks in [nanovdb/nanovdb/python/NanoVDBModule.cc](nanovdb/nanovdb/python/NanoVDBModule.cc), [PySampleFromVoxels.cc](nanovdb/nanovdb/python/PySampleFromVoxels.cc), [PyCreateNanoGrid.cc](nanovdb/nanovdb/python/PyCreateNanoGrid.cc), and [PyGridHandle.h](nanovdb/nanovdb/python/PyGridHandle.h) with a single registration loop driven by a `NANOVDB_PY_FOR_EACH_BUILDT(MACRO)` X-macro list. Each TU can include `BuildTypes.def` once.

2. **Install headers as Python package data**. Extend [nanovdb/nanovdb/python/CMakeLists.txt](nanovdb/nanovdb/python/CMakeLists.txt) so under `SKBUILD`:

```cmake
if(SKBUILD)
  install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/../../nanovdb/
          DESTINATION nanovdb/include/nanovdb
          FILES_MATCHING PATTERN "*.h" PATTERN "*.cuh")
endif()
```

3. **Add `nanovdb.get_include()`** to [nanovdb/nanovdb/python/__init__.py](nanovdb/nanovdb/python/__init__.py):

```python
def get_include():
    return os.path.join(os.path.dirname(__file__), 'include')
```

4. **Pre-existing bug fixes**:
   - `GridHandle.__bool__` returns `None` instead of `!empty()` — fix at [PyGridHandle.h:23-24](nanovdb/nanovdb/python/PyGridHandle.h).
   - Enum `__repr__` is commented out for `GridType` ([NanoVDBModule.cc:347-351](nanovdb/nanovdb/python/NanoVDBModule.cc)), `GridClass` ([NanoVDBModule.cc:366-370](nanovdb/nanovdb/python/NanoVDBModule.cc)), and `Codec` ([PyIO.cc:142-146](nanovdb/nanovdb/python/PyIO.cc)) — enable them.

5. **Move `sampleFromVoxels`** from [PySampleFromVoxels.cu](nanovdb/nanovdb/python/cuda/PySampleFromVoxels.cu) (`nanovdb.math.cuda`) into `nanovdb.tools.cuda`. Delete the phantom `math.cuda` submodule registration. No functional change to the kernel itself.

6. **Generate `.pyi` type stubs** via `nanobind-stubgen` for IDE/typing support.

## Phase 1 — Polymorphic & Type-Erased Access

1. **Drop type-specific grid accessors**. Replace `floatGrid()`, `doubleGrid()`, `int32Grid()`, `vec3fGrid()`, `rgba8Grid()` on `GridHandle` and the `device*` equivalents on `DeviceGridHandle` with `handle.grid(n=0)` and `handle.deviceGrid(n=0)`. Return type is selected at runtime via `callNanoGrid<OpT>` from [NanoVDB.h](nanovdb/nanovdb/NanoVDB.h) and is the correct typed Python `Grid` subclass per `gridType(n)`.

2. **Add a common `Grid` Python base class** that all typed grids inherit from. Lift shared methods (`memUsage`, `gridSize`, classification queries, world/index transforms, name/version/checksum) onto it.

3. **Bind `GridMetaData`** (the 768 B type-erased introspector defined in [NanoVDB.h](nanovdb/nanovdb/NanoVDB.h)). Mirrors `FileMetaData` accessors but works on any in-memory grid without knowing `BuildT`. Useful for generic loaders, viewers, and converters.

4. **Blind data API**:
   - `Grid.blindDataCount()`, `Grid.blindMetaData(n)`, `Grid.findBlindData(name)`, `Grid.findBlindDataForSemantic(semantic)`
   - `GridBlindMetaData` struct (data offset, value count/size, semantic, class, type, name)
   - `Grid.getBlindData(n)` returns a zero-copy NumPy view typed by the blind metadata (dtype, shape inferred from value size and count)
   - Enums: `GridBlindDataClass`, `GridBlindDataSemantic`

5. **`PointAccessor<AttT, BuildT>`** so the existing `PointGrid` is actually usable: `gridPoints()`, `leafPoints()`, `voxelPoints()` yielding zero-copy NumPy attribute views.

6. **`GridHandle` utilities**:
   - `splitGrids(handle)` → `list[GridHandle]`
   - `mergeGrids(handles)` → `GridHandle`
   - `handle.copy()` for cross-buffer-type copies where supported

## Phase 2 — Broader Grid Type Coverage

Driven by the Phase 0 code-gen. Expand from the 6 currently bound `BuildT` to the full `GridType` enum:

- **Scalars**: `bool` (Boolean), `int16_t`, `int64_t`, `uint32_t`, `uint8_t`, `Half`
- **Vectors**: `Vec3d`, `Vec4f`, `Vec4d`, `Vec3u8`, `Vec3u16`
- **Quantized**: `Fp4`, `Fp8`, `Fp16`, `FpN` (read-only value access, no `setVoxel`)
- **Index/mask**: `ValueIndex`, `ValueOnIndex`, `PointIndex`, `ValueMask` (no scalar value; `getValue` returns index/state)

Each adds: `Grid` subclass, `ReadAccessor`, `NodeInfo` where applicable, `Sampler` specializations, and a `grid()`/`deviceGrid()` dispatch arm.

**Special handling**:
- Quantized types decode to `float` on read; the Python `Grid` subclass exposes `getValue()` returning float but does not bind `setVoxel`.
- Index types expose `getValue()` returning the `uint64` voxel index; the actual scalar value lives in a blind-data channel reachable via `getBlindData(n)`.
- `ValueMask` exposes only active-state queries.

## Phase 3 — Tree, Node, NodeManager, VoxelBlockManager

1. **`Grid.tree()`** returning `NanoTree<T>` with `root()`, `background()`, `nodeCount(level)`, `activeVoxelCount()`, `activeTileCount(level)`, `getFirstNode(LEVEL)`, `extrema()`.

2. **Node types** (one templated set per `BuildT`):
   - `NanoRoot<T>`: tile/child iterators (`cbeginChild`, `cbeginValueOn`), `background()`, `tileCount()`, stats (`minimum`/`maximum`/`average`/`stdDeviation`)
   - `NanoUpper<T>` (32³), `NanoLower<T>` (16³): origin, bbox, stats, child & value masks, iterators
   - `NanoLeaf<T>` (8³): origin, bbox, stats, active mask, **`values()` returning a zero-copy 512-element NumPy view**

3. **`NodeManager` + `NodeManagerHandle` + `createNodeManager()`** from [NodeManager.h](nanovdb/nanovdb/NodeManager.h):
   - `nm.leafCount()`, `lowerCount()`, `upperCount()`, `leaf(i)`, `lower(i)`, `upper(i)`
   - `isLinear()`, `memUsage()`

4. **`VoxelBlockManager`** (host build + decode) from [tools/VoxelBlockManager.h](nanovdb/nanovdb/tools/VoxelBlockManager.h): `VoxelBlockManagerHandle`, `buildVoxelBlockManager()`, host-side `decodeInverseMaps()`. Enables sequential active-voxel iteration over an `OnIndexGrid` without tree traversal.

5. **High-level bulk extractor**: `grid.leaf_values()` returning an `(N_leaves, 512)` zero-copy NumPy array. The single most useful operation for in-Python analytics and ML feature extraction on sparse volumes.

## Phase 4 — Build, Stats, and Validation

1. **`tools.build.Grid<T>`** for direct, voxel-by-voxel construction (currently only the bbox-functor path is exposed via `createFloatGrid`):
   - Constructor with `(background, name, gridClass)`
   - `setValue(ijk, v)`, `getValue(ijk)`, `setActive(ijk, on)`, `isActive(ijk)`, `background()`, `setTransform()`
   - `getAccessor()` / `getWriteAccessor()` for amortized cached access
   - `.to_nanovdb(sMode, cMode, buffer)` returning a `GridHandle`
   - Bind `ValueAccessor` and `WriteAccessor` proxy classes for cached batched writes

2. **Stats**: bind `tools::updateGridStats(grid, mode)` and `tools::getExtrema(grid, bbox)` from [tools/GridStats.h](nanovdb/nanovdb/tools/GridStats.h). Expose `Extrema<T>` and `Stats<T>` as read-only views with `min()`/`max()`/`avg()`/`var()`/`std()` accessors.

3. **Validation/checksum**: bind missing functions from [tools/GridValidator.h](nanovdb/nanovdb/tools/GridValidator.h) and [tools/GridChecksum.h](nanovdb/nanovdb/tools/GridChecksum.h):
   - `tools.validateGrid(handle, gridID, mode, verbose)` (single-grid; complements the existing `validateGrids`)
   - `tools.checkGrid(grid, mode)` → `(bool, error_str)`
   - `tools.isValid(grid, mode, verbose)`
   - `tools.evalChecksum(grid, mode)` (compute without writing)
   - `tools.validateChecksum(grid, mode)` (verify against stored)

## Phase 5 — Missing Primitives & Conversion

1. **Remaining `Create*` primitives** in [tools/CreatePrimitives.h](nanovdb/nanovdb/tools/CreatePrimitives.h) — only 4 of 12+ are bound today:
   - `createLevelSetBox`, `createLevelSetBBox` (wireframe), `createLevelSetOctahedron`
   - `createFogVolumeBox`, `createFogVolumeOctahedron`
   - `createPointSphere`, `createPointTorus`, `createPointBox`
   - `createPointScatter` (generic scatter on any level set)

2. **Quantized `createNanoGrid` overloads** from [tools/CreateNanoGrid.h](nanovdb/nanovdb/tools/CreateNanoGrid.h):
   - FpN with `AbsDiff` / `RelDiff` oracle classes (bind both)
   - Fp4/Fp8/Fp16 with `ditherOn`
   - Index/OnIndex with `channels` / `includeStats` / `includeTiles` options

3. **Generic `createNanoGrid` from build grids**: bind `tools::createNanoGrid<SrcGridT, DstBuildT>(src, sMode, cMode, ...)` so a `tools.build.Grid<float>` constructed in Python can be converted to any compatible `BuildT` (including quantized variants).

## Phase 6 — Documentation, Examples, Migration

1. **Auto-generated `.pyi` stubs** in the wheel (rolled in from Phase 0).

2. **Example scripts/notebooks**:
   - "Loading a `.nvdb` file and inspecting any grid type"
   - "Walking the tree with `NodeManager`"
   - "Building a grid in pure Python" (using `tools.build.Grid`)
   - "Bulk leaf-value access for analytics" (NumPy)
   - "Creating quantized grids" (Fp4/Fp8/Fp16/FpN)
   - "Round-tripping with OpenVDB"
   - "Validating and checksumming a grid"
   - "Working with PointGrid and blind data"

3. **API reference docs** generated from docstrings (Sphinx or mkdocs).

4. **Migration guide** for users on the old type-specific accessor API (`floatGrid()` → `grid()`).

---

## Suggested ordering / dependencies

```mermaid
graph LR
  P0[Phase 0: Foundation] --> P1[Phase 1: Polymorphic]
  P1 --> P2[Phase 2: Type coverage]
  P2 --> P3[Phase 3: Tree/nodes]
  P1 --> P4[Phase 4: Build/stats/validate]
  P0 --> P5[Phase 5: Primitives]
  P3 --> P6[Phase 6: Docs/examples]
  P4 --> P6
  P5 --> P6
```

Phase 0 unlocks everything else; Phase 1 unlocks Phases 2-4 (anything that returns a `Grid` benefits from the polymorphic accessor); Phase 2 unlocks Phase 3 (tree/node bindings must cover all `BuildT`); Phase 5 is independent of Phases 1-4 and can proceed in parallel; Phase 6 ties everything together.

## Risks and mitigations

- **NumPy buffer lifetime for zero-copy views**: nanobind's ndarray binding takes a `nb::keep_alive` on the source grid, but document the lifetime model clearly to avoid use-after-free on the consumer side.
- **Index/quantized grid semantics in Python**: these are confusing types in C++ too; lean on the existing C++ docstrings and add Python-specific explanatory text where the indirection through blind data or decoded values is non-obvious.

swahtz:
Fixes **issue #2208**: brings the NanoVDB Python bindings to parity with the C++ `nanovdb::*` surface. This is the merge of `feature/nanovdb_python` into `master` after nine sub-PRs (#2209–#2218) landed against the branch.

## Summary

Before this PR, Python exposed ~6 of 24 `GridType`s, a few callback grid factories, basic math, and a CUDA helper or two — no tree walking, no type-erased introspection, no mutable builder. After it, the module mirrors the full C++ surface: a single polymorphic grid accessor over all 24 grid types, walkable tree/node/leaf interiors, voxel-by-voxel construction in pure Python, the quantized (`Fp4`/`Fp8`/`Fp16`/`FpN`) and index (`ValueIndex`/`ValueOnIndex`) BuildTs, `GridMetaData` + blind-data (reading *and* authoring), `ChannelAccessor`, the per-grid stats/validation/checksum surface, the full primitive library, host `VoxelBlockManager`, plus shipped C headers and `.pyi` stubs.

Bindings are generated from a single X-macro BuildT list (`python/BuildTypes.def`), so a new BuildT is one line rather than a repeated `defineGrid<T>` block.

## What's now reachable

**Polymorphic grid access** — `handle.grid(n)` / `device_handle.deviceGrid(n)` return the correct typed subclass for whatever `gridType(n)` reports (`FloatGrid`, `Fp16Grid`, `OnIndexGrid`, …); `isinstance` dispatches. The old typed accessors are removed (see *Breaking changes*).

**Multi-grid handles** — `splitGrids(handle)` explodes a multi-grid handle into one handle per grid, `mergeGrids(handles)` combines a sequence into one buffer (without consuming the inputs), and `handle.copy()` deep-copies.

**18 additional BuildTs** — scalars (`bool`, `int16`, `int64`, `uint8`, `uint32`), vectors (`Vec3d`, `Vec4f`, `Vec4d`, `Vec3u8`, `Vec3u16`), quantized read-only (`Fp4`/`Fp8`/`Fp16`/`FpN`, decode to float), index/mask (`ValueIndex`/`ValueOnIndex`/`ValueMask`), and `PointIndex`. (`Half` stays unbound — upstream `class Half{}` is still a placeholder.)

**Introspection & data** — `GridMetaData(grid)` (768-byte header view, no compile-time BuildT); the blind-data API (`blindDataCount`, `blindMetaData`, `findBlindData[ForSemantic]`, zero-copy `getBlindData`); tree/node/leaf walking per BuildT (`tree()` → root/upper/lower/leaf with stats, masks, per-voxel `getValue`/`isActive`/`probeValue`); `createNodeManager(grid)`. Every grid also exposes the `worldToIndex`/`indexToWorld` transform family (`Dir`/`Grad` and single-precision `F` variants) alongside the `applyMap`/`applyJacobian`/`applyIJT` names.

**Index-grid workflows** — `IndexGrid`/`OnIndexGrid` expose `valueCount()` (and `PointGrid` exposes `pointCount()`), and `ChannelAccessor` closes the read loop for indexed channels: `createChannelAccessor(grid, channelID)` inspects the channel's recorded dataType and returns the matching typed accessor (`{Index,OnIndex}{Float,Double,Int32,Vec3f}ChannelAccessor`), so `acc(ijk)` reads channel values directly by coordinate.

**Blind-data authoring** — `tools.CreateNanoGrid(src)` mirrors the C++ converter class: declare channels with `addBlindData(name, count, dataType=…, dataSemantic=…, dataClass=…)`, bake with `getHandle()`, then fill the zero-initialized channel through the writable NumPy view from `grid.getBlindData(n)`:

```python
conv = nanovdb.tools.CreateNanoGrid(src_grid)
conv.addBlindData("uv", count=1000, dataType=nanovdb.GridType.Vec3f)
handle = conv.getHandle()
grid = handle.grid()
grid.getBlindData(grid.findBlindData("uv"))[:] = uv_array  # zero-copy write
```

**Bulk NumPy interop** — `grid.leaf_values()` returns a zero-copy `(N_leaves, 512)` view of every leaf's `mValues` (BuildTs that carry one); writes mutate the grid. Highest-bandwidth path for analytics/ML on sparse volumes.

**Mutable construction** — `tools.build.<Suffix>Grid` builds voxel-by-voxel in pure Python, with cached `getAccessor()`, thread-safe buffered `getWriteAccessor()`, and `.to_nanovdb(sMode=...)` to bake a host `NanoGrid`:

```python
g = nanovdb.tools.build.FloatGrid(0.0, "demo", nanovdb.GridClass.FogVolume)
g.setValue(nanovdb.math.Coord(1, 2, 3), 4.5)
handle = g.to_nanovdb(sMode=nanovdb.tools.StatsMode.All)
```

**Stats / validation / checksum** — per-BuildT `Extrema` / `Stats`, `updateGridStats`, `getExtrema`; single-grid `validateGrid` / `checkGrid` / `isValid`; `evalChecksum` / `validateChecksum`; `Checksum` exposes `isEmpty()` / `isHalf()` / `isFull()` / `mode()`.

**Primitives** — 9 added on top of the original 4: `createLevelSetBox`/`BBox`/`Octahedron`, `createFogVolumeBox`/`Octahedron`, `createPointSphere`/`Torus`/`Box`/`Scatter`.

**Quantization / conversion** — `createNanoGridFp4`/`Fp8`/`Fp16`, `createNanoGridFpN(src, oracle, ...)` with `AbsDiff`/`RelDiff` oracles, `createNanoGridIndex`/`OnIndex`. All accept either a `NanoGrid<T>` or a `tools.build.Grid<T>` source.

**Sampler derivatives** — the host trilinear sampler exposes `gradient(xyz)` and the trilinear/triquadratic samplers expose `zeroCrossing(xyz)`, gated to floating-point BuildTs exactly as the C++ templates are.

**`VoxelBlockManager` (host)** — `tools.buildVoxelBlockManager(grid, log2_block_width=6, ...)` with zero-copy `firstLeafID()`/`jumpMap()` views and `decodeBlock`, polymorphic over block width 64/128/256/512.

**Packaging** — wheel ships the C headers under `nanovdb/include/`, `nanovdb.get_include()` returns the path, and auto-generated `.pyi` stubs give Pylance/Pyright/mypy coverage.

**Docs/examples** — 14 runnable example scripts under `python/examples/`; 9 of them are Python ports of the host-capable C++ examples (`ex_write_nanovdb_grids`/`ex_read_nanovdb_sphere*` as one io round-trip, `ex_make_funny_nanovdb`, `ex_make_typed_grids`, the host paths of `ex_raytrace_level_set`/`ex_raytrace_fog_volume`/`ex_collide_level_set` re-expressed with the bound samplers, the host halves of `ex_index_grid_cuda`/`ex_nodemanager_cuda`, and `openvdb_interop`). All 14 are smoke-tested by a new ctest (`pytest_nanovdb_examples`, ~5 s) gated on `NANOVDB_BUILD_PYTHON_UNITTESTS`. Not ported: the pool-buffer examples (HostBuffer pool API unbound), the MagicaVoxel converter, device-only examples, and a dedicated PointAccessor example. Docstring sweep took the binding surface from ~374 undocumented sites to 0.

## Breaking changes for downstream Python users

No deprecation cycle — the restructure explicitly replaces the typed-grid accessors.

| Before | After |
| --- | --- |
| `handle.floatGrid(i)` (+ `double`/`int32`/`vec3f`/`rgba8`/`point`) | `handle.grid(i)` |
| `dh.deviceFloatGrid(i)` (+ typed device variants) | `dh.deviceGrid(i)` |
| `nanovdb.math.cuda.sampleFromVoxels(...)` | `nanovdb.tools.cuda.sampleFromVoxels(...)` (no C++ `math::cuda` namespace; `math.cuda` removed) |
| `if handle:` raised | `if handle:` ≡ `not handle.empty()` |
| `repr(GridType.Float)` → `'<GridType.Float: 1>'` | `'Float'` |

Find affected call sites with `\.\(float\|double\|int32\|vec3f\|rgba8\|point\)Grid\b` and `nanovdb\.math\.cuda`.

## Sub-PRs merged into `feature/nanovdb_python`

| Sub-PR | Landed |
| --- | --- |
| #2209 | Foundation: `BuildTypes.def` X-macro, ship headers + `get_include()`, fix `__bool__` / enum `__repr__`, relocate `sampleFromVoxels`, `.pyi` stubs. |
| #2210 | Polymorphic `handle.grid(n)`, `GridMetaData`, blind-data, `PointAccessor`. |
| #2211 | Broaden BuildT coverage to 24 grid types. |
| #2212 | Tree / nodes / NodeManager / `leaf_values`. |
| #2213 | `VoxelBlockManager` (host); Windows CI numpy step. |
| #2214 | `tools.build.Grid<T>` mutable builder + Value/Write accessors + `.to_nanovdb()`. |
| #2215 | Stats; single-grid validation; checksum eval/verify. |
| #2216 | 9 primitives; quantized `Fp*` conversion; `Index`/`OnIndex` conversion. |
| #2218 | Example scripts; full docstring sweep (374 → 0). |

## Test plan

- [x] `TestNanoVDB.py` grew 14 → 161 cases; green on a CUDA build except two pre-existing `BLOSC ... disabled` errors gated on `NANOVDB_USE_BLOSC=ON`.
- [x] CI green across linux/windows/macos/nanovdb-lite (CUDA + non-CUDA, OpenVDB on + off) on the latest sub-PR merges.
- [x] CPU-only / driverless runners gated via `isCudaAvailable()` + `isGpuAvailable()`.

Closes #2208.


nanovdb/nanovdb/python/PyGridHandle.h
```diff
@@ -7,62 +7,194 @@
...
+             "Return the n-th grid as a typed Grid subclass selected by "
+             "gridType(n), or None if the BuildT is not bound in Python. "
+             "The returned grid keeps this handle alive.")
+        .def("isPadded", &nanovdb::GridHandle<BufferT>::isPadded,
+            "True iff this handle's buffer is aligned past the natural "
+            "GridData alignment (used by the I/O code path).")
+        .def("gridCount", &nanovdb::GridHandle<BufferT>::gridCount,
+            "Number of grids stored in this handle.")
+        .def("gridSize", &nanovdb::GridHandle<BufferT>::gridSize, nb::arg("n") = 0,
+            "Byte size of the n-th grid (without padding).")
```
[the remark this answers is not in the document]
swahtz: Fixed in 88a4f905 — `gridSize` is now a bounds-checked lambda that raises `IndexError` when `n >= gridCount()` (covering the empty-handle case) before touching the unchecked `mMetaData[n]`.

nanovdb/nanovdb/python/PyGridHandle.h
```diff
@@ -7,62 +7,194 @@
...
+        .def("gridType", &nanovdb::GridHandle<BufferT>::gridType, nb::arg("n") = 0,
+            "GridType enumerator of the n-th grid (e.g. GridType.Float). "
+            "Cheap to query — does not require materializing the grid.")
```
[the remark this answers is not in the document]
swahtz: Fixed in 88a4f905 — `gridType` now raises `IndexError` for out-of-range `n` (including any `n` on an empty handle) before indexing `mMetaData[n]`.

nanovdb/nanovdb/python/PyGridHandle.h
```diff
@@ -7,62 +7,194 @@
...
         .def(
             "gridData",
             [](nanovdb::GridHandle<BufferT>& handle, uint32_t n) { return nb::bytes(handle.gridData(n), handle.gridSize(n)); },
             nb::arg("n") = 0,
-            nb::rv_policy::reference_internal)
-        .def("write", nb::overload_cast<const std::string&>(&nanovdb::GridHandle<BufferT>::write, nb::const_), nb::arg("fileName"))
-        .def("write", nb::overload_cast<const std::string&, uint32_t>(&nanovdb::GridHandle<BufferT>::write, nb::const_), nb::arg("fileName"), nb::arg("n"))
+            nb::rv_policy::reference_internal,
+            "Raw byte contents of the n-th grid as a Python bytes object. "
```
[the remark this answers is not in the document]
swahtz: Fixed in 88a4f905 — the `gridData` lambda now bounds-checks `n` and raises `IndexError` up front. As you noted, `gridData(n)` itself returns nullptr for bad `n`, but the adjacent `gridSize(n)` is the call that read `mMetaData[n]` out of bounds; the guard covers both.

nanovdb/nanovdb/python/PyVoxelBlockManager.cc
```diff
@@ -0,0 +1,534 @@
...
+        "src_grid"_a,
+        "channels"_a = 0u,
+        "include_stats"_a = true,
+        "include_tiles"_a = true,
+        "verbose"_a = 0,
+        "Convert a source grid into a NanoGrid<ValueOnIndex> "
+        "(OnIndexGrid). Accepts FloatGrid / DoubleGrid / Int32Grid / "
+        "Vec3fGrid. Required for constructing inputs to "
+        "buildVoxelBlockManager. The broader createNanoGrid<SrcGridT, "
+        "DstBuildT> surface lands in a later phase.");
```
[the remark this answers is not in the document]
swahtz: Fixed in 88a4f905 — the docstring no longer references a future phase. It now notes this is a narrow helper kept alongside `buildVoxelBlockManager` and points users to the canonical `nanovdb.tools.createNanoGridOnIndex` for general index conversion.

nanovdb/nanovdb/python/PyGridHandle.h
```diff
@@ -7,62 +7,218 @@
...
+            for (uint32_t n = 0; n < h->gridCount(); ++n) {
+                const uint64_t gs = h->gridSize(n);
+                std::memcpy(dst, src, gs);
+                auto* gd = reinterpret_cast<nanovdb::GridData*>(dst);
+                nanovdb::tools::updateGridCount(gd, writeIndex++, totalGrids);
+                dst += gs;
+                src += gs;
+            }
+        }
+        return HandleT(std::move(buffer));
```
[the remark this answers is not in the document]
swahtz: Fixed in 7a1e6639.

**Empty input** — valid, fixed. `mergeGrids([])` (or only-empty handles) hit `BufferT::create(0)`, which for `HostBuffer` returns a non-null `data()` over a zero-byte region; the `GridHandle(buffer&&)` ctor then reads a full `GridData` header out of it (heap-overflow read → "invalid host buffer" throw). Now returns an empty handle up front when `totalGrids == 0`.

**Per-grid pointer** — adopted your suggestion (copy from `h->gridData(n)`), though for the record the contiguity assumption was actually safe here: `cpyGridHandleMeta` builds `mMetaData[n].offset` as a running sum of `mGridSize`, so grids inside a handle are always tightly packed and `isPadded()` only ever reports *trailing* bytes after the last grid (never gaps between grids) — which is also why the canonical `nanovdb::mergeGrids` advances by `gridSize`. That said, `gridData(n)` is the authoritative accessor and drops the manual pointer arithmetic, so the binding no longer bakes in that invariant.

nanovdb/nanovdb/python/PyVoxelBlockManager.cc
```diff
@@ -0,0 +1,535 @@
...
+                        firstLeafID,
+                        self.handle.hostJumpMap() + block_index * JumpMapLength,
+                        blockFirstOffset);
+                });
+            },
+            "grid"_a, "block_index"_a,
+            "Decode the inverse maps for the block_index-th block of this "
+            "VBM. Returns (leaf_index, voxel_offset) uint32 / uint16 NumPy "
+            "arrays of length BlockWidth = 1<<log2_block_width, using the "
+            "log2_block_width the handle was built with.");
```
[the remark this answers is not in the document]
swahtz: Fixed in 4d923768. `decodeBlock` delegates to `pyDecodeInverseMapsImpl`, which now scopes an `nb::gil_scoped_release` around the `VBM::decodeInverseMaps` compute kernel. I used a scoped release rather than a blanket `nb::call_guard<nb::gil_scoped_release>()`: this entry point casts the grid (`castOnIndexGrid` → isinstance/cast) and allocates the NumPy result arrays inside the body, so dropping the GIL across the whole call would touch the Python C API without it. The scoped release covers exactly the GIL-free kernel, matching the `tryIndexify`/`tryQuantizeFpX` pattern in PyCreateNanoGrid.cc.

nanovdb/nanovdb/python/PyVoxelBlockManager.cc
```diff
@@ -0,0 +1,535 @@
...
+        "first_offset"_a = 0,
+        "last_offset"_a = 0,
+        "n_blocks"_a = 0,
+        "Build a host-side VoxelBlockManager from an OnIndexGrid. "
+        "log2_block_width selects the per-block active-voxel count "
+        "(6=64, 7=128, 8=256, 9=512). Pass 0 for first_offset / "
+        "last_offset / n_blocks to use the full grid (first active "
+        "voxel through grid.activeVoxelCount(), minimum block count). "
+        "first_offset, if nonzero, must satisfy first_offset == 1 "
+        "(mod BlockWidth).");
```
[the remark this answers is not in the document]
swahtz: Fixed in 4d923768 — scoped an `nb::gil_scoped_release` around the `buildVoxelBlockManager<LBW>` call (the part that may parallelize via util::forEach). Used a scoped release rather than a blanket `call_guard` because the body first calls `castOnIndexGrid` (isinstance/cast) and throws Python `value_error`s for invalid args, which must run under the GIL; the release covers only the pure-C++ build.

nanovdb/nanovdb/python/PyVoxelBlockManager.cc
```diff
@@ -0,0 +1,535 @@
...
+        },
+        "grid"_a,
+        "first_leaf_id"_a,
+        "jump_map"_a,
+        "block_first_offset"_a,
+        "log2_block_width"_a = 6,
+        "Decode the inverse maps for a single voxel block of an OnIndexGrid. "
+        "Returns a (leaf_index, voxel_offset) tuple of fresh NumPy arrays of "
+        "length BlockWidth = 1<<log2_block_width. jump_map must have length "
+        "BlockWidth/64. first_leaf_id must be in [0, grid.tree().nodeCount(0)).");
```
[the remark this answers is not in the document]
swahtz: Fixed in 4d923768 via the shared `pyDecodeInverseMapsImpl` helper, which now releases the GIL around the `VBM::decodeInverseMaps` kernel and re-acquires it before constructing the NumPy arrays. Scoped release rather than blanket `call_guard` for the same reason as the sibling comments — the body casts the grid, reads the `jump_map` ndarray, and allocates the output arrays, all of which need the GIL.

kmuseth (changes requested):
I'm about to review this PR, but several of the CI tests are failing. Please fix those :)

swahtz:
> I'm about to review this PR, but several of the CI tests are failing. Please fix those :)

Seems that a change on `master` that was misaligned to the new python changes broke those tests.  One line fix, they should all pass now.

kmuseth:
This is a massive PR on restructuring of our Python bindings - something that's outside of my comfort zone. Can @apradhana  or @matthewdcong please take a look at this PR and give their feedback?

nanovdb/nanovdb/python/PyBuildGrid.cc
```diff
@@ -0,0 +1,239 @@
...
+            nb::keep_alive<0, 1>(),
+            "Return a WriteAccessor for thread-safe writes; the accessor "
+            "buffers changes into a private root and merges them into the "
+            "parent grid on destruction (or on an explicit merge() call). "
+            "Held by nanobind on the heap so the accessor's internal "
+            "references stay valid.")
+        // Baking a large grid is the most expensive operation on this
+        // class; release the GIL so other Python threads can run during
+        // the conversion (the lambda only touches C++ state).
+        .def("to_nanovdb",
```
matthewdcong: Should this be `toNanoVDB` to match conventions?

nanovdb/nanovdb/python/PyCreateNanoGrid.cc
```diff
@@ -35,20 +39,553 @@ GridHandle<HostBuffer> createNanoGridFromFunc(const BuildT&
...
+        const auto& src = nb::cast<const BuildSrcT&>(py_src);
+        GridHandle<HostBuffer> handle;
+        {
+            nb::gil_scoped_release release;
+            handle = tools::createNanoGrid<BuildSrcT, DstBuildT, HostBuffer>(
+                src, sMode, cMode, ditherOn, verbose);
+        }
+        return nb::cast(std::move(handle));
+    }
+    return nb::object();
```
matthewdcong, line 111: Should this be `nb::none()` instead? Same below in `tryQuantizeFpN`, etc.

nanovdb/nanovdb/python/PyGridHandle.h
```diff
@@ -7,62 +7,226 @@
...
+    }
+}
+
+// Free functions splitGrids / mergeGrids exposed at module scope. Templated
+// on BufferT so the same definitions work for both the host GridHandle and
+// the device GridHandle bindings.
+template<typename BufferT> void defineGridHandleUtilities(nb::module_& m)
+{
+    using HandleT = nanovdb::GridHandle<BufferT>;
+    m.def("splitGrids", [](const HandleT& handle) {
```
matthewdcong, line 65: Is there a way to use the already defined C++ versions of `mergeGrids` and `splitGrids`? We spent a while making that right for a lot of corner cases so I'm not keen on creating another version here.

nanovdb/nanovdb/python/PyIO.cc
```diff
@@ -119,38 +151,46 @@ void defineDeviceReadWriteGrid(nb::module_& m)
...
         .value("ZIP", io::Codec::ZIP)
         .value("BLOSC", io::Codec::BLOSC)
-        .export_values();
-        // .def("__repr__", [](const io::Codec& codec) {
-        //     char str[strlen<io::Codec>()];
-        //     toStr(str, codec);
-        //     return std::string(str);
-        // });
+        .export_values()
+        .def("__repr__", [](const io::Codec& codec) {
```
matthewdcong, line 189: Haha glad that works now. 🙂  It wasn't available in the early versions of nanobind.

matthewdcong (commented):
Overall, this look pretty good. Just a few comments.

nanovdb/nanovdb/python/PyBuildGrid.cc
swahtz: Absolutely, that's an oversight and maybe a stream-cross with a function in fvdb called `to_nanovdb`.  I've renamed that.

nanovdb/nanovdb/python/PyCreateNanoGrid.cc
swahtz: So this is intentional. The default-constructed (invalid) `nb::object` is a sentinel meaning "the source grid didn't match this `SrcBuildT`", and the callers test it with `r.is_valid()` before falling through to try the next source type. `nb::none()` is going to wrap a real `Py_None`, so `is_valid()` would return true and the dispatch stops at the first `SrcBuildT` and return None to Python instead of trying the remaining types. This doesn't create ambiguity with the success because it always returns a cast GridHandle, never None. I added a one-line comment at each of these return sites to make the sentinel pattern explicit.

nanovdb/nanovdb/python/PyGridHandle.h
swahtz: Yes, good point, I don't want two copies of this logic either. Two aspects:

`splitGrids` already calls `nanovdb::splitGrids` directly; the binding just repackages the returned std::vector as a Python list.

`mergeGrids` was reimplemented because the existing overload takes a `const std::vector<GridHandle>&` by value-of-element, and `GridHandle` is move-only — building that vector from Python-owned wrappers would either move from (and silently empty) the caller's Python objects, or require deep-copying every input buffer just to read them. Rather than keep the duplicate logic going forward, I've now added a non-owning overload to `GridHandle.h` if you want to have a look and let me know what you think — `mergeGrids(const GridHandle<BufferT>* const* handles, size_t count, pool)` — moved the merge logic there, and made the vector overload forward to it. The Python binding now just collects pointers and calls it, so there's a single implementation again.

While consolidating `mergeGrids`, I folded in two small fixes the binding had rolled in: empty input now returns an empty handle instead of constructing from `BufferT::create(0)`, and the source walk uses `gridData(n)` (which applies the per-grid offset) rather than assuming grids are contiguous from `data()`, so padded handles merge correctly. Existing callers of the vector overload are unaffected.

matthewdcong: Looks good, thanks for adding that

nanovdb/nanovdb/python/PyCreateNanoGrid.cc
matthewdcong: Ah ok, gotcha. Thanks for adding the comment.

matthewdcong:
Cool, looks good to me after the changes.

matthewdcong (approved):

kmuseth (approved):
after fixing the failed test and addressing Matts comments I approve

PR #2220, opened 2026-05-27
NanoVDB `ReadAccessor::getDimAndActive`

swahtz:
## Summary

`ReadAccessor::getDimAndActive` fuses NanoVDB's `getDim` and `isActive` into a single root-to-leaf descent. For callers that read active unconditionally — fvdb-core's `HDDASegmentIterator` and `HDDAVoxelIterator` are the applicable consumers — this collapses two tree walks per iteration into one.

Measured on a Blackwell RTX 6000 Pro GPU across four NanoVDB example grids (dragon SDF, crawler SDF, emu SDF, WDAS cloud fog volume; 1024×1024 ray casts, median of 50 iters):

#### HDDASegmentIterator

| Scene | Unfused | Fused | Speedup |
|---|---|---|---|
| dragon | 0.972 ms | 0.536 ms | **1.81×** |
| crawler | 4.064 ms | 2.953 ms | **1.38×** |
| emu | 1.879 ms | 1.055 ms | **1.78×** |
| wdas_cloud | 6.493 ms | 4.661 ms | **1.39×** |

#### HDDAVoxelIterator

| Scene | Unfused | Fused | Speedup |
|---|---|---|---|
| dragon | 2.645 ms | 1.235 ms | **2.14×** |
| crawler | 8.016 ms | 5.252 ms | **1.53×** |
| emu | 5.731 ms | 2.194 ms | **2.61×** |
| wdas_cloud | 4.174 ms | 2.508 ms | **1.66×** |

Looking at ncu metrics, numbers drop in parallel to wall time improvements: fusion drops dynamic instructions by 26–47% and long-scoreboard stall by 8–9 percentage points on dragon. Register pressure drops 4–6 regs vs unfused (72 → 66–68) — no occupancy regression. The packed `DimAndActive` encoding (uint32 storage) keeps the return boundary at one register.

## Details on Trying to Use the Fused API in `HDDA::ZeroCrossing`

The fused API is intentionally **not** used by `HDDA::ZeroCrossing`. That kernel was benchmarked with the fused call across two register configurations on my Blackwell RTX 6000 Pro GPU — once on master (the existing kernel sits at exactly 64 regs/thread, on the occupancy cliff) and once on top of a separate refactor I did I'll call 'Lean HDDA' which was to free up register headroom (for additional registers needed by the dim/active fusion) by removing the `mT1, mDelta, mStep` members from the `HDDA` class and deriving them from `Ray` components on each call in order to reduce register pressure (63 regs/thread, with 3 regs of headroom).

On master (without 'lean HDDA'), active/dim fusion grew the kernel to 66–67 regs and dropped occupancy 50% → 37.5% (2 → 1 blocks/SM at 256 threads). On 'lean HDDA', the fusion fits under the cliff (61 regs with the packed encoding) without an occupancy hit (so we could measure fusion without an occupancy regression).

Wall time, however, is unchanged in both regimes — variants land within 0–1.3% of each other across sphere / dragon / crawler / emu, with the fused versions consistently a hair slower. The reason isn't the register/occupancy cliff: it's that `ZeroCrossing`'s `if (hdda.dim() > 1 || !acc.isActive(ijk)) continue;` makes the active read dead at upper levels, and the L1 cache absorbs the second descent at ~87% hit rate on the leaf-level remainder. Eliminating it saves essentially no cycles, while the fused call's slightly heavier per-call overhead nets out marginally worse.

The fused API is retained because it's a substantial win for callers that read active unconditionally — fvdb's HDDA iterators see −27% to −62% wall time and −26% to −47% dynamic instructions. `ZeroCrossing`'s pattern just isn't one of them.


nanovdb/nanovdb/NanoVDB.h
```diff
@@ -1787,6 +1787,56 @@ struct ProbeValue;
...
+{
+    int32_t packed;
+
+    __hostdev__ DimAndActive() = default;
+    __hostdev__ DimAndActive(uint32_t d, bool a)
+        : packed(int32_t((d & 0x7FFFFFFFu) | (uint32_t(a) << 31))) {}
+
+    __hostdev__ uint32_t dim()    const { return uint32_t(packed) & 0x7FFFFFFFu; }
+    __hostdev__ bool     active() const { return packed < 0; }
+};
```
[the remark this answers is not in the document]
swahtz: Good catch — fixed in cdebf065. The packed field is now `uint32_t packed{0u}` instead of `int32_t`, so there is no longer any unsigned→signed conversion (the pack and both decoders operate entirely in `uint32_t`), and the defaulted constructor now zero-initializes. Size/ABI is unchanged (still one 32-bit word, one register at the return boundary).

nanovdb/nanovdb/examples/ex_raytrace_level_set/nanovdb.cu
```diff
@@ -18,6 +20,32 @@ using BufferT = nanovdb::HostBuffer;
...
+    float sum = 0;
+    for (float s : samples) sum += s;
+    const float meanMs = sum / float(samples.size());
+    std::cout << label
+              << " min=" << minMs << " ms"
+              << "  median=" << medianMs << " ms"
+              << "  mean=" << meanMs << " ms"
+              << "  max=" << maxMs << " ms"
+              << "  (n=" << samples.size() << ")" << std::endl;
+}
```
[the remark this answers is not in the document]
swahtz: Addressed in cdebf065. `reportStats` now takes `const std::vector<float>&`, early-returns on an empty vector (so `front()`/`back()` are never called on empty input), and sorts a local copy instead of reordering the caller's data.

nanovdb/nanovdb/examples/ex_raytrace_level_set/common.h
```diff
@@ -44,11 +44,41 @@ inline __hostdev__ void mortonEncode(uint32_t& code, uint32_t x, uint32_t y)
...
+            });
+
+        NANOVDB_CUDA_CHECK_ERROR(cudaEventRecord(stopEv, 0), __FILE__, __LINE__);
+        NANOVDB_CUDA_CHECK_ERROR(cudaEventSynchronize(stopEv), __FILE__, __LINE__);
+
+        float ms = 0.f;
+        NANOVDB_CUDA_CHECK_ERROR(cudaEventElapsedTime(&ms, startEv, stopEv), __FILE__, __LINE__);
+
+        cudaEventDestroy(startEv);
+        cudaEventDestroy(stopEv);
```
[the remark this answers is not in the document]
swahtz: Addressed in cdebf065. Both `cudaEventDestroy` calls are now wrapped in `NANOVDB_CUDA_CHECK_ERROR`, consistent with the create/record/synchronize calls in the same timing block.

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -1787,6 +1787,55 @@ struct ProbeValue;
...
+/// live across other work. Aggregate-initialization style `return {dim,
+/// active}` is preserved via the explicit two-argument constructor; accessors
+/// @c dim() and @c active() decode on read.
+struct DimAndActive
+{
+    uint32_t packed{0u};
+
+    __hostdev__ DimAndActive() = default;
+    __hostdev__ DimAndActive(uint32_t d, bool a)
+        : packed((d & 0x7FFFFFFFu) | (uint32_t(a) << 31)) {}
```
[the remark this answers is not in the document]
swahtz: Added in 1d725586. `NANOVDB_ASSERT(d < (uint32_t(1) << 31))` now documents and checks the invariant before packing. It guards a pathological tree config (`TOTAL >= 31`); the default hierarchy tops out at dim 4096, so it never fires in practice, but it makes the truncation non-silent in debug builds. Release builds are unaffected — `NANOVDB_ASSERT` maps to `assert`, which compiles out under `NDEBUG` (host and device), and the 31-bit mask is kept so the `active` bit stays intact even when the assert is gone.

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -4572,6 +4710,30 @@ class LeafNode : public LeafData<BuildT, CoordT, MaskT, Log2Dim>
         return ChildNodeType::dim();
     }
 
+    /// @brief Fused variant of getDimAndCache that additionally reports the
```
kmuseth: This comment seems excessively detailed and verbose. Please shorten it.

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -3632,6 +3709,67 @@ class InternalNode : public InternalData<ChildT, Log2Dim>
...
+    ///   - @c ActiveOnLeafOnly: cheap-path variant. When the skip-flag is
+    ///     set we short-circuit to {this->dim(), false} just like a
+    ///     standalone getDimAndCache would (no child read, no leaf
+    ///     descent). The returned @c active is unspecified in that case
+    ///     and the caller is contractually required to gate on @c dim
+    ///     before reading @c active. This matches the pattern used by
+    ///     ZeroCrossing-style code: `if (hdda.dim() > 1 || !active)
+    ///     continue;`. When the skip-flag is not set the behaviour is
+    ///     identical to ActiveExact -- we still need the descent to
+    ///     compute the leaf-level active bit.
```
kmuseth: This block of documentation seems excessively detailed and verbose. Please shorten it.

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -1808,6 +1808,58 @@ struct ProbeValue;
...
+///
+/// Storage is packed into a single uint32_t: low 31 bits hold @c dim
+/// (NanoVDB's max getDim return value is 4096 for the upper internal node
+/// in the default FloatTree hierarchy, well below the 2^31 limit), and the
+/// top bit (bit 31) encodes @c active (set = active). This packing saves one
+/// register at the function-return boundary vs. the otherwise natural
+/// {uint32_t, bool} layout, which matters for callers that hold the result
+/// live across other work. Aggregate-initialization style `return {dim,
+/// active}` is preserved via the explicit two-argument constructor; accessors
+/// @c dim() and @c active() decode on read.
```
kmuseth: please shorten

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -1808,6 +1808,58 @@ struct ProbeValue;
...
+///        the returned @c dim corresponds to a true leaf-level descent --
+///        i.e. when @c dim is no larger than the dim() of the deepest level
+///        that was actually visited. When the descent short-circuits via
+///        the InternalNode skip-flag, @c active is unspecified (the caller
+///        MUST gate on @c dim before reading @c active). This recovers the
+///        cheap skip-flag fast path of standalone getDim while still fusing
+///        the leaf-level isActive read into the same descent on the rays
+///        that actually need it. Use this when the caller's logic is of the
+///        form `if (hdda.dim() > 1 || !active) continue;` -- the dim>1
+///        short-circuit makes the unspecified active value dead.
```
kmuseth: please shorten

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -3136,6 +3188,31 @@ class RootNode : public RootData<ChildT>
...
+    /// active state of @a ijk in the same traversal. Shares the Root→Internal
+    /// →Leaf descent, the accessor cache update, and (at the terminating
+    /// level) the ValueMask / tile-state read between the two queries.
+    ///
+    /// The @c Policy tag controls how @c active is computed when an
+    /// InternalNode along the descent has its skip-flag set; see
+    /// ActiveExact and ActiveOnLeafOnly for the two contracts.
+    /// At the root level the skip-flag does not apply, so both policies
+    /// produce identical results here -- @c Policy is only forwarded to the
+    /// child descent.
```
kmuseth: shorten please

kmuseth (changes requested):
Please address the inlined comments (related to very verbose documentation)

Other than that it looks good

Here's Calude's summary:

What it does: adds ReadAccessor::getDimAndActive(ijk, ray), fusing the HDDA step-size query (getDim) and the active-state query (isActive) into one root→leaf descent, returned as a bit-packed DimAndActive (dim in low 31 bits, active in bit 31 — keeps the return in one register). Two policy tags control behavior when an InternalNode's skip-flag is set: ActiveExact (byte-for-byte matches the separate getDim+isActive calls, at the cost of still descending past skip-flagged nodes to get the real active bit) and ActiveOnLeafOnly (short-circuits like plain getDim and leaves active contractually unspecified — for callers whose logic already discards active when dim > 1). Purely additive: no existing method's signature changes.

Verification performed:

Hand-traced getDimAndActiveAndCache at all three node levels (Root, Internal, Leaf) against the existing getDimAndCache + getAndCache<GetState> it replaces, confirming algebraically that ActiveExact reproduces the exact original semantics (including the subtlety that the skip-flag short-circuits dim but must not short-circuit the active descent, since skip is set from value-range vs. render-delta, not from activeness) and that ActiveOnLeafOnly correctly mirrors getDim's cheap early-return when it takes that path.
Checked the DimAndActive packing for UB — none; the shift/mask are well-defined for uint32_t, and the constructor asserts dim < 2^31 (max real dim() is 4096).
Wrote a standalone reproduction of the PR's own GetDimAndActive unit test (not just read it) plus an additional dense cross-check test, compiled them against the PR's headers, and ran them on host: ~5.6M coordinate probes across both policies against directly-computed getDim/isActive, 0 mismatches, including the forced-skip-flag scenario where ActiveExact correctly reports the true active state while ActiveOnLeafOnly correctly reports the unspecified/false value — confirming the one place these two policies are supposed to diverge actually does.
Compiled the device-code path (__global__ kernel calling getDimAndActive under both policies) to real SASS for sm_89 — clean, no errors.
Reviewed the ex_raytrace_level_set example/benchmark-infrastructure changes (warmup iterations, min/median/mean/max reporting, GPU event-based timing) — these back the PR's performance claims but don't call the new API themselves (consistent with the PR body: ZeroCrossing deliberately doesn't adopt it, since register-pressure/occupancy experiments showed no win there). Low risk either way since it's example code, not library surface.
Confirmed the pendingchanges entry accurately and concisely describes the change.
Findings: none. The implementation is a faithful, carefully-reasoned fusion of the existing traversal logic, the two-policy design honestly documents (and the code honestly implements) where the fast path trades away exactness, and my independent testing corroborates the correctness claims rather than just the PR's own numbers.

One minor observation, not a defect: at the LeafNode level, ActiveOnLeafOnly doesn't actually take a cheaper path when the leaf's own skip-flag is set — it always reads mValueMask.isOn(n) regardless of policy. This is explicitly documented as a deliberate tradeoff (a leaf-level branch on Policy was measured to cost 2 registers and cross an occupancy cliff, while leaf skip-flags are rare in practice) and stays within the "unspecified" contract (a correct value is a valid value for "unspecified"), so it's a documented engineering choice rather than a bug.

nanovdb/nanovdb/NanoVDB.h
swahtz: Done

nanovdb/nanovdb/NanoVDB.h
swahtz: Done

swahtz:
> Please address the inlined comments (related to very verbose documentation)
> 
> Other than that it looks good
> 
> Here's Calude's summary:
> 
> What it does: adds ReadAccessor::getDimAndActive(ijk, ray), fusing the HDDA step-size query (getDim) and the active-state query (isActive) into one root→leaf descent, returned as a bit-packed DimAndActive (dim in low 31 bits, active in bit 31 — keeps the return in one register). Two policy tags control behavior when an InternalNode's skip-flag is set: ActiveExact (byte-for-byte matches the separate getDim+isActive calls, at the cost of still descending past skip-flagged nodes to get the real active bit) and ActiveOnLeafOnly (short-circuits like plain getDim and leaves active contractually unspecified — for callers whose logic already discards active when dim > 1). Purely additive: no existing method's signature changes.
> 
> Verification performed:
> 
> Hand-traced getDimAndActiveAndCache at all three node levels (Root, Internal, Leaf) against the existing getDimAndCache + getAndCache it replaces, confirming algebraically that ActiveExact reproduces the exact original semantics (including the subtlety that the skip-flag short-circuits dim but must not short-circuit the active descent, since skip is set from value-range vs. render-delta, not from activeness) and that ActiveOnLeafOnly correctly mirrors getDim's cheap early-return when it takes that path. Checked the DimAndActive packing for UB — none; the shift/mask are well-defined for uint32_t, and the constructor asserts dim < 2^31 (max real dim() is 4096). Wrote a standalone reproduction of the PR's own GetDimAndActive unit test (not just read it) plus an additional dense cross-check test, compiled them against the PR's headers, and ran them on host: ~5.6M coordinate probes across both policies against directly-computed getDim/isActive, 0 mismatches, including the forced-skip-flag scenario where ActiveExact correctly reports the true active state while ActiveOnLeafOnly correctly reports the unspecified/false value — confirming the one place these two policies are supposed to diverge actually does. Compiled the device-code path (**global** kernel calling getDimAndActive under both policies) to real SASS for sm_89 — clean, no errors. Reviewed the ex_raytrace_level_set example/benchmark-infrastructure changes (warmup iterations, min/median/mean/max reporting, GPU event-based timing) — these back the PR's performance claims but don't call the new API themselves (consistent with the PR body: ZeroCrossing deliberately doesn't adopt it, since register-pressure/occupancy experiments showed no win there). Low risk either way since it's example code, not library surface. Confirmed the pendingchanges entry accurately and concisely describes the change. Findings: none. The implementation is a faithful, carefully-reasoned fusion of the existing traversal logic, the two-policy design honestly documents (and the code honestly implements) where the fast path trades away exactness, and my independent testing corroborates the correctness claims rather than just the PR's own numbers.
> 
> One minor observation, not a defect: at the LeafNode level, ActiveOnLeafOnly doesn't actually take a cheaper path when the leaf's own skip-flag is set — it always reads mValueMask.isOn(n) regardless of policy. This is explicitly documented as a deliberate tradeoff (a leaf-level branch on Policy was measured to cost 2 registers and cross an occupancy cliff, while leaf skip-flags are rare in practice) and stays within the "unspecified" contract (a correct value is a valid value for "unspecified"), so it's a documented engineering choice rather than a bug.

Thanks @kmuseth, I addressed the long comments and reduced the total comment lines by ~60.

kmuseth (approved):
thanks for addressing my comments

PR #2221, opened 2026-05-27
Skip build CI if a PR is a draft.

apradhana:
Per discussion with @danrbailey , we may want to only run the CIs when a PR is NOT a draft PR. The logic is run CI if it is `push`, `non-draft PR`, or `manual run`.


danrbailey (approved):
Thanks @apradhana!

PR #2223, opened 2026-05-28
Remove Boost dependency from OPENVDB_USE_DELAYED_LOADING

dgovil:
### Motivation

  When `OPENVDB_USE_DELAYED_LOADING` is enabled (the default), the core OpenVDB
  library requires `Boost.Interprocess` and `Boost.Iostreams` as link-time
  dependencies. This is the only remaining Boost dependency in the core
  `openvdb/openvdb` library, and it covers functionality that is straightforwardly
  available through OS-native APIs and the C++ standard library.

  Removing it reduces the dependency surface for all downstream consumers of
  OpenVDB like OpenUSD, simplifies builds where Boost is not otherwise present, and brings the
  library closer to having no Boost dependency at all in its core.

  ---

  ### Changes

  #### `io/Archive.cc` — `MappedFile::Impl`

  Replaced `boost::interprocess::file_mapping` + `mapped_region` with a minimal
  cross-platform mmap wrapper:
  - **POSIX:** `open()` / `mmap()` / `munmap()` / `close()`
  - **Windows:** `CreateFileMapping()` / `MapViewOfFile()` / `UnmapViewOfFile()`

  Replaced `boost::iostreams::stream_buffer<array_source>` with `ArrayStreambuf`,
  a small `std::streambuf` subclass backed by a memory range. Crucially this
  overrides `seekoff`/`seekpos` since leaf node loading and point stream
  decompression both seek into the mapped buffer (`LeafBuffer.h`,
  `StreamCompression.cc`).

  #### `io/TempFile.cc`

  Replaced `boost::iostreams::file_descriptor_sink` (Unix path) with
  `std::filebuf`. `mkstemp()` is still used to guarantee a unique path; the fd it
  returns is closed immediately and the file is reopened via `std::filebuf`. The
  Windows path already used `std::filebuf` so both platforms now share the same
  implementation type.

  #### `io/Stream.cc` / `io/File.cc`

  Replaced `boost::iostreams::copy(src, dst)` with `dst << src.rdbuf()`.

  In `Stream.cc`, the `TempFile` is now explicitly closed before the path is
  handed to `MappedFile`. This is necessary because `std::filebuf` has user-space
  buffering — unlike the old `file_descriptor_sink` which wrote directly through
  the fd — so the file must be flushed before `mmap()` is called or the mapping
  may see incomplete data.

  #### `CMakeLists.txt` / `cmake/FindOpenVDB.cmake`

  - Removed `find_package(Boost COMPONENTS iostreams)` from the core library build
  - Removed `Boost::iostreams` from `OPENVDB_CORE_DEPENDENT_LIBS` and from
    downstream visible dependencies in `FindOpenVDB.cmake`
  - Removed the now-dead `OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING` CMake option
    and its associated Windows link block
  - Removed the `boost_iostreams` prerequisite heuristic used to auto-detect
    delayed loading in installed builds (detection still works via the version
    header)

  #### `unittest/TestAttributeArray.cc` / `unittest/TestStreamCompression.cc`

  Removed stale Boost includes that were copied from the implementation files but
  never actually used in the test bodies.

  ---

  ### New test

  `TestFile.testMappedFileSeek` directly exercises `MappedFile::createBuffer()`
  and verifies that the returned `streambuf` supports seeking from the beginning,
  current position, and end — and that two independent buffers from the same
  mapping maintain independent read positions. Finishes with a full delayed-load
  round-trip to confirm voxel values survive the path end-to-end.

  ---

  ### Testing

  Built and tested on macOS (Apple Silicon) with TBB 2020.3 and Blosc 1.21.5.
  All 738 unit tests pass.

I do not have a Windows or Linux box to test on, so I'm hoping the CI can catch issues on those platforms, but would love some eyes from people who primarily use those platforms to test this out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,8 +160,6 @@ preceeding flags. If you need more granular control you can provide your
 desired flags via CXX_FLAGS. Only applicable which compiling for x86 targets.
 ]=])
 
-cmake_dependent_option(OPENVDB_DISABLE_BOOST_IMPLICIT_LINKING
-  "Disable the implicit linking of Boost libraries on Windows" ON "WIN32" OFF)
 option(USE_STATIC_DEPENDENCIES [=[
 Only search for and use static libraries. If OFF the shared versions of requried libraries are prioritised, falling
 back to static libraries. Forcing individual static dependencies can be enabled by setting XXX_USE_STATIC_LIBS
--- a/cmake/FindOpenVDB.cmake
+++ b/cmake/FindOpenVDB.cmake
@@ -603,10 +603,6 @@ elseif(NOT OPENVDB_USE_STATIC_LIBS)
       set(OpenVDB_USES_IMATH_HALF ON)
     endif()
 
-    string(FIND ${PREREQUISITE} "boost_iostreams" _HAS_DEP)
-    if(NOT ${_HAS_DEP} EQUAL -1)
-      set(OpenVDB_USES_DELAYED_LOADING ON)
-    endif()
   endforeach()
 
   unset(_OPENVDB_PREREQUISITE_LIST)
@@ -644,7 +640,6 @@ endif()
 set(_OPENVDB_VISIBLE_DEPENDENCIES "")
 
 if(OpenVDB_USES_DELAYED_LOADING)
-  list(APPEND _OPENVDB_VISIBLE_DEPENDENCIES Boost::iostreams)
   list(APPEND OpenVDB_DEFINITIONS OPENVDB_USE_DELAYED_LOADING)
 endif()
 
--- a/openvdb/openvdb/CMakeLists.txt
+++ b/openvdb/openvdb/CMakeLists.txt
@@ -117,19 +117,6 @@ endif()
 
 # Collect and configure lib dependencies
 
-if(OPENVDB_USE_DELAYED_LOADING)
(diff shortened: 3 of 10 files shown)


danrbailey (commented):
Hi @dgovil, actually we are eliminating delayed loading altogether in the next release.

For now, the changes are all on the `feature/io` branch while it undergoes additional production testing, but the goal is to merge this branch in soon.

See this PR for the details of the removal - https://github.com/AcademySoftwareFoundation/openvdb/pull/2164

dgovil:
@danrbailey ah that's awesome , Dan. I'll close this PR out then.

nvmkuruc:
Thanks @danrbailey.  Would it be advisable to disable `OPENVDB_USE_DELAYED_LOADING` in earlier versions of OpenVDB (say 12)?

This is OpenUSD's usage which appears to load the grids immediately after the stream is loaded.  If I'm understanding the feature, perhaps OpenUSD isn't even benefitting from delayed loading?

https://github.com/PixarAnimationStudios/OpenUSD/blob/32ef0b2d9a301167dd1016e42cf1c0f5194ef0fa/pxr/imaging/hioOpenVDB/utils.cpp#L83-L86

PR #2217, opened 2026-05-21
Add docs quality CI gates and clear the doxygen warning backlog

swahtz:
Fixes #2161 — the docs site showed `nanovdb::createNanoGrid` after the symbol had moved to `nanovdb::tools::`. Markdown code in our docs is never compiled or tested, so this kind of rot goes unnoticed.

**Two CI gates** (`.github/workflows/docs.yml`):
- `validate_docs_references.py` — every `#include` and `github.com/.../blob/<ref>/` URL in doc code blocks must resolve to a real file.
- Doxygen runs with its native `WARN_AS_ERROR=FAIL_ON_WARNINGS`, so any new doc warning fails the PR.

**Clears the entire Doxygen warning backlog at the source** — the **117** warnings fixed directly:
- `doc/changes.txt` — 26 historical dead `\link`s to removed/renamed APIs → `@c name` (or plain prose); `createLevelSetSphere()` requalified `tools::`.
- `ext/imath/.../Half.h` (vendored Imath) — stale `@file` fixed, internal macros wrapped in `@cond`, and the file description made to render (ASCII diagrams → `<pre>`, double-backticks → single, which broke Doxygen 1.14's markdown).
- `nanovdb/.../ex_util/ComputePrimitives.h` — `cxx14::make_index_sequence` shim → `std::` (C++17 baseline), dropping a recursive-inheritance template Doxygen mis-flagged.
- `doc/CMakeLists.txt` — removed an over-broad `*Internal*` `EXCLUDE_SYMBOLS` glob (it hid the public `InternalNode` classes) and predefined `OPENVDB_API=`.

**Also:** `HelloWorld.md` pulls its example via `\snippet` from the compiled `ex_*` sources (the structural #2161 fix); AX math migrated LaTeX → MathJax (drops the TeX toolchain dependency).

PR #2226, opened 2026-06-03
Add NanoVDB TEACHME interactive tutorial + docs-compile test harness

harrism:
Adds `doc/nanovdb/TEACHME/` — an interactive, LLM-guided tutorial for the
NanoVDB **user** API, plus a test harness that keeps it from drifting from the
real headers.

- **User lesson** (`nanovdb_user_lesson.md`) — 10 modules: sparse-grid
  foundations, `GridHandle`/IO, `ReadAccessor`, `NodeManager`, math + sampling +
  HDDA, GPU kernels, GPU grid builders (`PointsToGrid`/`MeshToGrid`), topology
  operators + data re-homing (`Injection`), `IndexGrid`/`VoxelBlockManager`,
  and OpenVDB ↔ NanoVDB conversion.
- **Cheat sheet** (`nanovdb_user_cheatsheet.md`) — one-page API reference.
- **Capstone** (`capstone/`) — a GPU level-set ray-march renderer (scaffold +
  reference + CMake), rendering an SDF sphere on the device.

## Why a test harness

Markdown code in docs is never compiled, so it silently rots as the API
changes. `test/` extracts every C++/CUDA block and compiles it (compile-only)
against the real NanoVDB headers, and validates every `#include` / prose header
reference. ~64 blocks compile green; run `test/run.sh`. This is in the same
spirit as the doc-quality CI gates being added in #2217.

A CI workflow (`.github/workflows/teachme.yml`) runs the harness whenever the
NanoVDB headers (`nanovdb/nanovdb/**`) or the tutorial change, so an API change
that breaks a lesson is caught automatically. It's compile-only (~90s on a
standard runner, no GPU) and reuses the existing `aswf/ci-openvdb` CUDA image.

## Notes

- Docs + harness + a CI workflow that runs it — no changes to NanoVDB headers.
- The harness is compile-only by design (catches API drift); the capstone is
  the runnable end-to-end check.
- Authored interactively with Claude (see commit co-author trailer).


kmuseth (changes requested):
This looks great - nice work! Please update pendingchanges/nanovdb.txt 

A few high-level questions: it's unclear exactly what and how this documentation is intended for. More to the point, is it for novice developers or AI agents? I was expecting to see this in a top-level README, but I couldn't find it.

doc/nanovdb/TEACHME/capstone/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Build + editor + render artifacts (not source)
```
kmuseth, line 1: did you mean to include this file?

harrism:
Thanks Ken — pendingchanges added in 94c04690, as a New Features bullet in `pendingchanges/nanovdb.txt`.

On the audience question: **it's for human developers who are new to NanoVDB, not for AI agents.** The LLM is the delivery mechanism, not the reader. The lesson file is written as an instructor prompt — the agent reads it and then teaches you the API interactively, module by module, quizzing as it goes and adapting to your answers. You are the student; the agent is the teacher. It is explicitly *not* agent-facing reference material like a `CLAUDE.md` or an MCP schema.

The reason for that shape is that the agent can read both the lesson *and* the NanoVDB headers in the same session, so it can verify API details against the real source and debug your exercises live — which a static tutorial can't do, and which is also why the tutorial can't drift silently (the harness compiles every code block against the real headers).

That framing is in `doc/nanovdb/TEACHME/README.md`, but you're right that it isn't discoverable — you had to find the directory first. The PR currently links it from `nanovdb/nanovdb/Readme.md` and `doc/nanovdb/doc.md`, which are the two NanoVDB doc entry points; the repo-root `README.md` has no per-component documentation index, so there was no obvious slot there.

Happy to fix the discoverability whichever way you prefer — tell me which you meant by "top-level README" and I'll do it:
- lead `doc/nanovdb/TEACHME/README.md` with an explicit "who this is for" paragraph;
- promote the summary into `nanovdb/nanovdb/Readme.md` rather than a one-line bullet;
- add a documentation section to the repo-root `README.md` and list it there.

kmuseth (approved):
thanks for the changes - approved

PR #2227, opened 2026-06-03
Adding semantics for VoxelBVH.

areidmeyer:


kmuseth (approved):
This all looks good, but I'm surprised to see no actual changes to the NanoVDB tree or classes themselves. Is that really correct - are the only changes related to labels and meta data?

areidmeyer:
This is correct, since the underlying structure is currently just an Int64 Grid with particular metadata.

areidmeyer:
Thinking on this, what makes sense to add later is an Iterator to walk the list of primitives for a given voxel.
But there is no ABI impact with that, so wanted to do that as a followup.
The Iterator is slightly more complicated because in practice I use cooperation in the thread block.

PR #2205, opened 2026-05-18
Minor AX unit test fix

Idclip:
ASWF docker containers seem to have changed and LLVM builds now build with asserts - caught a small issue in a unit test, no impact to core lib

PR #2229, opened 2026-06-09
Bug fix and adding dilation to VDB blend

apradhana:
- Bug fix in VDB Blend: only do blending when the voxel is active for the two level-set inputs.
- Add dilation using `maskSDF` from `FastSweeping` in a support region of the blend.
Screenshot 2026-06-12 at 4 49 15 PM


danrbailey (approved):
Changes look good to me. I think we should do a round of feedback on the parameter names in particular - alpha, beta, gamma are not the most intuitive names (@jmlait).

Happy to do that on the feature branch though. Note that I created a new `feature/vdb_blend` branch and merged `vdb_blend` into that one. We have slightly different permissions around feature branches. 

The AX test that is failing is I believe one that @Idclip addressed here - https://github.com/AcademySoftwareFoundation/openvdb/pull/2205. Now that `feature/vdb_blend` contains that fix, I _think_ the CI will now succeed, so I retried the failed job on this PR.

apradhana:
Thanks for the feedback, @danrbailey. I should make time next week to address the parameters' name in Houdini SOP and to make improvements on the efficiency of the algorithm.

PR #2231, opened 2026-06-11
NanoVDB: add stream-ordered async memory-resource seam (CUDA)

harrism:
Part of #2232 (NanoVDB injectable CUDA memory resources — design & roadmap).

## What

Adds an instance-based async memory-resource concept to NanoVDB's CUDA path so downstream code (PyTorch extensions, custom CUDA apps, …) can inject its own allocator **without forking headers** — with zero change to existing call sites at the default.

- **`cuda::DeviceResource`** — adds instance `allocate_async`/`deallocate_async` alongside the existing static methods, plus `default_resource<R>()` (a program-lifetime singleton) and an `is_async_resource<R>` detection trait.
- **`cuda::PinnedResource`** (new) — host-pinned resource wrapping `cudaMallocHost`/`cudaFreeHost`. Since those are synchronous, it models the synchronous `Resource` concept (`allocate`/`deallocate`, **no stream**) — `is_async_resource<PinnedResource>` is false — rather than the stream-ordered `async_resource` shape.
- **`cuda::TempPool`** — routes allocations through a resource *instance* and frees on the stream of the most recent `reallocate()` at destruction (previously the null stream). This stream-retained free reaches every `TempDevicePool` user (TopologyBuilder, MeshToGrid, DistributedPointsToGrid, radixSort) automatically.
- **`tools::cuda::PointsToGrid`** — routes its internal scratch and `mDeviceData` through an injected resource instance via a defaulted `ResourceT&` constructor argument.

The method shapes (`allocate_async(bytes, alignment, stream)`) deliberately mirror CCCL's `cuda::mr::async_resource`, so a future type-erased `resource_ref` adapter is a thin shim.

## Why

Downstream projects currently fork NanoVDB headers to route allocations through their own pool (e.g. PyTorch's caching allocator). This adds the injection seam so they don't have to.

## Non-CUDA builds

This PR has **no impact on non-CUDA (`NANOVDB_USE_CUDA=OFF`) usage of NanoVDB.** Everything it adds is CUDA-domain: the headers live under `nanovdb/cuda/` and are reachable only behind `#if defined(__CUDACC__)` (the same gate `GridHandle.h` already uses for `cuda/GridHandle.cuh`), and the new unit test + test target are added only inside `if(NANOVDB_USE_CUDA)`. The CUDA-free core (`NanoVDB.h`, `GridHandle.h`, `HostBuffer.h`) is untouched and gains no new include.

Verified with a clean `NANOVDB_USE_CUDA=OFF` build (host compiler, no CUDA toolkit): it configures with no CUDA/CCCL pulled, and the CPU unit tests pass 152/152. CI also exercises this configuration via the `nanovdb-lite` (Linux) and `macos-nanovdb` jobs.

## Testing

New `unittest/TestMemoryResource.cu` (`TestMemoryResource` suite, 8 tests): concept satisfaction (`static_assert`), instance alloc/free, pinned page-lock + async H↔D copy, stream-retained free, no-leak-across-growth, default-ctor backward-compat, and `PointsToGrid` resource-injection (counts allocations through an injected mock, asserts no leak). Existing `nanovdb_test_cuda` suite is unchanged (52/52). Built and run on `sm_89` / CUDA 12.6.

## Scope / follow-ups (intentionally not in this PR)

- `TopologyBuilder` / `MeshToGrid` allocate through the dual `cuda::DeviceBuffer`; routing those — and the pinned-host `cudaMallocHost` stall fix — needs a resource-parameterized buffer (`ResourceDeviceBuffer`) and is deferred.
- `DistributedPointsToGrid` (multi-GPU) deferred.
- Converting `PointsToGrid`'s raw scratch to RAII containers (`DeviceVector<T,R>`) is a planned follow-up.
- `PointsToGrid<Point, CustomResource>` point-encoding is a pre-existing limitation (the `PointsToGrid<Point>` specialization is keyed to the default resource); tracked separately.

### DIFF
--- a/nanovdb/nanovdb/cuda/DeviceResource.h
+++ b/nanovdb/nanovdb/cuda/DeviceResource.h
@@ -7,10 +7,23 @@
 #include <cuda_runtime_api.h>
 #include <nanovdb/util/cuda/Util.h>
 
+#include <cstddef>
+#include <type_traits>
+#include <utility>
+
 namespace nanovdb {
 
 namespace cuda {
 
+/// @brief Default stream-ordered device memory resource. Allocations are made
+///        with cudaMallocAsync and freed with cudaFreeAsync via the
+///        util::cuda wrappers.
+/// @note Exposes both the legacy static methods (allocateAsync /
+///       deallocateAsync, retained for source compatibility) and the
+///       instance methods (allocate_async / deallocate_async) that model
+///       the AsyncResource concept. The instance methods forward to the
+///       static ones, so a default-constructed instance is stateless and
+///       adds no overhead.
 class DeviceResource
 {
 public:
@@ -26,7 +39,49 @@ public:
     static void deallocateAsync(void *p, size_t, size_t, cudaStream_t stream) {
         cudaCheck(util::cuda::freeAsync(p, stream));
     }
+
+    /// @brief Instance allocation entry point modelling the AsyncResource concept.
+    /// @param bytes number of bytes to allocate
+    /// @param alignment requested alignment (cudaMallocAsync always 256B-aligns)
+    /// @param stream cuda stream the allocation is ordered on
+    void* allocate_async(size_t bytes, size_t alignment, cudaStream_t stream) {
+        return allocateAsync(bytes, alignment, stream);
+    }
+
+    /// @brief Instance deallocation entry point modelling the AsyncResource concept.
+    /// @param p pointer previously returned by allocate_async
+    /// @param bytes size passed to the matching allocate_async
+    /// @param alignment alignment passed to the matching allocate_async
+    /// @param stream cuda stream the deallocation is ordered on
+    void deallocate_async(void* p, size_t bytes, size_t alignment, cudaStream_t stream) {
+        deallocateAsync(p, bytes, alignment, stream);
+    }
 };
+
+/// @brief Returns a program-lifetime, address-stable reference to a default
+///        instance of resource @c R.
+/// @details The instance is a function-local static, so it outlives every
+///          caller and is safe to bind through a default function/constructor
+///          argument. @c R must be default-constructible.
+template <class R>
+inline R& default_resource()
+{
+    static R sResource;
+    return sResource;
+}
+
+/// @brief Detection trait: @c is_async_resource<R>::value is true iff @c R
+///        models the stream-ordered AsyncResource concept, i.e. exposes
+///        allocate_async(size_t, size_t, cudaStream_t) and
+///        deallocate_async(void*, size_t, size_t, cudaStream_t).
+template <class R, class = void>
+struct is_async_resource : std::false_type {};
+
+template <class R>
+struct is_async_resource<R, std::void_t<
+    decltype(std::declval<R&>().allocate_async(size_t{0}, size_t{0}, cudaStream_t{0})),
+    decltype(std::declval<R&>().deallocate_async(std::declval<void*>(), size_t{0}, size_t{0}, cudaStream_t{0}))>>
+    : std::true_type {};
 
 }
 
--- a/nanovdb/nanovdb/cuda/DeviceResource.h
+++ b/nanovdb/nanovdb/cuda/DeviceResource.h
@@ -74,6 +74,19 @@ inline R& default_resource()
 ///        models the stream-ordered AsyncResource concept, i.e. exposes
 ///        allocate_async(size_t, size_t, cudaStream_t) and
 ///        deallocate_async(void*, size_t, size_t, cudaStream_t).
+/// @details Use it to dispatch between a stream-ordered resource and a
+///          synchronous one (which exposes allocate/deallocate without a
+///          stream argument):
+/// @code
+/// template<typename R>
+/// void* allocate(R& resource, size_t bytes, size_t alignment, cudaStream_t stream)
+/// {
+///     if constexpr (nanovdb::cuda::is_async_resource<R>::value)
+///         return resource.allocate_async(bytes, alignment, stream); // stream-ordered
+///     else
+///         return resource.allocate(bytes, alignment);               // synchronous
+/// }
+/// @endcode
 template <class R, class = void>
 struct is_async_resource : std::false_type {};
 
--- a/nanovdb/nanovdb/cuda/PinnedResource.h
+++ b/nanovdb/nanovdb/cuda/PinnedResource.h
@@ -17,44 +17,37 @@ namespace cuda {
 ///        host memory (cudaMallocHost) that is both host-accessible and
 ///        device-accessible, so it can serve as the host side of an
 ///        asynchronous host<->device copy.
-/// @note The default implementation is *synchronous*: cudaMallocHost /
-///       cudaFreeHost ignore the supplied stream and synchronize the context.
-///       The stream parameter is accepted purely so this type models the same
-///       AsyncResource shape as cuda::DeviceResource; the reason the resource
-///       is a seam at all is so callers can substitute a pooled / genuinely
-///       stream-ordered pinned-host resource and avoid that synchronization.
+/// @note This resource is *synchronous*: cudaMallocHost / cudaFreeHost
+///       synchronize the context, so it models the synchronous Resource
+///       concept (allocate / deallocate, no stream) rather than the
+///       stream-ordered AsyncResource concept. It is a seam so callers can
+///       substitute a pooled / genuinely stream-ordered pinned-host resource
+///       and avoid that synchronization.
 class PinnedResource
 {
 public:
-    // cudaMallocHost over-aligns (to at least the page size); 256 is the
-    // nominal default this resource advertises through the AsyncResource API.
+    // cudaMallocHost over-aligns (to at least the page size), so the requested
+    // alignment is always satisfied; 256 is the nominal default advertised here.
     static constexpr size_t DEFAULT_ALIGNMENT = 256;
 
-    static void* allocateAsync(size_t bytes, size_t, cudaStream_t) {
+    /// @brief Synchronous allocation of page-locked host memory.
+    /// @param bytes number of bytes to allocate
+    /// @param alignment requested alignment (ignored; cudaMallocHost is page-aligned)
+    void* allocate(size_t bytes, size_t alignment) {
+        (void)alignment;
         void* p = nullptr;
         cudaCheck(cudaMallocHost(&p, bytes));
         return p;
     }
 
-    static void deallocateAsync(void* p, size_t, size_t, cudaStream_t) {
+    /// @brief Synchronous deallocation.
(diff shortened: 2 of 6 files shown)


nanovdb/nanovdb/cuda/DeviceResource.h
```diff
@@ -26,8 +39,50 @@ class DeviceResource
     static void deallocateAsync(void *p, size_t, size_t, cudaStream_t stream) {
         cudaCheck(util::cuda::freeAsync(p, stream));
     }
+
+    /// @brief Instance allocation entry point modelling the AsyncResource concept.
+    /// @param bytes number of bytes to allocate
+    /// @param alignment requested alignment (cudaMallocAsync always 256B-aligns)
+    /// @param stream cuda stream the allocation is ordered on
+    void* allocate_async(size_t bytes, size_t alignment, cudaStream_t stream) {
```
kmuseth, line 47: function name should be camel-case like DeviceResource::DeviceResource defined above. This is, `allocate_async` -> `allocateAsync`

nanovdb/nanovdb/cuda/DeviceResource.h
```diff
@@ -26,8 +39,50 @@ class DeviceResource
...
+        return allocateAsync(bytes, alignment, stream);
+    }
+
+    /// @brief Instance deallocation entry point modelling the AsyncResource concept.
+    /// @param p pointer previously returned by allocate_async
+    /// @param bytes size passed to the matching allocate_async
+    /// @param alignment alignment passed to the matching allocate_async
+    /// @param stream cuda stream the deallocation is ordered on
+    void deallocate_async(void* p, size_t bytes, size_t alignment, cudaStream_t stream) {
```
kmuseth, line 56: please change `deallocate_async` to `deallocateAsync`

nanovdb/nanovdb/cuda/DeviceResource.h
kmuseth: how come the cuda stream arguments don't have default values in these function signatures?

nanovdb/nanovdb/cuda/DeviceResource.h
```diff
@@ -26,8 +39,50 @@ class DeviceResource
...
+inline R& default_resource()
+{
+    static R sResource;
+    return sResource;
+}
+
+/// @brief Detection trait: @c is_async_resource<R>::value is true iff @c R
+///        models the stream-ordered AsyncResource concept, i.e. exposes
+///        allocate_async(size_t, size_t, cudaStream_t) and
+///        deallocate_async(void*, size_t, size_t, cudaStream_t).
```
kmuseth, line 76: please provide an examples (in doxygen) demonstrating how to use this trait

nanovdb/nanovdb/cuda/PinnedResource.h
```diff
@@ -0,0 +1,65 @@
...
+
+    static void deallocateAsync(void* p, size_t, size_t, cudaStream_t) {
+        cudaCheck(cudaFreeHost(p));
+    }
+
+    /// @brief Instance allocation entry point modelling the AsyncResource concept.
+    /// @param bytes number of bytes to allocate
+    /// @param alignment requested alignment (cudaMallocHost is page-aligned)
+    /// @param stream cuda stream (ignored by the synchronous default impl)
+    void* allocate_async(size_t bytes, size_t alignment, cudaStream_t stream) {
```
kmuseth: please change `allocate_async` to `allocateAsync`

Also, why to provide default arguments for the stream?

nanovdb/nanovdb/cuda/PinnedResource.h
```diff
@@ -0,0 +1,65 @@
...
+        return allocateAsync(bytes, alignment, stream);
+    }
+
+    /// @brief Instance deallocation entry point modelling the AsyncResource concept.
+    /// @param p pointer previously returned by allocate_async
+    /// @param bytes size passed to the matching allocate_async
+    /// @param alignment alignment passed to the matching allocate_async
+    /// @param stream cuda stream (ignored by the synchronous default impl)
+    void deallocate_async(void* p, size_t bytes, size_t alignment, cudaStream_t stream) {
```
kmuseth: please change `deallocate_async` to `deallocateAsync`

Also, why to provide default arguments for the stream?

nanovdb/nanovdb/cuda/TempPool.h
```diff
@@ -20,13 +23,21 @@ template <class Resource>
...
+    explicit TempPool(Resource& resource) : mResource(&resource), mData(nullptr), mSize(0), mRequestedSize(0), mStream(nullptr) {}
 
-    /// @brief Destructor
+    /// @brief Destructor. Frees the managed memory on the stream of the most
+    ///        recent reallocate(), so the stream-ordered free is ordered after
+    ///        the work that used the memory (rather than on the null stream).
     ~TempPool() {
         mRequestedSize = 0;
-        Resource::deallocateAsync(mData, mSize, Resource::DEFAULT_ALIGNMENT, nullptr);
+        mResource->deallocate_async(mData, mSize, Resource::DEFAULT_ALIGNMENT, mStream);
```
kmuseth, line 40: `deallocate_async` -> `deallocateAsync`

nanovdb/nanovdb/cuda/TempPool.h
```diff
@@ -41,20 +52,27 @@ class TempPool {
...
+    cudaStream_t stream() const {return mStream;}
+
     /// @brief Re-allocation of the data managed by this instance. Only has affect if the pool in empty or
     ///        the requested memory is larger than the existing size.
     /// @param stream cuda stream used for asynchronous de-allocation and allocation.
     void reallocate(cudaStream_t stream) {
         if (!mData || mRequestedSize > mSize) {
-            Resource::deallocateAsync(mData, mSize, Resource::DEFAULT_ALIGNMENT, stream);
-            mData = Resource::allocateAsync(mRequestedSize, Resource::DEFAULT_ALIGNMENT, stream);
+            mResource->deallocate_async(mData, mSize, Resource::DEFAULT_ALIGNMENT, stream);
```
kmuseth, line 64: deallocate_async -> deallocateAsync

nanovdb/nanovdb/cuda/TempPool.h
```diff
@@ -41,20 +52,27 @@ class TempPool {
...
+            mData = mResource->allocate_async(mRequestedSize, Resource::DEFAULT_ALIGNMENT, stream);
```
kmuseth, line 65: allocate_async -> allocateAsync

kmuseth (changes requested):
overall it looks good, but I have some questions and requests for changes

kmuseth:
one more request @harrism  - please add a file "pendingchanges/nanovdb.h" with sentence describing this change

harrism:
@kmuseth I would like to maintain the snake_case spelling of `allocate`, `allocate_async`, `deallocate`, `deallocate_async`.  These spellings *are* the contract for compatibility with outside memory resources. This contract is defined by CCCL's `cuda::mr` (modeled on `std::pmr::memory_resource`). 

If we spell ours differently, callers who have existing memory resources that adhere to this contract will have to wrap them in an adapter to satisfy NanoVDB's contract.  The identifier must be exactly `allocate_async` because C++ concept/trait matching is by exact spelling — there's no aliasing, so `allocateAsync` simply isn't the same symbol an external resource exposes.  This design aims deliberately to widen compatibility and increase alignment with standards, and enable sharing memory resources across an application and all its dependencies.

Regarding spelling in particular: NanoVDB core already uses STL-style lowercase names where it models standard-library concepts — data (303 uses), size (128), end (30), empty (29), begin (24) across HostBuffer.h/GridHandle.h/NanoVDB.h. allocate_async/deallocate_async is the same situation: it's the spelling of an external concept (cuda::mr / the std::pmr-style allocator interface), not a free stylistic choice. Note also that the static legacy methods (retained for source compatibility) in these files are already CamelCase (allocateAsync/deallocateAsync, DeviceResource.h:33) — the PR deliberately reserves snake_case for the interop-concept instance methods only.

Regarding default values for stream parameters: for asynchronous / stream-ordered allocation and deallocation, I deliberately omit a default for reasons of safety. Defaulting to stream 0 is one of the hazards that this PR aims to eliminate.
 - *A defaulted stream invites callers to omit it and silently allocate/free on the null stream*.  Under the semantics of stream-ordered allocation, a null-stream free of memory last used on another stream is a potential use-after-free/race, not just a serialization cost.
 - *The null stream's meaning is also build-mode-dependent* (legacy default-stream implicitly syncs with all streams; --default-stream per-thread doesn't), so a defaulted stream is ambiguous across compile configs.
 - The "defaults" instinct comes from the legacy surface (`DeviceBuffer(size, device, cudaStream_t stream = 0)` and friends default to 0) — but that's the null-stream behavior we should move away from. The static `allocateAsync` in these resources already does _not_ default the stream, so internal consistency actually favors no-default.
 - CCCL / RMM do not default it. `cuda::mr::async_resource::allocate_async(size_t, size_t, cuda::stream_ref)` takes the stream as a required argument — stream-ordered allocation is meaningless without knowing the stream.
 
PinnedResource is synchronous (cudaMallocHost), so it shouldn't take a stream at all — a synchronous resource exposes allocate/deallocate with no stream, rather than a defaulted one. In this PR it currently exposes the async shape with an ignored stream; I'll switch it to a synchronous resource here, and we can add a stream-ordered pinned pool later as a separate async resource.

Will add a doxygen `@code` example showing `is_async_resource` used to dispatch `allocate_async` vs `allocate`.

nanovdb/nanovdb/cuda/DeviceResource.h
harrism: Done in 74d6c4db — added a `@code` example showing `is_async_resource` used to dispatch (`if constexpr`) between a stream-ordered `allocate_async(…, stream)` and a synchronous `allocate(…)`.

nanovdb/nanovdb/cuda/PinnedResource.h
harrism: Resolved in 74d6c4db: `cudaMallocHost` / `cudaFreeHost` are synchronous, so `PinnedResource` is now a synchronous resource — `allocate(bytes, alignment)` / `deallocate(ptr, bytes, alignment)` with no stream, and the static `allocateAsync` / `deallocateAsync` helpers are removed. So there's no `_async` method to rename and no stream to default here.

nanovdb/nanovdb/cuda/PinnedResource.h
harrism: Same resolution as above (74d6c4db): `PinnedResource` is synchronous now (`allocate` / `deallocate`, no stream), so `deallocate_async` is removed rather than renamed.

kmuseth (approved):

PR #2233, opened 2026-06-11
Houdini 22 Sync

jmlait:
Sync with (unreleased) Houdini 22.

Camera changes, behind version guards.
New option on filter for avoiding size drift.
format() supports the VDB Enums
Reference files updated.

### DIFF
--- a/openvdb_houdini/openvdb_houdini/reference/GEO_PrimVDB.h
+++ b/openvdb_houdini/openvdb_houdini/reference/GEO_PrimVDB.h
@@ -48,13 +48,16 @@ using ::GEO_PrimVDB;
 #include <UT/UT_BoundingBox.h>
 #include "UT_VDBUtils.h"
 
+#include <IMX/IMX_VDB.h>
+
 #include <openvdb/Platform.h>
 #include <openvdb/openvdb.h>
 
 
 class   GEO_Detail;
 class   GEO_PrimVolume;
-class   GEO_PrimVolumeXform;
+class   GA_PrimVolumeXform;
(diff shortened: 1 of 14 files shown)


openvdb_houdini/openvdb_houdini/reference/GEO_PrimVDB.h
apradhana, line 388: The definition of `calcVolume` was deleted in the corresponding `.cc`. Perhaps we should also delete the definition?

apradhana, line 391: The same comment as above.

apradhana, line 486: The definition of `calcPositiveDensity` was deleted, so perhaps we should also delete the declaration.

apradhana: Dropping the comment since it was clarified by @jmlait that Houdini 22 split the implementation across multiple translation units.

apradhana: Dropping the comment since it was clarified by @jmlait that Houdini 22 split the implementation across multiple translation units.

apradhana: Dropping the comment since it was clarified by @jmlait that Houdini 22 split the implementation across multiple translation units.

apradhana (approved):

PR #2234, opened 2026-06-12
Fix Windows Static CI

danrbailey:
When OpenVDB is built against a static Blosc library with `BLOSC_USE_EXTERNAL_SOURCES` enabled, `FindBlosc.cmake` loops over Blosc's external dependencies (lz4 snappy zlib zstd) and tries to `find_library()` each one to add it as an
interface link dependency on `Blosc::blosc`. 

However, OpenVDB also links against zlib directly, so this change falls back to using that one if not found. I have confirmed that this resolves the static Windows CI failing in the weekly build.


apradhana (approved):
Looks good to me. Thanks for fixing this issue, @danrbailey !

PR #2237, opened 2026-06-15
Bugfix for GridHandle::read().

areidmeyer:
seekg math was not quite accurate.

### DIFF
--- a/nanovdb/nanovdb/GridHandle.h
+++ b/nanovdb/nanovdb/GridHandle.h
@@ -410,7 +410,7 @@ void GridHandle<BufferT>::read(std::istream& is, const BufferT& pool)
             sum += data.mGridSize;
         }
         auto buffer = BufferT::create(size + sum, &pool);
-        is.seekg(-int64_t(sum + sizeof(GridData)), std::ios::cur);// rewind to start
+        is.seekg(-int64_t(size + sum - data.mGridSize + sizeof(GridData)), std::ios::cur);// rewind to start
         is.read((char*)(buffer.data()), buffer.size());
         *this = GridHandle(std::move(buffer));
     } else {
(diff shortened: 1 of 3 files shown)


matthewdcong (approved):
Looks good to me, nice catch!

swahtz (approved):

apradhana (approved):

apradhana:
@areidmeyer : is it possible to add something like `pendingchanges/gridhandle_read_bugfix.txt`just to record `NanoVDB Bug Fixes: GridHandle::read(std::istream&) was over-counting the size of multi-grid raw buffers.` for Release Notes purposes.

kmuseth (changes requested):
Good catch, but I prefer this simpler (only adding one line and modifying one) and more intuitive (no weird math) version:

   ```
    const std::streampos start = is.tellg();// remember where the raw buffer begins
    GridData data;
    is.read((char*)&data, sizeof(GridData));
    if (data.isValid()) {
        uint64_t size = data.mGridSize, sum = 0u;
        while(data.mGridIndex + 1u < data.mGridCount) {// loop over remaining raw grids in stream
            is.seekg(data.mGridSize - sizeof(GridData), std::ios::cur);// skip grid
            is.read((char*)&data, sizeof(GridData));
            sum += data.mGridSize;
        }
        auto buffer = BufferT::create(size + sum, &pool);
        is.seekg(start);// rewind to the start of the raw buffer
        is.read((char*)(buffer.data()), buffer.size());
        *this = GridHandle(std::move(buffer));
    } else {
        is.seekg(-sizeof(GridData), std::ios::cur);// rewind
        throw std::logic_error("This stream does not contain a valid raw grid buffer");
    }
```

areidmeyer:
Weird math purged.

kmuseth:
> Weird math purged.

still impressed that you caught and fixed this embarrassing bug!

kmuseth (approved):
thanks for the code change

PR #2238, opened 2026-06-16
NanoVDB: generalize `tools::cuda::MergeGrids` to an N-ary union

swahtz:
`tools::cuda::MergeGrids` produced the morphological (active-mask) union of **exactly two** device OnIndex grids. This PR generalizes it to merge **any number** of grids in a single pass, while keeping the binary form fully backward compatible.

### Motivation
The two-grid limit forces callers who need to union many grids to chain the binary merge into a tree of pairwise calls, which materializes ~`log2(N)` intermediate grids and re-reads data across levels.

The existing algorithm is already structured as an *accumulate-into-shared-masks* union: it builds a merged root from all inputs, then each input scatter-ORs its node masks into a shared, union-sized accumulator via atomics. That operation is commutative/associative and keyed by spatial position, so it extends to N inputs **without any change to the kernels** — only the host-side orchestration was two-grid-specific.

### Changes
- Add an N-ary constructor `MergeGrids(const std::vector<const GridT*>&, stream)`. The existing binary `MergeGrids(grid1, grid2, stream)` constructor now delegates to it, so current callers (the `ex_merge_nanovdb_cuda` example and the existing unit test) are unchanged.
- Replace the duplicated `grid1`/`grid2` members and the paired code blocks in `mergeRoot` / `mergeInternalNodes` / `mergeLeafNodes` with loops over the input vector; take output `GridData` metadata from the first input.
- Guard against an empty input list.

### Performance
A single N-ary pass allocates **one** accumulator (sized to the final union) and reads each input grid **once**, instead of materializing ~`log2(N)` intermediate grids when chaining the binary form — fewer allocations and less data movement, with identical output.

### Backward compatibility
No API breakage: the binary constructor is retained and delegates to the N-ary path. Output is byte-identical to chaining binary merges (verified in tests).

### Testing
Adds `TestNanoVDBCUDA.MergeGridsNary_ValueOnIndex`, which merges three grids (two overlapping, one disjoint in a separate upper node) and verifies the N-ary result is identical to chaining the binary merge (leaf/lower/upper node counts and active-voxel count), matches the expected 69 unique voxels, that a single-element list copies the grid, and that an empty list throws.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/MergeGrids.cuh
+++ b/nanovdb/nanovdb/tools/cuda/MergeGrids.cuh
@@ -17,6 +17,9 @@
 
 #include <cub/cub.cuh>
 
+#include <map>
+#include <vector>
+
 #include <nanovdb/NanoVDB.h>
 #include <nanovdb/GridHandle.h>
 #include <nanovdb/tools/cuda/TopologyBuilder.cuh>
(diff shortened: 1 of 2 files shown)


kmuseth (approved):
This PR looks good to me, with one exception - the naming (which is arguably not part of this specefic PR so I wont block it). I would much prefer we renamed the MergeGrids to TopoloyUnion since it's more descriptive (merging of grids is ill-defined) and more consistent with the existing terminology in OpenVDB (openvdb::Grid and openvdb::Tree define topologyUnion).

swahtz:
> This PR looks good to me, with one exception - the naming (which is arguably not part of this specefic PR so I wont block it). I would much prefer we renamed the MergeGrids to TopoloyUnion since it's more descriptive (merging of grids is ill-defined) and more consistent with the existing terminology in OpenVDB (openvdb::Grid and openvdb::Tree define topologyUnion).

That makes sense, I'll create a ticket and talk to @sifakis

PR #2241, opened 2026-06-24
Massive improvements to vdb tool

kmuseth:
New Features:

-   Added new action "-files", a control-flow scope (closed by -end, like -for and -each) that iterates over the files in a directory, binding each matching path to the loop variable "file" so the body can reference it as "{$file}" (e.g. "-files path=$HOME/data ext=obj,stl -read '{$file}' -mesh2ls -print -clear -end"). The mandatory "path" option sets the directory where the search begins; "recursive=1" descends into sub-directories. Matches are filtered by "extension" (a comma-separated list, e.g. ext="obj,ply,vdb"), by "include"/"exclude" filename patterns, and by file size via "min_size"/"max_size" which accept plain byte counts or human-readable suffixes (1B, 1KB, 1MB, 1GB, 1TB). As with the other loops, "quiet=true" suppresses the per-iteration "Processing: ..." output for this scope only.
-   Added new action "-mesh2udf" (alias "-soup2udf"), which converts a polygon mesh or polygon soup into an unsigned distance field, i.e. a VDB grid storing the non-negative distance to the input surface within a symmetric narrow band on both sides of it. Unlike a signed level set this does not require the input to be watertight or consistently oriented, so it works on open meshes and arbitrary triangle/quad soups. The output resolution is set by "dim" (largest bbox dimension in voxels, default 256) or by "voxel" (explicit world-space voxel size, which takes precedence), while "width" controls the half-width of the narrow band in voxel units (default 3). The transform can instead be inherited from an existing grid via "vdb" (its stack age); "geo" selects which geometry to process.
-   Added new action "-quad2tri" (alias "-q2t"), which triangulates every quadrilateral face in a polygon mesh, replacing each quad with two triangles so the result contains triangles only. This is useful before exporting to formats or feeding downstream tools that accept triangles exclusively. The split is a naive fan from the quad's first vertex (the diagonal between vertices 0 and 2), which is correct as long as each quad is planar and convex; the action assumes this and does not test for it, so non-planar or concave quads may produce a skewed or self-overlapping pair of triangles. Existing triangles in the mesh are left untouched. The "geo" option selects which geometry on the stack to process (defaults to the most recently inserted) and "keep" toggles whether that input is preserved or deleted after the conversion.
-   Added new action "-vol2mesh" (alias "-vdb2mesh"), which extracts a polygon mesh from any scalar VDB grid by meshing one of its iso-surfaces, producing a surface of triangles and quads at the chosen value. Unlike "-fog2mesh" (which defaults to the 0.5 iso-value of a normalized fog volume) it defaults to the zero iso-value, making it the general-purpose mesher for level sets and scalar fields. The "iso" option sets the iso-value, and "adapt" controls adaptive meshing on a normalized 0 -> 1 scale where 0 yields a uniform tessellation and higher values simplify flat regions into larger polygons. An optional grid given via "mask" restricts meshing to the region it covers, and "invert=true" meshes the complement of that mask. "vdb" selects the grid, "name" sets the output name, and "keep" toggles whether the input is preserved.
-   Added new action "-fog2mesh", which extracts a polygon mesh from a fog-volume VDB grid by meshing one of its iso-surfaces, producing a watertight surface of triangles and quads at the chosen density value. The "iso" option sets the iso-value defining the surface (default 0.5, suitable for a normalized fog volume), and "adapt" controls adaptive meshing on a normalized 0 -> 1 scale where 0 yields a uniform tessellation and higher values progressively simplify flat regions into larger polygons. An optional grid given via "mask" restricts meshing to the region it covers, and "invert=true" instead meshes the complement of that mask. "vdb" selects which grid to mesh, "name" sets the output geometry's name, and "keep" toggles whether the input volume is preserved (the mask is never removed).
-   Added new action "-slice", which generates one or more images of axis-aligned planar slices through a scalar VDB grid, writing each as a PPM image file. Slice positions are given as normalized coordinates in the range 0 -> 1 across the grid's active bounding box via the "X", "Y" and "Z" options (each accepts one or more values, defaulting to 0.5, i.e. the mid-plane), so a single invocation can emit many slices along any combination of axes. The "file" option sets the output base name, "scale" sets the image pixel size (e.g. 512 or 1920x1080, with the aspect ratio derived from the grid when only one dimension is given), and "force" recomputes the true min/max for color mapping instead of assuming the expected ranges for level-set and fog-volume grids. "vdb" selects which grid to slice.
-   Added new action "-img2mpeg", which converts multiple image files to an mpeg movie file when vdb_tool is optionally built with ffmpeg.
-   Added new action "-log", which mirrors the tool's diagnostic output (timing reports, warnings, non-fatal errors and -print output) to a log file for the remainder of the pipeline. It redirects all three standard text streams (std::clog, std::cerr and std::cout) so nothing normally printed to the terminal is lost, and the file is unit-buffered so "tail -f" or "watch -n 0.5 vdb_tool.log" reflect progress in real time. A self-describing header (date, version and the full command line) is written first so a saved log is readable later. The "file" option names the log (a timestamped default is used when omitted), "append" adds to an existing file instead of truncating it, and "tee" (default true) also echoes to the terminal so interactive feedback is preserved; set tee=false to route output exclusively to the file.
-   Added new action "-sdf2udf", which converts a signed distance field (a level set, negative inside the surface and positive outside) into an unsigned distance field by taking the absolute value of every voxel, so the result stores the non-negative distance to the surface on both sides. Because the inside/outside sign is discarded, the grid class is changed from level set to UNKNOWN, signalling to downstream actions that it is no longer a valid signed level set (so sign-dependent operations such as CSG booleans or signed-flood-fill should not be applied to it). This is useful when only proximity to the surface matters. The "vdb" option selects which grid (or grids) on the stack to process, "name" sets the output grid's name, and "keep" toggles whether the input volume is preserved or deleted after the conversion.
-   Added new action "-forAllValues/forOnValues,forOffValues", which applies a math kernel to all, active, or inactive values in a grid. The reserved variable "v" is bound to the current voxel value; every other identifier is looked up once in the Processor's string memory (the same namespace used by -eval / -calc) and bound as a per-voxel constant, throwing if any referenced name is undefined. This lets kernels mix the voxel value with scalars set earlier in the pipeline, e.g. -eval '{2:@scale}' -forOnValues 'scale*v + 1'. The kernel option accepts both the explicit form ("-forOnValues kernel='sin(v)+1'") and a bare positional form ("-forOnValues 'sin(v)+1'"); the bare form is enabled by the new greedy-anonymous parser mode (see Action::Action) and does not interfere with other named options of the same action such as keep= or class=. Stencil kernels: writing "v(dx,dy,dz)" with integer-literal offsets reads a relative neighbor voxel through a per-thread ConstAccessor (e.g. "v(1,0,0)-v(-1,0,0)" computes a finite-difference x-derivative or "v(1,0,0)+v(-1,0,0)+v(0,1,0)+v(0,-1,0)+v(0,0,1)+v(0,0,-1)-6*v" computes a discrete Laplacian); the output grid is internally deep-copied so reads come from a stable snapshot. Multi-grid kernels: use= and vdb= accept comma-separated lists so the kernel can read from multiple grids in one pass; the first entry is the output (iterated and written), the rest are read-only inputs (e.g. "-forOnValues 'a - b' use=a,b vdb=0,1" computes pointwise differences).
-   Added new class Calculator, a compact bytecode interpreter for math expressions over arbitrarily named float inputs. Accepts RPN (e.g. kernel="$x:abs:1:+"), single infix (e.g. kernel="abs(x)+1"), and infix multi-statement programs with assignment (e.g. kernel="t = x*x; t + sin(t)"). Input variables are discovered automatically and exposed via Calculator::variables(); the forAllValues/forOnValues/forOffValues actions only bind "x" and throw with a clear message on any other identifier.
-  Added new action "-calc", which runs a Calculator expression at command-line scope. Input variables are read from the Processor's string memory (the same namespace used by -eval); intermediate slot values and trailing-LHS names are written back so subsequent -eval / -calc / -for actions can read the results by name (e.g. "-calc 'a=1+2;b=a*3' -eval str='{$b}'"). Read-only inputs (e.g. a -for loop variable) are left untouched. Accepts both the explicit form ("-calc kernel='x=1+2'") and a bare positional form ("-calc 'x=1+2'"); the latter is enabled by the new greedy-anonymous parser mode (see Action::Action). Echoes the result only when the kernel ends in a plain expression; assignment-terminated kernels are silent since their outputs are already in memory.
-   Added new actions "-switch" and "-case", a multi-way control-flow construct analogous to -if but selecting one of N sub-pipelines. "-switch on=<selector>" opens a scope closed by -end; each enclosed "-case key=<value>" runs its body only if its key matches the selector, and "key=*" (or "key=default") is a catch-all that fires when no earlier case matched. Numeric keys are compared by value (so key=1 matches a selector of 1.0) and otherwise strings are compared verbatim. The selector accepts {...} template expressions, so e.g. "-for c=0,4 -switch on={$c} -case key=0 -sphere -end -case key=* -platonic -end -end -end" picks a primitive per iteration.
-  Added new optional action "-ax", which runs an OpenVDB AX expression over one or more grids on the stack, e.g. "-ax '@density = max(@density, 0.0f);'". AX is OpenVDB's JIT-compiled, C-like language for editing grid values and point attributes; it complements vdb_tool's built-in Calculator (-calc / -forValues) by adding the standard AX language, native point-attribute access, and LLVM-backed performance on large grids. The "code" snippet may be given bare (the "code=" prefix is optional) and "vdb" selects which grids to process ("*", the default, passes the whole stack so the code can address any grid by name via @gridname). By default the selected grids are edited in place; "keep=true" instead runs AX on deep copies pushed onto the stack, leaving the originals untouched. The code may alternatively be read from a file via "file=prog.ax" (handy for longer programs), and "bindings=axname:gridname,..." remaps the names used in the AX code to actual grid names (e.g. bindings=x:density lets a kernel written against @x operate on the grid named "density" without editing it). Because AX depends on the openvdb_ax library and LLVM, the action is gated behind the CMake option OPENVDB_TOOL_USE_AX (OFF by default) and is absent unless explicitly enabled.
-   Added triangulation of N-gons to Geometry::readOBJ and Geometry::readSTL, assuming they are convex and planar.
-   Added the string-similarity helpers vdb_tool::levenshtein (the classic edit-distance between two strings) and vdb_tool::fuzzyMatch (which ranks a list of candidate names against a query by combining substring matching with edit distance) to power "Did you mean ...?" suggestions throughout the tool. A mistyped action, option, or expression function/constant no longer just fails: the error now lists the closest known names, e.g. "-sphre" suggests "-sphere", "-erode raidus=2" suggests "radius=", and a kernel calling "tanj(x)" suggests "tan" or "atan". The same routine backs action-name matching in the Parser, option-name matching within an Action, and identifier matching in the Calculator, so the suggestions stay consistent across all three.

 Improvements:

 - Action "-read" now has optional support for reading UsdGeomMesh and UsdGeomPoints prims from OpenUSD files (.usd, .usda, .usdc, .usdz). Walks every prim, bakes each one's world transform into the vertex positions, preserves triangles and quads as-is, and fan-triangulates n-gons (n>4). UsdGeomPoints contribute vertex positions only. Instancing, subdivision, animation, and per-point widths are not supported by this minimal reader. Enable with VDB_TOOL_USE_USD.
 - Action "-read" now has optional support for reading glTF (.gltf) and binary glTF (.glb) files. Walks every mesh primitive in TRIANGLES mode (POINTS/LINES/STRIPS/FANS are skipped with a warning), and consumes both indexed and non-indexed primitives with UBYTE/USHORT/UINT index buffers. Node-graph transforms are not applied, and materials, normals, UVs, and vertex colors are ignored by this minimal reader. Uses the tinygltf header-only library, fetched at configure time. Enable with VDB_TOOL_USE_GLTF.
- Action "-read" now supports ASCII XYZ point files (.xyz), where each line holds the x, y and z coordinates of a single point separated by whitespace. Blank lines and lines beginning with '#' are skipped as comments; any other line that does not parse as three numbers raises an error identifying the offending text. The points are accumulated as a vertex-only primitive (no faces), suitable for conversion to a VDB via -points2ls or -points2vdb. The data can also be streamed from standard input using the name "stdin.xyz".
 - Action "-mesh2ls/mesh2sdf" now supports generation of asymmetric narrow bands.
 - Improved error reporting when using unsupported actions and options.
 - Actions can now have multiple names (or aliases), e.g. "-read", "-import", "-load", "-i".
 - Added cutoff option to "-ls2fog" action.
 - Improved error reporting in vdb_tool (prints diagnostic error messages).
 - Improved error handling in vdb_tool (will recover from non-fatal errors).
 - Improved parsing of configuration files (tabs and trailing comments are now ignored).
 - Improved unit-tests and Doxygen documentation for all files related to the vdb_tool.
 - Improved format and information for the -print action: output is now a column-aligned Unicode table, and a new "level" option controls detail. level=0 is the base table (age, name, type, class, dim, voxels, dx, size), now including the grid class (level set, fog volume, etc.). level=1 adds the active-voxel bounding box in both index space and world space and a per-level tree node-count column labeled "32->16->8" (e.g. "2->10->58" = 2 upper-internal, 10 lower-internal, 58 leaf nodes). level=2 additionally reports the background value, the active value range, and -- for level sets -- the world-space surface area and enclosed volume (via tools::levelSetArea / tools::levelSetVolume).
 - Action "-if" now accepts infix and RPN test expressions (via the Calculator), in addition to the existing 0/1/true/false literals and {...} expressions. For example "-if 'n > 0'" works where the variable n is read from the Processor's string memory, so "-for n=0,10 -if 'n > 5' ... -end" branches per iteration.
 - The Calculator (used by -calc and the -forAllValues/forOnValues/forOffValues kernels) now supports the compound-assignment operators +=, -=, *=, /= and %=, which desugar to "x = x <op> (rhs)" with the right-hand side parenthesized so precedence is preserved (e.g. "x *= a + b" means "x = x * (a + b)"). This makes in-place updates read naturally, e.g. "-forOnValues 'v += 1'" or "-calc 'a = 5; a += 3'". A compound operator with a non-identifier target (e.g. "a + b *= c") still reports an invalid-assignment-target error.
 - The Calculator now accepts OpenVDB AX-style attribute/grid sigils, so a simple scalar kernel is portable between -forValues and -ax: a leading "@" plus any single-character AX type prefix (b/i/l/f/d/v/m/s) is stripped, mapping "f@sphere" or "@density" to the plain variable "sphere"/"density". In -forValues the iterated grid can now be addressed by its own name (in addition to the "use=" variable), so "-forOnValues 'f@sphere += 1'" (on a grid named "sphere") and "-ax 'f@sphere += 1;'" run the same kernel through two different engines. Neighbor access via the grid name (e.g. "sphere(1,0,0)") is likewise recognized. Note this is a one-directional convergence of the common scalar subset; AX-only features (control flow, functions, vectors, point attributes) remain exclusive to -ax.
 - The expression-based actions "-calc", "-forAllValues/forOnValues/forOffValues" and "-ax" gained a "file=" option that reads the kernel/expression/code from a file instead of the inline option (useful for longer programs). If both the inline option and "file" are given, "file" takes precedence.
 - Actions "-for", "-each" and "-files" gained a "quiet" option (default false) that suppresses the per-iteration "Processing: ..." line for that loop without silencing the rest of the pipeline (unlike the global -quiet).
 - Action "-log" now redirects std::cerr and std::cout in addition to std::clog (so warnings/errors and -print output also land in the log file), uses unit-buffered output so "watch"/"tail -f" see updates in real time, writes a self-describing header (date, version, full command line), and gained "append" (append rather than truncate) and "tee" (also echo to the terminal, default true) options.

  Fixes:
 - Fixed issues in Geometry::readSTL so it works on all files in Thingi10K.
 - Fixed bug when reading ascii ply files and off files with normals.
 - Fixed out-of-bounds read in -slice when the grid's active values had been modified (e.g. by -forOnValues) so that sampled values fell outside the precomputed extrema range; the LUT index is now clamped to [0, 255].
 - Fixed several typos in the documentation of vdb_tool.
   Fixed vdb_tool::uuid so it generates a 36-character RFC 4122 v4 UUID.
 - Fixed corner cases in getFile/getName/getPath/getExt/replacePath/replaceExt by adopting <filesystem>.
 - Fixed handling of whitespace in greedy-anonymous action values (e.g. "-calc a = 1+2" or an -if/-switch expression with spaces) so that shell- or config-file-tokenized arguments are space-joined rather than comma-joined, which previously corrupted the expression.


richhones:
Very high level comment, the design of the kernel and forallvalues (and other iterators) actions looks like it provide a similar interface to what you might do with VDB AX, can you not chain the use of vdb_tool through the vdb_ax cli to provide the same functionality? I can see that some things that are missing in current AX functionality are supported (user-def functions for example) but would love to get efforts towards getting these working in AX instead of making a tool that overlaps in functionality.

kmuseth:
> Very high level comment, the design of the kernel and forallvalues (and other iterators) actions looks like it provide a similar interface to what you might do with VDB AX, can you not chain the use of vdb_tool through the vdb_ax cli to provide the same functionality? I can see that some things that are missing in current AX functionality are supported (user-def functions for example) but would love to get efforts towards getting these working in AX instead of making a tool that overlaps in functionality.

Great observation @richhones, and that has definitely crossed my mind. In fact, that is the goal, but there are two reasons why it shouldn't block this PR.

First, this PR is carefully crafted to introduce a lot of new features without any new mandatory dependencies. As you know, AX requires LLVM as a major new dependency. Second, while this implementation of stream processing does conceptually overlap with AX, it is a lightweight stand-in for fast prototyping. It has no JIT compilation—so it's presumably much slower—and is by no means a replacement for AX.

So, yes, I completely agree AX should be integrated into vdb_tool moving forward, but I'm not convinced it should block this already massive PR.

However, I'm happy to prioritize AX integration as the next improvement (though the memory management between scoped variables in AX and the vdb_tool will be nontrivial)

kmuseth:
CI is failing on the Windows build, but strangely it problem appears to be unrelated to changed made in this PR (vdb_view doesn't compile due so some GL issues). Any ideas?

kmuseth:
@richhones  - can you let me know what you think about my response (to your comment) above? More to the point, do you want me to add AX support to the AX tool in this PR or later? I don't mind either way, but it might make the review harder :)

kmuseth:
@richhones  - never mind - it turned out to be pretty straight forward to integrate AX into this tool, which I did in the latest commit. As expected it is faster that the existing -forOnValues action, but also requires LLVM - so I made it optional.

nanovdb/nanovdb/math/HDDA.h
swahtz, line 1: **❓ question:** ‏Should these changes in nanovdb be in a separate PR?  As I go reading through the changes in vdb tool, I think all these changes in `nanovdb/` (renaming ZeroCrossing, additional raytrace_iso_surface, etc.) are unrelated to the vdb tool work and are not mentioned in the `pendingchanges`.  It probably makes the most sense to put this in a separate PR

openvdb_cmd/vdb_tool/include/Calculator.h
```diff
@@ -0,0 +1,2184 @@
...
+                "Calculator: assignment ('=') and multi-statement (';') "
+                "require infix syntax; remove ':'/'$' or drop the assignment");
+        }
+        this->compileStatements(expr_);
+    } else if (hasRPNMarker) {
+        this->compileRPN(expr_);
+    } else {
+        this->compileInfix(expr_);
+    }
+    this->foldConstants();// must run BEFORE lazifyBranches: fold rewrites the
```
swahtz, line 664: **🚩 issue:** ‏I believe there should be an additional `verify()` before the optimization passes.  `lazifyBranches()` runs assuming the input notation produces a well-formed stack and when the notation is ill-formed, this can cause silent corruption of the heap or hangs before `verify()` can validate this and produce a meaningful error.  I compiled with ` -fsanitize=address` and tried various incorrectly formed `-calc` arguments like `-calc '1:+'` or `'1:if'` or `'min(1)'` and got heap overflow crashes in `lazifyBranches`.  In our regular optimization mode, I sometimes get a hang with these incorrectly formed arguments and sometimes I get through to the `verify()` message that there's a stack underflow on the op.


```suggestion
    this->verify();// validate raw front-end bytecode BEFORE the optimization
                   // passes: lazifyBranches indexes originStack assuming the
                   // stack is well-formed, which only verify establishes.    
     this->foldConstants();// must run BEFORE lazifyBranches: fold rewrites the
```

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -2091,58 +2895,331 @@ void Tool::levelSetToMesh()
...
-        for (size_t i = 0, I = poly.numQuads(); i < I; ++i) quad[qIdx++] = poly.quad(i);
-        for (size_t i = 0, I = poly.numTriangles(); i < I; ++i) tri[tIdx++] = poly.triangle(i);
+
+      // Per-voxel lambda. Each TBB worker reuses its thread-local cached
+      // accessors via accessors[k]->local() for fast sequential reads.
+      auto kernel_fn = [&calc, &base, &centers, &neighbors, &accessors](auto &it) {
+          constexpr size_t kMaxVars = 64;
+          OPENVDB_ASSERT(base.size() <= kMaxVars);
+          float values[kMaxVars];
+          for (size_t i = 0; i < base.size(); ++i) values[i] = base[i];
```
swahtz: **🚩 issue:** ‏ `OPENVDB_ASSERT` is only going to guard if the number of variables is <=64 in debug builds, otherwise we'll have an overflow here.  To not impact performance and keep this check out of the kernel function, I believe we could alternatively safely check after `calc.compile(kernel);` on line 3005 if the number of variables exceeds the max limit and throw an appropriate error there.

openvdb/openvdb/tools/LevelSetUtil.h
```diff
@@ -2599,6 +2628,150 @@ segmentSDF(const GridOrTreeType& volume, std::vector<typename GridOrTreeType::Pt
...
+    const ValueType interiorWidth = mm.min();
+
+    distTree.root().setBackground(exteriorWidth, /*updateChildNodes=*/false);
+    signedFloodFillWithValues(distTree, exteriorWidth, interiorWidth);
+
+    grid.setGridClass(GRID_LEVEL_SET);
+
+    // Optionally rebuild a symmetric narrow band via PDE renormalization.
+    if (rebuildNarrowBand) {
+        const int halfWidth = 3; // standard narrow band half-width in voxels
```
swahtz: **💡 suggestion:** ‏I think it would make sense to take the halfWidth value as an argument for future use.

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -2091,58 +2895,331 @@ void Tool::levelSetToMesh()
...
-        auto &polygons = polygonPoolList[i];
-        numTri  += polygons.numTriangles();
-        numQuad += polygons.numQuads();
+      // Snapshot the OUTPUT grid only if the kernel reads non-zero offsets
+      // from it: otherwise parallel writes to the iterator's grid would
+      // race with neighbor reads from the same grid. Input grids are
+      // read-only so they don't need a snapshot.
+      const bool needSnapshot = std::any_of(neighbors.begin(), neighbors.end(),
+          [](const NeighborBinding &n) {
+              return n.gridIdx == 0 && (n.dx || n.dy || n.dz);
```
swahtz: **🚩 issue:** ‏`needSnapshot` will trigger for `gridIdx == 0` but won't catch all aliases.  When running `-forOnValues 'a + b(1,0,0)' vdb=0,0 use=a,b` would cause the same data race issue.  We could check the grids' pointer identity here instead of index to trigger `needSnapshot`.

nanovdb/nanovdb/math/HDDA.h
kmuseth: > **❓ question:** ‏Should these changes in nanovdb be in a separate PR? As I go reading through the changes in vdb tool, I think all these changes in `nanovdb/` (renaming ZeroCrossing, additional raytrace_iso_surface, etc.) are unrelated to the vdb tool work and are not mentioned in the `pendingchanges`. It probably makes the most sense to put this in a separate PR

good question. These changes are used in nanovdb/examples/ex_raytrace_iso_surface, which in turn have been instrumental in debugging and testing the new tools/PolySoupToLevelSet.h. So while at first glance it might seem unrelated, I would still argue that they belong in this PR. Regardless, you're definitely right that the change-log should be updated to reflect these changes (I'll fix that now).

openvdb/openvdb/python/app/shrink_wrap.py
apradhana, line 1: I now think the shrinkwrap script can fit well as part of a broader Python examples effort.

The most valuable part of the addition is that `convertPolygonSoupToLevelSet()` is the real library feature. It is a low-level binding: polygon soup arrays go in, level-set grids come out. That belongs in the OpenVDB Python API.

`shrink_wrap.py` is different. It is a workflow example. It shows how someone might use the binding in a real USD pipeline: read a USD file, collect meshes, bake transforms, triangulate faces, convert to a level set, optionally convert back to mesh, and write output files. That is useful, but it should not look like an installed or officially supported OpenVDB app unless the project intends to test and maintain it that way. 

Ideally, in the light of agentic coding, we add a `vdb_examples` folder sitting at the same level of `openvdb_cmd` folder. We then add `vdb_examples/python/shrink_wrap.py` in that folder along with other examples that uses both the Openvdb-python-binding and Nanovdb-python-binding. I'd go one step further and scope it out as collecting examples that uses Open/NanoVDB-python-binding + USD + Numpy workflows. A good source of such examples is the Houdini toolkit example hip file contributed by Jeff Budsberg listed in https://www.openvdb.org/download/. This is useful because we can add this into the CI and users can point their agents to this folder as examples workflow when they use VDB in Python. However, this is a tall order. 

As a middle ground, I would prefer moving it from python/app into something like:
`openvdb/openvdb/python/examples/shrink_wrap/`
with a clear README saying:
```This is example code for learning and experimentation. It is not installed as part of the OpenVDB Python package and is not a stable API.```
That makes the script valuable without quietly expanding what OpenVDB maintainers are responsible for. The binding stays library-shaped, while the shrinkwrap script becomes a practical example.

File: openvdb_cmd/vdb_tool/include/Geometry.h [hunk 5]
```diff
@@ -784,7 +1051,11 @@ void Geometry::readPLY(std::istream &is)
                 mQuad.emplace_back(vtx);
                 break;
             default:
-                throw std::invalid_argument("Geometry::readPLY: binary " + std::to_string(n) + "-gons are not supported");
+                if (n > nGon) throw std::invalid_argument("Geometry::readPLY: binary " + std::to_string(n) + "-gons are not supported");
+                if (mVerbose) std::clog << "Geometry::readPLY: binary triangulating " << n << "-gon\n";
```
swahtz, line 1055: **❓ question:** ‏Could `readPLY` also use `Geometry::triangulate` like `readOBJ` does?

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -961,42 +1417,49 @@ void Tool::clear()
...
 #endif
-      throw std::invalid_argument("File \""+fileName+"\" has an invalid extension");
-      break;
+        throw std::invalid_argument("File \""+fileName+"\" has an invalid extension");
+      }// end switch
+    }// end for loop over files
+  } catch (const std::exception& e) {
+    if (mErrorOnWarning) {
+      throw std::invalid_argument(action_name + ": " + e.what());
+    } else {
```
swahtz: **💡 suggestion:** ‏This try/catch block appears duplicated each of the `Tool::xxx()` functions, sometimes with slightly different text (like 'do to' or 'due to'), could we just move the whole block  to the call site in `Parser.h` and remove the duplication?

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -386,58 +630,199 @@ void Tool::init()
...
+     {"use", "v", "v|x|x,y", "name(s) of the kernel variable(s) bound to the voxel value(s) of the input grid(s). Defaults to \"v\". Accepts a comma-separated list matching vdb=: use=x,y vdb=0,1 makes \"x\" the output grid and \"y\" a read-only input. Each name is excluded from the Processor-memory lookup and may be called as a function (e.g. \"x(1,0,0)\") to read a relative neighbor through a per-thread ConstAccessor."},
+     {"class", "", "ls", "class label of the output volume."},
+     {"background", "", "1.5,2.0", "background value(s) of the output volume. If two values are provided they are assumed to be outside, inside"},
+     {"name", "", "foo-bar", "name assigned to the output volume"},
+     {"file", "", "prog.txt", "read the kernel from a file instead of the \"kernel\" option (useful for longer kernels). If both are given, \"file\" takes precedence."}},
+     [&](){mParser.setDefaults();}, [&](){this->forValues();},
+     /*anonymous=*/2, /*greedy=*/true);// kernel (index 2) may itself contain '='; accept bare "x+1" or "t=x*x; t+1" alongside kernel='...'
+
+  mParser.addAction(
```
swahtz: **💅 polish:** ‏forAllValues/forOnValues/forOffValues all have the same options, we could collapse these `addAction` registrations down into a loop which would help prevent any drift.

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -386,58 +630,199 @@ void Tool::init()
...
+    {{"dim", "", "256", "largest dimension in voxel units of the mesh bbox (defaults to 256). If \"vdb\" or \"voxel\" is defined then \"dim\" is ignored"},
+     {"voxel", "", "0.01", "voxel size in world units (by defaults \"dim\" is used to derive \"voxel\"). If specified this option takes precedence over \"dim\""},
+     {"width", "", "3.0", "half-width in voxel units of the output narrow-band level set (defaults to 3 units on either side of the zero-crossing)"},
+     {"geo", "0", "0", "age (i.e. stack index) of the geometry to be processed. Defaults to 0, i.e. most recently inserted geometry."},
+     {"vdb", "-1", "0", "age (i.e. stack index) of reference grid used to define the transform. Defaults to -1, i.e. disabled. If specified this option takes precedence over \"dim\" and \"voxel\"!"},
+     {"keep", "", "1|0|true|false", "toggle wether the input geometry is preserved or deleted after the conversion"},
+     {"name", "", "mesh2udf_input", "specify the name of the resulting vdb (by default it's derived from the input geometry)"}},
+     [&](){mParser.setDefaults();}, [&](){this->meshToUnsignedDistanceField();});
+
+#ifdef VDB_TOOL_USE_SHRINKWRAP
```
swahtz, line 747: **❓ question:** ‏Is it intentional to keep this feature behind this gate that's not exposed via CMake?  I think we should just remove this gating if this feature is meant to be shipped in this PR.

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+
+//////////////////////////////////////////////////////////////////////////
+
+template<typename GridType>
+auto PolySoupToLevelSet<GridType>::shrinkWrap(GridType &grid, const GridType &gridB, float &d)
+{
+    const float maxDist = 2.0f;
+    LevelSetFilter<GridType> filter(grid);
+    filter.setNormCount(3);// halfWidth
+#if 1//first-order
```
swahtz, line 428: **💡 suggestion:** ‏These seem like user-drivable options, it seems like they should be selectable by an enum.

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////////
+
+template<typename GridType>
+template<class ShrinkWrapT, class ProgressT>
+void PolySoupToLevelSet<GridType>::process(const ShrinkWrapT &D, ProgressT *progress, int offset_mode)
+{
+    auto myProgress = [&](const std::string &s){if constexpr(!std::is_same<ProgressT,void>::value) if (progress) (*progress)(s);};
+
+    if (progress) std::cerr << std::endl;
```
swahtz: **☑️ todo:** ‏This should likely just be removed (a development debug?).

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz: Actually, I'm wondering if we can DRY this code with the same logic that appears in `distanceFieldToSDF` in `LevelSetUtils.h` in the narrow-band rebuild which has the same `#if` block over first-order/higher-order.  I'd propose both functions could use a same common `rebuildSymmetricNarrowband` (or some similar name) in `LevelSetUtils.h`

swahtz (changes requested):

nanovdb/nanovdb/examples/ex_raytrace_iso_surface/main.cc
```diff
@@ -0,0 +1,51 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: Apache-2.0
+
+#include <algorithm>
+#include <iostream>
```
apradhana, line 6: I believe we need #include <cstring> for `std::strcmp`.

openvdb_cmd/CMakeLists.txt
```diff
@@ -79,6 +80,10 @@ if (OPENVDB_BUILD_VDB_VIEW)
   add_subdirectory(vdb_view)
 endif()
 
+if (OPENVDB_BUILD_VDB_VIZ)
+  add_subdirectory(vdb_viz)
```
apradhana: Did you mean to add `vdb_viz` directory as part of this PR?

nanovdb/nanovdb/math/HDDA.h
swahtz: All good, whatever is gong to be easiest to look at in the history for retrospect.  Adding to the PR descrition/change-log would be great.

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+    switch (mode) {
+    case 0:// algorithm presented in the paper, using mesh<-> VDB round-trip
+        grid = meshToUnsignedDistanceField<GridType>(*xform, mPoly.vtx, mPoly.tri, mPoly.quad, mHalfWidth);// mesh -> UDF
+        volumeToMesh(*grid, mPoly.vtx, mPoly.tri, mPoly.quad, /*iso*/dx, /*adapt*/0.0);// UDF -> mesh (clears and re-allocates mesh)
+        grid = meshToLevelSet<GridType>(*xform, mPoly.vtx, mPoly.tri, mPoly.quad, mHalfWidth);// mesh -> SDF
+        break;
+    case 1:// algorithm using Mihai's signed flood-fill algorithm
+        grid = meshToUnsignedDistanceField<GridType>(*xform, mPoly.vtx, mPoly.tri, mPoly.quad, mHalfWidth + 1);// mesh -> UDF
+        tools::foreach(grid->beginValueOn(), [dx](const typename GridType::ValueOnIter& it){it.setValue(*it - dx);}, /*threaded*/true, /*share functor*/true);
+        //tools::changeBackground(grid->tree(), mHalfWidth*dx);
```
swahtz, line 392: **🧹 chore:** ‏Can remove these commented out lines of code here and below

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+template<typename GridType, class ShrinkWrapT = ShrinkWrapLimit, class ProgressT = void>
+typename GridType::Ptr
+polySoupToLevelSet(
+    PolySoup &&poly,
+    int dim = 256,
+    float voxelSize = 0.0,// invalid so use dim instead
+    const ShrinkWrapT &D = ShrinkWrapT(),
+    float halfWidth = float(LEVEL_SET_HALF_WIDTH),
+    ProgressT *progress = nullptr,
+    int offset_mode = 0);
```
swahtz, line 91: **☑️ todo:** ‏`offset_mode` isn't described in the docstrings of this function (or in the overloads below).  Also, since there are 3 modes, I'd suggest we change the type to an enum with descriptive names.

openvdb/openvdb/python/app/shrink_wrap.py
```diff
@@ -0,0 +1,369 @@
+#!/usr/bin/env python3
```
swahtz: **☑️ todo:** ‏Should include the SPDX identifier at the top


```suggestion
#!/usr/bin/env python3
# Copyright Contributors to the OpenVDB Project
# SPDX-License-Identifier: Apache-2.0
```

openvdb/openvdb/python/app/usd_utils.py
```diff
@@ -0,0 +1,273 @@
+import numpy as np
```
swahtz: **☑️ todo:** ‏Should include the SPDX identifier at the top


```suggestion
# Copyright Contributors to the OpenVDB Project
# SPDX-License-Identifier: Apache-2.0
import numpy as np
```

nanovdb/nanovdb/math/HDDA.h
kmuseth: done

openvdb/openvdb/python/app/shrink_wrap.py
swahtz: Fwiw, in updating NanoVDB python bindings in those PRs #2219  #2225, I've been pushing out a bunch of example python scripts (and updating the ones that exist) as I go along (into `nanovdb/nanovdb/python/examples`. Would love to get your review on how helpful those are… the goal was to either mirror the utility in the C++ examples (to show how you'd implement the same things in Python) or demonstrate the usage of the novel Python API surfaces.

I think one PR that's key for docs is getting #2217 merged because right now code like the NanoVDB HelloWorld up on the website/docs won't compile… that code has drifted from the actual APIs.  #2217 also includes a test for Doxygen warnings/failures and changes how we pull example code into docs by referencing from live code that gets compiled and tested in CI to hopefully prevent drift and keep docs accurate.

kmuseth: @apradhana  - let's talk. I didn't add this python script - I suspect @swahtz  did.  More to the point, I'm not sure what you're asking me to change in this PR :)

openvdb_cmd/vdb_tool/include/Calculator.h
kmuseth: Good catch - [I fixed this in a new commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/9df426f576a9c0c3b2e630bb2a8276c30006ea66): Swapped the order — verify() now runs before lazifyBranches() so malformed input is caught early with a clear error message (e.g., "stack underflow on binary op").  I couldn't use your suggested change since the correct order is foldConstants(), verify(), and finally0 lazifyBranches().

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: This issue, while legit, was address in another commit. Thanks tho!

openvdb/openvdb/tools/LevelSetUtil.h
kmuseth: good call - [merged in your suggested change to the API](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/1125e2e299ecd75843df8284537695339e0dc27c). Also changed the halfWidth to a float vs int argument, which is more consistent with other functions, eg. meshToVolume.

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: another great catch! [I've fixed this](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/72269cc77a380a1a56607385e5434b65000d2443) by useing pointer identity comparison to catch all aliases of the same grid, not just the primary index.

openvdb_cmd/vdb_tool/include/Geometry.h
kmuseth: good idea - refactored Geometry::readPLY() to use Geometry::triangulate() in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/7714a7507c158b0038957f1c14609a7634e3f65b).

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -2389,17 +3506,168 @@ void Tool::scatter()
...
+  const std::string &action_name = mParser.getAction().names[0];
+  OPENVDB_ASSERT(action_name == "movie");
+  try {
+    const int fps = mParser.get<int>("fps");
+    const std::string input = mParser.get<std::string>("input");
+    const std::string output = mParser.get<std::string>("output");
+    const VecI scale = mParser.getVec<int>("scale", "x");
+    //const bool keep = mParser.get<bool>("keep");
+    const std::string flip = mParser.get<std::string>("flip");
+    if (mParser.verbose) mTimer.start(action_name);
```
apradhana: Please consider using VDB_TOOL_FFMPEG_PATH defined in the CMakeLists.txt. Perhaps something like:
```cpp
std::string cmd("\"" VDB_TOOL_FFMPEG_PATH "\""), log("log.txt");
```
To accomplish this, we pass resolved path in the CMakeLists 
```cmake
if(OPENVDB_TOOL_USE_MPEG)
  target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_MPEG")
  find_program(FFMPEG_PATH NAMES ffmpeg REQUIRED)
  target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_FFMPEG_PATH=\"${FFMPEG_PATH}\"")
endif()
```
into the header file `Tool.h` and provide a fallback like so on top of the header file:
```cpp
#ifndef VDB_TOOL_FFMPEG_PATH
#define VDB_TOOL_FFMPEG_PATH "ffmpeg"
#endif
```

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: [Pushed up a major consolidation](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/c881959160eec4a8a2e1d076b8d800f4d0b3785a) of error handling from individual actions in vdb_tool. Removed 253 lines of duplicated error-handling code while preserving identical behavior.

openvdb/openvdb/python/app/shrink_wrap.py
swahtz: Mmmm, nope I don't recall writing that, I don't believe I've worked on this branch.

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: good idea - [made this simplification here](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/dca4ccb3a53c56d9cea55696725c740525dab928)

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: This is very much intentional - I'll explain in person

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+
+template<typename GridType>
+PolySoupToLevelSet<GridType>::PolySoupToLevelSet(PolySoup &&poly, int dim, float width)
+    : mPoly(poly), mHalfWidth(width)
+{
+    if constexpr(!std::is_floating_point<typename GridType::ValueType>::value) {
+        OPENVDB_THROW(TypeError, "polySoupToLevelSet: supported only for scalar floating-point grids");
+    }
+    if (!mPoly.bbox) mPoly.bbox = PolySoupToLevelSet::getBBox(mPoly.vtx);
+    const float maxLength = mPoly.bbox.extents()[mPoly.bbox.maxExtent()];
```
apradhana, line 275: Please consider sanitizing with `OPENVDB_THROW` for inputs violating:
- halfWidth is  positive
- bbox is non-empty and has positive extent
- computed mMinVoxelSize / mMaxVoxelSize are finite and positive

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -0,0 +1,484 @@
...
+
+template<typename GridType>
+PolySoupToLevelSet<GridType>::PolySoupToLevelSet(PolySoup &&poly, float voxelSize, float width)
+    : mPoly(poly), mMinVoxelSize(voxelSize), mHalfWidth(width)
+{
+    if constexpr(!std::is_floating_point<typename GridType::ValueType>::value) {
+        OPENVDB_THROW(TypeError, "polySoupToLevelSet: supported only for scalar floating-point grids");
+    }
+    if (!mPoly.bbox) mPoly.bbox = PolySoupToLevelSet::getBBox(mPoly.vtx);
+    const float maxLength = mPoly.bbox.extents()[mPoly.bbox.maxExtent()];
```
apradhana, line 309: Same as above, please consider sanitizing bbox is non-empty.

openvdb/openvdb/python/app/shrink_wrap.py
kmuseth: interesting ... I wonder who ... maybe Rohan? Let me check

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: The #if blocks in PolySoupToLevelSet.h and LevelSetUtils.h are un-related (numerically speaking), and as such should be independent options. They are not exposed in the API since we are still researching the optimal settings for these algorithm. In the future the #if blocks will be removed, but for now they have proven super useful. So, in short I prefer to keep the code as is and then revisit it in the near future.

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: fix this issue (including removed the line) in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/19f54774ca80e21175d425dd19ee2a0f1bbf574f)

nanovdb/nanovdb/examples/ex_raytrace_iso_surface/main.cc
kmuseth: nice one - addressed in t[his commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/dd6d6bddddd0b77e165eb722ace00557ebc87aba)

openvdb_cmd/CMakeLists.txt
kmuseth: Good catch - this was left in by mistake. I've removed it in t[his commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/fe859fb015fc2b11502ac30440299a54ec8e7eb2).

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: I would very much like to keep these lines of code for a little longer since we are still experimenting with the optimal implementation. The current version has proven useful (and it is already in use), but there is still work left to potentially optimize it, which is why I'd like to leave this code for now. In general, I of course agree that dead code should be removed, but I'd argue this is a special case where it does make sense, and more to the point, it will eventually be removed.

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: It's not a bad idea at all, but the problem is that long term I expect to remove this argument once the optimal mode (algorithm) has been identified. Adding an enum feels like a step in the wrong direction. How about instead changing the documentation to reflect that mode is an experimental argument that will soon be deprecated and therefore mode should only be used by experts? [Done in this commit.](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/ab09f62cfaa69dbe2b38a54ce108d69fd5b7cffe)

openvdb/openvdb/python/app/shrink_wrap.py
kmuseth: After talking to Andre, we decided to simply leave out all files in openvdb/python/app for now. They will be moved to another location in another PR. @swahtz, are you okay with that? Rohan added them, with my blessing, but I forgot, so it's my fault that it ended up in this PR by mistake. The problem is it breaks with the lower-level Python bindings we have at this location (it's more of an App), so I agree with Andre that it belongs elsewhere. I'll talk to Rohan, and the two of you (@apradhana and @swahtz) should coordinate on the low-level Python bindings. I've removed these files in[ this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/157cf39b4d96b428c45c646dc9f151dd594978d8).

openvdb/openvdb/python/app/shrink_wrap.py
kmuseth: as mentioned above this file has been removed from this PR

openvdb/openvdb/python/app/usd_utils.py
kmuseth: as mentioned above this file has been removed from this PR

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: great idea Andre - implemented in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/3f00a94f64baf87d8a234fab309e834e8cb1316a)

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz: Thanks

openvdb/openvdb/python/app/shrink_wrap.py
apradhana: Thank you for removing the `usd` dependency from this PR, @kmuseth ! I will work with @rohan-sawhney and @swahtz to include the example and more python bindings in separate PRs.

openvdb_cmd/vdb_tool/include/Tool.h
apradhana: Thank you!

openvdb/openvdb/tools/PolySoupToLevelSet.h
apradhana: This is non-blocking for me.

openvdb/openvdb/tools/PolySoupToLevelSet.h
apradhana: This is non-blocking for me.

apradhana:
Thanks for addressing the reviews. The rest of the comments I made are non-blocking. As long as DCO is passing, we can merge this PR. 

There are two ways to address this:
(1) Ask @sifakis , @areidmeyer , and @rohan-sawhney to add the exact Signed-off-by: lines they approve. The commits that are not passing DCOs are listed [here](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/checks?check_run_id=86492498462)
(2) @kmuseth  can squash/re-author those commits as his own commits and add:
```
```
The attributions for the contributions can be preserved with:
```
Co-authored-by: Rohan Sawhney <...>
```

apradhana (approved):

openvdb/openvdb/python/app/usd_utils.py
swahtz: OK thanks

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz: OK, thanks, I thought the selection of algorithm was going to be a user-drivable one depending on their preference.  If that's the case, then just the docstring change clarifying is great.

openvdb/openvdb/python/app/shrink_wrap.py
swahtz: Oh ok, great.  Removing them is fine with me, happy to keep this PR more focused, and fitting them into a separate python focused PR of examples/apps is all good.

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz: OK, I assumed this was in its final form and was going off the code style guide's "don't leave historical code in comments" guideline.

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz: I see, ok, I didn't know the settings/mode are an area of active changes.

swahtz (approved):
Thanks for addressing my notes, lgtm!

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: [addressed here](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/dd018b7e8e41113fdb1b724263503c4d49a893bd)

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: resolved in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2241/commits/dd018b7e8e41113fdb1b724263503c4d49a893bd)

rohan-sawhney:
I certify this contribution under the DCO and authorize Ken to add my Signed-off-by line.

areidmeyer:
I certify this contribution under the DCO and authorize Ken to add my Signed-off-by line.

sifakis:
I certify this contribution under the DCO and authorize @kmuseth  to add my Signed-off-by line.

Issue #2232, opened 2026-06-11
NanoVDB: injectable CUDA memory resources — design & roadmap


harrism:
## Summary

NanoVDB's CUDA path hard-wires its allocations — grid storage via `GridHandle<BufferT>`, scratch via `cuda::TempPool`/`cuda::DeviceResource`, and raw `cudaMallocAsync`/`cudaMallocHost` in builders. This tracks a layered design to let downstream inject its own allocator wherever NanoVDB allocates, aligned with CCCL's [`cuda::mr`](https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/memory_resource.html) / [`cuda::buffer`](https://nvidia.github.io/cccl/unstable/libcudacxx/runtime/buffer.html) model — without a hard [CCCL](https://github.com/NVIDIA/cccl) dependency. New APIs land alongside the existing ones; the legacy dual buffer + handle surface are deprecated and removed a release or two later, so there is no hard break at introduction.

## Why

- **Inject your own allocator without forking headers** — downstream currently forks NanoVDB headers to route through their pool (e.g. PyTorch's [`c10::cuda::CUDACachingAllocator`](https://pytorch.org/docs/stable/notes/cuda.html#memory-management); cf. openvdb/fvdb-core#655, which vendored three headers).
- **Fix the pinned-host stall** — the synchronous `cudaMallocHost` inside `cuda::DeviceBuffer` serializes streams where a dual buffer is actually constructed host-side. Note this is *not* every `DeviceBuffer` use: `init` allocates host or device but never both, so buffers created with a real device id never touch `cudaMallocHost` (see the correction under B1).
- **Share one pool** across tensors, scratch, and grid buffers (caching; fewer fragmentation OOMs).
- **Familiar + future-proof** — method/type shapes match CCCL, so a `resource_ref` / `cuda::buffer` adapter is a thin shim later, while NanoVDB keeps its low CUDA floor.

## Design — three pieces, each a minimal NanoVDB analog of a CCCL standard

[...]

PR #2244, opened 2026-07-02
NanoVDB: encode points for any resource in PointsToGrid (CUDA)

harrism:
Follow-up to #2231, part of #2232 (NanoVDB injectable CUDA memory resources).

## What
`tools::cuda::PointsToGrid<Point, ResourceT>` now encodes points for **any** resource. Previously the point-encoder lived only in the `PointsToGrid<Point>` full specialization (= `PointsToGrid<Point, DeviceResource>`), so a **Point** build with a **custom** resource missed it and fell through to the generic no-op — silently producing a grid with no encoded point data, no error. This was the known limitation called out in #2231's follow-ups.

The encoder is folded into the generic `PointsToGrid<BuildT, ResourceT>::processPoints`, gated by `if constexpr (util::is_same<BuildT, Point>::value)`; the now-redundant full specialization is removed.

## Why existing paths are unchanged
- **Non-Point builds:** the `if constexpr` block is discarded, leaving the prior no-op body verbatim.
- **Default-resource Point builds:** `ResourceT == DeviceResource`, so the same encode kernels run with the same alignment — byte-for-byte the previous path.

The resource governs allocation only; the encode kernels operate on device data, so making the encoder resource-agnostic is correct.

## Testing
New `TestMemoryResource.PointsToGrid_PointEncodedWithCustomResource` builds a `NanoGrid<Point>` via `PointsToGrid` with a custom resource and asserts the encoded coordinates match the input — red on current code (captures the no-op), green after the fix. RTX 6000 Ada / CUDA 12.6: memory-resource suite 9/9, CPU `nanovdb_test_nanovdb` 153/153, `nanovdb_test_cuda` 53/53.

## Non-CUDA builds
No impact on `NANOVDB_USE_CUDA=OFF` — the change is confined to `tools/cuda/PointsToGrid.cuh` (`#if defined(__CUDACC__)`-gated) and a CUDA-only test.


harrism:
@kmuseth friendly ping — this has been open since July 2 with no review yet.

It's the smaller of the two follow-ups from #2231 (+117/-65, 2 files) and fixes a silent-wrong-output case: `PointsToGrid<Point, CustomResource>` currently falls through to the generic no-op and produces a grid with **no encoded point data and no error**, because the encoder only existed in the `PointsToGrid<Point>` full specialization. That was the known limitation called out in #2231's follow-up list.

Both existing paths are unchanged by construction — non-`Point` builds discard the `if constexpr` block, and default-resource `Point` builds run the same code that used to live in the specialization. New test covers the custom-resource `Point` combination that was previously silent.

Happy to rebase if it's gone stale.

kmuseth (approved):
looks good to me - good catch

Issue #2245, opened 2026-07-06
NanoVDB: two multi-GPU DistributedPointsToGrid failures (RadixSort merge sync; DenseLeaf illegal access)


harrism:
Two pre-existing multi-GPU failures in `nanovdb_test_mgpu` (`nanovdb/nanovdb/unittest/TestMultiGPU.cu`), independent of the memory-resource work. Both reproduce on a 2-GPU host and are **NanoVDB-side bugs in `tools/cuda/DistributedPointsToGrid.cuh`**, not a CCCL/CUB regression — CUB's `DeviceMerge::MergePairs` and the `CUB_LAUNCH` two-phase pattern were verified correct in isolation under the vendored CCCL 3.5.

### Environment
- 2× NVIDIA RTX 6000 Ada, driver 595.71.05, CUDA 12.6 (nvcc), CCCL/CUB 3.5.0 (vendored).
- Note: NanoVDB CUDA/mgpu tests are compiled but never executed on a GPU in CI (runners are GPU-less; every CUDA job runs `ctest -E ".*cuda.*|.*mgpu.*"`), so these only manifest on dev machines.

### Bug 1 — `TestNanoVDBMultiGPU.RadixSort` (deterministic)
Sortedness checks (`EXPECT_LE`/`EXPECT_GE`) fail — the output is not globally sorted. Single-GPU (`CUDA_VISIBLE_DEVICES=0`) **passes**, isolating the fault to the 2-GPU merge. The right-device merge under-produces: the right half of the output shows inversions and many never-written slots.

Root cause: a cross-device event-ordering hazard in `radixSortAsync` (~L217–267). `postEvents[deviceId]` is reused to signal both "per-device sort done" and "mergePathKernel done", so the right-device `DeviceMerge::MergePairs` can launch before the peer device's sorted output is actually visible.

### Bug 2 — `TestNanoVDBMultiGPU.DenseLeaf_DistributedCudaPointsToGrid_UnifiedBuffer` (flaky / order-dependent)
Illegal memory access (CUDA error 700). Passes in-suite, fails in isolation — order/state dependent. `compute-sanitizer --tool memcheck` pinpoints the real fault:

```
Invalid __global__ read of size 4 bytes
  at nanovdb::tools::cuda::VoxelKeyFunctor<...>::operator()(...)+0xe0 in tools/cuda/PointsToGrid.cuh:562
[...]

PR #2248, opened 2026-07-07
NanoVDB: handle empty partitions in the distributed merge path search

harrism:
## What

`tools::cuda::DistributedPointsToGrid`'s merge path search (`mergePath`) mishandles an empty partition. Its bounds checks compare against `keys1Count - 1` / `keys2Count - 1` on `size_t` counts, so a zero count underflows to `SIZE_MAX` and the guards never fire; the binary search then reads the empty array (and out of bounds past either array) and returns wrong or garbage intervals — an inverted split in one direction, underflow garbage in the other.

An empty partition is reachable from `radixSortAsync`'s ceil-split whenever a device receives no items — e.g. sorting a single element across two devices gives counts `[1, 0]`.

The fix returns the trivial split before the search: with an empty side, every element up to the diagonal comes from the non-empty side.

## Testing

New test `TestNanoVDBMultiGPU.MergePathEmptyPartition` — runs on a **single** GPU (`mergePathKernel` is a plain kernel over pointers), covering empty-right, empty-left, both-empty, and a non-empty control at the median diagonal. The adjacent memory is filled with controlled sentinels so the out-of-bounds search fails deterministically rather than by luck:

- Without the fix: empty-right returns the inverted split `(0, 2)` instead of `(2, 0)`; empty-left returns underflow garbage in both intervals.
- With the fix: all cases pass, and `compute-sanitizer --tool memcheck` reports 0 errors.
- Full `nanovdb_test_mgpu` suite: 8/8 pass (single GPU; RTX 6000 Ada, CUDA 12.6).

Found while investigating the (environmental, since closed) multi-GPU failures in #2245 — this defect is independent of that machine issue and affects any multi-GPU sort with an empty partition.

## Non-CUDA builds

No impact on `NANOVDB_USE_CUDA=OFF` — the change is confined to a CUDA-only header and the multi-GPU test.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
@@ -60,6 +60,17 @@ void mergePath(KeyIteratorIn keys1, size_t keys1Count, KeyIteratorIn keys2, size
     using key_type = typename ::cuda::std::iterator_traits<KeyIteratorIn>::value_type;
 
     const size_t combinedIndex = intervalIndex * (keys1Count + keys2Count) / 2;
+
+    // An empty partition makes the merge trivial: every element up to the
+    // diagonal comes from the non-empty side. Handle it before the binary
+    // search, which reads both arrays and whose unsigned (keysNCount - 1)
+    // bounds checks underflow when a count is zero.
+    if (keys1Count == 0 || keys2Count == 0) {
+        *key1Intervals = static_cast<ptrdiff_t>(combinedIndex < keys1Count ? combinedIndex : keys1Count);
+        *key2Intervals = static_cast<ptrdiff_t>(combinedIndex < keys2Count ? combinedIndex : keys2Count);
+        return;
+    }
+
     size_t leftTop = combinedIndex > keys1Count ? keys1Count : combinedIndex;
     size_t rightTop = combinedIndex > keys1Count ? combinedIndex - keys1Count : 0;
     size_t leftBottom = rightTop;
--- a/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
@@ -52,7 +52,21 @@ private:
     const T* mValue;
 };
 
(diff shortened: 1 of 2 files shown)


nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
```diff
@@ -60,6 +60,17 @@ void mergePath(KeyIteratorIn keys1, size_t keys1Count, KeyIteratorIn keys2, size
     using key_type = typename ::cuda::std::iterator_traits<KeyIteratorIn>::value_type;
 
     const size_t combinedIndex = intervalIndex * (keys1Count + keys2Count) / 2;
+
```
matthewdcong: Do we need to launch the kernel if either `keys1Count` or `keys2Count` is 0? I think this condition could be moved into the host code prior to kernel launch.

harrism: Good call — done in 3a8f7607. The launch site now skips the kernel when either count is zero and computes the trivial split on the host (`mergePathTrivial`, kept as a small named function so the unit test can exercise it directly); `mergePath` reverts to the unguarded search with the non-empty precondition documented. The test's empty cases are now pure host calls, with the device search kept as the non-empty control. Verified: `nanovdb_test_mgpu` 8/8 (single GPU), compute-sanitizer clean.

harrism: Indeed. Fixed in 3a8f760

harrism:
@kmuseth friendly ping — open since July 7, no review yet.

This one is a latent out-of-bounds read rather than a feature: `mergePath` compares against `keys2Count - 1` on a `size_t`, so a zero count underflows to `SIZE_MAX`, the guards never fire, and the binary search reads past the end of the array and returns garbage intervals.

An empty partition is reachable in practice — `radixSortAsync`'s ceil-split hands one device nothing whenever the item count is small relative to the device count (sorting a single element across two GPUs gives `[1, 0]`).

Worth noting for scheduling: the new test `MergePathEmptyPartition` runs on a **single** GPU, so it isn't blocked on multi-GPU hardware. It's +67/-1 across 2 files.

kmuseth (approved):
looks good

Issue #1, opened 2013-04-17
Can openvdb and openvdb_houdini be split?


davvid:
It seems like these two repos should be made into separate projects.  There may be situations where we want to release patch versions of the openvdb code (or Houdini plugins) without revving the other side; keeping them separate makes this easier to manage.

PR #2249, opened 2026-07-14
NanoVDB: correctness and determinism fixes

swahtz:
## Summary

Five correctness/determinism fixes to the NanoVDB CUDA tools, plus regression tests. Bug fixes only — no API changes, no behavior change for already-correct callers.

## Fixes

**1. GridChecksum — out-of-bounds device read + host stack corruption**
`evalChecksum` copied `sizeof(GridData)+sizeof(TreeData)` (736 B) out of a device buffer that only holds a single 4-byte CRC, into the 4-byte head/tail of a stack `Checksum`. That's both an out-of-bounds device read and a host stack smash (compute-sanitizer flags it; some configs hard-fail with a CUDA "invalid argument"). Now copies `sizeof(uint32_t)` at all four sites, matching `validateChecksum`.

**2. Stream propagation — work silently escaping to the default stream**
The 26-neighbour (`NN_FACE_EDGE_VERTEX`) leaf-dilation kernel and SignedFloodFill's root pass launched and synchronized on the default stream instead of the caller's stream, silently corrupting output for callers on a non-blocking stream. Both now use the operation's stream.

**3. Grid-name copy — up to 255 bytes of host heap over-read into the output**
The name copy read a fixed `MaxNameSize` (256) bytes from a possibly shorter `std::string` — whose `.data()` is never null, so the old memset fallback was dead code — over-reading host heap into the grid and into any written `.nvdb`. Now zeroes the field and copies only `min(name.size()+1, MaxNameSize)`. Applied to PointsToGrid, DistributedPointsToGrid, and MeshToGrid.

**4. Deterministic output — recycled pool bytes leaking into grids/files**
PointsToGrid and IndexToGrid left alignment padding and stats fields (absent when the source has no stats) holding whatever the recycled device pool last contained, making output nondeterministic and leaking host heap into files.
- **PointsToGrid:** the grid buffer is zeroed once after allocation, which also makes the three dense background-fill passes (upper/lower value tables, and inactive-leaf-values for every build type except Point) redundant, so they're dropped — net-neutral on runtime, measured against master (same harness, 30 iters / 10 warmup, RTX PRO 6000 sm_120, median):

  | op | dataset | master | this PR | delta |
  |---|---|--:|--:|--:|
  | `p2g_onindex_n23347893` | dragon | 10.700 ms | 10.581 ms | −1.1% |
  | `p2g_onindex_n93391572_dup4` | dragon | 42.949 ms | 42.402 ms | −1.3% |
  | `p2g_onindex_n373566288_dup16` | dragon | 169.011 ms | 168.292 ms | −0.4% |
  | `p2g_onindex_n96956688` | emu | 47.592 ms | 46.996 ms | −1.3% |

  (marginally favourable rather than exactly neutral, by about the width of run-to-run noise; the master side needs `--checksum=skip` to benchmark at all, since it trips the crash fixed in 1 above)
- **IndexToGrid:** the initialization is scoped to only the bytes actually left unwritten — a memset of the non-leaf region `[0, node[0])` plus a per-leaf zeroing of the stats/padding gap in `processLeafsKernel` (every `mValues[i]` is written anyway). This is **byte-identical** to a full zero-init (all `indextogrid` goldens pass; compute-sanitizer memcheck clean) while avoiding the redundant multi-GB memset a whole-buffer zero-init would incur — which measures as a 10–22% regression vs. master on this op, eliminated here.

**5. DeviceBuffer — stream-ordered free (latent use-after-free)**
The destructor, move-assignment, and `clear()` freed device memory on the default stream regardless of what work was still outstanding on the buffer — a latent use-after-free for non-blocking-stream callers, since the default stream is only implicitly ordered after *blocking* streams.

An earlier revision of this PR fixed that by freeing on the stream the buffer was last used on. [@matthewdcong pointed out](https://github.com/AcademySoftwareFoundation/openvdb/pull/2249#discussion_r3716141426) that this is insufficient when a buffer is used on more than one stream: the free is then ordered after the last one only. That is reproducible — a device-only buffer plus a user kernel on a second blocking stream has the freed block recycled and fully overwritten by the late kernel, where freeing on the default stream passes.

The buffer now tracks its uses with a per-device **event**: each use waits on the event before issuing work and re-records it afterwards, so one event transitively covers every stream the buffer has been used on, and the frees wait on it. This is correct for any mix and number of blocking and non-blocking streams, and it removes any requirement that callers keep their streams alive until the buffer is destroyed, since the event is owned by the buffer.

Worth noting for reviewers: `clear()` calls `cudaFreeHost(mCpuData)` before the device frees, and `cudaFreeHost` implicitly synchronizes. That masks this whole class of hazard for buffers that own pinned host memory — it is **device-only** buffers (`create(size, nullptr, device, stream)`, e.g. in `buildVoxelBlockManager`) where `mCpuData` is null that are actually exposed. The fix here makes the ordering explicit rather than leaving correctness resting on that implicit synchronization, which is not in CUDA's documented list.

## Tests

New gtest coverage in `unittest/TestNanoVDB.cu`:
- cross-stream dilation determinism (default stream vs. a non-blocking stream while the default stream is occupied)
- grid-name round-trip (empty / short / maximum-length)
- PointsToGrid byte-for-byte output determinism
- DeviceBuffer allocate/use/free lifetime on a non-blocking stream

### DIFF
--- a/nanovdb/nanovdb/cuda/DeviceBuffer.h
+++ b/nanovdb/nanovdb/cuda/DeviceBuffer.h
@@ -38,7 +38,7 @@ class DeviceBuffer
     uint64_t mSize; // total number of bytes managed by this buffer (assumed to be identical for host and device)
     void *mCpuData, **mGpuData; // raw pointers to the host and device buffers
     int   mDeviceCount, mManaged;// if mManaged is non-zero this class is responsible for allocating and freeing memory buffers. Otherwise this is assumed to be handled externally
-    cudaStream_t *mStreams = nullptr;// per-device stream each managed allocation was made on (parallel to mGpuData, length mDeviceCount). Frees (destructor, move-assign, clear) are ordered on the owning device's stream instead of the default stream 0, otherwise a buffer last used on a non-blocking stream could be freed while that stream's work is still in flight (use-after-free). One stream per device is required because a single DeviceBuffer can hold allocations on several devices (multi-GPU), each on its own device-bound stream. The caller must keep these streams alive until the buffer is destroyed.
+    cudaStream_t *mStreams = nullptr;// per-device stream each managed device buffer was last used on - set at allocation, then updated by every deviceUpload/deviceDownload (parallel to mGpuData, length mDeviceCount). Frees (destructor, move-assign, clear) are ordered on the owning device's stream instead of the default stream 0, otherwise a buffer last used on a non-blocking stream could be freed while that stream's work is still in flight (use-after-free). One stream per device is required because a single DeviceBuffer can hold allocations on several devices (multi-GPU), each on its own device-bound stream. The caller must keep these streams alive until the buffer is destroyed.
 
     /// @brief Initialize buffer
     /// @param size byte size of buffer to be initialized
@@ -145,8 +145,8 @@ public:
     }
 
      /// @brief Destructor frees memory on both the host and device
-    /// @note Each managed device allocation is freed on the stream it was
-    ///       allocated on (mStreams[device]), not the default stream, so device
+    /// @note Each managed device allocation is freed on the stream it was last
+    ///       used on (mStreams[device]), not the default stream, so device
     ///       frees are ordered after the last work issued on that stream.
     ~DeviceBuffer() { this->clear(); };
 
@@ -316,7 +316,7 @@ public:
 
 inline DeviceBuffer& DeviceBuffer::operator=(DeviceBuffer&& other) noexcept
 {
-    if (mManaged) {// first free all the managed data buffers, each on the stream its device was allocated on
+    if (mManaged) {// first free all the managed data buffers, each on the stream its device was last used on
         cudaCheck(cudaFreeHost(mCpuData));
         for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStreams ? mStreams[i] : cudaStream_t{0}));
     }
@@ -350,7 +350,7 @@ inline void DeviceBuffer::init(uint64_t size, int device, cudaStream_t stream)
     } else {
         cudaCheck(util::cuda::mallocAsync(mGpuData+device, size, stream)); // un-managed memory on the device, always 32B aligned!
         checkPtr(mGpuData[device], "cuda::DeviceBuffer::init: failed to allocate device buffer");
-        mStreams[device] = stream;// remember this device's allocation stream so its free is ordered on it, not stream 0
+        mStreams[device] = stream;// record the stream this device buffer is used on so its free is ordered on it, not stream 0
     }
     mSize = size;
     mManaged = 1;// i.e. this instance is responsible for allocating and delete memory
@@ -403,7 +403,7 @@ inline void DeviceBuffer::deviceDownload(void* stream, bool sync)
 
 inline void DeviceBuffer::clear(cudaStream_t stream)
 {
-    if (mManaged) {// free all the managed data buffers, each on the stream its device was allocated on
+    if (mManaged) {// free all the managed data buffers, each on the stream its device was last used on
         cudaCheck(cudaFreeHost(mCpuData));
         for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStreams ? mStreams[i] : stream));
     }
--- a/nanovdb/nanovdb/cuda/DeviceBuffer.h
+++ b/nanovdb/nanovdb/cuda/DeviceBuffer.h
@@ -38,7 +38,7 @@ class DeviceBuffer
     uint64_t mSize; // total number of bytes managed by this buffer (assumed to be identical for host and device)
     void *mCpuData, **mGpuData; // raw pointers to the host and device buffers
     int   mDeviceCount, mManaged;// if mManaged is non-zero this class is responsible for allocating and freeing memory buffers. Otherwise this is assumed to be handled externally
-    cudaStream_t mStream = 0;// stream the managed device allocations are associated with. Frees (destructor, move-assign, clear) are ordered on this stream instead of the default stream 0, otherwise a buffer last used on a non-blocking stream could be freed while that stream's work is still in flight (use-after-free). The caller must keep this stream alive at least until the buffer is destroyed.
+    cudaStream_t *mStreams = nullptr;// per-device stream each managed allocation was made on (parallel to mGpuData, length mDeviceCount). Frees (destructor, move-assign, clear) are ordered on the owning device's stream instead of the default stream 0, otherwise a buffer last used on a non-blocking stream could be freed while that stream's work is still in flight (use-after-free). One stream per device is required because a single DeviceBuffer can hold allocations on several devices (multi-GPU), each on its own device-bound stream. The caller must keep these streams alive until the buffer is destroyed.
 
     /// @brief Initialize buffer
     /// @param size byte size of buffer to be initialized
@@ -123,11 +123,11 @@ public:
         , mGpuData(other.mGpuData)
         , mDeviceCount(other.mDeviceCount)
         , mManaged(other.mManaged)
-        , mStream(other.mStream)
+        , mStreams(other.mStreams)
     {
         other.mCpuData = other.mGpuData = nullptr;
+        other.mStreams = nullptr;
         other.mSize = other.mDeviceCount = other.mManaged = 0;
-        other.mStream = 0;
     }
 
     /// @brief Copy-constructor from a HostBuffer
@@ -145,10 +145,10 @@ public:
     }
 
      /// @brief Destructor frees memory on both the host and device
-    /// @note Frees on the stream the buffer is associated with (mStream), not
-    ///       the default stream, so device frees are ordered after the last
-    ///       work issued on that stream.
-    ~DeviceBuffer() { this->clear(mStream); };
+    /// @note Each managed device allocation is freed on the stream it was
+    ///       allocated on (mStreams[device]), not the default stream, so device
+    ///       frees are ordered after the last work issued on that stream.
+    ~DeviceBuffer() { this->clear(); };
 
     /// @brief Static factory method that return an instance of this buffer
     /// @param size byte size of buffer to be initialized
@@ -316,23 +316,24 @@ public:
 
 inline DeviceBuffer& DeviceBuffer::operator=(DeviceBuffer&& other) noexcept
 {
-    if (mManaged) {// first free all the managed data buffers on the stream they are associated with
+    if (mManaged) {// first free all the managed data buffers, each on the stream its device was allocated on
         cudaCheck(cudaFreeHost(mCpuData));
-        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStream));
+        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStreams ? mStreams[i] : cudaStream_t{0}));
     }
     delete [] mGpuData;
+    delete [] mStreams;
     mSize    = other.mSize;
     mCpuData = other.mCpuData;
     mGpuData = other.mGpuData;
     mDeviceCount = other.mDeviceCount;
     mManaged = other.mManaged;
-    mStream  = other.mStream;
+    mStreams = other.mStreams;
     other.mCpuData = nullptr;
     other.mGpuData = nullptr;
+    other.mStreams = nullptr;
     other.mSize = 0;
     other.mDeviceCount = 0;
     other.mManaged = 0;
-    other.mStream = 0;
     return *this;
 }
 
@@ -341,6 +342,7 @@ inline void DeviceBuffer::init(uint64_t size, int device, cudaStream_t stream)
     if (size==0) return;
     cudaCheck(cudaGetDeviceCount(&mDeviceCount));
     mGpuData = new void*[mDeviceCount]();// NULL initialization
+    mStreams = new cudaStream_t[mDeviceCount]();// zero (default-stream) initialization
     NANOVDB_ASSERT(device >= cudaCpuDeviceId && device < mDeviceCount);
     if (device == cudaCpuDeviceId) {
         cudaCheck(cudaMallocHost((void**)&mCpuData, size)); // un-managed pinned memory on the host (can be slow to access!). Always 32B aligned
@@ -348,10 +350,10 @@ inline void DeviceBuffer::init(uint64_t size, int device, cudaStream_t stream)
     } else {
         cudaCheck(util::cuda::mallocAsync(mGpuData+device, size, stream)); // un-managed memory on the device, always 32B aligned!
         checkPtr(mGpuData[device], "cuda::DeviceBuffer::init: failed to allocate device buffer");
+        mStreams[device] = stream;// remember this device's allocation stream so its free is ordered on it, not stream 0
     }
     mSize = size;
     mManaged = 1;// i.e. this instance is responsible for allocating and delete memory
-    mStream = stream;// remember the allocation stream so managed frees are ordered on it, not stream 0
 } // DeviceBuffer::init
 
 inline void DeviceBuffer::deviceUpload(int device, cudaStream_t stream, bool sync)
@@ -361,7 +363,7 @@ inline void DeviceBuffer::deviceUpload(int device, cudaStream_t stream, bool syn
     if (mGpuData[device] == nullptr) {
         if (mManaged==0) throw std::runtime_error("DeviceBuffer::deviceUpload called on externally managed memory that wasn\'t allocated.");
         cudaCheck(util::cuda::mallocAsync(mGpuData+device, mSize, stream)); // un-managed memory on the device, always 32B aligned!
-        mStream = stream;// remember the allocation stream so this managed device buffer is freed on it, not stream 0
+        mStreams[device] = stream;// remember this device's allocation stream so it is freed on it, not stream 0
     }
     checkPtr(mGpuData[device], "uninitialized gpu destination data");
     cudaCheck(cudaMemcpyAsync(mGpuData[device], mCpuData, mSize, cudaMemcpyHostToDevice, stream));
@@ -397,17 +399,18 @@ inline void DeviceBuffer::deviceDownload(void* stream, bool sync)
 
 inline void DeviceBuffer::clear(cudaStream_t stream)
 {
-    if (mManaged) {// free all the managed data buffers on the requested stream
+    if (mManaged) {// free all the managed data buffers, each on the stream its device was allocated on
         cudaCheck(cudaFreeHost(mCpuData));
-        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], stream));
+        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStreams ? mStreams[i] : stream));
     }
     delete [] mGpuData;
+    delete [] mStreams;
     mCpuData = nullptr;
     mGpuData = nullptr;
+    mStreams = nullptr;
     mSize = 0;
     mDeviceCount = 0;
     mManaged = 0;
-    mStream = 0;
 } // DeviceBuffer::clear
 
 }// namespace cuda
--- a/nanovdb/nanovdb/cuda/DeviceBuffer.h
+++ b/nanovdb/nanovdb/cuda/DeviceBuffer.h
@@ -363,10 +363,13 @@ inline void DeviceBuffer::deviceUpload(int device, cudaStream_t stream, bool syn
     if (mGpuData[device] == nullptr) {
         if (mManaged==0) throw std::runtime_error("DeviceBuffer::deviceUpload called on externally managed memory that wasn\'t allocated.");
         cudaCheck(util::cuda::mallocAsync(mGpuData+device, mSize, stream)); // un-managed memory on the device, always 32B aligned!
-        mStreams[device] = stream;// remember this device's allocation stream so it is freed on it, not stream 0
     }
     checkPtr(mGpuData[device], "uninitialized gpu destination data");
     cudaCheck(cudaMemcpyAsync(mGpuData[device], mCpuData, mSize, cudaMemcpyHostToDevice, stream));
+    // This device buffer was last used on 'stream', so free it there (not on
+    // stream 0). Updated on every upload - not just first allocation - so a
+    // re-upload on a different stream still orders the free after its work.
+    if (mStreams) mStreams[device] = stream;
     if (sync) cudaCheck(cudaStreamSynchronize(stream));
 } // DeviceBuffer::deviceUpload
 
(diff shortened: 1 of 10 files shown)


nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
```diff
@@ -938,11 +944,16 @@ inline void PointsToGrid<BuildT, ResourceT>::processGridTreeRoot(const PtrT poin
...
     char *dst = mData.getGrid().mGridName;
-    if (const char *src = mGridName.data()) {
-        cudaCheck(cudaMemcpyAsync(dst, src, GridData::MaxNameSize, cudaMemcpyHostToDevice, mStream));
-    } else {
-        cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, mStream));
+    cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, mStream));
+    if (!mGridName.empty()) {
+        const size_t nameSize = std::min<size_t>(mGridName.size() + 1, GridData::MaxNameSize);
+        cudaCheck(cudaMemcpyAsync(dst, mGridName.c_str(), nameSize, cudaMemcpyHostToDevice, mStream));
     }
```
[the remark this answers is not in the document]
swahtz: Fixed in 465599dda. The copy is now clamped to `min(size, MaxNameSize-1)` bytes, so the preceding `memset` always supplies the terminating `\0` and an over-long name is truncated rather than left unterminated. Also extended `GridName_CudaPointsToGrid` with an over-long (>= MaxNameSize) case asserting truncation + null-termination.

nanovdb/nanovdb/tools/cuda/MeshToGrid.cuh
```diff
@@ -884,12 +884,13 @@ void MeshToGrid<BuildT>::processGridTreeRoot()
...
+    // and copy only the actual string.
     char *dst = mBuilder.data()->getGrid().mGridName;
+    cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, mStream));
     if (!mGridName.empty()) {
-        cudaCheck(cudaMemcpyAsync(dst, mGridName.data(), GridData::MaxNameSize, cudaMemcpyHostToDevice, mStream));
-    } else {
-        cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, mStream));
+        const size_t nameSize = std::min<size_t>(mGridName.size() + 1, GridData::MaxNameSize);
+        cudaCheck(cudaMemcpyAsync(dst, mGridName.c_str(), nameSize, cudaMemcpyHostToDevice, mStream));
     }
```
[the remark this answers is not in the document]
swahtz: Fixed in 465599dda. The copy is now clamped to `min(size, MaxNameSize-1)` bytes, so the preceding `memset` always supplies the terminating `\0` and an over-long name is truncated rather than left unterminated. Also extended `GridName_CudaPointsToGrid` with an over-long (>= MaxNameSize) case asserting truncation + null-termination.

nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
```diff
@@ -909,11 +909,12 @@ inline void DistributedPointsToGrid<BuildT>::processGridTreeRoot(const PtrT poin
...
     char *dst = mData->getGrid().mGridName;
-    if (const char *src = mGridName.data()) {
-        cudaCheck(cudaMemcpyAsync(dst, src, GridData::MaxNameSize, cudaMemcpyHostToDevice, stream));
-    } else {
-        cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, stream));
+    cudaCheck(cudaMemsetAsync(dst, 0, GridData::MaxNameSize, stream));
+    if (!mGridName.empty()) {
+        const size_t nameSize = std::min<size_t>(mGridName.size() + 1, GridData::MaxNameSize);
+        cudaCheck(cudaMemcpyAsync(dst, mGridName.c_str(), nameSize, cudaMemcpyHostToDevice, stream));
     }
```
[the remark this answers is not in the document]
swahtz: Fixed in 465599dda. The copy is now clamped to `min(size, MaxNameSize-1)` bytes, so the preceding `memset` always supplies the terminating `\0` and an over-long name is truncated rather than left unterminated. Also extended `GridName_CudaPointsToGrid` with an over-long (>= MaxNameSize) case asserting truncation + null-termination.

nanovdb/nanovdb/unittest/TestNanoVDB.cu
```diff
@@ -3607,6 +3607,132 @@ TEST(TestNanoVDBCUDA, DilateInjectPrune_ValueOnIndex)
...
+    auto inHandle = conv.getHandle(static_cast<nanovdb::Coord*>(inBuf.deviceData()), pts.size());
+    auto* inGrid = inHandle.deviceGrid<BuildT>();
+    EXPECT_TRUE(inGrid);
+    cudaCheck(cudaDeviceSynchronize());
+
+    auto dilateOn = [&](cudaStream_t stream, bool occupyDefault) {
+        if (occupyDefault) {// keep the default stream busy for ~20 ms
+            int clockKHz = 0;
+            cudaCheck(cudaDeviceGetAttribute(&clockKHz, cudaDevAttrClockRate, 0));
+            streamBusyWaitKernel<<<1, 1, 0, 0>>>(static_cast<unsigned long long>(clockKHz) * 20ull);
```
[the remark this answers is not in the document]
swahtz: Fixed in 465599dda. The test now queries the active device via `cudaGetDevice(&dev)` and passes `dev` to `cudaDeviceGetAttribute`, instead of hard-coding ordinal 0.

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -310,21 +316,23 @@ class DeviceBuffer
 
 inline DeviceBuffer& DeviceBuffer::operator=(DeviceBuffer&& other) noexcept
 {
-    if (mManaged) {// first free all the managed data buffers
+    if (mManaged) {// first free all the managed data buffers on the stream they are associated with
         cudaCheck(cudaFreeHost(mCpuData));
-        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], 0));
+        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStream));
     }
```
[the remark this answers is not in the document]
swahtz: Fixed in 5bb2eba69. Confirmed this is a real regression for the multi-GPU path: `deviceUpload` overwrote the single `mStream`, so the destructor/move-assign freed every device's allocation on the last device's stream (a cross-device `cudaFreeAsync`). Now tracked per device — a `cudaStream_t` array parallel to `mGpuData`, set at `mStreams[device]` in `init`/`deviceUpload`, with each `mGpuData[i]` freed on its own `mStreams[i]` in `clear`/move-assign. Single-GPU path unchanged. Caveat: single-GPU verified here (DeviceBuffer lifetime gtest + byte-exact validation); the multi-GPU free path is correct by construction but not runtime-tested (single-GPU environment) — worth a confirm on a multi-GPU box.

nanovdb/nanovdb/tools/cuda/GridChecksum.cuh
```diff
@@ -234,7 +234,7 @@ inline Checksum evalChecksum(const GridData *d_gridData, CheckMode mode, cudaStr
     if (mode != CheckMode::Empty) {
         auto d_lut = util::cuda::createCrc32Lut(1, stream);
         crc32Head(d_gridData, d_lut.get(), d_lut.get() + 256, stream);
-        cudaCheck(cudaMemcpyAsync(&(cs.head()), d_lut.get() + 256, headSize, cudaMemcpyDeviceToHost, stream));
+        cudaCheck(cudaMemcpyAsync(&(cs.head()), d_lut.get() + 256, sizeof(uint32_t), cudaMemcpyDeviceToHost, stream));
```
kmuseth, line 237: good catch

kmuseth (approved):
it looks good but I would like Matthew Cong to take a look at it as well

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -38,6 +38,7 @@ class DeviceBuffer
     uint64_t mSize; // total number of bytes managed by this buffer (assumed to be identical for host and device)
     void *mCpuData, **mGpuData; // raw pointers to the host and device buffers
     int   mDeviceCount, mManaged;// if mManaged is non-zero this class is responsible for allocating and freeing memory buffers. Otherwise this is assumed to be handled externally
+    cudaStream_t *mStreams = nullptr;// per-device stream each managed device buffer was last used on - set at allocation, then updated by every deviceUpload/deviceDownload (parallel to mGpuData, length mDeviceCount). Frees (destructor, move-assign, clear) are ordered on the owning device's stream instead of the default stream 0, otherwise a buffer last used on a non-blocking stream could be freed while that stream's work is still in flight (use-after-free). One stream per device is required because a single DeviceBuffer can hold allocations on several devices (multi-GPU), each on its own device-bound stream. The caller must keep these streams alive until the buffer is destroyed.
```
matthewdcong: Would this have the potential to introduce bugs in existing code with blocking streams? For example, if you download the same buffer to multiple blocking streams, the clear operation would now only synchronize on the last blocking stream as opposed to all of the blocking streams (via the default stream) leading to early destruction.

matthewdcong:
Looks good to me for the most part, minus the change to the DeviceBuffer which I raised a question about above.

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -387,13 +403,15 @@ inline void DeviceBuffer::deviceDownload(void* stream, bool sync)
 
 inline void DeviceBuffer::clear(cudaStream_t stream)
 {
-    if (mManaged) {// free all the managed data buffers
+    if (mManaged) {// free all the managed data buffers, each on the stream its device was last used on
         cudaCheck(cudaFreeHost(mCpuData));
-        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], stream));
+        for (int i=0; i<mDeviceCount; ++i) cudaCheck(util::cuda::freeAsync(mGpuData[i], mStreams ? mStreams[i] : stream));
```
harrism: **💡 suggestion (non-blocking):** For a managed buffer, `clear(stream)`'s argument is now dead — `mStreams[i]` overrides it. Every current caller passes the stream the buffer already lives on, so nothing changes today, but the signature reads as if the argument is honored. A doc line on `clear` saying frees are ordered on the recorded per-device streams (argument only a fallback when none was recorded) would keep callers honest.

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -310,18 +316,21 @@ class DeviceBuffer
 
 inline DeviceBuffer& DeviceBuffer::operator=(DeviceBuffer&& other) noexcept
```
harrism, line 379: **🤓 nitpick (non-blocking):** Pre-existing, but since this function is being touched: no self-move guard, and it now frees and then reads its own fields on self-move. One line fixes it: `if (this == &other) return *this;` (`cuda::Buffer` carries the equivalent guard).

nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
```diff
@@ -792,6 +793,12 @@ inline BufferT PointsToGrid<BuildT, ResourceT>::getBuffer(const PtrT, size_t poi
 
     mData.d_bufferPtr = buffer.deviceData();
     if (mData.d_bufferPtr == nullptr) throw std::runtime_error("Failed to allocate grid buffer on the device");
+    // Zero the whole grid buffer up front. This (a) makes the dense background
+    // fills (upper/lower value tables, inactive leaf values - all zero in this
+    // builder) redundant, so those kernels are skipped, and (b) makes the
+    // output bit-deterministic: alignment padding and stats fields no longer
+    // carry recycled pool bytes (which previously leaked into written files).
+    cudaCheck(cudaMemsetAsync(mData.d_bufferPtr, 0, mData.size, mStream));
```
harrism, line 801: **❓ question (non-blocking):** Was "net-neutral on runtime" measured the way the IndexToGrid regression was? A full-buffer memset versus the three dropped fill passes is a bandwidth trade that varies with density and build type — a number in the description would make the claim checkable later.

harrism (commented):
All five are real bugs, and the fixes are the right shape. Three brief inline notes, none blocking.

On the buffer stream fixes specifically:

- The `freeAsync(ptr, 0)` free is a genuine use-after-free for non-blocking-stream callers — the legacy default stream does not order against them — so this is a correctness fix, not defensive extra synchronization.
- Tracking the allocation/last-use stream and freeing on it is the same retained-stream contract `cuda::Buffer` (#2251) uses and the roadmap (#2232) records. The legacy buffer converging on it independently is good validation of the rule, and the per-device stream array is the right extension for a buffer that can hold allocations on several devices.
- I swept `tools/cuda/` on this head for kernel launches without an explicit stream argument: none remain — the two fixed here (DilateGrid 26-neighbour pass, SignedFloodFill root pass) were the last.

The IndexToGrid scoped zero-init — measured against the naive whole-buffer version and engineered to be byte-identical without the 10–22% cost — is exactly how a change like that should be justified.

Sequencing note: this conflicts textually with #2269/#2270 in `MeshToGrid.cuh` and `PointsToGrid.cuh`. This PR should land first; I'll rebase that stack over it.

nanovdb/nanovdb/cuda/DeviceBuffer.h
swahtz: Good catch — you're right, and it reproduces. I wrote a repro and ran it against three states of the code:

| scenario | master | this PR (as reviewed) | with fix |
|---|:--:|:--:|:--:|
| **A)** device-only buffer, user kernel on a second **blocking** stream | pass | **FAIL** — 64 MB clobbered | pass |
| **B)** host buffer, upload on a **non-blocking** stream | pass | pass | pass |

Scenario A is exactly the mechanism you described: the free is ordered on the last-used stream only, the block is recycled, and the other stream's still-queued work lands in the next allocation. Two things were needed to observe it at all — a warm-up pass (on a cold context the first launches serialize and hide the race) and a spin loop the compiler can't elide.

**A second finding while chasing this, which I think reframes the change.** `clear()` calls `cudaFreeHost(mCpuData)` *before* the device frees, and `cudaFreeHost` implicitly synchronizes — I verified directly that a busy stream goes to `done` across it, while `cudaFreeAsync` does not. So for any buffer that owns pinned host memory that call already masks the hazard, including the non-blocking-stream case this hunk was written to fix, which is why scenario B passes on master too. The regression only escapes because **device-only** buffers (`create(size, nullptr, device, stream)`, e.g. in `buildVoxelBlockManager`) leave `mCpuData` null, so `cudaFreeHost(nullptr)` is a no-op.

So as written, the hunk fixed a hazard that is currently masked and introduced one that isn't. Rather than revert it, I've made the ordering explicit instead of leaving correctness resting on `cudaFreeHost`'s implicit synchronization, which isn't in CUDA's documented implicit-synchronization list.

**The fix** replaces the per-device *stream* with a per-device *event*: each use waits on the event before issuing work and re-records it afterwards, so a single event transitively covers every stream the buffer has been used on, and the frees wait on it. That is correct for any mix and number of blocking/non-blocking streams, and it also drops the "caller must keep these streams alive until the buffer is destroyed" requirement the old comment conceded, since the event is owned by the buffer.

Both scenarios now pass, and the VBM/topology/maintenance goldens are unchanged (46/46 on dragon and emu). Pushed, along with a corrected `pendingchanges` entry — the old wording described the superseded "free on the last-used stream" mechanism.

Happy to add the repro as a regression test if you'd like it, though I left it out deliberately: it needs a ~350 ms spin, and if the allocator declines to recycle the block it would pass vacuously, so it risks being a slow test that can silently stop testing anything.

swahtz: Follow-up: I went ahead and added the regression test after all — `DeviceBufferMultiStreamFreeOrdering` in `TestNanoVDB.cu`.

The concern about it passing vacuously turned out to be fixable. My first attempt gated the assertion on having staged the race (work still pending *and* the block recycled), which was backwards: with a correctly ordered free the allocator cannot hand the block out again until the other stream drains, so both of those observations are the fix *working* rather than a reason to skip. They are now reported in the failure message instead of gating the check.

Verified in both directions — it fails against the pre-fix header (64 MB clobbered) and passes with the fix. It uses a device-only buffer deliberately, so `clear()`'s `cudaFreeHost` cannot mask the problem, and it needs the warm-up pass because on a cold context the first launches serialize and the race never forms.

Also relabelled the pre-existing `DeviceBufferNonBlockingStreamLifetime` as a smoke test: it only asserts `cudaSuccess`, so it passes with or without a correctly ordered free, and its comment still described the superseded "free on the stream it was allocated on" mechanism.

nanovdb/nanovdb/cuda/DeviceBuffer.h
swahtz: Good catch on the reviewed revision — the argument really was dead there, since the recorded per-device stream overrode it.

It has since become live again. Fixing the multi-stream issue [matthewdcong raised](https://github.com/AcademySoftwareFoundation/openvdb/pull/2249#discussion_r3716141426) replaced the per-device *stream* with a per-device *event*, so `clear(stream)` now issues the frees on the stream you pass and the event supplies the ordering:

```cpp
this->orderAfterPriorUses(i, stream);                  // wait on the tracking event
cudaCheck(util::cuda::freeAsync(mGpuData[i], stream)); // ...then free on the caller's stream
```

So the contract is the inverse of what you'd have had to document before: `stream` selects where the free is *enqueued*, not what it is *ordered against*. I've added that as a doc line on `clear`, since it is worth stating either way — and it means any stream is safe to pass regardless of where the buffer was used.

nanovdb/nanovdb/cuda/DeviceBuffer.h
swahtz: Agreed, and it is worse than a nitpick — it aborts. Confirmed by running `buf = std::move(buf)` against the unguarded version:

```
Assertion `device >= 0 && device < mDeviceCount' failed.
Aborted (core dumped)
```

The free path zeroes `mDeviceCount` before the self-assignment reads it back. Added `if (this == &other) return *this;` — the same test then passes with the buffer intact (pointer and size unchanged).

One correction on the reference: I could not find a `cuda::Buffer` to match against — `grep "this == &other"` returns nothing anywhere in nanovdb, and the closest analogue, `UnifiedBuffer::operator=(&&)`, has the identical flaw (frees `mPtr`, then reads `other.mPtr`). So there is no existing precedent to follow, which if anything strengthens the point. I have left `UnifiedBuffer` alone here since it is outside this PR's scope, but it wants the same one-liner.

nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
swahtz: Fair — it was not, and it should have been. Measured now, master vs this branch, same harness, 30 iters / 10 warmup, RTX PRO 6000 (sm_120), median of samples:

| op | dataset | master | this PR | delta |
|---|---|--:|--:|--:|
| `p2g_onindex_n23347893` | dragon | 10.700 ms | 10.581 ms | −1.1% |
| `p2g_onindex_n93391572_dup4` | dragon | 42.949 ms | 42.402 ms | −1.3% |
| `p2g_onindex_n373566288_dup16` | dragon | 169.011 ms | 168.292 ms | −0.4% |
| `p2g_onindex_n96956688` | emu | 47.592 ms | 46.996 ms | −1.3% |

So the claim holds and is marginally favourable rather than exactly neutral — the single zero-init is slightly cheaper than the three dense fill passes it replaces at these densities, by about the width of run-to-run noise. I will put the table in the PR description so it is checkable later, as you suggest.

Worth noting for anyone reproducing: the master side needs `--checksum=skip` to benchmark at all, because it trips the checksum out-of-bounds crash that fix #1 in this PR addresses.

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -46,6 +47,44 @@ class DeviceBuffer
     /// @warning size is expected to be non-zero. Use clear() clear buffer!
     void init(uint64_t size, int device, cudaStream_t stream);
 
+    /// @brief Order work subsequently issued on @a stream after every prior use of this
+    ///        device buffer, whichever stream those uses were issued on.
+    void orderAfterPriorUses(int device, cudaStream_t stream) const
+    {
+        if (mEvents && mEvents[device]) cudaCheck(cudaStreamWaitEvent(stream, mEvents[device], 0));
+    }
```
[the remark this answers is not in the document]
swahtz: All four points from this review (the comment above plus the three suppressed ones) evaluated and addressed in 457b671c3. For the humans following along:

**Raw-pointer uses are untracked (this comment) — correct, and the most important of the four.** The event only covers `deviceUpload`/`deviceDownload`; kernels launched against `deviceData()`'s raw pointer are invisible to it. Two clarifications on scope: work on *blocking* streams is still covered (the free is issued on the default stream, which is implicitly ordered after them — the same coverage master provided), and the uncovered case — raw-pointer work on a *non-blocking* stream — was equally uncovered on master, where no API could even express the dependency. Fixed as suggested by making the contract explicit: `recordUse(device, stream)` is now public, and `deviceData()` documents that raw-pointer work on non-blocking streams must either be registered through it or synchronized before the buffer is cleared/destroyed.

**Test's `user` stream is blocking (suppressed, TestNanoVDB.cu) — deliberate, but the observation exposed a gap.** The blocking scenario is what discriminates the last-used-stream revision this thread started with; simply flipping it to non-blocking would create a test no implementation can pass, because the late write never goes through the buffer's API. With `recordUse` public, the constructive version exists, so the test is now two scenarios, verified per revision:

| revision | blocking + unregistered write | non-blocking + `recordUse` |
|---|:--:|:--:|
| this PR (event fix) | pass | pass |
| earlier revision (free on last-used stream) | **FAIL** | **FAIL** |
| master (free on stream 0, no event) | pass | **FAIL** |

**One stream for every device's free (suppressed, DeviceBuffer.h) — agreed in principle.** A stream belongs to one device, and nothing in the SOMA documentation blesses freeing one device's pool allocation on another device's stream ("deallocation can be performed in any stream" is stated in a same-device context). `freeDeviceBuffers` now frees each allocation on its own device — the caller's stream for the current device, the owning device's default stream otherwise, switching devices as needed. Honesty note: master's loop had the same single-stream shape, and this machine has one GPU, so the multi-device path is verified by inspection and compilation only.

**SignedFloodFill test coverage (suppressed, SignedFloodFill.cuh) — agreed, added** as `NonBlockingStreamSignedFloodFill`, mirroring the dilation regression (occupied default stream, non-blocking run, readback on the producing stream, identical output required). One scope caveat, verified rather than assumed: it locks the cross-stream contract and would catch node passes escaping to the default stream, but it cannot discriminate `processRoot`'s internal ordering on constructible inputs — I swapped the pre-fix header in and it passes (3/3), because that path only does work when interior root-level tiles exist (a level set thousands of voxels across) and the pre-fix code was accidentally host-synchronous otherwise via its blocking legacy-stream memcpy. The test comment says so.

Goldens unchanged (46/46), full test file compiles, self-move check still passes.

PR #2251, opened 2026-07-16
NanoVDB: add cuda::Buffer, cuda::BufferView, and the synchronous resource concept (CUDA)

harrism:
Step 2 (part A) of #2232 (NanoVDB injectable CUDA memory resources). New types and tests only — no existing call site changes; the scratch retrofit that consumes these lands separately.

## What

- **`cuda::Buffer<T, R>`** — typed, resource-aware, stream-ordered container modeled on CCCL's [`cuda::buffer<T>`](https://nvidia.github.io/cccl/unstable/libcudacxx/runtime/buffer.html). Constructor order and shape follow `cuda::buffer` exactly — `(stream, resource, count, noInit)`, with defaulted-resource conveniences and stream-less synchronous forms. `size()` counts elements, `size_bytes()` bytes. Move-only with an explicit `copy()`. The retained-stream rule throughout: frees are ordered on the stream of the last allocation (or `setStream`), and `resize` orders everything — including the free of the old block, whose last use is the prefix copy — on the passed stream.
- **`cuda::BufferView<T>`** — typed non-owning view with span semantics (element constness in `T`, shallow const, trivially copyable, no resource, no stream) satisfying the same static interface, so `GridHandle<BufferView<std::byte>>` wraps externally owned memory with zero copies (tested end-to-end against a `createLevelSetSphere` grid — pointer-equal, no double free).
- **`is_resource<R>`** (synchronous tier) beside `is_async_resource<R>`, which now enforces CCCL's refinement: an async resource provides the stream-ordered *and* the synchronous interface, so `is_async_resource<R>` implies `is_resource<R>`. `DeviceResource` gains the delegating sync pair; `PinnedResource` remains synchronous-only.
- **`DeviceResource`'s legacy static `allocateAsync`/`deallocateAsync` are `[[deprecated]]`** (the implementation moved into the instance methods, so nothing internal trips the warning). Migration note for v13-era custom resources is on #2232.

## Design decisions (rationale on #2232)

- **No implicitly initializing count constructor** — the `noInit` tag is mandatory, matching `cuda::buffer` (whose count constructor requires `cuda::no_init`) and the RMM `device_uvector` lesson: implicit fill of freshly allocated memory is a hidden cost the dominant allocate-then-overwrite pattern wastes. Enforced by `is_constructible` static_asserts.
- **A synchronous-`R` `Buffer` has no stream API at all** (member, constructor parameter, `stream()`, `setStream` — all absent, verified by detection-idiom static_asserts): a synchronous allocation is immediately valid on every stream, so no parameter exists to ignore.
- **CamelCase type names, standard member names**: type names are aliasable (`using Buffer = ::cuda::buffer<T>` remains available at CUDA ≥ 13.2), member names are structural and keep CCCL's spelling.
- **CUDA graph capture**: the async path is capture-safe — allocation/use/free inside a stream capture records, instantiates, and relaunches (tested). The sync tier synchronizes and is excluded from capture by construction.
- A failed `resize` (overflow or allocation error) leaves the buffer fully untouched, including its retained stream (tested).

## Testing

New `unittest/TestBuffer.cu` (`TestBuffer` suite, 20 tests; ctest `nanovdb_cuda_buffer_unit_test`): trait satisfaction for both tiers and the refinement; exact allocation sizes through a counting resource; stream retention, `setStream` redirection, and resize stream ordering via a stream-recording resource; move/copy semantics without double-free; byte-size overflow throws; pinned page-lock; graph capture; `BufferView` semantics incl. the zero-copy `GridHandle` round-trip. Full suites green: CPU 153/153, `nanovdb_test_cuda` 53/53, memory-resource 8/8; compute-sanitizer memcheck clean (sm_89, CUDA 12.6).

## Non-CUDA builds

No impact on `NANOVDB_USE_CUDA=OFF`: new headers live under `nanovdb/cuda/`, and the test target is registered inside `if(NANOVDB_USE_CUDA)`.

### DIFF
--- a/nanovdb/nanovdb/cuda/Buffer.h
+++ b/nanovdb/nanovdb/cuda/Buffer.h
@@ -1,0 +1,341 @@
+// Copyright Contributors to the OpenVDB Project
+// SPDX-License-Identifier: Apache-2.0
+//
+/// @file nanovdb/cuda/Buffer.h
+///
+/// @brief Typed containers for CUDA memory: the owning, resource-aware,
+///        stream-ordered cuda::Buffer and the non-owning cuda::BufferView.
+
+#ifndef NANOVDB_CUDA_BUFFER_H_HAS_BEEN_INCLUDED
+#define NANOVDB_CUDA_BUFFER_H_HAS_BEEN_INCLUDED
+
+#include <cuda_runtime_api.h>
+#include <nanovdb/cuda/DeviceResource.h>
+#include <nanovdb/util/cuda/Util.h>
+
+#include <cstddef>
+#include <limits>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+
+namespace nanovdb {
+
+namespace cuda {
+
+/// @brief Tag type selecting the Buffer constructors that skip element
+///        initialization, leaving the contents indeterminate.
+struct NoInit {};
+inline constexpr NoInit noInit{};
+
+namespace detail {
+
+/// @brief Conditional stream storage for Buffer. The async specialization
+///        retains the stream of the most recent allocation; the synchronous
+///        specialization is an empty base, so a Buffer over a synchronous
+///        resource carries no stream state and exposes no stream API.
+template<bool IsAsync>
+struct StreamHolder {};
+
+template<>
+struct StreamHolder<true> { cudaStream_t mStream = 0; };
+
+} // namespace detail
+
+/// @brief Owning, typed container of @c T elements allocated from a memory
+///        resource @c R held by value.
+/// @tparam T element type; sizes are expressed in elements, not bytes
+/// @tparam R memory resource, either stream-ordered (AsyncResource concept,
+///         see is_async_resource) or synchronous (Resource concept, see
+///         is_resource). When @c R provides both interfaces the stream-ordered
+///         one is used.
+/// @details With a stream-ordered resource the Buffer retains the stream of
+///          the most recent allocation (or the one supplied via setStream)
+///          and orders its deallocation on that stream. Buffer is move-only.
+template<typename T, typename R = DeviceResource>
+class Buffer : private detail::StreamHolder<is_async_resource<R>::value>
+{
+    static_assert(is_async_resource<R>::value || is_resource<R>::value,
+                  "Buffer requires R to model the AsyncResource or the Resource concept");
+    static_assert(std::is_trivially_copyable<T>::value,
+                  "Buffer requires a trivially copyable T: elements are copied bytewise and never constructed or destroyed");
+
+    static constexpr bool IsAsync = is_async_resource<R>::value;
+
+    R      mResource;
+    T*     mData = nullptr;
+    size_t mSize = 0; // element count
+
+public:
+    /// @brief Default c-tor of an empty buffer; performs no allocation.
+    Buffer() = default;
+
+    /// @brief C-tor allocating @c count uninitialized elements, stream-ordered
+    ///        on @c stream. Parameter order follows cuda::buffer:
+    ///        (stream, resource, count, no_init).
+    /// @param stream cuda stream the allocation is ordered on
+    /// @param resource resource instance the buffer takes ownership of
+    /// @param count number of elements
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    explicit Buffer(cudaStream_t stream, R resource, size_t count, NoInit)
+        : detail::StreamHolder<true>{stream}, mResource(std::move(resource))
+    {
+        this->allocate(count, stream);
+    }
+
+    /// @brief Convenience c-tor using a default-constructed resource.
+    /// @note There is deliberately no count c-tor without NoInit: implicit
+    ///       initialization of freshly allocated memory costs a hidden fill
+    ///       pass that the dominant allocate-then-overwrite pattern wastes,
+    ///       so initialization is always explicit (matching cuda::buffer,
+    ///       whose count c-tor likewise requires cuda::no_init).
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    Buffer(cudaStream_t stream, size_t count, NoInit) : Buffer(stream, R(), count, noInit) {}
+
+    /// @brief C-tor allocating @c count uninitialized elements from a
+    ///        synchronous resource: the stream-less analog of
+    ///        (stream, resource, count, no_init).
+    template<typename S = R, std::enable_if_t<!is_async_resource<S>::value && is_resource<S>::value, int> = 0>
+    explicit Buffer(R resource, size_t count, NoInit)
+        : mResource(std::move(resource))
+    {
+        this->allocate(count, cudaStream_t{0});
+    }
+
+    /// @brief Convenience c-tor using a default-constructed resource.
+    template<typename S = R, std::enable_if_t<!is_async_resource<S>::value && is_resource<S>::value, int> = 0>
+    Buffer(size_t count, NoInit) : Buffer(R(), count, noInit) {}
+
+    Buffer(const Buffer&) = delete;
+    Buffer& operator=(const Buffer&) = delete;
+
+    /// @brief Move c-tor; steals the allocation (and retained stream, if any)
+    ///        and leaves @c other empty.
+    Buffer(Buffer&& other) noexcept
+        : detail::StreamHolder<IsAsync>(other)
+        , mResource(std::move(other.mResource))
+        , mData(other.mData)
+        , mSize(other.mSize)
+    {
+        other.mData = nullptr;
+        other.mSize = 0;
+    }
+
+    /// @brief Move assignment; frees the current allocation first, then steals
+    ///        from @c other and leaves it empty. Self-move is a no-op.
+    Buffer& operator=(Buffer&& other) noexcept
+    {
+        if (this != &other) {
+            this->clear();
+            static_cast<detail::StreamHolder<IsAsync>&>(*this) = other;
+            mResource = std::move(other.mResource);
+            mData = other.mData;
+            mSize = other.mSize;
+            other.mData = nullptr;
+            other.mSize = 0;
+        }
+        return *this;
+    }
+
+    /// @brief Returns a deep copy of this buffer, allocated from a copy of the
+    ///        resource; the allocation and element copy are ordered on @c stream,
+    ///        which becomes the copy's retained stream.
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    Buffer copy(cudaStream_t stream) const
+    {
+        Buffer out(stream, mResource, mSize, noInit);
+        if (mData) cudaCheck(cudaMemcpyAsync(out.mData, mData, this->size_bytes(), cudaMemcpyDefault, stream));
+        return out;
+    }
+
+    /// @brief Returns a deep copy of this buffer, allocated from a copy of the
+    ///        synchronous resource.
+    template<typename S = R, std::enable_if_t<!is_async_resource<S>::value && is_resource<S>::value, int> = 0>
+    Buffer copy() const
+    {
+        Buffer out(mResource, mSize, noInit);
+        if (mData) cudaCheck(cudaMemcpy(out.mData, mData, this->size_bytes(), cudaMemcpyDefault));
+        return out;
+    }
+
+    /// @brief D-tor. A stream-ordered resource frees on the retained stream;
+    ///        a synchronous resource frees immediately.
+    ~Buffer() { this->clear(); }
+
+    /// @brief Returns the retained stream, i.e. the stream the buffer's memory
+    ///        will be freed on.
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    cudaStream_t stream() const { return this->mStream; }
+
+    /// @brief Replaces the retained stream without synchronizing; subsequent
+    ///        deallocation (and destruction) is ordered on @c stream instead.
+    /// @warning The caller is responsible for ordering @c stream after any
+    ///          in-flight work that uses the buffer's memory.
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    void setStream(cudaStream_t stream) { this->mStream = stream; }
+
+    /// @brief Resizes the buffer to @c count elements, preserving the leading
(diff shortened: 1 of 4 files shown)


nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+
+    /// @brief C-tor allocating @c count uninitialized elements, stream-ordered
+    ///        on @c stream. Parameter order follows cuda::buffer:
+    ///        (stream, resource, count, no_init).
+    /// @param stream cuda stream the allocation is ordered on
+    /// @param resource resource instance the buffer takes ownership of
+    /// @param count number of elements
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    explicit Buffer(cudaStream_t stream, R resource, size_t count, NoInit)
+        : detail::StreamHolder<true>{stream}, mResource(std::move(resource))
```
kmuseth: very minor but I prefer:

```
explicit Buffer(cudaStream_t stream, R resource, size_t count, NoInit)
    : detail::StreamHolder<true>{stream}
    , mResource(std::move(resource))
{
```

I see you use the same layout below

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+    explicit Buffer(R resource, size_t count, NoInit)
+        : mResource(std::move(resource))
+    {
+        this->allocate(count, cudaStream_t{0});
+    }
+
+    /// @brief Convenience c-tor using a default-constructed resource.
+    template<typename S = R, std::enable_if_t<!is_async_resource<S>::value && is_resource<S>::value, int> = 0>
+    Buffer(size_t count, NoInit) : Buffer(R(), count, noInit) {}
+
```
kmuseth, line 114: add:

/// Explicitly disallow copy construction and assignment operation

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /// @brief D-tor. A stream-ordered resource frees on the retained stream;
+    ///        a synchronous resource frees immediately.
+    ~Buffer() { this->clear(); }
+
+    /// @brief Returns the retained stream, i.e. the stream the buffer's memory
+    ///        will be freed on.
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    cudaStream_t stream() const { return this->mStream; }
```
kmuseth, line 176: is there a reason this is not called getStream()? I see setStream(..) below so it seems a little strange that the get method uses a different semantic.

kmuseth (commented):
looks good - only non-blocking changes 

how much VRAM does the unit-test require? I've recently notice that tests would fail my my Nvidia laptop with only 8GB.

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+    {
+        this->allocate(count, stream);
+    }
+
+    /// @brief Convenience c-tor using a default-constructed resource.
+    /// @note There is deliberately no count c-tor without NoInit: implicit
+    ///       initialization of freshly allocated memory costs a hidden fill
+    ///       pass that the dominant allocate-then-overwrite pattern wastes,
+    ///       so initialization is always explicit (matching cuda::buffer,
+    ///       whose count c-tor likewise requires cuda::no_init).
```
kmuseth, line 94: please use "@param count" etc

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+            mData = newData;
+            mSize = count;
+        }
+        else {
+            this->mStream = stream; // no reallocation: setStream semantics
+        }
+    }
+
+    /// @brief Resizes the buffer to @c count elements through the synchronous
+    ///        resource, preserving the leading min(old, new) elements.
```
kmuseth, line 227: please use @param count

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+    static size_t checkedBytes(size_t count)
+    {
+        if (count > std::numeric_limits<size_t>::max() / sizeof(T))
+            throw std::runtime_error("nanovdb::cuda::Buffer: element count overflows the byte size");
+        return count * sizeof(T);
+    }
+
+    /// @brief Allocates @c count elements through the resource and records the
+    ///        new extent; @c stream is used by the stream-ordered form and
+    ///        ignored by the synchronous one. A zero count allocates nothing.
```
kmuseth: instead of documenting all the arguments in @brief please use "@param count" etc

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,341 @@
...
+        }
+        else {
+            this->mStream = stream; // no reallocation: setStream semantics
+        }
+    }
+
+    /// @brief Resizes the buffer to @c count elements through the synchronous
+    ///        resource, preserving the leading min(old, new) elements.
+    template<typename S = R, std::enable_if_t<!is_async_resource<S>::value && is_resource<S>::value, int> = 0>
+    void resize(size_t count)
```
kmuseth, line 230: The async resize wraps its cudaMemcpyAsync in a try/catch to deallocate newData on failure, leaving the buffer untouched. The synchronous overload ([Buffer.h:228-237](vscode-webview://0gd03c4greh2ma0mkbnva5a4k82fdtiii3de2beloeqmutck61tq/nanovdb/nanovdb/cuda/Buffer.h)) has no such guard. If cudaCheck throws (e.g. cudaErrorInvalidValue), newData is leaked and the old block is never freed.

kmuseth (changes requested):
The async resize wraps its cudaMemcpyAsync in a try/catch to deallocate newData on failure, leaving the buffer untouched. The synchronous overload ([Buffer.h:228-237](vscode-webview://0gd03c4greh2ma0mkbnva5a4k82fdtiii3de2beloeqmutck61tq/nanovdb/nanovdb/cuda/Buffer.h)) has no such guard:

nanovdb/nanovdb/cuda/Buffer.h
harrism: Fixed in 0583b27b — the synchronous overload now mirrors the async one.

Two clarifications on the diagnosis, because they change what the fix is actually buying us:

**The old block was never at risk.** On the throwing path `mData`/`mSize` are untouched, so the buffer still owns the old allocation and the destructor frees it. Only `newData` leaked.

**More importantly, `cudaCheck` doesn't throw — it calls `exit()`.** `util/cuda/Util.h:49` expands to `gpuAssert`, which does `fprintf` + `exit(code)`. So neither overload could actually reach the handler: the async `try`/`catch` was already dead code, and the synchronous one had no live leak.

I've added the guard anyway, because the asymmetry was the real defect — one invariant, two overloads, and it stays correct if `cudaCheck` ever becomes throwing or if a custom resource's `allocate`/`deallocate` throws. But it's defense-in-depth, not a live bug fix, and I didn't want to claim otherwise in the commit message.

That does leave a genuine open question: should `cudaCheck` throw instead of `exit()`? Terminating the process from a library header is harsh for anything embedding NanoVDB, and `Buffer` is the first place we've written code that assumes otherwise. Happy to take that on separately if you agree — it's too broad to fold into this PR.

(Unrelated, but adjacent: the comment above `cudaCheck` says "No-op in release builds", while the guard is `#if 1 || defined(DEBUG) || defined(_DEBUG)` — so it's always on. Worth fixing whenever that file is next touched.)

nanovdb/nanovdb/cuda/Buffer.h
harrism: Done in 0583b27b — split onto one member per line, matching the move c-tor below.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Added in 0583b27b, with your wording.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Done in 0583b27b.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Done in 0583b27b.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Done in 0583b27b. I applied the same treatment everywhere the pattern appeared rather than just at the three you flagged, so it's consistent across the header: `copy(stream)`, `setStream`, the async `resize`, `checkedBytes`, `deallocate`, and the `BufferView` c-tor all document their arguments with `@param` now instead of describing them inside `@brief`.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Deliberate, though I agree the asymmetry looks odd in isolation.

`stream()` follows the precedent already in the tree: `cuda::TempPool::stream()` (`nanovdb/cuda/TempPool.h:57`), which landed in #2231. It also matches `cuda::buffer`'s accessor, which this class is shaped after — same reasoning we used in #2231 for keeping `allocate`/`allocate_async` in snake_case rather than renaming them to NanoVDB style: where a name is part of a contract we're deliberately mirroring, matching it exactly is worth more than local consistency.

`setStream()` isn't a plain setter — it re-targets the retained stream without synchronizing, and the caller takes on the ordering obligation — so I wanted the verb there to make the mutation obvious at the call site.

That said, this is your call and it's cheap either way. Say the word and I'll rename to `getStream()`/`setStream()`; if I do, I'd want to rename `TempPool::stream()` in the same commit so we don't end up with both spellings.

harrism:
> how much VRAM does the unit-test require? I've recently noticed that tests would fail on my Nvidia laptop with only 8GB.

Measured it: **~490 MiB peak for the whole process**, and essentially none of that is the test's own data.

`TestBuffer` allocates almost nothing. The element counts across the suite are 64, 96, 100, 128, 256, 300 and 1000, of `int`/`float` — the largest single buffer in the file is 1000 ints, i.e. **4 KB**. The two `oversized` cases throw in `checkedBytes` before any allocation is attempted, so they never reserve anything either. The ~490 MiB is CUDA context and runtime overhead that any CUDA process on this machine pays before it allocates a byte.

Method: built `TestBuffer.cu` against CUDA 12.6 on an RTX 6000 Ada, ran it under `--gtest_repeat=40`, and sampled `nvidia-smi --query-compute-apps=used_memory` for the test PID throughout. Context overhead does vary with driver, architecture and how many modules get loaded, so the absolute number will differ on a laptop part — but the test's own footprint is ~4 KB no matter where it runs.

So this suite shouldn't be what's failing on your 8 GB laptop. If you can tell me which test binary and which case, I'm happy to dig — `TestNanoVDB.cu` builds real grids and is a far more plausible candidate. Worth noting that nothing would catch a VRAM regression for us anyway, since NanoVDB CI never runs the CUDA tests on a GPU.

Also: all 20 tests pass on the updated branch (0583b27b), and I've pushed fixes for the other seven comments — replies are on the individual threads.

nanovdb/nanovdb/cuda/Buffer.h
harrism: FWIW, this should be dictated by automatic formatting (clang-format) so people never need to think about it.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Previous comment was Claude, this is Mark: cudaCheck should definitely NOT exit(). As Claude says, a library header terminating a process because of a CUDA error is bad. We need to make failures handle-able.  This is the kind of behavior that would stop real applications from using our software. I will file an issue.

harrism: Followed up as #2265. Short version: the `try`/`catch` here becomes live if `cudaCheck` throws, which is what I've proposed there — with the companion that teardown paths need a separate *non*-throwing check, since throwing from a destructor or `noexcept` move calls `std::terminate`. RMM solved the same problem with exactly that split.

nanovdb/nanovdb/cuda/Buffer.h
kmuseth: I accept you argument for keeping it as is

nanovdb/nanovdb/cuda/Buffer.h
kmuseth: thx

kmuseth (approved):
thanks for the changes

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -0,0 +1,368 @@
...
+            throw std::runtime_error("nanovdb::cuda::Buffer: element count overflows the byte size");
+        return count * sizeof(T);
+    }
+
+    /// @brief Allocates @c count elements through the resource and records the
+    ///        new extent. A zero count allocates nothing.
+    /// @param count number of elements
+    /// @param stream cuda stream the allocation is ordered on; used by the
+    ///        stream-ordered form and ignored by the synchronous one
+    void allocate(size_t count, cudaStream_t stream)
```
sifakis, line 290: This helper (and the three other allocation sites) always requests `R::DEFAULT_ALIGNMENT` and then `static_cast<T*>`s the result — `alignof(T)` is never consulted. `checkedBytes(count)` accounts for `sizeof(T)`, but nothing accounts for the element type's alignment.

So `Buffer<T,R>` is correctly aligned for `T` only if `R::DEFAULT_ALIGNMENT >= alignof(T)`, and nothing enforces that. `Buffer<uint64_t, R>` over a resource with `DEFAULT_ALIGNMENT == 4` can yield a 4-byte-aligned pointer — UB on the host, `misaligned address` on the device.

Would something like this be better?

```cpp
static constexpr size_t ALIGNMENT =
    R::DEFAULT_ALIGNMENT > alignof(T) ? R::DEFAULT_ALIGNMENT : alignof(T);
```

used in place of `R::DEFAULT_ALIGNMENT` ...

harrism: @sifakis you commented on an already merged PR. Can you comment on #2301?

harrism: Otherwise, file an issue.

PR #2254, opened 2026-07-21
Fix builds with newer TBB versions

mmalecki:
Such as `TBB_INTERFACE_VERSION`.
Without it, the `#else` statement always triggers, breaking build for newer TBB version (albeit rare at this point).

### DIFF
--- a/openvdb/openvdb/thread/Threading.h
+++ b/openvdb/openvdb/thread/Threading.h
@@ -25,6 +25,7 @@
 ///   software doesn't need to provide compile time defines.
 #include <tbb/task_arena.h>
 #include <tbb/task_group.h>
+#include <tbb/version.h>
 
 namespace openvdb {
 OPENVDB_USE_VERSION_NAMESPACE
--- a/openvdb/openvdb/thread/Threading.h
+++ b/openvdb/openvdb/thread/Threading.h
@@ -13,19 +13,24 @@
     #define OPENVDB_THREAD_THREADING_RESTORE_DEPRECATION_MESSAGE_TASK
 #endif
 
+/// @warning  the below changes what includes are available from this header
+///   depending on whether version.h is available
+#if __has_include(<tbb/version.h>)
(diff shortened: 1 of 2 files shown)


openvdb/openvdb/thread/Threading.h
```diff
@@ -25,6 +25,7 @@
 ///   software doesn't need to provide compile time defines.
 #include <tbb/task_arena.h>
 #include <tbb/task_group.h>
+#include <tbb/version.h>
```
Idclip: The reason we don't do this is to support older TBB versions which didn't deploy with the version header (see the comment above - our current minimum supported TBB version is 2020.3). However, as we are now a minimum C++17 project, we might be able to use the __has_include preproc directive to achieve this:

```
/// @warning  the below changes what includes are available from this header depending on whether version.h is available
#if __has_include(<tbb/version.h>)
    #include <tbb/version.h>
#else
    /// @note tbb/task_arena.h is the ONLY include that persists from TBB 2020
    ///   to TBB 2021 to TBB 2023 that itself includes the TBB specific version
    ///   header files.
    ///   In TBB 2020, the version header was called tbb/stddef.h. In 2021+, it's
    ///   called tbb/version.h. We include tbb/task_arena.h here to indirectly
    ///   access the version defines in a consistent way so that downstream
    ///   software doesn't need to provide compile time defines. 
    #include <tbb/task_arena.h>
#endif

#include <tbb/task_group.h>
```

Would this work in your case?

mmalecki: Ah, reading comprehension when frantically trying to get something working: 1/10.
Apologies for the noise! I'll make this a conditional include - although most likely not today, but during the weekend.

Idclip: No worries, I've just made this change!

Idclip (approved):

apradhana (approved):

danrbailey (approved):

Issue #2255, opened 2026-07-21
Issue with DeviceResource (`cudaMallocAsync`) on vGPU without memory pool support


swahtz:
When attempting to use the AWS EC2 `g6f` instance type for CI in `fvdb-core` (working prototype on [this branch](https://github.com/openvdb/fvdb-core/tree/ci/g6f_instances)), we receive an error at the first call to `cudaMallocAsync` at `nanovdb/cuda/DeviceResource.h:22`:  `CUDA error 801: operation not supported`

Through some investigation, it appears that the stream-ordered memory pools capability is not exposed by the vGPU (`g6f` is a fractional L4 GPU instance).  Confirmed when running:

```
> #include <cuda_runtime.h>
> #include <stdio.h>
>
> int main() {
>   int supported = 0, driver = 0;
>   cudaDriverGetVersion(&driver);
>   cudaError_t e = cudaDeviceGetAttribute(
>       &supported, cudaDevAttrMemoryPoolsSupported, 0);
>   printf("driver=%d, query=%s, memoryPoolsSupported=%d\n",
>          driver, cudaGetErrorString(e), supported);
>   return e != cudaSuccess || !supported;
> }
```
`driver=13000, query=no error, memoryPoolsSupported=0`

These instances are running the latest AWS-supplied GRID driver (version 19.5/580.159.03) and can confirm the GPU device is available:
```
>nvidia-smi
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.159.03             Driver Version: 580.159.03     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L4-3Q                   On  |   00000000:31:00.0 Off |                    0 |
| N/A   N/A    P0            N/A  /  N/A  |       0MiB /   3072MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
[...]

PR #1807, opened 2024-05-08
NanoVDB v32.7.0

kmuseth:
Introducing NanoVDB v32.7.0, which is a major refactoring of v32.6, primarily related to the introduction of new namespaces and an accompanying restructuring of the directory layout of header files. This will likely require changes to client code! To this end I have included a shell script (`cmd/updateFiles.sh`) that will scan files and make (most) of the required changes.

The motivation for there these changes is twofold: First we want to use the same practice in terms of name spaces as OpenVDB and second the old version (32.6) was getting cluttered and messy.  Specifically the main changes are:

1. Virtually everything in NanoVDB is no using `nanovdb` as a namespace
2. Additionally the following nested namespaces are introduced: `math`, `tools`, `cuda`, `io`
3. these namespaces are reflected in matching restructures in the directories of header files as well as names of functions, kernels and a classes.

Here are some examples of changes:

```
nanovdb::CpuTimer -> nanovdb::util::Timer
nanovdb::GpuTimer -> nanovdb::util::cuda::Timer
nanovdb::gridStats -> nanovdb::tools::gridStats
nanovdb::Coord -> nanovdb::math::Coord
```

Issue #1798, opened 2024-04-24
[REQUEST] make nanoVDB CUDA async allocation optional so it can be used on vGPU


w0utert:
### Is your feature request related to a problem? Please describe.
Current nanoVDB implementation uses functions like `cudaMallocAsync` and `cudaMemcpyAsync`, for example in `CudaDeviceBuffer` when allocating or uploading data to the GPU. These functions are not available when using a vGPU that does not have unified memory enabled, which is common for example for GPU-enabled Azure VM's where the GPU is shared/sliced between multiple instances. Trying to run nanoVDB code on such a VM will result in CUDA 801 'not supported' exceptions.

### Describe the solution you'd like
Projects such as PyTorch usually implement the async code paths using a switch to enable/disable them, plus a fallback path that uses synchronous functions. If nanoVDB had something similar, that would be the perfect solution, save for potential efficiency disadvantages the asynchronous paths could have.

### Describe alternatives you've considered
For my situation there is not really an alternative, as I am not in a position to change hypervisor settings to enable unified memory support or use some other deployment target for the code I want to use with nanoVDB. The only option is to switch to a VM that uses passthrough GPU instead of vGPU, but again this is not something under my control.

w0utert:
Some more information/corrections: 
- This is only about `cudaMallocAsync` and `cudaMallocFree`, not `cudaMcpyAsync` etc.
- Upon closer inspection of `nanovdb/util/cuda/CudaUtils.h` I found out there is already a fallback path when building with CUDA versions prior to 11.2 (the version that introduced async malloc functions)

Based on this, I created PR #1799 that introduces macros `CUDA_MALLOC` and `CUDA_FREE`, and a define `NANOVDB_USE_SYNC_CUDA_MALLOC` that can be set by the host build system to force synchronous CUDA allocations.

harrism:
Closing the loop on this: the ask here led to PR #1799, whose approach was reimplemented (macros → namespaced inline functions) and shipped in #1807 as `NANOVDB_USE_SYNC_CUDA_MALLOC` in `nanovdb/util/cuda/Util.h` — a compile-time switch that redirects all of NanoVDB's device allocation to `cudaMalloc`/`cudaFree`. So the capability requested here has existed since v32.7.0, just under-documented.

[...]

PR #2256, opened 2026-07-22
NanoVDB: fail with an actionable diagnostic when CUDA memory pools are unavailable

harrism:
Addresses the failure mode in #2255 and complements #1798 / #1807's `NANOVDB_USE_SYNC_CUDA_MALLOC`. Related roadmap: #2232 (whose step 2 adds an injectable synchronous resource — the explicit, per-call form of the backend choice this PR keeps out of a hidden runtime path).

## What

Fractional vGPU configurations (e.g. AWS `g6f` L4 instances) expose no stream-ordered memory pools — `cudaDevAttrMemoryPoolsSupported == 0` — so NanoVDB's first `cudaMallocAsync` fails with `cudaErrorNotSupported` (801) with no indication of why or what to do.

Every device allocation in NanoVDB routes through the `util::cuda::mallocAsync`/`freeAsync` wrappers, so this PR improves the diagnostic at that single point **without silently changing allocation semantics**:

- On a device without memory pools, `mallocAsync` emits an actionable message naming `NANOVDB_USE_SYNC_CUDA_MALLOC` and returns `cudaErrorNotSupported` — which every caller already surfaces through `cudaCheck`. `freeAsync` mirrors the capability check.
- **It does not silently fall back to `cudaMalloc`.** A silent fallback would make an async-typed resource (`DeviceResource`, which advertises `is_async_resource`) quietly perform synchronous allocation — the type would misrepresent its runtime semantics, and a multi-stream caller would get unexpected serialization it never asked for. The choice of a synchronous backend belongs to the caller, explicitly: today via the macro, and (roadmap) via an injected synchronous resource under #2232.
- `NANOVDB_USE_SYNC_CUDA_MALLOC` remains the caller's compile-time opt-in to synchronous allocation, documented with the three-way selection and the one-definition-rule requirement (the wrappers are `inline`, so it must be defined identically in every TU). `memoryPoolsSupported()` exposes the cached per-device query.

## Deploying on a pool-less device

Define `NANOVDB_USE_SYNC_CUDA_MALLOC` project-wide. A CI probe can automate this: compile-and-run a one-line `cudaDevAttrMemoryPoolsSupported` check on the target instance and set the flag when it reports 0 — capability detection performed by the deployer, not hidden in the library.

## Testing

New `unittest/TestUtilCuda.cu`, built as **two binaries** — default (ctest `nanovdb_cuda_util_unit_test`) and `-DNANOVDB_USE_SYNC_CUDA_MALLOC` (ctest `nanovdb_cuda_util_sync_unit_test`) — covering both allocation modes. Tests: the cached query agrees with a direct per-device attribute query (and rejects out-of-range ids); and an allocate/use/free round-trip through the wrapper in whichever mode the build selects. The default binary's async round-trip skips on a device without memory pools (where those wrappers fail by design). Full suites green (CPU 153/153, `nanovdb_test_cuda` 53/53, memory-resource 8/8; compute-sanitizer clean; sm_89 / CUDA 12.6).

## Non-CUDA builds

No impact — `util/cuda/Util.h` is CUDA-domain and unreachable from `NANOVDB_USE_CUDA=OFF` builds.

### DIFF
--- a/nanovdb/nanovdb/util/cuda/Util.h
+++ b/nanovdb/nanovdb/util/cuda/Util.h
@@ -16,6 +16,7 @@
 
 #include <cuda.h>
 #include <cuda_runtime_api.h>
+#include <vector>
 #include <nanovdb/util/Util.h> // for stderr and NANOVDB_ASSERT
 
 // change 1 -> 0 to only perform asserts during debug builds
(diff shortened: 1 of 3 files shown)


swahtz:
I can confirm that this works on a `g6f` vGPU instance.

harrism:
Thanks for validating on `g6f` — that confirms the detection and the synchronous path both work on the real hardware.

Heads up that I've since changed the behavior in this PR, so it's worth a re-test: instead of *silently* falling back to `cudaMalloc` on a pool-less device, the wrappers now **fail with an actionable diagnostic** and return `cudaErrorNotSupported`. The reasoning is that a silent fallback makes `DeviceResource` — which advertises itself as an async resource — quietly perform synchronous allocation, so the type would misrepresent its own semantics; the choice of a synchronous backend should be the caller's, explicitly.

So on `g6f` you'll now define `NANOVDB_USE_SYNC_CUDA_MALLOC` (which the error message names). To keep that zero-touch in CI, a configure-time probe can set it automatically — details on #2255. If you're able to re-run on `g6f` with the macro defined, that would confirm the deploy path end-to-end.

swahtz:
I've rebuilt `fvdb-core` using the NanoVDB from this latest commit with `NANOVDB_USE_SYNC_CUDA_MALLOC` defined and can confirm that setting this flag works as expected on a `g6f` vGPU instance.

swahtz:
To test the expected behaviour without `NANOVDB_USE_SYNC_CUDA_MALLOC` defined, I recompiled and got the appropriate messge:

```
unit/test_accessors.py::TestAccessors::test_read_from_dense_cminor_0_cpu PASSED
unit/test_accessors.py::TestAccessors::test_read_from_dense_cminor_1_cuda NanoVDB: device 0 does not support stream-ordered CUDA memory pools required by cudaMallocAsync. Define NANOVDB_USE_SYNC_CUDA_MALLOC to allocate synchronously with cudaMalloc/cudaFree instead.
CUDA error 801: operation not supported (/home/ubuntu/fvdb-core/build/cp312-cp312-linux_x86_64-Release/_deps/nanovdb-src/nanovdb/nanovdb/cuda/DeviceResource.h:35)
```

My only note about the message should be to say "Define NANOVDB_USE_SYNC_CUDA_MALLOC **when building** …" to make it clear that this isn't a runtime environment variable that's respected.

swahtz (approved):
This looks like the right sized fix for this point in time for situations where a hardware configuration can't support the stream-ordered memory pool feature (time-share vGPUs) until we get the injectable memory allocator feature. 

My only note is at the end changing the message to make it clear that the variable needs to be defined when building so the end-user doesn't assume we're referring to a runtime environment variable.

harrism:
Thanks for validating both paths on `g6f` — macro-defined and the fail-loud message.

Good call on the wording; fixed in 3d9fb02c. The message now reads:

> Define NANOVDB_USE_SYNC_CUDA_MALLOC **when building** to allocate synchronously with cudaMalloc/cudaFree instead.

so it's clear it's a compile-time macro, not a runtime environment variable.

nanovdb/nanovdb/util/cuda/Util.h
```diff
@@ -70,42 +71,96 @@ namespace nanovdb {// =========================================================
...
+///        cudaErrorNotSupported rather than silently allocating synchronously.
 /// @param d_ptr Device pointer to allocated device memory
 /// @param size  Number of bytes to allocate
 /// @param stream The stream establishing the stream ordering contract and the memory pool to allocate from
 /// @return Cuda error code
-inline cudaError_t mallocAsync(void** d_ptr, size_t size, cudaStream_t stream){return cudaMallocAsync(d_ptr, size, stream);}
+inline cudaError_t mallocAsync(void** d_ptr, size_t size, cudaStream_t stream)
+{
+    int device = 0;
+    if (const cudaError_t err = cudaGetDevice(&device); err != cudaSuccess) return err;
```
kmuseth: for clarity I much prefer:

```
const cudaError_t err = cudaGetDevice(&device);
if ( err != cudaSuccess) return err;
```

nanovdb/nanovdb/util/cuda/Util.h
```diff
@@ -70,42 +71,96 @@ namespace nanovdb {// =========================================================
...
+///        device without memory pools reports cudaErrorNotSupported rather than
+///        calling cudaFreeAsync where it cannot succeed.
 /// @param d_ptr Device pointer that will be freed
 /// @param stream The stream establishing the stream ordering promise
 /// @return Cuda error code
-inline cudaError_t freeAsync(void* d_ptr, cudaStream_t stream){return cudaFreeAsync(d_ptr, stream);}
+inline cudaError_t freeAsync(void* d_ptr, cudaStream_t stream)
+{
+    int device = 0;
+    if (const cudaError_t err = cudaGetDevice(&device); err != cudaSuccess) return err;
```
kmuseth: same here

kmuseth (approved):
looks good with the exception of two minor cosmetic code changes

nanovdb/nanovdb/util/cuda/Util.h
harrism: Done in 9e3f4d59 — agreed, it reads better with the query, the test and the early return on separate lines.

nanovdb/nanovdb/util/cuda/Util.h
harrism: Done in 9e3f4d59 (same change in `freeAsync`).

nanovdb/nanovdb/util/cuda/Util.h
harrism: Answering my own question above, having gone looking: NanoVDB does have one — `cudaCheck(ans)` in `nanovdb/util/cuda/Util.h`, alongside `cudaCheckError()`, `cudaSync()` and `checkPtr()`. It's used in ~450 places across the library.

It isn't usable here, though. `cudaCheck` expands to `gpuAssert`, which does `fprintf` + **`exit(code)`** — it terminates the process rather than returning or throwing. `mallocAsync`/`freeAsync` are contracted to *return* `cudaError_t` so the caller can decide, which is the whole point of the diagnostic in this PR. So the explicit `return err` is right here, and your suggestion was purely about readability — which is now applied.

That `exit()` behaviour is worth a wider look, along with the stale comment above the macro ("No-op in release builds", while the guard is `#if 1 || defined(DEBUG) || defined(_DEBUG)`, so it is always on). I'll open a separate issue rather than widen this PR.

harrism: Filed as #2265 — it turned out to be more than a style question: `cudaCheck` in a destructor during static teardown sees `cudaErrorCudartUnloading` (a benign shutdown condition) and calls `exit()`. RMM's split between throwing and shutdown-safe checks is written up there.

PR #2258, opened 2026-07-23
VBM select-based inverse-map decode

swahtz:
## Summary

This PR replaces the inverse-map decode in the NanoVDB **VoxelBlockManager** with an output-centric **rank + select**, and consolidates the decode API down to a single **thread-local, per-slot** function.

The old `decodeInverseMaps` had every thread in the block sweep all 512 voxel slots of every leaf overlapping the block — **O(nLeaves × 512)** work per block, cooperative by necessity (scatter into shared memory behind `__syncthreads`). The new `decodeInverseMap` runs **one thread per output slot** and computes that slot's `(leafIndex, voxelOffset)` **entirely in registers**: a jumpMap popcount ranks the slot into its leaf, then the leaf's precomputed 9-bit `mPrefixSum` directory plus an in-word `__fns` bit select recover the voxel — **O(1) per slot**, no shared memory, no barrier, callable from divergent threads.

diagram1

## What changed

The decode is a **rank + select** over the VBM's bit-vectors:

- **rank** into the block's jumpMap (a popcount of leaf-start bits at or before the slot) selects which leaf owns the slot;
- **select** into that leaf's 512-bit value mask — the precomputed 9-bit prefix-sum directory finds the containing 64-bit word, an in-word `__fns` finds the bit — recovers the voxel offset.

This removes the old sweep's per-leaf full scan and its data-dependent inner branch, making the decode's work **independent of leaf sparsity**. And since each slot's result is fully thread-local, the shared-memory materialization the sweep required is no longer part of the algorithm — so it is no longer part of the API.

## API changes

- **Added** `decodeInverseMap` — per-slot decode into registers (composed from the also-public `jumpMapRank` and `selectVoxelInLeaf` helpers). No shared memory, no `__syncthreads`, divergent-safe.
- **Removed** `decodeInverseMaps` — the cooperative shared-memory form. It performed no cooperative *computation* (only cooperative materialization), and every block-level fact a consumer read from the materialized maps derives directly from the VBM metadata:
  - the block's first leaf **is** `firstLeafID`;
  - slot `p` starts a new leaf **iff** jumpMap bit `p` is set;
  - the spanned-leaf count is **1 + popcount(jumpMap)**.
- `computeBoxStencil` now takes the decoded `(leafIndex, voxelOffset)` by value instead of shared-memory arrays.

Migration:

```cpp
// before                                          // after
__shared__ uint32_t li[BW];                        uint32_t leafIndex;
__shared__ uint16_t vo[BW];                        uint16_t voxelOffset;
VBM::decodeInverseMaps(grid, firstLeafID,          VBM::decodeInverseMap(grid, firstLeafID,
    jumpMap, blockFirstOffset, li, vo);                jumpMap, blockFirstOffset, threadIdx.x,
VBM::computeBoxStencil(grid, li, vo, taps);            leafIndex, voxelOffset);
                                                   VBM::computeBoxStencil(grid, leafIndex,
                                                       voxelOffset, taps);
```

A consumer that genuinely needs cross-slot results can materialize them at the call site (store + `__syncthreads`) at exactly the old wrapper's cost — nothing is lost, the cost just isn't imposed on consumers that don't need it.

## A/B vs master — decode across leaf sparsity

Level-set spheres at narrow-band half-widths 1–32 voxels sweep per-leaf occupancy from 16% to 86% (occupancy is the old sweep's cost driver: sparser leaves mean more leaves — each costing a full 512-slot scan — per block). RTX PRO 6000 (sm_120), median of 50 iterations, decode-only:

| leaf occupancy      | width 64  | width 128 | width 512 |
| ------------------- | --------- | --------- | --------- |
| 16% (half-width 1)  | **2.20×** | 2.04×     | 1.72×     |
| 36% (half-width 3)  | 2.04×     | 1.97×     | 1.52×     |
| 60% (half-width 8)  | 2.00×     | 1.94×     | 1.40×     |
| 86% (half-width 32) | 2.00×     | 1.55×     | **1.24×** |

image

The sweep's per-voxel cost roughly doubles as occupancy falls (top panel, blue); the select stays flat (orange), so the win grows exactly where grids are sparsest. Width 64 is the strongest case at every occupancy — there the sweep needs 8 rounds of 64 threads to scan each 512-slot leaf, so it is slow even on dense grids and the select holds a uniform ~2× (2.20× peak). Consumer-side, the register decode also removes a barrier and 6 bytes/slot of shared memory from every stencil kernel's budget; on a streaming cached box-stencil consumer this measured an additional 8–15% over the shared-memory decode, independent of the decode speedup itself.

## Validation

- **Byte-identical outputs**: decode maps *and* all 27 box-stencil taps compared bytewise against the master sweep at widths 64/128/512, across 16–86% leaf occupancy (11 sphere configurations per width) — identical everywhere, including the final partial block and the beyond-last-active-voxel sentinels.
- The `TestNanoVDBCUDA.VoxelBlockManager_*` unit tests are updated to the new API and pass.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/VoxelBlockManager.cuh
+++ b/nanovdb/nanovdb/tools/cuda/VoxelBlockManager.cuh
@@ -19,7 +19,7 @@
         jumpMap on the GPU from a device-resident NanoGrid.
       - decodeInverseMaps (device): per-block SIMT decode of the inverse maps
         (sequential active-voxel index -> leaf ID + intra-leaf voxel offset),
-        executed cooperatively across a CUDA thread block.
+        with one thread per output slot across a CUDA thread block.
 */
 
 #ifndef NANOVDB_VOXELBLOCKMANAGER_CUH_HAS_BEEN_INCLUDED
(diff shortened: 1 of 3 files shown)


kmuseth (approved):
Looks good to me but please have @sifakis review this as well

sifakis:
Discussed this with @swahtz . Method is 100% sound. My recommendation (for now) is to just retain both APIs (i.e. both writing results into buffers, and not) for symmetry with the CPU case, where true cooperative decoding likely has a material benefit (avoids popcnts which are not SIMDified, other than in AVX512POPCNT). But both APIs (on CUDA) can use the new decoding approach.

swahtz:
> Discussed this with @swahtz . Method is 100% sound. My recommendation (for now) is to just retain both APIs (i.e. both writing results into buffers, and not) for symmetry with the CPU case, where true cooperative decoding likely has a material benefit (avoids popcnts which are not SIMDified, other than in AVX512POPCNT). But both APIs (on CUDA) can use the new decoding approach.

I've restored the `decodeInverseMaps` function with the same signature (which is just a wrapper over `decodeInverseMap`) so the API can retain both as well as test coverage.  I added a note to `decodeInverseMaps` that just explains when you'd choose to use either

kmuseth:
any comments from Efty?

swahtz:
> any comments from Efty?

Yes see above. We talked and I addressed the note

PR #2259, opened 2026-07-23
Added new actions to vdb tool

kmuseth:
New actions in vdb_tool:

Stack management
-copy — deep-copies VDB grids and/or Geometry by index onto the top of their stacks. Supports comma-separated indices, * wildcard, and an optional prefix= to rename copies. Defaults to vdb=0 (top grid).
-rename — renames a VDB grid and/or Geometry on the stack by age index.
-swap — swaps two VDB grids and/or two Geometry entries. O(1), no data copied.

Composite math (via tools::compMul / tools::compDiv)
-multiply (alias -mul) — per-voxel product of two scalar grids.
-divide — per-voxel quotient of two scalar grids.

Inspection
-stats — prints a value-derived table per grid: background, min, max, mean, std. dev., and for level sets: world-space surface area and enclosed volume. Intentionally complementary to -print which is structural-only.
-diagnose — runs tools::checkLevelSet / tools::checkFogVolume / NaN+Inf checks per grid. Prints OK or FAILED with indented messages. Options: vdb=, checks=N (default 9), fatal=true.

Changes to existing actions
-print — removed level=2 (background, value range, area, volume moved to -stats). Now level=0 and level=1 only.
-copy — vdb defaults to 0 so -copy alone duplicates the top grid.
-diagnose — gained checks=N option to control how many checks to run.

Bug fixes
Diagnostics.h — fixed a floating-point false positive in CheckLevelSet::checkBackground() where background / voxelSize could compute to just below LEVEL_SET_HALF_WIDTH due to IEEE 754 precision, causing a correctly-created level set to fail check #4. Fixed with a relative tolerance of 1e-6. Regression test added to TestDiagnostics.cc.

Config file improvement
Backslash line continuation in config files — a line ending with \ (trailing whitespace is ignored) is joined to the next line before tokenization, allowing long action invocations to be split across multiple lines. Unit test added (ToolConfigLineContinuation).

Documentation
README.md — added actions to the table (copy, diagnose, divide, multiply, rename, stats, swap); updated -print and -stats descriptions to reflect the structural vs value-derived distinction; added a new "Inspecting grids: -print vs -stats" section; documented backslash line continuation in the config file format section.
pendingchanges/openvdb_cmd.txt — added entries for all new actions, updated -print and -stats descriptions, added the Diagnostics.h float fix entry.


openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -642,6 +658,26 @@ void Tool::init()
...
+
+  mParser.addAction(
+     {"swap"}, "Swap two VDB grids and/or two Geometry entries on their respective stacks",
+    {{"vdb", "", "0,1", "comma-separated pair of age indices of VDB grids to swap (omit to skip VDB)"},
+     {"geo", "", "0,1", "comma-separated pair of age indices of Geometry to swap (omit to skip Geometry)"}},
+     [](){}, [&](){this->swap();});
+
+  mParser.addAction(
+     {"copy"}, "Deep-copy VDB grids and/or Geometry by index onto the top of their respective stacks",
+    {{"vdb",    "0", "0|0,1,2|*", "comma-separated age index/indices of VDB grids to copy, or \"*\" for all (default: 0, i.e. the most recently added VDB)"},
```
harrism: **Blocking.** Defaulting `vdb` to `"0"` here makes it impossible to skip the VDB stack, and makes `-copy geo=...` fail outright when no grids are loaded.

`Option::value` (Parser.h) *is* the per-action default, so `vdb_str` is never empty and the guard in `Tool::copy()`

```cpp
if (vdb_str.empty() && geo_str.empty()) { throw ... }
```

is unreachable for this action. Two consequences:

1. `-copy geo=0` silently deep-copies grid 0 as well, contradicting the "omit to skip" semantics that the `geo` docstring and the changelog entry both describe.
2. With an empty grid stack, `vdb_tool -read mesh.obj -copy geo=0` throws `copy: vdb index 0 is out of range (stack size 0)`.

Note `-rename` and `-swap` both default `vdb` to `""`, so `-copy` is the odd one out and the three actions don't behave consistently.

The bare-`-copy`-duplicates-the-top-grid convenience is worth keeping — it just needs to distinguish "user omitted the option" from "user passed `vdb=0`" rather than encoding the default as a value. E.g. leave the default `""` and treat *both* options empty as "vdb=0", which preserves `-copy` alone while letting `-copy geo=1` mean only geometry.

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -1404,6 +1464,268 @@ void Tool::clear()
...
+                          << std::setw(w)  << "volume"
+            << "\n" << std::string(aw + 2 + nw + 7*w, '-') << "\n";
+
+  for (size_t i = 0; i < grids.size(); ++i) {
+    auto& base = grids[i];
+    const int age = vdb_str == "*" ? (int)i : mParser.getVec<int>("vdb")[i];
+    auto grid = gridPtrCast<FloatGrid>(base);
+    if (!grid) {
+      std::clog << std::right << std::setw(aw) << age << "  "
+                << std::left  << std::setw(nw) << base->getName() << "  (not a FloatGrid, skipped)\n";
```
harrism: **Blocking.** `-stats` handles `FloatGrid` only, but it is replacing `-print level=2`, which handled considerably more — so this is a silent functional regression for anyone using the old option.

The `formatBackground` / `formatRange` helpers that `level=2` called dispatch over `FloatGrid`, `DoubleGrid`, `Int32Grid`, `Int64Grid`, `BoolGrid` (and `Vec3SGrid` for the background). After this PR, all of those report `(not a FloatGrid, skipped)` from `-stats` and have no other way to get their background or value range.

`levelSetArea`/`levelSetVolume` are legitimately float-level-set-only, but background/min/max/mean aren't — `tools::statistics()` is templated and works for the numeric grid types. Could `-stats` reuse the existing `gridBgStr`/`gridRangeStr` dispatch pattern (which already enumerates these types) so the removal of `level=2` isn't a net loss of capability?

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -4114,10 +4441,6 @@ void Tool::print(std::ostream& os) const
         if (level >= 1) row.push_back(formatBBox(bbox));
         if (level >= 1) row.push_back(formatBBoxd(grid.transform().indexToWorld(bbox)));
         if (level >= 1) row.push_back(formatNodes(grid));
```
harrism, line 4455: **Blocking (cleanup).** Removing the `level >= 2` rows here orphans roughly 100 lines of helper code further up in this file.

`formatBackground`, `formatRange`, `formatArea` and `formatVolume` each drop from two references to one (their own definition), and their helpers `gridBgStr` and `gridRangeStr` become transitively dead along with them. Because they're `inline` free functions, no compiler warning will fire — the dead code just accumulates.

Either delete them along with `level=2`, or reuse them from `Tool::stats()` (which would also address the non-float regression noted on that function).

openvdb_cmd/vdb_tool/src/unittest.cpp
```diff
@@ -3140,6 +3140,51 @@ TEST_F(Test_vdb_tool, ToolConfigHeader)
     std::remove(cfg.c_str());
 }
 
+TEST_F(Test_vdb_tool, ToolConfigLineContinuation)
```
harrism, line 3185: **Blocking.** This is the only test added by the PR, and it covers the config-file feature. None of the seven new actions has a functional test.

Worth covering at minimum:

- `-swap` actually reorders the stack (build two named grids, swap, assert the names are exchanged) — and, since the changelog motivates `-swap` with non-commutative CSG, that `-difference` after a swap gives the other result.
- `-copy` produces an independent deep copy rather than an alias (modify the copy, assert the original is unchanged), and that `prefix=` is applied.
- `-copy` validates all indices *before* mutating, which the implementation deliberately does — an out-of-range index should leave the stack untouched.
- `-rename` renames, and rejects an empty `name`.
- `-multiply` / `-divide` produce correct per-voxel results. `-divide` especially needs a division-by-zero / inactive-B-voxel case, which is currently entirely unexercised.
- `-diagnose fatal=true` throws on a deliberately corrupted grid, and does not throw on a clean one.

These are 337 lines of new behaviour; a regression in any of it would currently ship silently.

openvdb_cmd/vdb_tool/src/unittest.cpp
```diff
@@ -3140,6 +3140,51 @@ TEST_F(Test_vdb_tool, ToolConfigHeader)
...
+        out << "vdb_tool " << Tool::version() << "\n";
+        out << "sphere \\\n";
+        out << "  radius=2.0 \\\n";
+        out << "  name=cont_sphere\n";
+        out << "stats\n";
+    }
+    ASSERT_TRUE(fileExists(cfg));
+
+    // The config must load and produce a grid named "cont_sphere".
+    EXPECT_NO_THROW({
```
harrism, line 3238: **Non-blocking, but the test is weaker than its comment claims.** The comment above says "The config must load and produce a grid named `cont_sphere`", but `EXPECT_NO_THROW` never checks the name — or that the continuation joined anything at all.

An implementation that ignored `\` entirely would still pass this, as long as it didn't throw. Consider asserting on the resulting grid, e.g. run the tool and then check the stack has one grid named `cont_sphere` with `radius=2.0` applied, rather than only that no exception escaped.

pendingchanges/openvdb_cmd.txt
```diff
@@ -39,6 +45,7 @@ vdb_tool:
   - Action "-log" now redirects std::cerr and std::cout in addition to std::clog (so warnings/errors and -print output also land in the log file), uses unit-buffered output so "watch"/"tail -f" see updates in real time, writes a self-describing header (date, version, full command line), and gained "append" (append rather than truncate) and "tee" (also echo to the terminal, default true) options.
 
   Fixes:
+  - Fixed a floating-point false positive in "tools::CheckLevelSet::checkBackground()" in Diagnostics.h where the ratio "background / voxelSize" could compute to just below the expected half-width due to non-representable IEEE 754 values, causing a valid level set created with exactly LEVEL_SET_HALF_WIDTH voxels of narrow band to incorrectly fail check #4. The strict "<" comparison is now relaxed by a relative tolerance of 1e-6. A regression test covering this exact case was added to TestDiagnostics.cc.
```
harrism: **Blocking (changelog placement).** This entry documents a change to `openvdb/openvdb/tools/Diagnostics.h` — the core library — but it's filed in `openvdb_cmd.txt`, which is the vdb_tool changelog. `pendingchanges/openvdb_tools.txt` already exists and is the right home for it.

More broadly: is the `Diagnostics.h` fix in scope for this PR? It's a behaviour change to a public check in the core library, riding in a PR titled "Added new actions to vdb tool". Splitting it out would make it independently reviewable and bisectable, and would let it land on its own schedule.

The fix itself looks right, for what it's worth — `background / voxelSize` evaluated in `Real` from `float` operands can land just below `LEVEL_SET_HALF_WIDTH`, and a 1e-6 relative tolerance sits comfortably above float epsilon (~1.2e-7).

pendingchanges/openvdb_cmd.txt
```diff
@@ -30,7 +36,7 @@ vdb_tool:
...
+  - Improved format and information for the -print action: output is now a column-aligned Unicode table, and a new "level" option controls detail. "-print" reports only structural metadata that is independent of voxel values: level=0 shows the base table (age, name, type, class, dim, active voxel count, voxel size, memory usage); level=1 additionally shows the active-voxel bounding box in both index and world space and a per-level tree node-count column labeled "32->16->8" (e.g. "2->10->58" = 2 upper-internal, 10 lower-internal, 58 leaf nodes). Value-derived information (background, min, max, mean, stddev, surface area, enclosed volume) is intentionally not part of "-print" and is reported instead by the dedicated "-stats" action.
```
harrism: **Non-blocking.** Removing `-print level=2` is a user-facing breaking change, but it's written up under "Improvements". Any existing script or config file passing `level=2` will now fail option validation.

Suggest calling it out explicitly — either its own bullet under a breaking/removed heading, or at least a sentence here stating that `level=2` has been removed and pointing at `-stats` as the replacement. (The README section does explain the split well; it's just this changelog entry that reads as purely additive.)

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -1555,11 +1877,18 @@ void Tool::config()
...
             if (!header.isCompatible()) throw std::invalid_argument("readConf: incompatible version \""+line+"\"");
             std::vector<char*> args({&header.mMagic[0]});//parser is expecting first argument to the name of the executable
+            std::string accum;
             while (getline(file, line)) {
                 const size_t start = line.find_first_not_of(" \t"), stop = line.find_first_of("#%");
                 if (start >= stop) continue;// line is empty or starts with a comment
                 line = line.substr(start, stop - start);// remove leading whitespaces and tailing comments
                 line = line.substr(0, line.find_last_not_of(" \t") + 1);// remove tailing whitespaces
+                if (!line.empty() && line.back() == '\\') {
+                    accum += line.substr(0, line.size() - 1);// strip \ and accumulate
```
harrism, line 1931: **Non-blocking.** Two edge cases in the continuation logic:

1. **No separator is inserted when joining.** Trailing whitespace has already been stripped by the `find_last_not_of` line above, so a user who writes `radius=2.0\` without a space before the backslash gets `radius=2.0` concatenated directly onto the next line's first token — `radius=2.0name=x` — which then fails to parse with a confusing message. The test only covers the spaced form. Appending a space (or joining with `" "`) would make this robust.

2. **A trailing continuation at EOF is silently discarded.** If the last line of a config ends in `\`, the loop exits with `accum` non-empty and its contents are dropped without a diagnostic. Worth either flushing it or reporting `unterminated line continuation at end of file`.

Also note the `if (start >= stop) continue;` guard above skips blank and comment-only lines *without* flushing `accum`, so a comment line in the middle of a continuation is transparently absorbed. That's probably the behaviour you want, but it's worth a comment saying so explicitly, since it isn't obvious.

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -1404,6 +1464,268 @@ void Tool::clear()
...
+                          << std::setw(w)  << "max"
+                          << std::setw(w)  << "mean"
+                          << std::setw(w)  << "stddev"
+                          << std::setw(w)  << "area"
+                          << std::setw(w)  << "volume"
+            << "\n" << std::string(aw + 2 + nw + 7*w, '-') << "\n";
+
+  for (size_t i = 0; i < grids.size(); ++i) {
+    auto& base = grids[i];
+    const int age = vdb_str == "*" ? (int)i : mParser.getVec<int>("vdb")[i];
```
harrism: **Non-blocking.** `mParser.getVec<int>("vdb")` is re-evaluated on every row, making this O(n²) in the number of grids and re-running any `{}` expression substitution once per row.

The same vector is already computed above when building `grids`; hoisting it into a local before the loop (and indexing that) would be both cheaper and clearer.

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -111,6 +111,7 @@
 
 #ifdef VDB_TOOL_USE_MPEG
 #include <cstdlib>// for std::system
+#include <sstream>
```
harrism: **Non-blocking (nit).** This include is redundant — `Util.h` and `Parser.h` both `#include <sstream>` unconditionally, and `Tool.h` includes both.

Worth noting it's also in the wrong place for its apparent purpose: the new `Tool::diagnose()` uses `std::istringstream` on an unconditional code path, so if this were the only `<sstream>` it would break builds with `VDB_TOOL_USE_MPEG` off. It isn't, so nothing is broken today — but if the intent was to satisfy `diagnose()`, the include belongs outside the `#ifdef`. (Removing the paired `<sstream>` from `Geometry.h`'s PDAL block is a fine cleanup for the same reason.)

openvdb_cmd/vdb_tool/README.md
```diff
@@ -52,6 +55,7 @@ The vdb_tool is a versatile yet lightweight command-line utility that chains tog
 | **mesh2ls** | Convert a watertight polygon surface into a narrow-band level set, i.e. a narrow-band signed distance to a polygon mesh |
 | **min** | Given grids A and B, compute min(a, b) per voxel |
 | **movie** | Convert image and movie files to mpeg or animated gif files |
+| **multiply** | Given grids A and B, compute a * b per voxel |
```
harrism: **Non-blocking (nit).** The `-mul` alias isn't mentioned. Other aliased actions in this table document theirs (e.g. `soup2udf`), so worth matching: `**multiply** (alias **mul**)`.

harrism (changes requested):
[Claude assisted in this review.]

Thanks for this — the set of actions is a genuinely useful addition, and the `-print` (structural) vs `-stats` (value-derived) split is a good design call that's documented clearly in both the README and the changelog.

I've left inline comments marked **Blocking** or **Non-blocking**. Requesting changes for the blocking set; the rest are polish you're welcome to take or leave.

**Blocking**

1. **`-copy` can't skip the VDB stack, and throws on an empty one.** `copy`'s `vdb` option defaults to `"0"` while `-rename`/`-swap` default to `""`, so the "at least one of vdb/geo" guard in `Tool::copy()` is unreachable, `-copy geo=0` also copies a grid, and `-copy geo=0` with no grids loaded throws.
2. **`-stats` is a functional regression against the `-print level=2` it replaces.** The removed path handled Float/Double/Int32/Int64/Bool (and Vec3S backgrounds); `-stats` is `FloatGrid`-only, so those types lose background and value range with no replacement.
3. **Dead code left behind by the `level=2` removal** — `formatBackground`, `formatRange`, `formatArea`, `formatVolume` and their `gridBgStr`/`gridRangeStr` helpers, ~100 lines. Silent, since unused `inline` functions don't warn.
4. **No functional tests for any of the seven new actions.** The single new test covers config line continuation. `-divide`'s division-by-zero behaviour is entirely unexercised.
5. **The `Diagnostics.h` fix is documented in the wrong changelog** (`openvdb_cmd.txt` rather than `openvdb_tools.txt`), and is arguably out of scope for this PR — see the inline comment.

**One question, not inline anywhere**

The PR description says `vdb_tool/CMakeLists.txt` was changed to handle OpenEXR v3 vs v2 (`OpenEXR::OpenEXR` / `OpenEXR::IlmImf`) and to add `find_package(Imath)` + `Imath::ImathConfig`. That file isn't in the diff — this PR touches 7 files and `CMakeLists.txt` isn't one of them. Since it's described as fixing a real build failure on OpenEXR 3.x systems, it'd be easy to lose: was it dropped in a rebase, or is the description carried over from another branch?

**Separately, not a change request for this PR**

While reading `Tool.h` for this review I filed #2278 about the file being 4.2k lines with all definitions non-`inline` in the header. It's pre-existing and this PR is entirely consistent with the surrounding style — just noting it here so the two threads are connected.

kmuseth:
Thanks - lots of great feedback and comments. I believe [I've addressed all of them in this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2259/commits/caa379afa87a769c90aeec0dd1a68741cafa545d). Below is a detailed description of the changes:

Blocking issues:

- -copy: fixed the unreachable "at least one of vdb/geo" guard (vdb's default was "0", masking it) — vdb now defaults to "" like geo, matching rename/swap. -copy geo=0 no longer implicitly copies a grid, and no longer throws when the VDB stack is empty.
- -stats: fixed the functional regression vs. the old -print level=2 — now reports background/min/max/mean/stddev across Float, Double, Int32, Int64, Bool grids (plus background for Vec3S), instead of skipping every non-FloatGrid.
- Removed ~90 lines of dead code (formatBackground, formatRange, formatArea, formatVolume, gridBgStr, gridRangeStr) left behind by the -print level=2 removal.
- Added functional tests for all seven new actions (copy, rename, swap, diagnose, stats, multiply, divide), including edge cases raised in follow-up review (deep-copy independence, validate-before-mutate on out-of-range indices, empty-name rejection, swap actually reordering the stack + non-commutative -difference after swap, and -divide's zero/inactive-divisor behavior).
- Moved the Diagnostics.h changelog entry from openvdb_cmd.txt to openvdb_tools.txt (it fixes a tools:: bug, not a vdb_tool one).

Non-blocking / nits:

- Tool::config()'s line-continuation: joined lines now get a space separator (previously radius=2.0\ with no space before \ would concatenate directly onto the next line's token); a trailing \ at EOF now throws instead of silently dropping the accumulated line; added a comment documenting that blank/comment-only lines mid-continuation are intentionally absorbed.
- Tool::stats(): hoisted the vdb= age list into a local vector instead of re-evaluating mParser.getVec<int>("vdb") (and re-running any {} expression) on every row — was O(n²).
- Removed the redundant #include <sstream> from Tool.h's VDB_TOOL_USE_MPEG block (already pulled in unconditionally via Util.h/Parser.h).
- Strengthened the ToolConfigLineContinuation test to actually assert on the resulting grid's name and radius, rather than only checking tool.run() doesn't throw.
- Fixed the changelog write-up for -print/-stats to explicitly call out level=2's removal as a breaking change (was previously worded as purely additive), and updated the -stats/-copy changelog blurbs to match the corrected behavior.
- Fixed a systemic README/-help format=md gap: the markdown-table generator only ever printed each action's primary name, silently dropping aliases for ~25 actions (not just -multiply/-mul). Now emits (alias: x) / (aliases: x, y, z) generically; regenerated README.md's action table accordingly (also picked up two pre-existing unrelated drift issues: calc's math alias and the missing soup2ls row).

pendingchanges/openvdb_tools.txt
```diff
@@ -5,10 +5,4 @@ tools:
...
-    relative to the input mesh (larger than maxLength/2, where maxLength is the largest bounding-box
-    dimension). Such input left the internal offset-grid list empty, and process() then dereferenced
-    an empty vector, segfaulting in optimized builds where the guarding assertion is compiled out.
-    process() now checks for this and throws a descriptive ValueError (naming the voxel size and the
-    maximum allowed) instead of crashing. The check is confined to process() so that offset() -- used
-    by the offset-only path -- continues to accept coarser voxel sizes unaffected.
\ No newline at end of file
+  - Fixed a floating-point false positive in "tools::CheckLevelSet::checkBackground()" in Diagnostics.h where the ratio "background / voxelSize" could compute to just below the expected half-width due to non-representable IEEE 754 values, causing a valid level set created with exactly LEVEL_SET_HALF_WIDTH voxels of narrow band to incorrectly fail check #4. The strict "<" comparison is now relaxed by a relative tolerance of 1e-6. A regression test covering this exact case was added to TestDiagnostics.cc.
```
harrism, line 15: **Blocking — this is new, and it's my fault for sending you here.** Moving the entry into this file clobbered the `Fixes:` section that was already in it. The file shows `1+ / 7-`:

```diff
   Fixes:
-  - Fixed a crash in PolySoupToLevelSet::process() when the requested voxel size is too coarse
-    relative to the input mesh (larger than maxLength/2, ...)
-    ... segfaulting in optimized builds where the guarding assertion is compiled out.
-    ... throws a descriptive ValueError (naming the voxel size and the maximum allowed)
-    instead of crashing. ...
\ No newline at end of file
+  - Fixed a floating-point false positive in "tools::CheckLevelSet::checkBackground()" ...
```

That PolySoupToLevelSet entry is live on `master` — you added it yourself in #2266 on 2026-08-21. As it stands, merging this PR would silently drop a segfault fix from the 13.1 release notes.

Note the `\ No newline at end of file` marker on the old content: `master`'s copy of this file has no trailing newline, which is a common way for an append or a merge resolution to swallow the preceding paragraph. Since `cdc70439` is a merge from master, it may be worth a quick scan of that merge for the same failure mode elsewhere.

The fix is just restoring the PolySoup paragraph above the new one (and adding the trailing newline while you're here).

harrism (commented):
Thanks — I went through each item against the code at `cdc70439` rather than just the summary, and **10 of the 11 are genuinely fixed**. One new issue came in with the changelog move (commented inline), and one item from the last round is still outstanding.

**Verified fixed**

| Item | Evidence |
|---|---|
| `-copy` default | `vdb` now `""`, guard reachable, consistent with `-rename`/`-swap` |
| `-stats` regression | Dispatches Float/Double/Int32/Int64/Bool + Vec3S background; `(n/a)` only where genuinely undefined |
| Dead code | All six helpers (`formatBackground`, `formatRange`, `formatArea`, `formatVolume`, `gridBgStr`, `gridRangeStr`) at zero references |
| Tests | Six new `TEST_F`, +527 lines |
| Line continuation | Space separator, EOF now throws, absorption behaviour documented |
| `getVec` per row | Hoisted into an `ages` vector |
| Redundant `<sstream>` | Removed |
| Continuation test | Now asserts grid name *and* radius via world-space bbox |
| Breaking change | Explicit `BREAKING CHANGE:` in the `-print` entry |
| `-mul` alias | Fixed generically in the `format=md` generator |

The tests are substantive rather than box-ticking, which I appreciated. `ActionCompositeDivide` covers exactly the three cases raised — ordinary, active-zero divisor, inactive-B-voxel — asserting `std::isinf` on the latter two. `ActionCopy` proves deep-copy independence by mutating the copy's background and checking the original, and proves validate-before-mutate by confirming a single grid survives `-copy vdb=0,5`.

The alias fix is better than what was asked for: generalising the markdown generator caught ~25 actions rather than just `-mul`, plus two pre-existing drift issues.

**Still outstanding from the last round**

The `CMakeLists.txt` question — every inline comment got addressed, but this one was summary-only with no line to anchor to, so I suspect it just got missed. The PR description still advertises OpenEXR v3-vs-v2 detection (`OpenEXR::OpenEXR` / `OpenEXR::IlmImf`) and `find_package(Imath)` + `Imath::ImathConfig`, but the diff is 8 files and none is a `CMakeLists.txt`. If that fix exists somewhere it'd be good not to lose it; if it landed separately or was abandoned, the PR description should probably drop it.

**Non-blocking observations**

- Bare `-copy` now errors instead of duplicating the top grid. I'd suggested preserving that convenience by treating *both* options empty as `vdb=0`; strictness is a defensible call and the changelog and option docs are consistently updated, so I only want to confirm it was deliberate rather than a side-effect of the default change.
- `-stats` on a grid with no active voxels still prints `math::Stats` min/max as ±inf. Cosmetic, pre-existing in this PR, never previously raised — mentioning only for completeness.

Leaving this as a comment rather than a re-request: the one blocker is a two-line changelog restore, and the substance of the work is done. Happy to re-review once that's back.

pendingchanges/openvdb_tools.txt
kmuseth: Thanks for confirming the 10/11 — updated the PR description to remove the OpenEXR/CMakeLists.txt paragraph. That fix already landed separately in #2253 (commit 4775fa607) and is already an ancestor of this branch's CMakeLists.txt; there was nothing to restore in the diff, just a stale description.

updated change-log in openvdb_tools.txt

On the two non-blocking notes:

-copy strictness: deliberate, not a side-effect of the default change. vdb now defaults to "" (matching -rename/-swap's "omit to skip" convention) specifically so the "at least one of vdb/geo" guard is reachable and a bare -copy fails loudly instead of silently duplicating grid 0. Happy to revisit if you'd rather keep the old convenience, but I think strict is the safer default here.

-stats on an empty grid printing ±inf: fixed — min/max/mean/stddev now print (empty) for a grid with no active voxels, matching what the old -print level=2 path did via gridRangeStr. Added a regression test (ActionStats) covering it.

harrism (approved):

swahtz (approved):

PR #2260, opened 2026-07-23
Fix read accessor

kmuseth:
Branch: fix_ReadAccessor

Core fix — nanovdb/nanovdb/NanoVDB.h

Fixed a caching bug in nanovdb::ReadAccessor<0,1,2>. The new fixed accessor is enabled by default; the old (buggy) behaviour is preserved behind NANOVDB_USE_OLD_ACCESSOR.
CMake — nanovdb/nanovdb/CMakeLists.txt

Added NANOVDB_USE_OLD_ACCESSOR option (default OFF) with description "Use the old (buggy) ReadAccessor instead of the new fixed one". When OFF, passes -DNANOVDB_NO_OLD_ACCESSOR to suppress the fallback #define in NanoVDB.h.
CMake — openvdb_cmd/vdb_tool/CMakeLists.txt

Fixed build of vdb_tool with OpenEXR on macOS (the OpenEXR v2/v3 target name fix and Imath include path fix, same as on our other branch).
CI — .github/workflows/build.yml and nanovdb.yml

nanovdb/nanovdb/PNanoVDB.h — minor fix.
pendingchanges/nanovdb.txt — updated pending changes description.


nanovdb/nanovdb/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ option(NANOVDB_BUILD_PYTHON_MODULE "Build the nanovdb Python module" OFF)
 option(NANOVDB_USE_INTRINSICS "Build with hardware intrinsics support" OFF)
 option(NANOVDB_USE_CUDA "Build with CUDA support" OFF)
 option(NANOVDB_CUDA_KEEP_PTX "Keep CUDA PTX" OFF)
+option(NANOVDB_USE_OLD_ACCESSOR "Use the old (buggy) ReadAccessor instead of the new fixed one" OFF)
```
apradhana, line 34: Do you mind explaining why we need `NANOVDB_USE_OLD_ACCESSOR` CMake option at all? The pros of not having this is cleaner code and downstream users would immediately get the more efficient caching behavior. 

Also, a suggestion for the CMake option name: `NANOVDB_USE_READACCESSOR_LEAF_CACHE_ONLY`. IMHO, it's a better descriptor of the caching behavior that was used, where value lookups fall back to root after a leaf-cache miss

pendingchanges/nanovdb.txt
```diff
@@ -4,7 +4,7 @@ NanoVDB:
   - Added new _hostdev_ function named nanovdb::math::isoCrossing, which intersects a ray against a user-defined iso-surface.
 
   Improvements:
-  - TBD
+  - Added a new CMake option NANOVDB_USE_OLD_ACCESSOR (default ON) to control the ReadAccessor caching-bug fix without editing NanoVDB.h. The macro is still enabled by default in the header; setting -DNANOVDB_USE_OLD_ACCESSOR=OFF defines NANOVDB_NO_OLD_ACCESSOR, which suppresses it (a compiler -D cannot un-define a hard #define, hence the negative macro). When NANOVDB_USE_OLD_ACCESSOR if set to OFF ReadAccesor<0,1,2> and ReadAccessor<0,1> will behave like the openvdb::ValueAccessor, as oppose to merely using cached leaf nodes.
```
apradhana: As is, the default of this CMake option is `OFF` :)

openvdb_cmd/vdb_tool/CMakeLists.txt
```diff
@@ -123,8 +123,17 @@ endif()
...
+    target_link_libraries(vdb_tool_common INTERFACE OpenEXR::OpenEXR)
+    # Imath was split from OpenEXR in v3; ensure its headers are on the include
+    # path since OpenEXR's own headers still use bare <ImathBox.h> includes.
+    find_package(Imath QUIET)
+    if(TARGET Imath::ImathConfig)
+      target_link_libraries(vdb_tool_common INTERFACE Imath::ImathConfig)
+    endif()
+  else()
+    target_link_libraries(vdb_tool_common INTERFACE OpenEXR::IlmImf)
+  endif()
```
apradhana, line 132: I think a better solution is:
```
if(OPENVDB_TOOL_USE_EXR)
  target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_EXR")

  find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED)
  find_package(Imath ${MINIMUM_IMATH_VERSION} CONFIG REQUIRED)

  if(TARGET OpenEXR::OpenEXR)
    target_link_libraries(vdb_tool_common INTERFACE OpenEXR::OpenEXR Imath::Imath)
  else()
    target_link_libraries(vdb_tool_common INTERFACE OpenEXR::IlmImf Imath::Imath)
  endif()
endif()
```
This keeps the PR's proposed solution for the OpenEXR target naming, but still requires and links the real Imath target in both paths.

I see that   `find_package(Imath ${MINIMUM_IMATH_VERSION} CONFIG REQUIRED)` was deleted during merging with `master`.

nanovdb/nanovdb/CMakeLists.txt
kmuseth: The old code-path is included (but disabled by default) since there is speculation that some applications (notably WENO on GPU) might actually perform better with the old (buggy) version. Incidentally, the old ReadAccessor <0,1,2> should perform identically to the ReadAccessor<0>, so this PR still allows for the same behaviour (even if NANOVDB_USE_OLD_ACCESSOR is OFF). However, it's easier to make this performance evaluation by changing NANOVDB_USE_OLD_ACCESSOR in a CMake file than replacing ReadAccess<0,1,2> with ReadAccessor<0> through client code. In other words, this is a temporary solution and in the near future I will do exactly what you're suggesting - deleting the old code and completely removing NANOVDB_USE_OLD_ACCESSOR :) Hope that explains it?

pendingchanges/nanovdb.txt
kmuseth: good catch - I'll update the documentation

kmuseth: fixed in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2260/commits/ebba9eea26378a74a6216fc59867ab89b26efdbb)

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: made the suggested change in [this commit](https://github.com/AcademySoftwareFoundation/openvdb/pull/2260/commits/38d3c486e4de323d5683577e6a1ac0131e12049f)

nanovdb/nanovdb/CMakeLists.txt
apradhana: Yes, that makes sense. I was going to run a benchmark on, say on RayTracing for old v.s. new accessor this morning, but I was running out of time. I didn't know that WENO on GPU may perform better with old accessor.

harrism: I think it would be better to remove the old code now since there is a workaround available to regain the old behavior (`ReadAccessor<0>`), and you plan to delete it anyway. Otherwise it is confusing, error prone (there are already errors in your implementation and documentation), increases the support and testing surface, and risks silent ODR violation. Not to mention that the `#ifdef` interleaved into an `if constexpr` chain is hard to read and maintain.

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -158,6 +158,13 @@
 // Use this to switch between std::ofstream or FILE implementations
 //#define NANOVDB_USE_IOSTREAMS
 
+// Fix the caching bug in the ReadAccessor. Enabled by default; define
+// NANOVDB_NO_OLD_ACCESSOR (e.g. via CMake) to turn it off.
+#if !defined(NANOVDB_USE_OLD_ACCESSOR) && !defined(NANOVDB_NO_OLD_ACCESSOR)
+#define NANOVDB_USE_OLD_ACCESSOR
+#endif
+
```
harrism: **🚩 issue:** ‏With no macros defined, this defines `NANOVDB_USE_OLD_ACCESSOR` — i.e. the buggy path. Verified empirically against the branch:

```
$ g++ -std=c++17 -I. -fsyntax-only probe.cc
RESULT: OLD (buggy) accessor is ACTIVE
$ g++ -std=c++17 -I. -DNANOVDB_NO_OLD_ACCESSOR -fsyntax-only probe.cc
RESULT: NEW (fixed) accessor is ACTIVE
```

This contradicts the PR body ("The new fixed accessor is enabled by default"), `pendingchanges/nanovdb.txt` ("This is the new default behaviour"), and the header comment here ("Enabled by default...").

The fix only reaches consumers that link the in-tree `nanovdb INTERFACE` target. NanoVDB is not exported as a CMake target — nanovdb/nanovdb/CMakeLists.txt:376-383 is `install(FILES ...)` only, no `install(TARGETS nanovdb EXPORT ...)`. So every downstream consumer of installed or vendored headers (the dominant NanoVDB usage pattern) keeps the bug.

My suggested fix is to delete both macros and this block of code and get rid of the old behavior, since we have a workaround (`ReadAccessor<0>`).

If that is not satisfactory, then I suggest to at least delete this header block entirely and invert the CMake side:

```
if(NANOVDB_USE_OLD_ACCESSOR)
  target_compile_definitions(nanovdb INTERFACE -DNANOVDB_USE_OLD_ACCESSOR)
endif()
```

Undefined == fixed, defined == old. The `NANOVDB_NO_OLD_ACCESSOR` double-negative and the "a `-D` cannot un-define" comment both disappear, since nothing self-defines the macro anymore.

If you do NOT delete this block of code, then the comment needs to be fixed -- it is logically ambiguous. WHICH is enabled by default?

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -5391,9 +5414,14 @@ class ReadAccessor<BuildT, 0, 1, 2>
 #endif
         if constexpr(OpT::LEVEL <= 0) {
             if (this->isCached<LeafT>(dirty)) return ((LeafT*)mNode[0])->template setAndCache<OpT>(ijk, *this, args...);
-        } else if constexpr(OpT::LEVEL <= 1) {
+        }
+        if constexpr(OpT::LEVEL <= 1) {
```
harrism, line 5419: **🚩 issue:** ‏This one is missing the 

```
#ifdef NANOVDB_USE_OLD_ACCESSOR
        else
#endif
```

Therefore defining `NANOVDB_USE_OLD_ACCESSOR` will not actually recover the former behavior for `set<SetValue>`.

openvdb_cmd/vdb_tool/CMakeLists.txt
```diff
@@ -123,8 +123,18 @@ endif()
 if(OPENVDB_TOOL_USE_EXR)
   target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_EXR")
   find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED)
-  find_package(Imath ${MINIMUM_IMATH_VERSION} CONFIG REQUIRED)
-  target_link_libraries(vdb_tool_common INTERFACE OpenEXR::OpenEXR Imath::Imath)
+  if(OPENVDB_TOOL_USE_EXR)
+    target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_EXR")
+
+    find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED)
```
harrism: **🚩 issue:** First: this looks unrelated to the `ReadAccessor` fix, and therefore should be in a separate PR. Second: this looks like a copy/paste mistake, because lines 126-129 are duplicated from lines 123-125: 
```
if(OPENVDB_TOOL_USE_EXR)
  target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_EXR")
  find_package(OpenEXR ${MINIMUM_OPENEXR_VERSION} REQUIRED)
```
This causes the if() to be nested inside an if with the same condition and the `target_compile_definitions` and `find_package` to be executed twice. Note when the duplicated lines are deleted, the inner `endif` will also need to be deleted.

nanovdb/nanovdb/PNanoVDB.h
```diff
@@ -1088,7 +1088,7 @@ PNANOVDB_FORCE_INLINE void pnanovdb_write_vec3(pnanovdb_buf_t buf, pnanovdb_addr
 
 #define PNANOVDB_MAJOR_VERSION_NUMBER 32// reflects changes to the ABI
 #define PNANOVDB_MINOR_VERSION_NUMBER  9// reflects changes to the API but not ABI
-#define PNANOVDB_PATCH_VERSION_NUMBER  1// reflects bug-fixes with no ABI or API changes
+#define PNANOVDB_PATCH_VERSION_NUMBER  2// reflects bug-fixes with no ABI or API changes
```
harrism, line 1091: **💡 suggestion:** ‏This PR changes the default behavior of one of the most-used methods in the library. With this patch version bump the change will land silently for many. For discussion: should this be a minor version bump instead?

nanovdb/nanovdb/NanoVDB.h
```diff
@@ -158,6 +158,13 @@
 // Use this to switch between std::ofstream or FILE implementations
 //#define NANOVDB_USE_IOSTREAMS
 
+// Fix the caching bug in the ReadAccessor. Enabled by default; define
+// NANOVDB_NO_OLD_ACCESSOR (e.g. via CMake) to turn it off.
+#if !defined(NANOVDB_USE_OLD_ACCESSOR) && !defined(NANOVDB_NO_OLD_ACCESSOR)
+#define NANOVDB_USE_OLD_ACCESSOR
+#endif
+
+// Uncomment to use (slower) branched version of LeafData<FpN,...>::getValue
```
harrism: **🚩 issue:** ‏This comment is backwards. The following line defines `PFN_BRANCHLESS`, so the comment should read: 
```suggestion
// Comment out to use (slower) branched version of LeafData<FpN,...>::getValue
```

nanovdb/nanovdb/NanoVDB.h
harrism: **🚩 issue:** ‏Also, this is an ODR hazard. The macro changes the bodies of inline __hostdev__ member templates in a header. If one TU includes this header with `NANOVDB_USE_OLD_ACCESSOR` defined and another TU includes it without that definition (or with `NANOVDB_NO_OLD_ACCESSOR`, then that's a *silent* ODR violation and the linker will pick one arbitrarily. 

At a minimum, please document that the macros must be defined (or undefined) uniformly across the whole link. (This is another good reason to just delete the old code).

pendingchanges/nanovdb.txt
```diff
@@ -4,7 +4,7 @@ NanoVDB:
   - Added new _hostdev_ function named nanovdb::math::isoCrossing, which intersects a ray against a user-defined iso-surface.
 
   Improvements:
-  - TBD
+  - Fixed a caching bug in ReadAccessor<0,1,2> and ReadAccessor<0,1>: the accessors now cache nodes at all requested levels (matching the behaviour of openvdb::ValueAccessor) rather than caching only leaf nodes. This is the new default behaviour. The previous (buggy) behaviour can be restored by setting the CMake option -DNANOVDB_USE_OLD_ACCESSOR=ON, or by defining the macro NANOVDB_USE_OLD_ACCESSOR before including NanoVDB.h.
```
harrism: **🤓 nitpick (non-blocking):** ‏this note names only `<0,1,2>` and `<0,1>`; the generic `<LEVEL0,LEVEL1,-1>` specialization also covers `<1,2>` and `<0,2>`, which this PR changes too. Also this is a bug fix filed under "Improvements:" when a "Fixes:" section is just below it.

harrism (changes requested):
This is an important performance fix for some of the most-used device functions in the library. The fix itself (removing the `else` from the `if constexpr` chain to ensure caching is not skipped on a top-level miss) is good.

There are several problems. See comments on the code as well as the following.

1. No benchmarks. I know you did benchmarks in another branch, but they should be included in the fix PR to justify merging, since it's possible this could be a regression in some cases, so that should at least be documented.
2. No tests. A correctness fix to NanoVDB's hottest data structure would ship with zero test coverage, and now there are two compile-time configurations, neither exercised. `TEST_F(TestNanoVDB, RandomReadAccessor)` (unittest/TestNanoVDB.cc:5709) exercises values but asserts nothing about which cache level served them, so it passes identically before and after — it cannot regression-guard this.

Minimum viable coverage:
- Populate the accessor, then query a coord that misses the leaf cache but shares the parent internal node; assert acc.getNode<1>()/getNode<2>() are non-null and have the expected origin(). That assertion fails on master and passes here.
- Cover both set and get, and the <0,1> / <1,2> / <0,2> specializations, not just <0,1,2>.
- A CI job building with -DNANOVDB_USE_OLD_ACCESSOR=ON, otherwise the compat path bit-rots immediately (it's already broken per #2). (unless this is removed as suggested)

3. Unrelated vdb_tool CMake change should be split out.

openvdb_cmd/vdb_tool/CMakeLists.txt
kmuseth: fixed

nanovdb/nanovdb/NanoVDB.h
kmuseth: fixed

nanovdb/nanovdb/NanoVDB.h
kmuseth: fixed

nanovdb/nanovdb/CMakeLists.txt
kmuseth: I think we should keep the old code around for a little longer since many of us are still performing benchmark tests

nanovdb/nanovdb/NanoVDB.h
apradhana: Good catch, Mark. Thanks for catching this

apradhana (approved):

nanovdb/nanovdb/PNanoVDB.h
kmuseth: As stated in the documentation for our semantic versioning, it's a patch bump because neither the ABI nor the API are changing. I do, however, agree that it might have a significant impact on the performance, but that's not what our versioning is designed to communicate. That will hopefully be clear from the documentation itself. Regardless, this will all be packed up in the next major release of OpenVDB (version 14), so the version numbering of NanoVDB plays a small practical role (mostly its major number is used to track ABI changes).

kmuseth:
I will add more unit tests.

However, benchmark tests are not part of this PR, not because they haven't been written, more because it's unrelated to this PR that fixes an actual bug - regardless of its impact on performance. If it turns out that other caching schemes perform better for some applications, they should be using specializations like ReadAccessor<0> instead of a buggy ReadAccessor<0,1,2>.

openvdb_cmd/vdb_tool/CMakeLists.txt
harrism: This really should be in a separate PR...

nanovdb/nanovdb/NanoVDB.h
harrism: Uh, I mentioned multiple problems in this thread. Which did you fix?

harrism:
> I will add more unit tests.
> 
> However, benchmark tests are not part of this PR, not because they haven't been written, more because it's unrelated to this PR that fixes an actual bug - regardless of its impact on performance. If it turns out that other caching schemes perform better for some applications, they should be using specializations like ReadAccessor<0> instead of a buggy ReadAccessor<0,1,2>.

Yeah I was asking for the PR description to include the results of your benchmarks, not for you to add benchmark code.

kmuseth:
[RESULTS.pdf](https://github.com/user-attachments/files/30436179/RESULTS.pdf)

> Yeah I was asking for the PR description to include the results of your benchmarks, not for you to add benchmark code.

PR #2266, opened 2026-08-04
added vdb option to -soup2ls action and improved error handeling

kmuseth:
I've made two improvements to the vdb_tool:

- Better error handling when the voxel size exceeds the bbox of the input mesh
- Added an option, namesd"vdb", to the action -soup2ls. It is used to specify what vdb grids to output. By default "vdb=0", but you can add more, e.g. "vdb=0,1,2", or all "vdb=*".


openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -2193,13 +2197,48 @@ void Tool::soupToLevelSet()
...
+  } else {
+    levels = mParser.getVec<int>("vdb");
+    if (levels.empty()) levels.push_back(0);
+  }
+
+  // Validate every requested level BEFORE pushing any grid, so an out-of-range
+  // index produces a clean error with no partial output on the stack.
+  for (const int lvl : levels) {
+    if (lvl < 0 || lvl >= count) {
+      throw std::invalid_argument("soup2ls: requested output grid vdb=" + std::to_string(lvl) +
```
swahtz, line 2227: **💡 suggestion:** ‏Use `OPENVDB_THROW` to be consistent with other throws

openvdb/openvdb/tools/PolySoupToLevelSet.h
```diff
@@ -331,6 +337,17 @@ void PolySoupToLevelSet<GridType>::process(const ShrinkWrapT &D, ProgressT *prog
         mGrids.push_back(this->offset(dx, offset_mode));
     }
 
+    // The loop above produces no grids when mMinVoxelSize > mMaxVoxelSize (i.e. the
```
swahtz, line 346: **☑️ todo:** ‏We should have a test for this case now that it's fixed.

openvdb/openvdb/tools/PolySoupToLevelSet.h
swahtz, line 341: **❓ question:** ‏ This loop doesn't produce any grids when  `mMinVoxelSize>mMaxVoxelSize` but the assertions in the constructor fire when `mMinVoxelSize>mMaxVoxelSize/2`, is that the correct behaviour here?

swahtz (changes requested):
Changes look good, I would just request we have some tests for the fixed crash when the requested voxel size is too coarse for the input and a basic test for the output `vdb=*` and `vdb=0,1,2` new features.

openvdb/openvdb/tools/PolySoupToLevelSet.h
kmuseth: Good question - I see your confusion. The bound is mMaxVoxelSize/2, not mMaxVoxelSize, because a single hierarchy level is not enough — process() would skip all upsample/shrinkWrap iterations and return the raw coarse offset with no refinement. The assertion ensures at least two levels exist (fine dx = mMinVoxelSize and coarse dx = 2*mMinVoxelSize). The stricter empty-loop case (mMinVoxelSize > mMaxVoxelSize) is caught separately in process() with an explicit throw.

kmuseth: [I improved the documentation to explain this better](https://github.com/AcademySoftwareFoundation/openvdb/pull/2266/commits/7d5cc838adeab4a3b49315591ca6e3ef6dee3fc4)

swahtz (approved):
Thanks for clarifying my questions

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: OPENVDB_THROW adds two things over a raw throw: it embeds __FILE__/__LINE__ in the exception message, and it throws an openvdb::Exception subclass instead of a std:: type.

Neither matters in vdb_tool:

File/line info is only useful if someone catches and inspects the exception message programmatically. In a CLI tool the top-level handler just prints e.what() to stderr and exits — the file/line noise in the message would be confusing to end users, not helpful.
Typed openvdb::Exception is only useful if a caller needs to catch (openvdb::ValueError&) separately from catch (openvdb::RuntimeError&). In CLI code there's no such caller — everything collapses to a single generic catch.
The only scenario where it would help is debugging a crash during development (file/line in the message), but you'd use a debugger for that anyway. It's not worth the churn.

PR #2267, opened 2026-08-04
added build support on Ampere

kmuseth:
ex_make_mgpu_nanovdb uses TF32 WMMA which requires SM_80+ (Ampere or newer)

changed cmake to address this issue

### DIFF
--- a/nanovdb/nanovdb/examples/CMakeLists.txt
+++ b/nanovdb/nanovdb/examples/CMakeLists.txt
@@ -117,8 +117,17 @@ nanovdb_example(NAME "ex_coarsen_nanovdb_cuda" OPENVDB)
 nanovdb_example(NAME "ex_mesh_to_grid_cuda" OPENVDB)
 
 if(CUDAToolkit_FOUND)
-  nanovdb_example(NAME "ex_make_mgpu_nanovdb") # requires cuRAND
-  target_link_libraries(ex_make_mgpu_nanovdb PRIVATE CUDA::curand)
+  # ex_make_mgpu_nanovdb uses TF32 WMMA which requires SM_80+ (Ampere or newer)
+  execute_process(
+    COMMAND bash -c "nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | sort -n | tail -1 | tr -d '.'"
+    OUTPUT_VARIABLE _max_cuda_arch OUTPUT_STRIP_TRAILING_WHITESPACE)
+  if(_max_cuda_arch GREATER_EQUAL 80)
+    nanovdb_example(NAME "ex_make_mgpu_nanovdb") # requires cuRAND and SM_80+ (TF32 WMMA)
+    target_link_libraries(ex_make_mgpu_nanovdb PRIVATE CUDA::curand)
+  else()
+    message(STATUS "Skipping ex_make_mgpu_nanovdb: requires SM_80+ (TF32), detected SM_${_max_cuda_arch}")
+  endif()
+  unset(_max_cuda_arch)
 endif()
 
 if(NANOVDB_USE_MAGICAVOXEL)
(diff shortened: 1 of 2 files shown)


nanovdb/nanovdb/examples/CMakeLists.txt
```diff
@@ -117,8 +117,17 @@ nanovdb_example(NAME "ex_coarsen_nanovdb_cuda" OPENVDB)
 nanovdb_example(NAME "ex_mesh_to_grid_cuda" OPENVDB)
 
 if(CUDAToolkit_FOUND)
-  nanovdb_example(NAME "ex_make_mgpu_nanovdb") # requires cuRAND
-  target_link_libraries(ex_make_mgpu_nanovdb PRIVATE CUDA::curand)
+  # ex_make_mgpu_nanovdb uses TF32 WMMA which requires SM_80+ (Ampere or newer)
+  execute_process(
+    COMMAND bash -c "nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | sort -n | tail -1 | tr -d '.'"
```
swahtz: What if you're building on a machine that doesn't have a GPU and going to run it on an SM >=8.0 machine?  Or are we assuming all examples are built/run on the same machine and/or they're not used as CI tests (which might have this behaviour)?

apradhana: For `Windows`, assuming availability of bash is a hard requirement. Can we rely on something like `CMAKE_CUDA_ARCHITECTURES` (see, e.g. [this](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html)) instead of probing the local machine with `bash` and `nvidia-smi`?

apradhana: I asked Codex to scan `CMAKE_CUDA_ARCHITECTURES` and only add target if at least one architecture is `>=80`
```
if(CUDAToolkit_FOUND)
  set(_nanovdb_has_sm80 OFF)

  foreach(_cuda_arch IN LISTS CMAKE_CUDA_ARCHITECTURES)
    string(REGEX MATCH "^[0-9]+" _cuda_arch_number "${_cuda_arch}")
    if(_cuda_arch_number GREATER_EQUAL 80)
      set(_nanovdb_has_sm80 ON)
    endif()
  endforeach()

  if(_nanovdb_has_sm80)
    nanovdb_example(NAME "ex_make_mgpu_nanovdb") # requires cuRAND and SM_80+ (TF32 WMMA)
    target_link_libraries(ex_make_mgpu_nanovdb PRIVATE CUDA::curand)
  else()
    message(STATUS "Skipping ex_make_mgpu_nanovdb: requires CUDA architecture 80 or newer")
  endif()

  unset(_nanovdb_has_sm80)
  unset(_cuda_arch_number)
  unset(_cuda_arch)
endif()
```

kmuseth: Good catch, both of you — the nvidia-smi/bash probe was checking the wrong thing (the configure-time machine's GPU) in the wrong way (not portable to Windows). Replaced it with a check against CMAKE_CUDA_ARCHITECTURES instead, which already governs what SM architectures this build actually targets (and is what the rest of nanovdb/CMakeLists.txt uses for this purpose too, defaulting to 75 when unset) — so it correctly handles cross-compiling and GPU-less CI runners, and has no bash/nvidia-smi dependency at all.

@apradhana — used your Codex-suggested loop over CMAKE_CUDA_ARCHITECTURES as-is. Verified the arch-matching regex/comparison standalone against 75, 80, 86, the -real/-virtual suffix forms, native, and empty/multi-value lists — all resolve as expected.

swahtz (approved):
Thanks for addressing my notes

apradhana (approved):
Thanks for addressing the comment regarding `bash`+`nvidia-smi` approach, Ken. I tested the CMake with CUDA architecture: 75 skips, while 80, 75;80, and suffixed values like 80-real build as expected.

Issue #2264, opened 2026-07-31
NanoVDB: please run the CUDA unit tests on a GPU in CI


harrism:
### Request

Please add GPU-backed execution of the NanoVDB CUDA unit tests to CI. Today they are compiled and then skipped, so no device-side runtime behaviour in NanoVDB is verified before merge.

### Current state

All three `ctest` invocations in `.github/workflows/nanovdb.yml` carry the same filter:

```yaml
run: cd build && sudo ctest -V -E ".*cuda.*|.*mgpu.*"
```

Of the 7 registered tests, **5 never execute**:

| test | runs in CI |
|---|---|
| `nanovdb_unit_test` | yes |
| `nanovdb_openvdb_unit_test` | yes |
| `nanovdb_cuda_unit_test` | **no** |
| `nanovdb_mgpu_unit_test` | **no** |
| `nanovdb_cuda_memory_resource_unit_test` | **no** |
| `nanovdb_cuda_util_unit_test` | **no** |
| `nanovdb_cuda_util_sync_unit_test` | **no** |

#2251 adds a sixth (`nanovdb_cuda_buffer_unit_test`), which will also be skipped.

The exclusion is correct given the runners — `ubuntu-22.04-8c-32g-300h` and `windows-2022-8c-32g-300h` have no GPU, so these tests would fail with CUDA error 35. The Linux job does configure `-DNANOVDB_USE_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES="80"`, so **compilation** of device code is covered. Runtime behaviour is not.

### Why this matters now

Every device-side guarantee in NanoVDB is currently unverified by CI. Recent work that CI cannot check: stream-ordered allocation and stream-retained frees (#2231), memory-pool capability detection and its fallback (#2256), buffer resize/copy/move semantics and graph-capturability (#2251), and the distributed merge-path bounds fix (#2248). Regressions in any of these reach `master` green.

There are second-order effects too. There is no place to put a performance regression guard — relevant for #2260, where the `ReadAccessor` caching fix is invisible to value assertions and observable only as a timing difference, so a benchmark is the only possible guard. And correctness questions about GPU behaviour cannot be settled by CI at all; they currently require a maintainer to run by hand on their own hardware and report back, which is how the last several device-side issues were resolved.

### What would satisfy this

In rough order of preference, acknowledging that GPU runner time costs money:

[...]

PR #2268, opened 2026-08-04
NanoVDB: inject a resource into TopologyBuilder scratch, and align cuda::Buffer with cuda::buffer (CUDA)

harrism:
Follow-up to #2231 and #2251, part of #2232 (NanoVDB injectable CUDA memory resources). This is **B1** of step 2's retrofit; see the roadmap for the B1/B2/B3 split.

Two related changes: the `TopologyBuilder` scratch retrofit, and the `cuda::Buffer` naming alignment that the retrofit surfaced.

## 1. TopologyBuilder scratch from an injected resource

`tools::cuda::TopologyBuilder` gains a `ResourceT` template parameter, and its device-only scratch moves from the dual `cuda::DeviceBuffer` to the single-space `cuda::Buffer` added in #2251. This is an injectability and ownership change, not a performance one — see below.

An audit of host (`.data()`) vs device (`.deviceData()`) use across the builder's ten buffers:

Counts are call sites of `.data()` (host pointer) and `.deviceData()` (device pointer) on each member:

| member | host uses | device uses | converted |
|---|---|---|---|
| `mUpperMasks`   | 0 | 3 | yes |
| `mLowerMasks`   | 0 | 3 | yes |
| `mUpperOffsets` | 0 | 5 | yes |
| `mLowerOffsets` | 0 | 5 | yes |
| `mLeafOffsets`  | 0 | 4 | yes |
| `mVoxelOffsets` | 0 | 6 | yes |
| `mLowerParents` | 0 | 2 | yes |
| `mLeafParents`  | 0 | 2 | yes |
| `mProcessedRoot`| 1 | 1 | no — dual-space |
| `mData`         | 1 | 1 | no — dual-space |

The eight with **zero** host uses are converted (plus three function-local buffers, eleven `DeviceBuffer::create` calls in total). `mProcessedRoot` and `mData` are genuinely dual-space and are left alone; they are step 3's problem, when `DeviceBuffer` is re-implemented as a composition of two `cuda::Buffer`s.

### This is not a performance change

An earlier revision of this description claimed the conversion removes a synchronous `cudaMallocHost` per buffer. **That was wrong, and I have measured it.** `DeviceBuffer::init` allocates host memory *or* device memory, never both:

```cpp
if (device == cudaCpuDeviceId) {
    cudaCheck(cudaMallocHost((void**)&mCpuData, size));            // host only
} else {
    cudaCheck(util::cuda::mallocAsync(mGpuData+device, size, stream)); // device only
}
```

`TopologyBuilder` always passed a real device id from `cudaGetDevice()`, so every one of these buffers took the `else` branch. There was no host allocation on this path to remove.

Dilate benchmark, 50 iterations per run, three runs each, RTX 6000 Ada / CUDA 12.6:

| points | baseline (ms/dilate) | this PR (ms/dilate) |
|---|---|---|
| 1,000 | 4.304 / 4.113 / 4.101 | 4.322 / 4.091 / 4.095 |
| 20,000 | 4.233 / 4.243 / 4.276 | 4.226 / 4.232 / 4.210 |
| 200,000 | 5.178 / 5.189 / 5.253 | 5.207 / 5.215 / 5.211 |

Within run-to-run noise at every size, as expected once the above is understood.

### What it is worth

- **Injectability** — the eleven allocation sites now route through `ResourceT` instead of a hard-wired `DeviceResource`. That is the actual step 2 goal, and `TopologyBuilder` was the last builder without the seam.
- **Ownership** — scratch is freed by scope rather than by seven hand-written `clear(stream)` calls.
- **Dropping unused dual-space state** — each of these buffers carried a host pointer and a `new void*[deviceCount]` array it never used.

The builder's `TempPool` also becomes `TempPool<ResourceT>`, so cub scratch honors the injected resource too.

**Why `Data` moved.** `TopologyBuilder::Data` is hoisted to namespace scope as `TopologyBuilderData<BuildT>`, with an in-class alias so existing uses read unchanged. It carries no dependence on the resource, and leaving it nested would give every `ResourceT` instantiation its own incompatible `Data` type — which the device functors, templated on `BuildT` alone, could not name.

**Compatibility.** `ResourceT` defaults to `nanovdb::cuda::DeviceResource`, matching how `PointsToGrid` took its resource parameter in #2231, so all six existing consumers — `DilateGrid`, `MergeGrids`, `MeshToGrid`, `PruneGrid`, `RefineGrid`, `CoarsenGrid` — compile unchanged. Three of them reach into `mBuilder.mUpperMasks` / `mLowerMasks` directly and needed `.deviceData()` -> `.data()`.

`DeviceBuffer::clear(stream)` takes a stream where `cuda::Buffer` frees on its retained stream. These are the same stream on every path: the builder is constructed with the operator's stream and every subsequent call uses that same stream, so the translation preserves ordering exactly.

## 2. cuda::Buffer naming alignment

Translating those `clear(stream)` calls surfaced a gap, so this PR also closes it. A member-by-member comparison against `cuda::buffer` is written up on #2232; the actionable parts:

- **`destroy()` / `destroy(stream)`** — `cuda::buffer` names the free operation `destroy` and provides both a retained-stream and an explicit-stream overload. We had only the no-argument form, under a different name, which is why the `clear(stream)` translation above needed hand-verification. `clear()` stays as a **transitional** delegate to `destroy()`, and is marked as such in the header. It exists only because `GridHandle::reset()` still calls `clear()` on its buffer; it goes away in step 3's removal phase, when the legacy dual buffers are deleted and `GridHandle` moves to `destroy()`. `cuda::Buffer` is the type those buffers are being replaced by, so its interface should be the one we want to keep, not an accretion of theirs.
- **`setStream` -> `set_stream`** — member names cannot be aliased, which is the stated reason the resource concept kept `allocate_async` in snake_case against house style; the same argument applies here.
- **`swap`** — the generic `std::swap` already behaves correctly through the move operations (verified: no allocation, no free, pointers and sizes exchanged), but `std::vector` and `rmm::device_buffer` both provide a dedicated one, and the B3 retrofit leans on swapping owners as its central idiom.

Note on `set_stream` semantics: ours deliberately does not synchronize — and that matches `cuda::buffer::set_stream` in both name and contract. NVIDIA/cccl#5697 removed the synchronization deliberately (*"avoid implicit synchronization in fundamental primitives"*) but left the earlier "always synchronizes" `@note` in place; that stale doc is what NVIDIA/cccl#10649 reports. No divergence, no revisit needed.

## Testing

Three new tests in `TestBuffer.cu` cover the added API: `destroy()` frees and is idempotent; `destroy(stream)` frees on the given stream rather than the retained one and retains it afterwards; `swap` exchanges pointers and sizes with no allocation or free. The `TopologyBuilder` conversion itself is behavior-preserving and is already covered by the existing suite through all six consumers.

Verified locally on an RTX 6000 Ada, CUDA 12.6:

- `nanovdb_cuda_buffer_unit_test` — **23/23 pass** (20 existing + 3 new)
- `nanovdb_test_cuda` — **52/53 pass**. The single failure, `UnifiedBuffer_IO`, is a missing test-data file (`data/3_spheres.nvdb`) in an unrelated component; the unmodified baseline fails identically, so it is pre-existing and environmental.

Per #2264 the CUDA tests are excluded from CI (`ctest -E ".*cuda.*|.*mgpu.*"`), so CI will build this but not run it — the numbers above are from a local GPU run.

## Follow-ups

- **B2** — `TempPool`'s own bytes onto `Buffer`, and a `SyncFromAsync` mixin so a custom resource is two methods rather than four.
- **B3** — `PointsToGrid`'s 48 allocation sites, which need an ownership design first (device-visible raw pointers in an uploaded struct, ownership moved by `std::swap`, a `goto` retry loop).
- `DistributedPointsToGrid` stays deferred; #2248 should land before any work there.

### DIFF
--- a/nanovdb/nanovdb/cuda/Buffer.h
+++ b/nanovdb/nanovdb/cuda/Buffer.h
@@ -275,6 +275,10 @@ public:
     }
 
     /// @brief Frees the buffer memory (if any) and resets to the empty state.
+    /// @deprecated Use destroy(). Documentation-level only for now: the
+    ///             [[deprecated]] attribute would fire from GridHandle::reset
+    ///             and NodeManager::reset, which must keep calling clear()
+    ///             until every buffer type provides destroy().
     /// @note Transitional, and not the name to use: it exists only because
     ///       GridHandle::reset still calls clear() on its buffer. It goes away
     ///       when the legacy dual buffers do and GridHandle moves to destroy().
--- a/nanovdb/nanovdb/cuda/Buffer.h
+++ b/nanovdb/nanovdb/cuda/Buffer.h
@@ -50,7 +50,7 @@ struct StreamHolder<true> { cudaStream_t mStream = 0; };
 ///         is_resource). When @c R provides both interfaces the stream-ordered
 ///         one is used.
 /// @details With a stream-ordered resource the Buffer retains the stream of
-///          the most recent allocation (or the one supplied via setStream)
+///          the most recent allocation (or the one supplied via set_stream)
 ///          and orders its deallocation on that stream. Buffer is move-only.
 template<typename T, typename R = DeviceResource>
 class Buffer : private detail::StreamHolder<is_async_resource<R>::value>
@@ -133,7 +133,7 @@ public:
     Buffer& operator=(Buffer&& other) noexcept
     {
         if (this != &other) {
-            this->clear();
+            this->destroy();
             static_cast<detail::StreamHolder<IsAsync>&>(*this) = other;
             mResource = std::move(other.mResource);
             mData = other.mData;
@@ -168,7 +168,7 @@ public:
 
     /// @brief D-tor. A stream-ordered resource frees on the retained stream;
     ///        a synchronous resource frees immediately.
-    ~Buffer() { this->clear(); }
+    ~Buffer() { this->destroy(); }
 
     /// @brief Returns the retained stream, i.e. the stream the buffer's memory
     ///        will be freed on.
@@ -179,9 +179,11 @@ public:
     ///        deallocation (and destruction) is ordered on @c stream instead.
     /// @param stream cuda stream subsequent deallocation is ordered on
     /// @warning The caller is responsible for ordering @c stream after any
-    ///          in-flight work that uses the buffer's memory.
+    ///          in-flight work that uses the buffer's memory. This deliberately
+    ///          does not synchronize, matching cuda::buffer's
+    ///          set_stream_unsynchronized rather than its set_stream.
     template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
-    void setStream(cudaStream_t stream) { this->mStream = stream; }
+    void set_stream(cudaStream_t stream) { this->mStream = stream; }
 
     /// @brief Resizes the buffer to @c count elements, preserving the leading
     ///        min(old, new) elements. Every operation — the new allocation, the
@@ -219,7 +221,7 @@ public:
             mSize = count;
         }
         else {
-            this->mStream = stream; // no reallocation: setStream semantics
+            this->mStream = stream; // no reallocation: set_stream semantics
         }
     }
 
@@ -263,11 +265,44 @@ public:
     bool empty() const { return mSize == 0; }
 
     /// @brief Frees the buffer memory (if any) and resets to the empty state.
-    void clear()
+    ///        A stream-ordered resource frees on the retained stream.
+    /// @note Spelled destroy to match cuda::buffer. This is the name to use.
+    void destroy()
     {
         this->deallocate(mData, mSize);
         mData = nullptr;
         mSize = 0;
+    }
+
+    /// @brief Frees the buffer memory (if any) and resets to the empty state.
+    /// @note Transitional, and not the name to use: it exists only because
+    ///       GridHandle::reset still calls clear() on its buffer. It goes away
+    ///       when the legacy dual buffers do and GridHandle moves to destroy().
+    void clear() { this->destroy(); }
+
+    /// @brief Frees the buffer memory (if any) on @c stream and resets to the
+    ///        empty state. @c stream becomes the retained stream.
+    /// @param stream cuda stream the deallocation is ordered on
+    /// @warning The caller is responsible for ordering @c stream after any
+    ///          in-flight work that uses the buffer's memory.
+    template<typename S = R, std::enable_if_t<is_async_resource<S>::value, int> = 0>
+    void destroy(cudaStream_t stream)
+    {
+        this->mStream = stream;
+        this->destroy();
+    }
+
+    /// @brief Exchanges the contents of this buffer with @c other. Neither
+    ///        buffer allocates, frees, or copies element data.
+    /// @param other buffer to exchange contents with
+    void swap(Buffer& other) noexcept
+    {
+        auto& lhs = static_cast<detail::StreamHolder<IsAsync>&>(*this);
+        auto& rhs = static_cast<detail::StreamHolder<IsAsync>&>(other);
+        std::swap(lhs, rhs);
+        std::swap(mResource, other.mResource);
+        std::swap(mData, other.mData);
+        std::swap(mSize, other.mSize);
     }
 
 private:
--- a/nanovdb/nanovdb/cuda/Buffer.h
+++ b/nanovdb/nanovdb/cuda/Buffer.h
@@ -180,8 +180,8 @@ public:
     /// @param stream cuda stream subsequent deallocation is ordered on
     /// @warning The caller is responsible for ordering @c stream after any
     ///          in-flight work that uses the buffer's memory. This deliberately
-    ///          does not synchronize, matching cuda::buffer's
-    ///          set_stream_unsynchronized rather than its set_stream.
+    ///          does not synchronize, matching cuda::buffer's set_stream, which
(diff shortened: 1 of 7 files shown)


harrism:
@kmuseth ready for review — B1 of the #2232 retrofit: TopologyBuilder allocates scratch through an injected resource, plus the cuda::Buffer member-name alignment with cuda::buffer (destroy/set_stream/swap; details and the CCCL comparison are on #2232). All CUDA suites pass on a local GPU — numbers in the description; per #2264 CI builds but does not run them. #2269 stacks on this, so landing order matters.

harrism:
Ran Copilot review cycles on this PR via a fork mirror (requesting Copilot upstream requires write access). Net changes here: scratch is now released with `destroy(stream)` at all seven sites rather than relying on the retained stream matching, and four `cudaGetDevice` calls left dead by the conversion are removed (c7302bf4). Alignment concerns it raised were incorrect (`Buffer` allocates at `R::DEFAULT_ALIGNMENT`, not `alignof(T)`) but are now also enforced by a `static_assert` in the stacked follow-up. Full threads: https://github.com/harrism/openvdb/pull/1

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -263,13 +265,46 @@ class Buffer : private detail::StreamHolder<is_async_resource<R>::value>
...
         this->deallocate(mData, mSize);
         mData = nullptr;
         mSize = 0;
     }
 
+    /// @brief Frees the buffer memory (if any) and resets to the empty state.
+    /// @note Transitional, and not the name to use: it exists only because
+    ///       GridHandle::reset still calls clear() on its buffer. It goes away
+    ///       when the legacy dual buffers do and GridHandle moves to destroy().
+    void clear() { this->destroy(); }
```
kmuseth, line 285: shouldn't this have a deprecation warning?

nanovdb/nanovdb/unittest/TestBuffer.cu
```diff
@@ -288,7 +344,7 @@ TEST(TestBuffer, SetStreamRedirectsTheFree)
     StreamLog log;
     {
         nanovdb::cuda::Buffer<float, StreamRecordingResource> buf(a, StreamRecordingResource{&log}, 32, nanovdb::cuda::noInit);
-        buf.setStream(b); // member update only, no synchronization
+        buf.set_stream(b); // member update only, no synchronization
```
kmuseth, line 391: This is another one of these camel-case issues. Why the need for abandoning the camel-case which is used elsewhere in OpenVDB? I understand that CUDA itself uses the other case, but I don't understand why this means we need to adopt it as well. Even worse, currently we use both semantics, which is inconsistent and confusing.

kmuseth (changes requested):
added two requests for changes but otherwise it looks good

nanovdb/nanovdb/unittest/TestBuffer.cu
harrism: This is as we discussed before, and documented in #2232: 

> Naming: types are CamelCase per OpenVDB style (`cuda::Buffer`, `cuda::BufferView`) — type names are aliasable, so the opt-in `using Buffer = ::cuda::buffer<T>` at CUDA ≥ 13.2 is preserved. Member names keep the standard spelling (`data`, `size`, `size_bytes`, `allocate_async`) because member matching is structural and cannot be aliased.

This is analogous to using `std::vector` rather than building your own vector class.

If we went with `setStream` then the above `using` alias will not work in the future. My goal, since long before I started working on VDB, is to maximize compatibility and shareability of resource management across GPU libraries in C++. CCCL is pushing this goal forward by standardizing memory resource and storage interfaces. I would like NanoVDB to be as compatible with that as possible. By matching the exact interface of CCCL types, they become drop-in compatible. So in the future (once our minimum CUDA version is high enough), we can eliminate this code (all buffer types) from our library and just rely on what comes with CUDA.

If you disagree with this change in favor of that goal, let me know.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Deprecation is part of step 3 of the plan in #2232, and this PR is part of Step 2. Adding [[deprecated]] to this now would cause warnings on internal calls of this function that users have no way to act on. And if they compile with `-Werror` their TU will fail to compile.

In step 3 we will be removing all internal calls to methods that will be removed, and then deprecate them.

harrism: I added Doxygen @deprecated line to the docs to cover this for now.

harrism:
Fixed the nvcc `#177-D` warnings in 7d58a335 — `TestBuffer.cu` now builds with zero warnings. Root cause: the trait `static_assert`s detect `allocate`/`deallocate` through unevaluated contexts, which never odr-use them, so every concept-required member of the file-local test resources was genuinely unreferenced. `[[maybe_unused]]` turns out not to help — nvcc's front end ignores it for this diagnostic (verified empirically) — so the members are now exercised instead: one test verifies the synchronous halves of the counting/stream-recording doubles behave like their stream-ordered halves, and one pins the trait probes' stub behavior. Real assertions, not just references. The other test TUs build `#177`-clean as well.


Note on provenance: these warnings are not introduced by this PR — they shipped with #2251 and current master produces the same twelve (verified). Building main shows them today; merging this PR clears them there.

kmuseth (approved):
thanks for the changes

PR #2270, opened 2026-08-05
NanoVDB: own PointsToGrid's device arrays with cuda::Buffer (CUDA)

harrism:
Follow-up to #2231/#2251, part of #2232. This is **B3**, the last of step 2's scratch retrofit: `PointsToGrid` was the remaining builder with hand-paired `allocate_async`/`deallocate_async` calls — 47 of them.

> **Stacked on #2269** (which stacks on #2268) — review the last two commits: the loop prep and the conversion.

## Commit 1 — express the density search as a loop

The bisection search over voxel size was a backward `goto`, which made the lifetimes of the buffers it retries over non-lexical. It becomes `while (true)` with `continue` on retry and `break` on convergence. Behavior-preserving; the loop body re-indents, so **`git diff -w` shows the real change: 27 lines**. Two locals (`d_keys`, `d_node_count`) carry results past the loop and hoist above it, as does the copy event — which was created inside the retried region on every iteration but destroyed once at the end, so **each retry leaked an event handle**; it is now created once.

## Commit 2 — ownership

Design rule, refined from the note on the roadmap: **owners own; raw pointers are views, refreshed immediately after every owner event** (assignment, swap, destroy). The device-visible fields inside `mData` — which is `memcpy`'d to the device twice — are therefore never stale, replacing the previous unwritten invariant that nothing reads the device copy of `d_indx` in the ~185 lines between the two uploads. Released views are nulled, so a stale use faults instead of silently reading a freed block.

- **Members vs locals follow the pipeline.** The inventory showed 9 of the 10 largest allocations are created in `countNodes` and freed in a *different* member function (`processUpperNodes`, `processLeafNodes`, `processPoints`, `processBBox`), so their owners are members; contained lifetimes are locals. All owners are `Buffer<T, ResourceRef<ResourceT>>`, borrowing the injected resource per the ownership rule of record (#2232).
- **The index ping-pong becomes a swap of owners** (`mIndxBuf.swap(indxScratch)`) instead of a bare pointer swap between a local and an uploaded struct field.
- **The retry path keeps its free-before-reallocate order** via explicit `destroy()` calls, so peak device memory at retry is unchanged (move-assignment alone would briefly hold old + new).
- Hand-matched byte sizes at every free site disappear; frees that sat one line before scope exit now just leave scope.
- One behavior change, named: the too-many-points-per-leaf `throw` previously leaked its reduction scratch; ownership releases it during unwind.

## Verification

- **Alloc/free parity gate** (the roadmap's acceptance test for this conversion): a counting resource over three shapes — bulk segmented-sort branch (200k points, many tiles), serial per-tile branch (2k points, one tile), and the bisection retry engaged (`PointsToGrid<Point>(4, 0, 8)`, 50k points):

  | case | before (allocs/frees/bytes) | after |
  |---|---|---|
  | bulk | 18 / 18 / 16,499,187 | **identical** |
  | serial | 19 / 19 / 117,720 | **identical** |
  | retry | 38 / 38 / 6,579,539 | **identical** |

- Full CUDA suite **52/53** (the failure is the pre-existing `UnifiedBuffer_IO` missing-data-file issue, identical on baseline); memory-resource suite **9/9**; all instantiations (`Point`+custom, `float`+custom, `Point`+default) compile clean.
- Per #2264, CI builds but does not run these — the numbers above are from a local RTX 6000 Ada, CUDA 12.6.

`DistributedPointsToGrid` (49 sites) stays deferred behind #2248. The remaining small builders (`IndexToGrid` 4, `AddBlindData` 2, `GridStats` 2, `SignedFloodFill` 2, `VoxelBlockManager` 1) and the sync-resource dispatch (step 2's vGPU acceptance criterion) are follow-ups.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
@@ -610,112 +610,123 @@ void PointsToGrid<BuildT, ResourceT>::countNodes(const PtrT points, size_t point
         bool operator<(const Foo &rhs) const {return density < rhs.density || (density == rhs.density && dx < rhs.dx);}
     } min{0.0, 1}, max{0.0, 0};// min: as dx -> 0 density -> 1 point per voxel, max: density is 0 i.e. undefined
 
-jump:// this marks the beginning of the actual algorithm
-
-    mData.d_keys = static_cast<uint64_t*>(mResource->allocate_async(pointCount*sizeof(uint64_t), ResourceT::DEFAULT_ALIGNMENT, mStream));
-    mData.d_indx = static_cast<uint32_t*>(mResource->allocate_async(pointCount*sizeof(uint32_t), ResourceT::DEFAULT_ALIGNMENT, mStream));// uint32_t can index 4.29 billion Coords, corresponding to 48 GB
-    cudaCheck(cudaMemcpyAsync(mDeviceData, &mData, sizeof(PointsToGridData<BuildT>), cudaMemcpyHostToDevice, mStream));// copy mData from CPU -> GPU
-
-    if (mVerbose==2) mTimer.start("\nAllocating arrays for keys and indices");
-    auto *d_keys = static_cast<uint64_t*>(mResource->allocate_async(pointCount*sizeof(uint64_t), ResourceT::DEFAULT_ALIGNMENT, mStream));
-    auto *d_indx = static_cast<uint32_t*>(mResource->allocate_async(pointCount*sizeof(uint32_t), ResourceT::DEFAULT_ALIGNMENT, mStream));
-
-    if (mVerbose==2) mTimer.restart("Generate tile keys");
-    util::cuda::lambdaKernel<<<numBlocks(pointCount), mNumThreads, 0, mStream>>>(pointCount, TileKeyFunctor<BuildT, PtrT>(), mDeviceData, points, d_keys, d_indx);
-    cudaCheckError();
-    if (mVerbose==2) mTimer.restart("DeviceRadixSort of "+std::to_string(pointCount)+" tile keys");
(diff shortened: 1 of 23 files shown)


harrism:
Copilot cycles (fork mirror: https://github.com/harrism/openvdb/pull/3). Net changes: the copy event is owned by a scope guard so the too-many-points-per-leaf throw cannot leak the handle (49575823 — pre-existing leak), `PointsToGrid` asserts its stream-ordered-resource requirement at class scope, and one pre-existing bare `cudaCheckError()` gained its semicolon (2e1e2375).

harrism:
@kmuseth #2269 merged — thank you! Latest master is merged in (no rebase; existing commits unchanged) and the diff now shows only this PR's payload: PointsToGrid's ~47 hand-paired device allocations replaced with RAII `cuda::Buffer` members, the goto retry loop rewritten as a structured loop, and a per-retry event leak fixed along the way. Allocation parity with the old code is exact (18/18 allocs, 19/19 frees, byte-identical totals, including the retry-engaged path). Marking ready for review.

nanovdb/nanovdb/tools/cuda/PointsToGrid.cuh
```diff
@@ -611,154 +635,192 @@ void PointsToGrid<BuildT, ResourceT>::countNodes(const PtrT points, size_t point
...
+    uint32_t *d_indx = nullptr, *d_points_per_tile = nullptr, *d_node_count = nullptr;
+    // Owns the copy event so the too-many-points-per-leaf throw below cannot
+    // leak the handle.
+    struct EventGuard {
+        cudaEvent_t event;
+        EventGuard() { cudaCheck(cudaEventCreate(&event)); }
+        ~EventGuard() { cudaCheck(cudaEventDestroy(event)); }
+        EventGuard(const EventGuard&) = delete;
+        EventGuard& operator=(const EventGuard&) = delete;
+    } eventGuard;
```
kmuseth, line 654: cool

kmuseth (approved):
looks good - a clear progression from 2269 (I'm starting to get the bigger picture here)

PR #2273, opened 2026-08-07
NanoVDB: expose ResourceT on the five TopologyBuilder consumers (CUDA)

swahtz:
Follow-up to #2231/#2251, part of #2232. This is **B5**: the five `tools::cuda::TopologyBuilder` consumers get the same injectable-resource seam that #2268 gave the builder and #2269 gave `MeshToGrid`.

> **Stacked on #2269** (which stacks on #2268) — it branches from that PR, so the diff shown here includes its commits until they merge. Review only the last commit, titled *"NanoVDB: expose ResourceT on the five TopologyBuilder consumers"*.

## What / why

#2268 gave `tools::cuda::TopologyBuilder` a `ResourceT` template parameter so its device scratch + cub `TempPool` allocate through an injected resource. But the five classes that instantiate it — `DilateGrid`, `MergeGrids`, `PruneGrid`, `RefineGrid`, `CoarsenGrid` (all in `nanovdb/tools/cuda/`) — were left as unaffected callers: each held `TopologyBuilder<BuildT> mBuilder;` constructed as `mBuilder(stream)`, binding `default_resource<DeviceResource>()` with no injection path. Downstream (fvdb-core) needs to route these ops' scratch through PyTorch's `c10::cuda::CUDACachingAllocator`; without this seam it cannot delete its forked headers.

## The pattern (established by `MeshToGrid` in #2269)

`tools/cuda/MeshToGrid.cuh` is the reference. Each of the five headers receives the same conversion:

1. The class template gains `typename ResourceT = nanovdb::cuda::DeviceResource`, with an `is_async_resource` guard `static_assert` first in the class body.
2. The builder member becomes `TopologyBuilder<BuildT, ResourceT> mBuilder;`.
3. The constructor(s) gain a trailing defaulted `ResourceT& resource = nanovdb::cuda::default_resource<ResourceT>()`, forwarded as `mBuilder(stream, resource)`. Trailing position keeps every existing call source-compatible; `MergeGrids` gets the argument on both its N-ary and its binary-convenience constructor, and the delegating one forwards it.
4. Every out-of-class member definition moves to the two-parameter form (`template<typename BuildT, typename ResourceT>` / `ClassName<BuildT, ResourceT>::`), including the dual-template `getHandle` and the `} // ClassName<...>::method` comment trailers.

No nested-type hoisting is needed: `TopologyBuilder::Data` was already hoisted to `TopologyBuilderData` in #2268, and none of the five classes declares its own nested struct.

The change is **additive-in-behavior**: `ResourceT` defaults everywhere, so every existing caller compiles unchanged with identical allocation behavior.

## Deliberately untouched — Step 3 residue

Each operator's `mBuilder.mProcessedRoot` and `TopologyBuilder`'s internal `mData` are still `nanovdb::cuda::DeviceBuffer` (host+device dual-space) and are left out of scope here — they are Step 3's dual-space problem. The grid handle's output buffer likewise goes through `BufferT::create`. None of these is observed by the injected resource, which is expected and is called out in the new tests' comments.

## Testing

Adds one injected-resource test per operator to `TestMemoryResource.cu` (`DilateGrid_InjectedResourceSeam`, etc.), mirroring the existing `PointsToGrid` seam tests: build a small `ValueOnIndex` source grid, run the op with a `CountingResource` passed as the new trailing constructor argument, synchronize, and assert `allocs > 0 && allocs == deallocs`. `MergeGrids` merges two disjoint grids; `PruneGrid` supplies a single all-on `Mask<3>` sidecar over a one-leaf grid; `CoarsenGrid` spreads voxels over a 2×2×2 leaf block so coarsening is non-degenerate. A pendingchanges note covers the five new seams.

Verified locally on an RTX PRO 6000 Blackwell, CUDA 13.0:

- `nanovdb_test_cuda_memory_resource` — **14/14** (9 existing + 5 new)
- `nanovdb_test_cuda_buffer` — **27/27**, unchanged
- `nanovdb_test_cuda` — **52/53**; the failure is `UnifiedBuffer_IO`, a missing test-data file (`data/3_spheres.nvdb`) in an unrelated component, which fails identically on an unmodified baseline. The full CUDA suite also exercises all five ops through their defaulted parameter (dilate/merge/prune/refine/coarsen tests), confirming zero impact on default callers.

Per #2264 the CUDA tests are excluded from CI, so CI will build but not run these.


harrism (approved):
Looks good. Thanks for extending Resources to the topologyBuilder consumers.

kmuseth (approved):
looks good to me. Since Mark has also approved I'll do the same and merge

kmuseth:
@swahtz  - there seems to be a new conflict with cuda/Buffer.h

harrism:
This depends on 2268, 2269, 2270 and 2272 being merged first.

swahtz:
> This depends on 2268, 2269, 2270 and 2272 being merged first.

2272 seems to be orthogonal to these changes, so with the others merged recently, this one can merge next.

PR #2275, opened 2026-08-10
Fix race conditions and empty launches in DistributedPointsToGrid

matthewdcong:
1. In the case of a single voxel/tile/key range for N > 1 GPUs, DistributedPointsToGrid will error out due to zero block sizes and prefetch ranges. This PR adds the appropriate guards against empty kernel launches and zero sized prefetch ranges and allow the newly added single GPU test case to succeed.
2. The `leafPrefixSumKernel` also reads from the prior populated range so we need to add an explicit event dependency on the prior device's range in addition to the implicit stream ordering.
3. `processGridTreeRootEvent` was not recorded on the correct stream, but rather implicit synchronized using the default stream. Change this to the correct stream.
4. Calling `cudaMemAdvise` with `cudaMemSetPreferredLocation` and `cudaMemAdviseSetReadMostly` is not supported with memory pool allocations. Thus, we have removed these calls from the externally provided `coords` which would be a memory pool allocation (unlike the others which are member variables allocated via `cudaMallocManaged`).


kmuseth (approved):
looks good (I like the use of cuda::ceil_div :)

PR #2276, opened 2026-08-10
Improve load balancing in DistributedPointsToGrid

matthewdcong:
The existing DistributedPointsToGrid implementation sorts on each local range independently, followed by a pairwise merge tree to obtain a globally sorted range. The existing merge-tree approach halves the number of devices utilized at each step, which leaves the remaining GPUs idle.

This PR improves throughput by involving all the GPUs during the pairwise merge. At a high-level, instead of being limited to one GPU writing the left half of the output (sub)array and another GPU writing to the right half of the output (sub)array, we further subdivide the output (sub)array across all available GPUs so that each GPU writes a section of the output (sub)array. Moreover, this allows us to generalize the pairwise merge to non power-of- two GPU counts.

With 65,536 points sampled on a torus on two RTX 3090s, this improves end-to-end construction time by 8%.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
@@ -55,84 +55,35 @@ private:
     const T* mValue;
 };
 
-/// @brief Computes the trivial merge path split for the case where either
-///        input is empty: every element up to the diagonal comes from the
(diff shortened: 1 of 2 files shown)


kmuseth (approved):
looks okay to me but it's a lot of code changes for merely an 8% performance improvement. Can you try to run this on a more recent GPU thanks a 3090? I wonder if you have two GPUs connected via PCI-e will change the picture?

PR #2196, opened 2026-04-22
Add ScalarCodec, ValueMaskCodec and Read Conversion

danrbailey:
This adds new codecs for handling all integer, vector and valuemask codecs. It also provides options for read conversion (reading a FloatGrid as a HalfGrid for example).

There is also a small change to the TreeIterator to be able to support using it with MaskGrids (https://github.com/AcademySoftwareFoundation/openvdb/commit/669b2ef530bb9d38fc6b85ea8d527b3810bf35b4).

There are new unit tests and all existing unit tests have been updated to use these codecs where applicable.

PR #2281, opened 2026-08-12
Codec Framework and Delayed Loading Removal

danrbailey:
This introduces the new codec framework and completely removes delayed-loading / out-of-core into the master branch.

All of the components included in this PR have been reviewed independently:

https://github.com/AcademySoftwareFoundation/openvdb/pull/2155
https://github.com/AcademySoftwareFoundation/openvdb/pull/2164
https://github.com/AcademySoftwareFoundation/openvdb/pull/2166
https://github.com/AcademySoftwareFoundation/openvdb/pull/2179
https://github.com/AcademySoftwareFoundation/openvdb/pull/2180
https://github.com/AcademySoftwareFoundation/openvdb/pull/2184
https://github.com/AcademySoftwareFoundation/openvdb/pull/2185
https://github.com/AcademySoftwareFoundation/openvdb/pull/2197
https://github.com/AcademySoftwareFoundation/openvdb/pull/2196
https://github.com/AcademySoftwareFoundation/openvdb/pull/2230

These changes have been tested extensively at ILM.

For completeness, here are the new additions to the changelog on this branch from aggregating all the pending changes:

```
OpenVDB:
    Highlights:
    - Completely removed support for out-of-core / delayed loading across OpenVDB
      core, including removal of the OPENVDB_USE_DELAYED_LOADING build option,
      DelayedLoadMetadata, and TempFile. This was done to improve performance,
      reduce codebase complexity and to entirely eliminate the optional Boost
      dependency.
    New features:
    - Introduced new openvdb::codecs subsystem and CodecRegistry for stream
      encoding and decoding.
    - Added support for implicit codec conversion fallback (e.g. float to half,
      scalar to mask) and ReadDiagnostics.
    API changes:
    - Extracted Point Data I/O function overloads from PointDataGrid.h into new
      openvdb/points/PointDataIO.h header.
    - GridDescriptor::read() has been deprecated, use GridDescriptor::readHeader()
      followed by GridDescriptor::readStreamPos() instead.
    - Added optional trailing WriteOptions parameter to io::File::write() and
      io::Stream::write(). No changes in behavior.
    - Added optional trailing ReadOptions parameter to io::File::getGrids() and
      io::File::readGrid(). No changes in behavior.
    - TreeBase::readTopology() and TreeBase::writeTopology() are now pure virtual.
      Derived classes of TreeBase now need to read and write int32_t(1) to remain
      backwards-compatible.
    Improvements:
    - Large refactor of I/O classes, many private and protected member functions have
      been modified. No change in behavior for io::File and io::Stream.
    - Remove direct testing of Tree I/O methods in favor of indirect testing using
      the Grid I/O.
    Houdini:
    - Add new Vulkan viewport support for VDB Points primitives and remove support
      for delayed loading.
```


openvdb/openvdb/unittest/TestCodec.cc
```diff
@@ -0,0 +1,514 @@
...
+        EXPECT_EQ(*it, Half(1.5f));
+    }
+
+    // Background value must equal Half(0.0f)
+    EXPECT_EQ(halfGrid->background(), Half(0.0f));
+
+    // Cleanup
+    std::remove(floatPath.c_str());
+}
+
```
apradhana, line 350: Codex found an edge case of failing to read float as half when a file doesn't have a grid offsets
```
TEST_F(TestCodec, testReadModeHalfAppliesToFileWithoutGridOffsets)
{
    using namespace openvdb;
    using namespace openvdb::io;

    CodecRegistry::clear();
    io::internal::initialize();

    FloatGrid::Ptr srcGrid = FloatGrid::create(3.25f);
    srcGrid->setName("float_to_half");
    srcGrid->tree().setValue(Coord(0, 0, 0), 1.0f / 3.0f);

    const std::string path = "test_float_to_half_no_offsets.vdb";

    {
        std::ofstream os(path, std::ios_base::out | std::ios_base::binary);
        io::Stream(os).write(GridPtrVec{srcGrid});
    }

    ReadOptions readOptions;
    readOptions.readMode = ReadMode::Half;

    {
        io::File f(path);
        f.open();
        GridPtrVecPtr grids = f.getGrids(readOptions);
        ASSERT_TRUE(grids);
        ASSERT_EQ(size_t(1), grids->size());
        EXPECT_TRUE((*grids)[0]->isType<HalfGrid>());
        f.close();
    }

    {
        io::File f(path);
        f.open();
        GridBase::Ptr grid = f.readGrid(srcGrid->getName(), readOptions);
        ASSERT_TRUE(grid);
        EXPECT_TRUE(grid->isType<HalfGrid>());
        f.close();
    }

    std::remove(path.c_str());
}
```

The problem is `io::Stream(os).write(...)` creates a VDB file without grid offsets. For that kind of file, `File::open()` reads the grids using default options before the caller passes `ReadOptions`. So later calls to `getGrids(readOptions)` or `readGrid(name, readOptions)` return the cached original `FloatGrid`, not a converted `HalfGrid`.

danrbailey: Hi Andre, thanks for testing. This is true and I think it's possibly also an issue for instance VDBs too. We currently only offer on-the-fly conversion, as in the data is converted between types _while_ it is being read. The caching approach used for non grid offset VDBs doesn't align with that.

I don't think it's a good idea to add the `ReadOptions` to the `File::open()` method, but there are a few of other options to discuss:
* We just add a ReadDiagnostics warning that documents that on-the-fly grid conversion is not supported for non grid offset VDBs. I don't think it deserves to be an exception, but maybe we should consider that too.
* We do in-memory grid-to-grid conversion - have you explored that for float->half grids and is it a feature that is currently supported?
* We remove on-the-fly grid conversion altogether and support it properly for the new file format only. I don't intend to support writing non grid offset VDBs in the new file layouts (except for writing legacy files).

I do agree that quietly ignoring the hint and just returning the float grid is unexpected and should be resolved, so I'll have a think about which of these options might be most viable.

danrbailey: I opted to do the in-memory grid conversion to resolve this issue. See https://github.com/AcademySoftwareFoundation/openvdb/pull/2298

openvdb/openvdb/io/Archive.cc
```diff
@@ -209,9 +179,10 @@ struct StreamMetadata::Impl
     uint32_t mPass = 0;
     MetaMap mGridMetadata;
     AuxDataMap mAuxData;
-    bool mDelayedLoadMeta = DelayedLoadMetadata::isRegisteredType();
+    bool mDelayedLoadMeta = false;
     uint64_t mLeaf = 0;
     uint32_t mTest = 0; // for testing only
+    bool mAllocateLeafBuffers = false;
```
apradhana, line 185: This may be blocking because of ABI incompatibilities between 13.0 and 13. 1.

`StreamMetadata` is shared between coexisting OpenVDB versions and is part of the protected Grid ABI. A 13.1 reader can access or copy this field from an `Impl` created by 13.0, where this bool does not exist or was never initialized. Tail padding does not make that access safe. Could this flag use existing compatible storage such as `AuxDataMap`, or be restricted to ABI 14?

For example, `setAllocateLeafBuffers(true)` could store true under a private key in `auxData()`. `allocateLeafBuffers()` could look up that key with `std::any_cast<bool>`, and clearing the flag could erase it. This uses storage that already exists and is initialized in ABI 13, so no new `Impl` member is needed.

apradhana:
This is mainly a comment regarding the removal of `PImpl` and not a blocking issue. I just want us all to be aware that removing `PImpl` reduces ABI flexibility. With PImpl, private implementation members can be changed without changing the public layout of `File` or `Stream`. For example, a new version could add more fields to the hidden implementation while the public object remains just a pointer.

Without PImpl, those private members become part of the public object’s binary layout. Future internal changes may therefore require applications and plugins to be rebuilt. Mixing versions also becomes unsafe: old headers might allocate 72 bytes for File, while the new library expects 296 bytes, potentially causing crashes or memory corruption.

Again, this is non-blocking.

openvdb/openvdb/io/File.h
```diff
@@ -54,20 +54,16 @@ class OPENVDB_API File: public Archive
...
     /// @return @c true if the file's UUID has changed since it was last read.
-    /// @see setCopyMaxBytes
-    bool open(bool delayLoad = true, const MappedFile::Notifier& = MappedFile::Notifier());
-#else
-    bool open(bool /*delayLoad*/ = false);
-#endif
+    bool open();
+
+    OPENVDB_DEPRECATED_MESSAGE("Use File::open() instead.This method is deprecated and will be removed. Delayed loading is no longer supported.")
+    bool open(bool /*delayLoad*/) { return open(); }
```
apradhana, line 64: Should an explicit `open(true)` fail or emit a runtime diagnostic instead of silently falling back to non-delayed loading? The deprecation warning helps recompiled callers, but it can be disabled or missed, and the notifier overload also silently discards the notifier. Maybe using `OPENVDB_THROW` when `delayLoad` is `true` with a message saying `"Delayed Loading is no longer supported!"` can be a solution.

openvdb/openvdb/io/Stream.h
```diff
@@ -22,21 +22,30 @@ class OPENVDB_API Stream: public Archive
...
+    /// @param is           The input stream to read from
+    /// @param readOptions  Options controlling how grids are read (e.g. attribute
+    ///                     skipping for point data grids)
+    /// @throw ValueError if @a readOptions requests @c io::ReadMode::MetadataOnly
+    ///     or @c io::ReadMode::TopologyOnly. These modes leave a grid's data
+    ///     partially unread, which a sequentially read stream cannot skip over.
+    Stream(std::istream& is, const io::ReadOptions& readOptions);
+
+    OPENVDB_DEPRECATED_MESSAGE("Use Stream(std::istream&) instead. This method is deprecated and will be removed. Delayed loading is no longer supported.")
+    Stream(std::istream& is, bool /*delayLoad*/) : Stream(is) { }
```
apradhana, line 38: As with `File::open(bool)`, should this throw when `delayLoad` is true instead of silently loading all grid data into memory? For large streams, ignoring an explicit delayed-loading request could cause unexpectedly high memory use. The false case can continue to delegate to `Stream(is)`.

danrbailey:
> This is mainly a comment regarding the removal of `PImpl` and not a blocking issue. I just want us all to be aware that removing `PImpl` reduces ABI flexibility. With PImpl, private implementation members can be changed without changing the public layout of `File` or `Stream`. For example, a new version could add more fields to the hidden implementation while the public object remains just a pointer.
> 
> Without PImpl, those private members become part of the public object’s binary layout. Future internal changes may therefore require applications and plugins to be rebuilt. Mixing versions also becomes unsafe: old headers might allocate 72 bytes for File, while the new library expects 296 bytes, potentially causing crashes or memory corruption.
> 
> Again, this is non-blocking.

This is sort of a general comment about why pimpl might be useful. If we remove delayed loading, then the StreamMetadata is no longer part of the Grid ABI because you cannot pass around a partially created grid. Using the codec framework improves flexibility and does away with all the complexity of managing and maintaining StreamMetadata objects. That is the overall intention here.

openvdb/openvdb/io/File.h
danrbailey: My vote would be not to throw an exception here. Delayed loading has always been a kind of memory optimization technique. Also, based on all the conversations we have had, I believe it is very rarely used (if ever) in production. Breaking the API like this seems unnecessary given that the overall functionality itself remains unchanged.

openvdb/openvdb/io/Archive.cc
apradhana: From TSC meeting 8/26/2026 this is okay, because ABI changes is applied only for `Tree` and `Grid` contracts say that we want to be able to `reinterpret_cast` the pointer. In the discussion, it was mentioned that it's for the old grid reading workflow. It is used to be read during a topology-only read. So it seems okay.

For completeness, I'll also include a discussion all the way back from 2019: https://github.com/AcademySoftwareFoundation/openvdb/blob/master/tsc/meetings/2019-09-26.md?plain=1#L27-L45. So there may be a compatibility issue with Houdini.

openvdb/openvdb/io/File.h
apradhana: From TSC meeting 8/26/2026, we agree to keep the API compiling with compiler warning and make a lot of documentation to explain the behavior since we deprecate delayed loading.

openvdb/openvdb/io/Stream.h
apradhana: From TSC meeting 8/26/2026, we agree to keep the API compiling with compiler warning and make a lot of documentation to explain the behavior since we deprecate delayed loading.

apradhana:
> > This is mainly a comment regarding the removal of `PImpl` and not a blocking issue. I just want us all to be aware that removing `PImpl` reduces ABI flexibility. With PImpl, private implementation members can be changed without changing the public layout of `File` or `Stream`. For example, a new version could add more fields to the hidden implementation while the public object remains just a pointer.
> > Without PImpl, those private members become part of the public object’s binary layout. Future internal changes may therefore require applications and plugins to be rebuilt. Mixing versions also becomes unsafe: old headers might allocate 72 bytes for File, while the new library expects 296 bytes, potentially causing crashes or memory corruption.
> > Again, this is non-blocking.
> 
> This is sort of a general comment about why pimpl might be useful. If we remove delayed loading, then the StreamMetadata is no longer part of the Grid ABI because you cannot pass around a partially created grid. Using the codec framework improves flexibility and does away with all the complexity of managing and maintaining StreamMetadata objects. That is the overall intention here.

Thanks for clarifying this and for the discussion during TSC.

openvdb/openvdb/unittest/TestCodec.cc
apradhana: Thanks for addressing this issue this way.

openvdb/openvdb/codecs/TopologyCodec.h
```diff
@@ -0,0 +1,399 @@
...
+            diagnostics.addWarning(gridName, "multi-buffer trees are no longer supported");
+        }
+
+        // Delete the existing tree.
+        root.clear();
+
+        // Read a RootNode that was stored in the current format.
+
+        is.read(reinterpret_cast<char*>(&storageBackground), sizeof(StorageValueT));
+        background = static_cast<ValueT>(storageBackground);
```
apradhana, line 143: Codecs pointed out: For numeric-to-Mask conversion (`ReadMode::Mask`), the result should use `activity` rather than numeric values. This cast, and the tile-value casts below, can make the `MaskGrid` disagree with the source topology. A nonzero background or inactive `0.1f` tile becomes `true`, while an active zero tile can become `false`. Could this be special-cased for the Mask conversion so `inactive` entries and the background are `false`, and `active` entries are `true`?

danrbailey: This is a good suggestion, I like this. In fact, I think this should ultimately become a tool that lives in `tools/Mask.h` and turns a normal numeric grid into a mask grid that works with activity and ignores values. In lieu of doing that right now, I've implemented your suggestion here:

https://github.com/AcademySoftwareFoundation/openvdb/pull/2298/changes/ab58a58dd142adfb8f830e3b90adf3b298b7ffa0

I know you have already approved that PR, but let me know if you have any concerns about that, otherwise I'll merge that into this `feature/io` branch along with the other changes on that PR.

apradhana: Yes, please! Thanks so much, @danrbailey . 

Will approve this PR once [2298](https://github.com/AcademySoftwareFoundation/openvdb/pull/2298) is in.

apradhana (approved):
Thanks for adding the no-offset conversion and the activity-based Mask fixes. The `TopologyOnly` contract also reflects the agreed behavior. Approved!

PR #2283, opened 2026-08-18
NanoVDB: assign each upper-node tile to a single GPU in DistributedPointsToGrid

swahtz:
The device segment boundaries produced by the initial even split were rebalanced by adjusting only adjacent pairs of boundaries. That cannot consolidate an upper-node tile spanning three or more GPUs, since a fully-interior device lies entirely within the tile. Downstream construction assumes each tile - and therefore each lower node, leaf node and voxel - is owned by exactly one device, so multiple devices built the same leaf concurrently and raced on its value mask, silently dropping active voxels.

Replace the pairwise GPU rebalance with a global, monotonic snap performed on the host over the (small) set of device boundaries. The keys are globally sorted at that point, so a tile boundary is simply a position where the key changes; fully-interior devices are left empty, which the rest of the pipeline already handles.

Add two multi-GPU regression tests: SingleUpperNode, which places all input in one upper-node tile and asserts the exact unique active-voxel count, and MatchesSingleGpu, which cross-checks topology and occupancy against the single-GPU builder on the same input.

### DIFF
--- a/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
+++ b/nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
@@ -40,21 +40,6 @@ private:
     uint32_t* mNodeCounts;
 };
 
-/// @brief Indicator functor that returns 1 if the input value matches the member value, 0 otherwise
-template <typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
-struct EqualityIndicator
-{
-    EqualityIndicator(const T* value) : mValue(value) {}
-
-    __hostdev__
-    T operator()(const T& x) const
-    {
-        return x == (*mValue);
-    }
-private:
-    const T* mValue;
-};
-
 /// @brief Find a partition at an arbitrary diagonal in the conceptual merge of two sorted input arrays.
 template<typename KeyIteratorIn>
 __device__
@@ -84,34 +69,6 @@ __global__
 void mergePathKernel(KeyIteratorIn keys1, size_t keys1Count, KeyIteratorIn keys2, size_t keys2Count, ptrdiff_t* key1Intervals, ptrdiff_t* key2Intervals, size_t combinedIndex)
 {
     mergePath(keys1, keys1Count, keys2, keys2Count, key1Intervals, key2Intervals, combinedIndex);
-}
-
-/// @brief Extends or shortens the left end of an array interval
-template<typename DistanceIteratorIn, typename CountIteratorOut, typename OffsetIteratorOut>
-__global__
-void leftRebalanceKernel(DistanceIteratorIn leftDistance, DistanceIteratorIn rightDistance, CountIteratorOut leftCount, OffsetIteratorOut leftOffset)
-{
-    if (*leftDistance < *rightDistance) {
-        *leftCount -= *leftDistance;
-    }
-    else {
-        *leftCount += *rightDistance;
-    }
-}
-
-/// @brief Extends or shortens the right end of an array interval
-template<typename DistanceIteratorIn, typename CountIteratorOut, typename OffsetIteratorOut>
-__global__
-void rightRebalanceKernel(DistanceIteratorIn leftDistance, DistanceIteratorIn rightDistance, CountIteratorOut rightCount, OffsetIteratorOut rightOffset)
-{
-    if (*leftDistance < *rightDistance) {
-        *rightCount += *leftDistance;
-        *rightOffset -= *leftDistance;
-    }
-    else {
-        *rightCount -= *rightDistance;
-        *rightOffset += *rightDistance;
-    }
 }
 
 } // namespace kernels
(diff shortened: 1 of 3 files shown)


nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
```diff
@@ -650,60 +601,68 @@ void DistributedPointsToGrid<BuildT>::countNodes(const PtrT coords, size_t coord
...
-
-        if (deviceStripeCounts[deviceId] && deviceId < static_cast<int>(mDeviceMesh.deviceCount() - 1) && deviceStripeCounts[deviceId + 1]) {
-            cudaCheck(cudaStreamWaitEvent(stream, sortEvents[deviceId + 1]));
-            EqualityIndicator<uint64_t> indicator(deviceInputKeys + deviceStripeCount);
-            CUB_LAUNCH(DeviceReduce::TransformReduce, mTempDevicePools[deviceId], stream, deviceInputKeys, leftIntervals + deviceId, deviceStripeCount, ::cuda::std::plus(), indicator, 0);
-        } else {
-            leftIntervals[deviceId] = 0;
-        }
-        cudaCheck(cudaEventRecord(transformReduceEvents[deviceId], stream));
+        cudaCheck(cudaStreamSynchronize(stream));
```
matthewdcong: Is it possible to do this on the device instead of the host in order to avoid the stream sync? Alternatively, would it be possible to run a benchmark on an analytic example (e.g. a sampled torus) to show that performance isn't affected?

swahtz: Sure, I added a `snapBoundariesToRunsKernel` so we run the snapping computation on-device.

kmuseth (approved):
This looks good, but I'm worried that Claude detected a bug (unrelated to this PR) that prevented it from running and verifying this PR. I'll approve this PR but someone needs to do a deep dive and figure out what the blocking issue (in master). Let me know if you will do this our I should (or Matthew)

There's the summary (by Claude) on which I base my comments above:

The bug: the old code rebalanced device segment boundaries pairwise (only adjacent devices adjusted their shared boundary to avoid splitting an upper-node tile). That can't fix a tile spanning 3+ devices, since a fully-interior device's entire stripe sits inside the tile with no adjacent-boundary adjustment able to reach it. The result: multiple devices independently build the same leaf and race on its value mask, silently losing active voxels — a real correctness bug, not just a performance one.

The fix: replace the pairwise rebalance with snapBoundariesToRunsKernel, a single-thread kernel that walks all device boundaries left-to-right over the now-globally-sorted key array and snaps each one to the nearest edge of the "run" (tile) it falls inside, growing a fully-interior device's boundary through the whole straddling run when needed.

Verification performed:

Hand-traced the kernel's boundary-snapping algorithm step by step, including a worked example of a run spanning 3 devices (confirmed it correctly consolidates the whole run onto the device where it started, leaving the interior devices empty via the monotonic clamp).
Checked all the edge cases: boundary >= keyCount, boundary == previousBoundary (already-empty device), the runStart <= previousBoundary "give the whole tile away" branch, and the tie-breaking heuristic when a run is short enough to fit in one device's original split — none produce an out-of-bounds keys[] access or a non-monotonic boundary.
Verified the underlying data-visibility assumption the fix depends on: mKeys/mStripeOffsets/mStripeCounts are all cudaMallocManaged, and traced radixSortAsync's (unmodified) closing mergeStreams call to confirm it fans in every device's completion and fans it back out to every device stream (including device 0, where the new snap kernel runs) before returning — so the snap kernel is guaranteed to see the fully, globally sorted array without an extra explicit cross-device wait. This is the load-bearing correctness assumption in the PR's own code comment, and it checks out.
Confirmed the host-side cudaEventSynchronize after the snap kernel is what makes the corrected mStripeOffsets/mStripeCounts values safe to read on the host before they're baked into subsequent per-device kernel launch arguments — no further GPU-side dependency on the snap kernel is needed downstream, since the RLE kernels only touch mKeys (already synchronized by the sort) using host-supplied scalar bounds.
Reviewed the two new regression tests (SingleUpperNode, which forces all input into one tile and asserts an exact active-voxel count; MatchesSingleGpu, which cross-checks topology against the trusted single-GPU builder) — well-targeted at exactly this bug class.
Attempted live execution on this machine's 2 real GPUs (with working P2P per nvidia-smi topo). Compiled and linked TestMultiGPU.cu cleanly against both the PR and master, but hit an environment-level failure (RadixSort test fails, then a "misaligned address" crash in DistributedPointsToGrid.cuh) that reproduces identically on master, at the same line, before this PR's changes are even reached — this rules out the PR as the cause and points at a CUDA-toolkit/CCCL-version mismatch in this sandbox (nvcc 13.3 vs driver's CUDA 13.2) affecting radixSortAsync's cub::DeviceMerge::MergePairs step, unrelated to this PR. I was not able to get a clean live run of the new regression tests as a result, so my confidence rests on the manual trace above rather than an executed pass.

Findings: none that block this PR. The fix is logically sound, correctly reasons about the synchronization contract it depends on, and is well-tested in principle (even though I couldn't execute those tests live here due to an unrelated environment limitation). I'd approve this as-is, though I'd flag to the author/reviewers that I could not independently confirm the new tests pass on real hardware in this session — if CI can run it on a working multi-GPU environment, that's the stronger signal to rely on.

PR #2284, opened 2026-08-18
NanoVDB: undef CALL_CUBS at the end of MeshToGrid.cuh

sifakis:
## Summary

`MeshToGrid.cuh` defines the `CALL_CUBS` utility macro (used to call CUB device functions that
need dynamic temporary storage) but never undefs it, unlike its sibling CUDA tools. Both
`TopologyBuilder.cuh` and `PointsToGrid.cuh` undef `CALL_CUBS` right after their last use, so the
macro doesn't leak into whatever gets included next in the same translation unit. This adds the
same `#ifdef CALL_CUBS` / `#undef CALL_CUBS` block to `MeshToGrid.cuh`, in the same place and style
as those two files.


kmuseth (approved):

swahtz (approved):
Good catch, thanks

PR #2286, opened 2026-08-19
NanoVDB: own the small builders' device scratch with cuda::Buffer (CUDA)

harrism:
Part of #2232 (step 2). Completes the raw-allocation audit that PointsToGrid (#2270), TopologyBuilder/MeshToGrid (#2268/#2269) and the topology-op consumers (#2273) already went through: the last ten `mallocAsync`/`freeAsync` pairs in the GPU tools become RAII `cuda::Buffer` scratch over an injectable resource.

## What
- **IndexToGrid** (4 sites: the device NodeAccessor and the grid-name upload), **addBlindData** (2: byte-size scratch), **GridStats** (2: per-node statistics scratch), **SignedFloodFill** (2: node counts).
- Each gains a defaulted `ResourceT` template parameter following the established conventions: classes take a must-outlive resource instance in the constructor (as DilateGrid et al.), free functions take the template parameter and route through the per-type default instance (as `pointsToGrid`). Existing code is unaffected — every call site in the tree was traced and remains source-compatible.
- `IndexToGrid`'s nested `NodeAccessor` is hoisted to a namespace-scope `IndexToGridNodeAccessor<SrcBuildT>` so the kernels that name it do not acquire a spurious `ResourceT` dependency; a public alias preserves the nested spelling.
- Behavior parity: every replaced free keeps its position in stream order (`destroy(stream)` at the old `freeAsync` site), and GridStats' mode-dependent allocation becomes a zero-element buffer when the mode has no average — the resource sees identical traffic, asserted exactly by the tests.

## Tests
New seam tests in `TestMemoryResource.cu` with **exact** allocation counts per builder (IndexToGrid: 2, including the grid-name buffer; GridStats: 1, and 0 for the Extrema mode; SignedFloodFill: 1; addBlindData: 1 through the default-instance path), plus free-function `ResourceT` forwarding coverage.

## Two observations for the reviewer
1. **Pre-existing latent bug, preserved for parity**: `IndexToGrid::setGridName` uploads the name to the device, but no kernel reads `d_gridName` — the name never reaches the output grid. This PR keeps that behavior (the upload now goes through the resource); fixing or removing it should be its own change.
2. **Documented allocation boundary**: `GridStats::update`'s temporary NodeManager still allocates through the dual-space `DeviceBuffer` (`createNodeManager` is `hasDeviceDual`-gated), so that allocation bypasses an injected resource. Now called out in an `@note` on `update()`; the real fix belongs to the single-space handle work on #2232.

## Verification
- Zero warnings under the (now default) `--Werror=all-warnings`; full build of tests, tools and examples at `CMAKE_CUDA_ARCHITECTURES=80`.
- 7/7 ctests on a GPU runner (2× RTX 6000 Ada, CUDA 12.x).
- Reviewed by Codex (`gpt-5.6-terra`): one finding (the NodeManager boundary above), resolved; second pass clean with the reviewer independently building and running the test binary.


nanovdb/nanovdb/tools/cuda/IndexToGrid.cuh
```diff
@@ -99,8 +114,9 @@ private:
 //================================================================================================
 
 template<typename SrcBuildT>
-struct IndexToGrid<SrcBuildT>::NodeAccessor
+struct IndexToGridNodeAccessor
```
kmuseth, line 117: why is this no longer a member class of NodeAccesspr?

kmuseth (approved):
Seems like minor changes to adopt the new API for remaining tools. I am however confused as to why one member struct was changed to a namespaced struct

nanovdb/nanovdb/tools/cuda/IndexToGrid.cuh
harrism: Nested types inherit all of the enclosing class's template parameters, so keeping it nested would make the kernels' (below) argument type vary with `ResourceT` and that would cause every kernel to be re-instantiated for every resource type. This would impact compile time and binary size. 

Hoisting the struct outside the class keeps its type keyed on `SrcBuildT` only, with a member alias preserving the old spelling inside the class. 

This is the same as TopologyBuilderData in #2268.

PR #2288, opened 2026-08-19
NanoVDB: support single-space device buffers in GridHandle (CUDA)

harrism:
## What this PR is, and why

Grids on the GPU currently require a *dual-space* buffer: a host allocation mirrored by a device allocation, even in pipelines where the host copy is never read — doubling memory for device-resident workflows and pinning allocation to NanoVDB's built-in allocator instead of the application's pool. This PR (step 3 of #2232, part 1 of 3) adds a **device-only alternative**: `GridHandle<cuda::Buffer<std::byte, R>>`, a grid owned by a single device allocation made through any injected memory resource (an RMM-style pool, or downstream wrappers like fvdb's Torch caching allocator — the consumers the #2232 seam work exists for). Host mirror: gone. Host-facing APIs on such a handle: compile-time errors by design.

**Dual-space handles are untouched by this PR** — today they're still the only way to move grids between host and device. That changes in part 2, which will add explicit cross-space transfers (per-direction `copy()` between host-space and device-space handles, currently a named compile error). Once those cover the dual buffers' use cases, the dual buffers themselves will be deprecated and then removed: the end state is one buffer template class, with on type per memory space and explicit transfers between them.

## What
- **A new (detected) `hasDeviceSingle` buffer trait** selects a constructor in `cuda/GridHandle.cuh` that parses grid metadata through the device — validated header walk, metadata scratch through the buffer's own resource, kernel, and readback all ordered on the buffer's retained stream.
- **Device accessors work; host paths are compile-time errors.** `deviceData()`/`deviceGrid<T>()` are available; `data()`, `grid<T>()`, `gridData()`, `gridMetaData()` are SFINAE-removed; the `read`/`write` I/O members and `splitGrids`/`mergeGrids` stay addressable (the python bindings take `write`'s address via `overload_cast`) and instead fail with explanatory `static_assert`s when instantiated for a single-space handle. The python module (`PyGridHandle.h` bindings included) is compiled as part of the local gate.
- **`copy()` dispatches on the traits**: host buffers keep the memcpy path; a single-space handle deep-copies device-to-device through its own resource on the retained stream and reuses the host-resident metadata instead of re-parsing. `cuda::Buffer` gains `ElementType`/`ResourceType`, `resource()`, and a no-arg `copy()` for stream-ordered resources (= `copy(stream())`).
- **Hardening shared with the legacy path**: the device parse now validates the entire grid chain (per-header bounds checks) before launching the metadata kernel, so truncated buffers and forged `mGridCount`/`mGridSize` headers throw instead of reading out of bounds — the pre-existing dual-space parse had the same hole and now uses the same helper. Its scratch runs on `MallocResource`, so the long-standing `GridHandle<DeviceBuffer>` device parse keeps working on devices without memory-pool support. Six raw allocation sites in `cuda/GridHandle.cuh` (one an *unchecked* `cudaMalloc`) are replaced with resource-aware buffers.
- **Host accessibility is a resource property**: the new `is_host_accessible_resource` trait detects a `HOST_ACCESSIBLE` marker (on `PinnedResource`, forwarded through `ResourceRef`/`AsyncFromSync`). A pinned-resource `cuda::Buffer` handle is a *named* compile error for now — GridHandle's host paths need an allocation interface `cuda::Buffer` doesn't yet provide (the `create` mapping, scheduled for the step-3 completion PR) — rather than a half-working host path.
- **Compatibility**: both new trait members are detected, not required — `BufferHasDeviceSingle`/`BufferHasHostSingle` default to false when a `BufferTraits` specialization omits them, so pre-existing specializations in tree (`PoolBuffer` examples, `DeviceBuffer`, `UnifiedBuffer`) and out of tree compile unchanged. Every existing call site of the re-gated members was traced; none changes overload resolution.

## Tests
Seven new tests + compile-time classification asserts in `TestBuffer.cu`: device meta parse and typed/wrong-type `deviceGrid`; exact allocation accounting through a counting resource (grid bytes + meta scratch, freed via `reset()`); deep D2D copy with byte comparison; empty-handle and empty-copy; invalid-grid throw with leak check; synchronous-resource (`MallocResource`) construction and copy; multi-grid parse with a stream-recording resource proving every allocation lands on the retained (non-blocking) stream; forged-`mGridCount` rejection with an unpoisoned CUDA context.

## Verification
- Zero warnings under the default `--Werror=all-warnings`; full build (tests, tools, examples) at `CMAKE_CUDA_ARCHITECTURES=80`; 7/7 ctests on a GPU runner, plus `g++ -fsyntax-only` on `GridHandle.h` proving the header stays CUDA-free.
- Reviewed by an agent audit (consumer trace over every call site, two-phase-lookup check, stream/exception-path analysis) and four Codex review rounds — findings included the pinned-resource misclassification, the chain-validation P1, and a trait-query regression, all addressed; final round clean with the reviewer independently building and running the tests.


nanovdb/nanovdb/cuda/GridHandle.cuh
```diff
@@ -43,6 +44,54 @@ static __global__ void updateGridCount(GridData *d_data, uint32_t gridIndex, uin
     }
 }
 
+/// @brief Walks the grid chain with per-header device-to-host reads, checking
+///        that every header and every grid span lies inside the allocation, so
+///        the device-side metadata walk that follows can never read out of
+///        bounds from a truncated buffer or a forged header.
+/// @return the validated grid count
+inline uint32_t validGridChainCount(const GridData *d_head, uint64_t bytes, cudaStream_t stream)
```
swahtz: **🚩 issue:** I traced the consumers of the device-parse path, and I believe this new validation breaks `tools::cuda::addBlindData` for grids that came out of a multi-grid handle. `addBlindData` copies the source grid's `GridData` header verbatim (including `mGridIndex`/`mGridCount`) into its new single-grid buffer and never normalizes them before constructing `GridHandle<BufferT>` (AddBlindData.cuh:131, a device-only buffer, so it takes this path). With a source grid from `handle.deviceGrid<T>(n)` of a multi-grid handle: for `n > 0` the first iteration throws "inconsistent grid index/count" (`mGridIndex != 0`), and for `n == 0` with `mGridCount > 1` the walk runs past the single-grid buffer and throws "grid chain exceeds the device buffer". The old parse accepted these buffers (albeit with OOB metadata reads — which is exactly the hole this validation fixes).

I think the fix belongs in `addBlindData` rather than in weakening the validation: normalize the header to index 0 / count 1 before constructing the handle, the way `splitGridHandles` does with `detail::updateGridCount<<<1,1>>>`. Might also be worth a quick audit for any other tool that constructs a device-parse handle from a copied header.

nanovdb/nanovdb/cuda/GridHandle.cuh
```diff
@@ -122,28 +171,52 @@ mergeDeviceGrids(const VectorT<GridHandle<BufferT>> &handles, const BufferT* oth
...
-            GridData tmp;
-            cudaCheck(cudaMemcpy(&tmp, d_data, sizeof(GridData), cudaMemcpyDeviceToHost));
-            if (!tmp.isValid()) throw std::runtime_error("GridHandle was constructed with an invalid device buffer");
-            GridHandleMetaData *d_metaData;
-            cudaMalloc((void**)&d_metaData, tmp.mGridCount*sizeof(GridHandleMetaData));
-            cuda::detail::cpyGridHandleMeta<<<1,1>>>(d_data, d_metaData);
-            mMetaData.resize(tmp.mGridCount);
-            cudaCheck(cudaMemcpy(mMetaData.data(), d_metaData,tmp.mGridCount*sizeof(GridHandleMetaData), cudaMemcpyDeviceToHost));
-            cudaCheck(cudaFree(d_metaData));
+            const uint32_t count = cuda::detail::validGridChainCount(d_data, mBuffer.size(), cudaStream_t(0));
```
swahtz: **🚩 issue:** This long-standing constructor previously did a single header read; it now does one `cudaMemcpyAsync` + full `cudaStreamSynchronize` per grid header inside `validGridChainCount`. `cuda::mergeGridHandles` ends by constructing exactly this handle, so merging hundreds of grids now issues hundreds of serialized 672-byte D2H copies, each draining the stream, where the old code did one memcpy plus one kernel.

Two shapes that would fix it, either of which I'd be happy with: (a) since the validation walk already copies every full `GridData` header to the host, accumulate `mMetaData` directly in the walk — that deletes the scratch allocation, the `cpyGridHandleMeta<<<1,1>>>` launch, the second D2H copy, and the final sync in both constructors, so the validation *replaces* the old parse machinery instead of adding to it (the round trips stay O(N), but the whole second phase disappears); or (b) move the validation into a single device-side kernel with one status readback, restoring O(1) round trips for the multi-grid case.

nanovdb/nanovdb/GridHandle.h
```diff
@@ -169,6 +235,22 @@ class GridHandle
     typename util::enable_if<BufferTraits<U>::hasDeviceDual, NanoGrid<ValueT>*>::type
     deviceGrid(uint32_t n=0){return const_cast<NanoGrid<ValueT>*>(static_cast<const GridHandle*>(this)->template deviceGrid<ValueT>(n));}
 
+    //@{
+    /// @brief Return a pointer to the @a n'th grid of a single-space device buffer
+    /// @warning Note that the return pointer can be NULL if the GridHandle is empty, @a n is invalid,
+    ///          or if the template parameter does not match the specified grid.
+    template<typename ValueT, typename U = BufferT>
+    typename util::enable_if<BufferHasDeviceSingle<U>::value, const NanoGrid<ValueT>*>::type
+    deviceGrid(uint32_t n=0) const {
```
swahtz: **💅 polish:** This body is a token-for-token copy of `grid()` and the dual-space `deviceGrid()` — the same null/index/gridType checks and `util::PtrAdd`, differing only in the base pointer (`mBuffer.data()` vs `mBuffer.deviceData()`). A private ungated helper taking the base pointer (e.g. `gridAt<ValueT>(const void* base, uint32_t n)`) would serve all three, so a future change to the lookup contract doesn't need three edits — this inline copy sits far from the two out-of-line definitions and would be the one that gets missed.

nanovdb/nanovdb/GridHandle.h
```diff
@@ -371,14 +462,40 @@ template<typename BufferT>
 template <typename OtherBufferT>
 inline GridHandle<OtherBufferT> GridHandle<BufferT>::copy(const OtherBufferT& other) const
```
swahtz, line 452: **💡 suggestion:** Both `copy()` overloads carry an identical `if constexpr` guard and multi-line `static_assert`, and this overload's single-space branch just discards `other` and forwards to `copy<OtherBufferT>()` — which re-fires the same assert, so today every misuse emits the diagnostic twice. I believe the guard can live solely in the no-arg `copy()` with identical diagnostics for every (BufferT, OtherBufferT) combination. This matters for part 2: when cross-space transfers land, the "not supported yet" condition and message get relaxed, and with two copies of the guard the two spellings can silently drift to accepting different type combinations.

nanovdb/nanovdb/GridHandle.h
```diff
@@ -30,18 +32,52 @@ namespace nanovdb {
 
 struct GridHandleMetaData {uint64_t offset, size; GridType gridType;};
 
+/// @brief Detects whether @c BufferTraits<BufferT> defines @c hasDeviceSingle,
+///        i.e. whether the buffer manages a single device-resident allocation.
+/// @details Defaults to false when the trait member is absent, so pre-existing
+///          BufferTraits specializations (in or out of tree) that only define
+///          hasDeviceDual keep compiling unchanged.
+template<typename BufferT, typename = void>
+struct BufferHasDeviceSingle { static constexpr bool value = false; };
```
swahtz: **💡 suggestion:** Could we move `BufferHasDeviceSingle`/`BufferHasHostSingle` next to the `BufferTraits` primary in HostBuffer.h (or a small shared traits header)? They aren't GridHandle-specific — they're companions to the trait protocol itself — and there's a concrete second consumer lined up: `VoxelBlockManagerHandle` needs exactly the same dispatch to map `deviceFirstLeafID()`/`deviceJumpMap()` onto `data()` for single-space buffers (the VBM follow-up to this PR under #2232), and it shouldn't have to include GridHandle.h, or duplicate the detectors, to get it.

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -402,6 +421,36 @@ class BufferView
...
+///        owns exactly one allocation, resident on the device, so the handle
+///        parses metadata through a device read and exposes only the device
+///        accessors. Requires byte-addressed storage.
+/// @note A buffer whose trait sets hasDeviceSingle must provide the interface
+///       the single-space GridHandle constructor consumes: ElementType and
+///       ResourceType typedefs, data(), size_bytes(), resource(), stream()
+///       (stream-ordered resources), and the (stream, resource, count, noInit)
+///       constructor shape.
+template<typename T, typename R>
+struct BufferTraits<cuda::Buffer<T, R>>
```
swahtz, line 448: **💡 suggestion:** This @note is effectively the definition of a "single-space buffer concept" — it lists exactly the members a buffer must provide for `hasDeviceSingle` to be honored. Since other handle types will consume the same contract (`VoxelBlockManagerHandle` is next, per the #2232 plan), could we phrase it as such — name the concept, and state that any consumer of `hasDeviceSingle` may rely on exactly this interface? That lets the follow-up PRs cite it instead of reverse-engineering the GridHandle constructor.

nanovdb/nanovdb/cuda/Buffer.h
```diff
@@ -62,6 +62,13 @@ class Buffer : private detail::StreamHolder<is_async_resource<R>::value>
 
     static constexpr bool IsAsync = is_async_resource<R>::value;
 
+public:
+    /// @brief Element and resource types, for generic code that rebinds one
+    ///        or constructs sibling buffers over the same resource.
+    using ElementType  = T;
```
swahtz, line 68: **💅 polish:** Tiny one: a rebind alias here (`template<class U> using rebind = Buffer<U, R>;`) would let generic code construct sibling buffers without spelling the full type — `makeMetaScratch` spells `cuda::Buffer<GridHandleMetaData, ResourceT>` manually today, and the VBM follow-up will spell two more.

swahtz (changes requested):
Reviewed with a VBM follow-up as part of #2232 in mind. The design lands the right conventions — the detected traits, the named compile errors instead of half-working host paths, and never requiring default-constructible buffers are all things the next handle types can build on directly. Two issues inline: a correctness regression where the hardened chain validation now throws on addBlindData output from multi-grid sources, and the O(N) stream syncs the validation adds to the long-standing GridHandle<DeviceBuffer> constructor (with two possible shapes for fixing it). The rest are non-blocking polish.

nanovdb/nanovdb/cuda/GridHandle.cuh
harrism: Fixed in 5ea0b9e2 — normalized to index 0 / count 1 inside the existing header-fixup kernel (ahead of the tool's checksum recompute, which covers the change). Your audit hunch was right: `indexToGrid` copies the source header the same way (`*dstGrid.data() = *srcGrid.data()` and then patches only `mGridType`/`mData1`), fixed in the same commit. The other handle-returning tools are clean — the topology ops and MeshToGrid normalize in `TopologyBuilder`, and the point builders initialize fresh headers via `GridData::init`. Regression tests run grids 0 and 1 of a merged two-grid handle through both tools; verified failing before the fix with exactly the two throws you predicted.

nanovdb/nanovdb/cuda/GridHandle.cuh
harrism: Fixed in 0f4395d3 with shape (b) plus the deletion from (a): one host read of the head header (still needed to size the metadata scratch and bound a forged `mGridCount` before allocating), then a single kernel that validates the chain **and** fills the metadata scratch, one readback, one sync. So the whole second phase — the `cpyGridHandleMeta` launch, its D2H copy, and the final sync — is gone from both constructors, and constructing the handle `mergeGridHandles` returns costs two round trips regardless of grid count.

nanovdb/nanovdb/GridHandle.h
harrism: Fixed in e86e55e6 — a private `gridAt<ValueT>(base, n)` now backs `grid()` and both `deviceGrid()` families.

nanovdb/nanovdb/GridHandle.h
harrism: Fixed in e86e55e6 — the guard lives solely in the no-arg `copy()`; the pool overload forwards its single-space branch there uncommented, so a misuse diagnoses once and the two spellings cannot drift when cross-space transfers relax the condition.

nanovdb/nanovdb/GridHandle.h
harrism: Fixed in e86e55e6 — both detectors now sit beside the `BufferTraits` primary in HostBuffer.h, so `VoxelBlockManagerHandle` gets them without GridHandle.h.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Fixed in e86e55e6 — the @note now defines the single-space device-buffer concept and enumerates the exact interface a consumer of `hasDeviceSingle` may rely on. It also got tighter while writing it down: the metadata scratch now allocates through `resource()` as a `cuda::Buffer`, so the constructor shape is no longer part of the contract — a conforming buffer never needs to be constructible by a consumer.

nanovdb/nanovdb/cuda/Buffer.h
harrism: Added in e86e55e6. One deliberate non-use: the GridHandle constructor keeps spelling `cuda::Buffer<std::byte, ResourceT>` for its scratch rather than rebinding `BufferT`, so `rebind` stays out of the single-space concept — scratch only needs the buffer's resource, not a sibling of the buffer type.

swahtz (approved):
Great, thanks for addressing my comments

kmuseth (approved):

kmuseth:
[Claude assisted in this review.]

Five findings, ranked by severity:

1. GridHandle.h:446 — copy(pool) silently discards pool for single-space handles (CONFIRMED)
The pool-argument overload immediately does (void)other and delegates to the no-arg path. A caller passing a custom allocator/pool gets a valid copy but the allocation silently came from the wrong resource. The docstring says "ignored" but there's no static_assert or [[deprecated]] to surface this at the call site for generic template code that doesn't read comments. At minimum, a static_assert(false, "pool argument is ignored for single-space handles") on the specialisation would make this impossible to misuse.

2. cuda/GridHandle.cuh:590 — cudaMemcpyAsync to pageable host memory on a non-default stream (PLAUSIBLE)
validGridChainHead and parseGridChain copy GridData/ChainStatus into stack variables and std::vector storage via async copies on the retained (potentially non-default) stream. CUDA 12.2 deprecated async D→H copies to pageable memory on non-default streams; a future driver is not required to honour cudaStreamSynchronize as a visibility fence. The fix is a small pinned scratch buffer (e.g. cuda::Buffer<std::byte, cuda::PinnedResource>) for the readbacks.

3. cuda/Buffer.h:430 — hasDeviceSingle SFINAE fires before the actionable static_assert (CONFIRMED)
BufferTraits<cuda::Buffer<T,R>>::hasDeviceSingle is true for any non-host-accessible T, not just std::byte. When a user writes GridHandle<cuda::Buffer<int,DeviceResource>> and calls handle.data(), SFINAE removes the overload and the compiler emits a confusing "no matching member function" error. The helpful static_assert(sizeof(ElementType)==1) in the constructor is never reached. Narrowing the trait to sizeof(T)==1 && !is_host_accessible<R> would make it fire correctly.

4. cuda/GridHandle.cuh:613 — chain error messages omit the failing grid index (CONFIRMED)
ChainStatus has a gridIndex field populated by parseGridChainKernel, but none of the throw sites in the switch on ChainError include it in the message. A corrupt grid at index 14 in a 20-grid buffer throws the same generic message as a corrupt grid at index 0. Adding " at grid index " + std::to_string(status.gridIndex) to each message costs nothing and makes debugging large merged buffers tractable.

5. GridHandle.h:131 — copy() doc promises non-default-constructible support, else branch requires it (CONFIRMED)
The no-arg copy() docstring says it "also works for buffers that are not default-constructible" but the non-single-space else branch calls this->copy(OtherBufferT()), which requires OtherBufferT to be default-constructible. The claim is only true for the single-space path. The doc should either be corrected or the else branch should be constrained with a static_assert(std::is_default_constructible_v<OtherBufferT>) with a clear message.

kmuseth (approved):
Please address review comments (none are blocking but I'd like a response)

harrism:
Thanks Ken — responses per finding, fixes in 70fe676e:

**1. copy(pool) silently discards the pool — fixed as proposed.** The pool-argument overload is now a compile error for single-space handles (static_assert naming the no-arg overload). Bonus: that makes the overload statically host-only, so its `if constexpr` forward disappears — which also finishes the guard-consolidation from the earlier review round. The place where a caller-supplied allocation target becomes meaningful is the upcoming explicit `copyTo`, which takes a prototype buffer on purpose.

**2. Async D→H to pageable memory — I believe this one is incorrect.** I could not find any such deprecation in the [CUDA 12.2 release notes](https://docs.nvidia.com/cuda/archive/12.2.0/cuda-toolkit-release-notes/index.html) or the [runtime API docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html), and `cudaStreamSynchronize` is contractually a completion barrier ("blocks until stream has completed all operations") — completion of a D2H copy means the bytes are in the destination. Copies to pageable memory are documented to stage through the driver's internal pinned buffer, which costs *overlap*, not *correctness*. Two data points: this exact idiom (async D2H into a stack variable + stream sync) appears at 28 sites in the NanoVDB tree, including the `splitGridHandles` dirty readback that long predates this PR; and a pinned scratch here would add a synchronizing `cudaMallocHost` to a path that reads ~700 bytes twice — strictly worse. Happy to revisit with a doc citation if you have one.

**3. Confusing SFINAE error for non-byte buffers — confirmed, fixed with a different shape.** Narrowing the trait to `sizeof(T)==1` would be worse than the disease: `GridHandle<cuda::Buffer<int,R>>` would then classify as neither single-space nor dual nor host-single, fall through to the **host** constructor path, and dereference a device pointer on the host — the confusing compile error becomes a runtime crash. Instead the byte requirement moved to class scope (`static_assert(!hasDeviceSingle || byte-elements)`, with a detector that treats buffers exposing no ElementType as byte-addressed), so the actionable message fires at instantiation, before any accessor SFINAE.

**4. Error messages omit the grid index — fixed as proposed.** All chain-validation throws now carry "(grid i of N)".

**5. copy() doc overpromise — fixed.** The doc now states plainly that the host path default-constructs the pool argument and therefore requires a default-constructible OtherBufferT; the non-default-constructible freedom is (and was) the single-space path only. I skipped the suggested static_assert: the compiler error at `OtherBufferT()` already names the missing default constructor, and GridHandle.h stays free of `<type_traits>`.

harrism:
CI note: `linux-nanovdb:cxx:clang++-Release` failed on 70fe676e — `ex_nodemanager_cuda` hit an undefined reference to the `GridHandle<cuda::DeviceBuffer>` move constructor. Root cause (reproduced locally with a two-TU test): host-only `.cc` files construct that handle against the declaration alone and depend on a CUDA TU to provide the instantiation, which the header forced through an unused anonymous-namespace function — at -O3 clang dead-strips the function and the complete-object constructor with it, leaving only the base-object variant. A latent fragility this PR's header churn happened to expose, not a semantic change. Fixed in d93b4bcb by replacing the dummy function with an explicit instantiation definition, which emits both constructor variants as weak symbols in every CUDA TU (verified: the failing clang link now succeeds, two CUDA TUs merge cleanly, g++ unaffected).

PR #2292, opened 2026-08-23
NanoVDB: cuda::copyTo cross-space transfers and pinned-buffer GridHandles (CUDA)

harrism:
## What this PR is, and why

Second step of the single-space `GridHandle` arc under #2232 (follows #2288). A single-space handle owns memory in exactly one address space, so moving a grid between spaces needs an explicit API — this PR adds it, and completes the host-readable side of the story:

- **`cuda::copyTo`** — the explicit, stream-carrying deep copy between handles in different address spaces: single-space device → host-readable (HostBuffer or a pinned-resource `cuda::Buffer`) and back, and device → device across different resources. An optional prototype buffer routes the destination allocation through a caller-supplied resource (required for non-default-constructible resources such as `ResourceRef`). A host-readable destination synchronizes the transfer stream before returning; a device destination parses — and thereby validates — its metadata on the transferred bytes. The stream-less overload issues the copy on the source buffer's retained stream, so it needs no caller-side ordering; the stream-taking overload documents the caller-orders contract for cross-stream use.
- **Pinned-buffer handles** — `GridHandle` over a `cuda::Buffer` with a host-accessible resource (e.g. `PinnedResource`, or a stream-ordered pooled variant) is now a first-class host-readable handle instead of a named compile error: metadata parses on the host, the host accessors and `write()` apply, and `read()`/`copy()` allocate through the buffer's resource. Pinned handles are the natural staging half of a `copyTo` pipeline, keeping transfers genuinely asynchronous where a pageable `HostBuffer` degrades them.

**Dual-space handles are untouched**: they keep `deviceUpload`/`deviceDownload`, and `cuda::copyTo` rejects them with a named error — their migration and deprecation land together later in the sequence, as discussed on #2288.

## Testing

Six new tests in `TestBuffer.cu` (`nanovdb_cuda_buffer_unit_test`): pinned handle host-parse/copy/write-read round trip, the same over a stream-ordered host-accessible resource, device↔host and device↔pinned round trips with byte-equality checks, a multi-grid chain transfer, prototype-resource routing verified with exact allocation counts through a counting resource, and the empty-handle edge cases. Full CUDA/mgpu/memory-resource/CPU suites pass unchanged; python module and examples build against the header changes.


kmuseth (approved):
Approved as-is! 

Here's Claude's summary and tests.

What it does:

cuda::copyTo<Dst>(src, stream, proto=nullptr) — explicit deep copy between a device-single GridHandle and a host-readable one (or device↔device across different resources). Handles all four space combinations except host↔host (rejected — use GridHandle::copy for that) and dual-space (rejected — still uses deviceUpload/deviceDownload). An optional prototype buffer routes the destination allocation through a caller-supplied resource, needed for non-default-constructible resources like ResourceRef.
Pinned/host-accessible cuda::Buffers become first-class host-readable GridHandles instead of a hard compile-time rejection: metadata parses on the host, write()/read()/copy() all work, going through the buffer's own resource via a new detail::createHostStorage helper.

Verification performed (not just read-through):

Compiled and linked the PR's actual TestBuffer.cu (all 42 tests, including the 7 new ones covering pinned handles, device↔host↔pinned round trips, multi-grid chains, and prototype-resource allocation counting) against a real GPU (RTX 6000 Ada) — all 42 pass.
Ran the new tests under compute-sanitizer --tool memcheck, racecheck, and synccheck — 0 errors/hazards on all three, which specifically stresses the cross-stream ordering this PR is built around (the manual cudaStreamSynchronize calls gating host reads and sync-resource metadata parses).
Wrote and compiled negative test cases confirming the SFINAE/static_assert gates still correctly reject invalid usage (write() on a device-single handle now points at cuda::copyTo in its error message, as documented) and that the empty-handle/non-default-constructible-resource combination throws (GridHandleCopyToProtoResource exercises this and passed).
Traced the allocation/free bookkeeping in makeTransferStorage/copyTo by hand: the if (!proto) branch is deliberately split into two if constexpr arms rather than one, specifically so a non-default-constructible resource without a proto doesn't get silently instantiated on the untaken branch — verified this compiles and throws correctly rather than hard-erroring.
Checked the sync logic in copyTo: for a device destination with an async (stream-ordered) resource, no extra host sync is inserted (the metadata parse rides the same retained stream as the copy); for a synchronous resource or a host-readable destination, an explicit cudaStreamSynchronize is added first. This matches the documented contract and is exactly what the GridHandleCopyToHostAndBack test (mixing DeviceResource, HostBuffer, and MallocResource) exercises.
Confirmed GridHandle::copy(pool)/copy() correctly restrict pool-argument and cross-type copies away from device-single buffers via static_assert, so detail::createHostStorage is only ever reached for host-readable destinations.
Findings: none. This is a well-scoped, symmetric extension of the single-space buffer work, the stream-ordering reasoning holds up under sanitizer scrutiny (not just review), and the test coverage genuinely exercises the interesting cases (stream-ordered vs. synchronous destination resources, borrowed/non-default-constructible resources, empty handles, multi-grid chains). 

One very minor, non-blocking observation: GridHandle<BufferT>::copy<OtherBufferT>() (no-arg overload) falls back to this->copy(OtherBufferT()), which requires OtherBufferT to be default-constructible — for a host-single buffer over ResourceRef this fails with a generic constructor-not-found error rather than the friendly "requires a prototype" message copyTo/makeTransferStorage give. Not something this PR needs to fix (it's calling into existing, unchanged code), just worth knowing if it comes up in a future PR in this series.

harrism:
Thanks Ken. On the non-blocking observation: agreed, and it has a clean fix now — a void_t default-constructibility detector beside the other trait companions in HostBuffer.h lets the no-arg copy() carry a named "pass a prototype buffer to copy(other)" diagnostic without GridHandle.h taking a <type_traits> include (the reason I passed on the same assert during the #2288 review). I'll land it in the next PR in this series rather than re-cycling this approved one through CI.

PR #2293, opened 2026-08-23
NanoVDB: handle reset() via destroy(), single-space NodeManager, GridStats resource routing (CUDA)

harrism:
## What this PR is, and why

Completion sweep for the single-space handle work under #2232 (follows #2288; independent of #2292 — they can merge in either order). Two threads, both mechanical follow-through of decisions recorded on the earlier PRs:

**`reset()` dispatches to `destroy()`.** `GridHandle`, `NodeManagerHandle` and `tools::VoxelBlockManager` release storage through the buffer's `destroy()` when it provides one (detected by a `BufferHasDestroy` companion to the trait protocol) and fall back to the legacy `clear()` otherwise. With no in-tree callers left, the transitional `cuda::Buffer::clear()` / `cuda::BufferView::clear()` finally carry their `[[deprecated]]` attribute, exactly as their docs promised when they landed — `BufferView` gains `destroy()` with its detach semantics. The legacy `HostBuffer`/`DeviceBuffer` `clear()` methods are unaffected.

**NodeManager goes single-space, and GridStats' last bypass closes.** `cuda::createNodeManager` gains an overload taking a device-resident memory resource, returning a handle over `cuda::Buffer<std::byte, ResourceRef<ResourceT>>` whose storage and size scratch allocate through it (borrowed, must outlive the handle — the builders' ownership contract). `NodeManagerHandle` supports single-space buffers: `deviceMgr()` maps onto the buffer, host accessors become compile-time errors, and its constructors member-initialize so non-default-constructible buffers work. `tools::cuda::GridStats` builds its temporary NodeManager through its injected resource — deleting the documented boundary where that allocation bypassed `ResourceT`, the last one in the tools.

**Drive-by fix**: both `createNodeManager` overloads now synchronize before the host reads the device-computed size — the old code consumed the destination of an asynchronous copy with no ordering, which only worked because device-to-pageable copies degrade to synchronous behavior. The dual overload's raw `mallocAsync`/`freeAsync` size-scratch pair becomes an RAII buffer, retiring the last raw allocation outside DistributedPointsToGrid.

## Testing

New `SingleSpaceNodeManager` test (construction through a counting resource with exact allocation counts, typed/mistyped `deviceMgr`, reset-through-resource leak check); `GridStats_InjectedResourceSeam` expectations updated to assert the NodeManager allocations now route through the resource; buffer tests migrated from `clear()` to `destroy()`. Full CPU/CUDA/buffer/memory-resource/mgpu suites pass; python module and examples build against the header changes.


kmuseth (approved):
Approved after I reviewed it with the help of Claude's summary:

What it does: two mechanical follow-throughs on the single-space-buffer roadmap (#2232):

GridHandle::reset(), NodeManagerHandle::reset(), and VoxelBlockManager::reset() now dispatch to a buffer's destroy() when present (detected via a new BufferHasDestroy trait), falling back to legacy clear() otherwise. cuda::Buffer::clear()/cuda::BufferView::clear() finally get [[deprecated]].
cuda::createNodeManager gains a single-space overload taking a device memory resource directly (NodeManagerHandle<Buffer<std::byte, ResourceRef<ResourceT>>>), and GridStats::update now builds its temporary NodeManager through the injected resource instead of a bare CudaDeviceBuffer, closing the last allocation bypass.
Drive-by fix: both createNodeManager overloads now cudaStreamSynchronize before reading the device-computed size back on the host (previously relied on an unordered async copy that only worked by accident).
Verification performed:

Traced BufferHasDestroy/BufferHasDeviceSingle/BufferHasHostSingle SFINAE gates through NodeManagerHandle and confirmed the if constexpr branches mean the deprecated clear() is never instantiated on types with destroy(), so the new [[deprecated]] attributes won't trip NANOVDB_CUDA_WERROR anywhere in-tree — grepped for every remaining .clear() call on nanovdb buffer types and confirmed each one is on the fallback (no-destroy()) branch.
Compiled real usage (GridHandle over cuda::Buffer, the new createNodeManager overload, deviceMgr<T>(), GridStats::update) against the PR's headers — clean build.
Compiled a negative case (createNodeManager with PinnedResource, a host-accessible resource) and confirmed it's correctly rejected at compile time via the documented SFINAE exclusion, rather than silently miscompiling.
Checked ownership/ordering: the new overload's sizeBuf RAII-frees on the retained stream after the cudaMemcpyAsync+sync read it back, and the large-NodeManager reallocation path (buffer = std::move(tmp)) frees the old allocation on the correct stream via move-assignment, matching the existing dual-space overload's pattern.
Confirmed the updated test count expectations in TestBuffer.cu/TestMemoryResource.cu (2 allocs/1 dealloc for a linear NodeManager built through a resource; 3/3 for GridStats now that its NodeManager routes through the resource too) match the code paths.
Findings: none that need fixing. The one soft edge worth knowing about: GridStats<..., ResourceT> already requires is_async_resource<ResourceT>, and the new createNodeManager overload additionally excludes host-accessible resources — if someone ever instantiated GridStats with an async and host-accessible resource (e.g. a hypothetical pinned+stream-ordered type), they'd get a generic "no matching overload" error inside NodeManager.cuh rather than a message pointing at the actual constraint. Not a bug, just a slightly less friendly error message for a combination that doesn't appear to exist in the codebase today — not worth blocking on.

PR #2294, opened 2026-08-23
improved the vdb option in the vdb_tool

kmuseth:
Summary
Every vdb/geo option across vdb_tool's actions currently only accepts a stack age (an integer index, e.g. vdb=0). This adds support for referencing grids/geometry by name as well, e.g. -diagnose vdb=density instead of having to track which age density currently sits at.

Motivation
Age indices are a shifting reference — they change as the pipeline pushes/pops entries — so scripts that address a specific grid by index are fragile to reordering. Addressing by name is self-documenting and survives stack changes elsewhere in the pipeline.

Design
Implemented as a single, generic pre-processing step rather than touching every action: a new Parser::beforeActionRun hook (mirroring the existing onActionError hook) fires once per action, immediately before its run() callback.
Tool::resolveStackOptions() is wired to that hook. For any option literally named vdb or geo on the about-to-run action, it rewrites the option's value in place via Tool::resolveStackOption():
"" and "*" are left untouched (every action already special-cases these).
Each comma-separated token that parses as an integer is left untouched.
Any other token is looked up by exact name against the corresponding stack (mGrid for vdb, mGeom for geo).
A name matching more than one entry expands to all of their ages, ascending (e.g. vdb=0,density with two grids named density at ages 1 and 3 becomes vdb=0,1,3).
A name matching nothing throws std::out_of_range with the same "<action>: ... (stack size N)" message shape the existing out-of-range-index checks already use — so a typo'd name fails exactly as loudly and clearly as a typo'd/out-of-range index, rather than silently doing nothing.
Because the option's stored string is rewritten before the action body runs, this required zero changes to any of the ~25+ existing action implementations — all their existing age-list parsing, "*" handling, and out-of-range error reporting keep working unmodified downstream.

Testing
Added StackOptionNameResolution to unittest.cpp, covering: plain-name lookup, ambiguous-name expansion mixed with a numeric token, the not-found error, "*" passthrough, and geo= resolution independently of vdb=. Full suite passes (30/30).

Docs
Updated docs/introduction.md's option-semantics summary and added a changelog entry under pendingchanges/openvdb_cmd.txt.


openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -536,6 +553,58 @@ auto Tool::getGeom(size_t age) const
 
 // ==============================================================================================================
 
+void Tool::resolveStackOptions()
+{
+    for (auto &opt : mParser.getAction().options) {
+        if (opt.name != "vdb" && opt.name != "geo") continue;
+        // Read via Parser::get (not opt.value directly) so any {} expression in the
+        // option is fully expanded before name resolution sees it.
+        opt.value = this->resolveStackOption(opt.name, mParser.get<std::string>(opt.name));
```
swahtz: **🚩 issue:** `opt.value` is the parsed source expression, and loop execution revisits the same Action. Replacing it permanently with the first resolved age prevents subsequent iterations from re-evaluating expressions or names against the current stack.  For example, the existing documented pattern:

`  -for v=0,... -render vdb='{$v}' -end`

  will resolve `{$v}` to 0 on its first iteration, store "0", and render age 0 on every subsequent iteration.‏

swahtz: We should also add a loop regression test using `vdb='{$v}'`

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -536,6 +553,58 @@ auto Tool::getGeom(size_t age) const
...
+
+    const bool isVdb = optName == "vdb";
+    const size_t stackSize = isVdb ? mGrid.size() : mGeom.size();
+    const std::string &actionName = mParser.getAction().names[0];
+
+    VecS out;
+    for (const std::string &tok : tokenize(raw, "(),")) {
+        int age;
+        if (isInt(tok, age)) { out.push_back(tok); continue; }// already a stack age
+        // Name lookup: scan age 0 (top/most-recent) upward, collecting every match.
```
swahtz, line 577: **🚩 issue:** ‏Expanding duplicate names to ascending ages is unsafe for actions that mutate the stack while consuming the list. For grids named x at ages 0 and 2, `-clear vdb=x` erases age 0, shifts the stack, then fails on age 2 after a partial deletion. Singleton consumers such as rename also cannot consume the expanded "0,2" value.

swahtz (commented):
Overall looks good, just a pair of issues that might cause issues.

kmuseth:
thanks @swahtz  - I'll address your review. comments tomorrow

kmuseth:
Thanks @swahtz  — both issues were real, and I reproduced each before fixing. Addressed in three commits.

1. Loop expression clobbering. Confirmed exactly as described: -for v=0,2 -stats vdb={$v} -end reported age 0 on every iteration. Root cause was writing the resolved value back into Option::value, destroying the parsed source expression.

Resolution now runs per-read on a copy instead of mutating the option: Parser::beforeActionRun is replaced by Parser::onGetOption, invoked from Parser::getStr() after expression expansion. Every accessor (get<T>, getVec<T>, getVec3<T>) funnels through getStr(), so coverage is unchanged, but the stored expression survives and is re-expanded and re-resolved each iteration. Good catch that this would also have corrupted the source form -write emits to a config file, which reads Option::value directly.

2. Loop regression test. Added, using vdb={$v} as suggested, plus a second case covering a bare name in a loop body to confirm it re-resolves against the current stack each pass.

3. Unsafe multi-age expansion. Also confirmed, including the partial-deletion case — -clear vdb=x erased age 0 before failing on the shifted age 2. An audit settled the design: there are 29 singleton consumers (get<int>("vdb"/"geo")) versus 19 list consumers, so most actions physically cannot accept an expanded list.

Expansion is gone. A name must now identify exactly one entry; an ambiguous one is rejected with the matching ages listed:


clear: the name "x" is ambiguous -- it matches 2 VDB grid entries at ages 0,2. Use an explicit age index to select one.
Because that fires during option resolution, the action aborts before mutating anything — verified all three grids survive.

Two follow-ons your comment surfaced:

The -clear partial deletion is a pre-existing bug, independent of name resolution — -clear vdb=0,2 fails identically with plain integers, since ages are distances from the back of the stack and erasing one shifts the rest. Fixed in its own commit (under Fixes:, kept separate for bisect/revert granularity): every requested age is resolved to a list iterator, with duplicates removed, before anything is erased. An out-of-range index now leaves both stacks untouched — matching how -copy validates up front — and vdb=0,0 deletes one entry rather than erasing the same iterator twice (which was UB). Safe to resolve up front because mGrid/mGeom are std::lists.

-soup2ls was the one action needing an exemption, since its vdb option selected an output resolution level rather than a stack entry — which also forced the awkward "unlike other actions, here vdb selects outputs (not inputs)" caveat in its help text. Rather than special-case it, I renamed the option to levels=. The exemption is gone and every remaining vdb/geo option now uniformly means "an entry on one of the two stacks". The option has never shipped in a release (it exists only in pendingchanges), so I amended that entry in place rather than adding a deprecation.

swahtz (approved):
Thanks for addressing the notes.  I'd also agree with the note in #2296 that the entries in the changelog are going to be too verbose and could be shortened before this is merged.

kmuseth:
shortened change notes as requested

PR #2296, opened 2026-08-23
added histogram action to vdb_tool

kmuseth:
Add -histogram action to vdb_tool (and fix a heap overflow in math::Histogram it exposed)
-histogram (alias -hist)
Prints an ASCII bar chart of the distribution of active values for one or more VDB grids, binned in parallel via tools::histogram().

This complements -stats rather than overlapping it: summary statistics (min/max/mean/stddev) can't distinguish a uniform distribution from a strongly bimodal or saturated one, but the shape is immediately obvious in a histogram. Useful for sanity-checking the output of -forOnValues/-ax kernels, or spotting a level set whose narrow band isn't the expected ramp.

$ vdb_tool -sphere r=1 dim=64 -ls2fog -histogram bins=8
histogram: [0] "ls2fog_sphere" (float)
  101769 of 101769 active voxels in 8 bins over [0.00574875, 1]
  [      0.0057,       0.1300)       3720   3.7% ##
  [      0.1300,       0.2543)       3708   3.6% ##
  [      0.2543,       0.3786)       3654   3.6% ##
  [      0.3786,       0.5029)       3608   3.5% ##
  [      0.5029,       0.6272)       3480   3.4% ##
  [      0.6272,       0.7514)       3330   3.3% ##
  [      0.7514,       0.8757)       3456   3.4% ##
  [      0.8757,       1.0000]      76813  75.5% ########################################
(That 75.5% pile-up in the top bin is the fog volume's saturated interior — exactly the kind of structure -stats' mean/stddev hides.)

Options: vdb selects grids (comma-separated age indices, or * for all — the default); bins sets the bin count (default 10); cols sets the character width of the longest bar (default 40); min/max override the range, which otherwise spans the grid's own extrema (either bound may be given on its own); log=true switches to a logarithmic bar scale, which keeps highly peaked distributions readable.

Supported for Float, Double, Int32, Int64 and Bool grids. Degenerate cases are each reported explicitly rather than rendering an empty or misleading chart: non-scalar grid types, grids with no active voxels, grids whose active values are all identical (no range to bin), and explicit ranges that exclude every voxel.

Fix: heap buffer overflow in math::Histogram::add()
Adding -histogram immediately segfaulted on any fog volume, which turned out to be a pre-existing bug in core OpenVDB, not in the new action.

Histogram::add() computes the bin index as numBins*(val-min)/(max-min), while its range check admits values up to mMax, which is the upper bound plus a small epsilon. A sample exactly equal to the upper bound therefore scales to exactly numBins and writes one element past the end of the bin array — silently corrupting the heap in optimized builds.

This is trivially reachable through tools::histogram(), whose natural usage passes the data's own extrema as the range: any fog volume has an interior of exactly its maximum value, so every one of those voxels writes out of bounds.

Confirmed independently of vdb_tool with a standalone ASAN reproducer:

==96478==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000002a0
READ of size 8 at 0x6060000002a0 thread T0
    #0 openvdb::v13_0::math::Histogram::add(double, unsigned long long) Stats.h:312
0x6060000002a0 is located 0 bytes after 64-byte region [0x606000000260,0x6060000002a0)
The bin index is now clamped to the last bin, so boundary samples are counted in the top bin as intended. Note the sibling constructor Histogram(const Stats&, size_t) already derived mDelta from mMax and so was mostly protected; clamping in add() fixes both constructors and any remaining floating-point edge cases.

Testing
TestStats.cc: new regression case asserting a sample equal to the upper bound is counted once, in the last bin — which fails by construction against the old code, since that count landed out of bounds. Also covers the lower bound and out-of-range rejection. Full TestStats suite (8 tests) passes.
unittest.cpp: new ActionHistogram test covering the fog-volume boundary case, explicit/clipping/excluding ranges, constant-valued and non-scalar grids, invalid bins/cols/inverted range, out-of-range vdb=, empty stack, and a stream-state regression (per-bin rows switch std::clog to fixed/low precision, which must not leak into the next grid's summary). Full vdb_tool_test suite (30 tests) passes.

Docs
Added the action to the README table and a changelog entry under pendingchanges/openvdb_cmd.txt; the Stats.h fix is documented in pendingchanges/openvdb_tools.txt.


openvdb/openvdb/math/Stats.h
```diff
@@ -309,7 +309,13 @@ class Histogram
     inline bool add(double val, uint64_t n = 1)
     {
         if (val<mMin || val>mMax) return false;
```
swahtz: **🚩 issue:** ‏ Both range comparisons are false for NaN, after which NaN is converted to `size_t`. There should be a NaN check and a test to cover it

openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -1726,6 +1739,132 @@ void Tool::stats()
...
+  // Bin the active values of one typed grid. Returns null when there is nothing to
+  // plot, in which case @a n distinguishes the two degenerate cases for the caller:
+  // n==0 means no active voxels at all, whereas n>0 with lo>=hi means every active
+  // voxel shares a single value (tools::histogram requires a non-empty range).
+  auto build = [&](const auto &grid, double &lo, double &hi, uint64_t &n)
+      -> std::unique_ptr<math::Histogram> {
+    const math::Extrema ex = tools::extrema(grid.cbeginValueOn());
+    n = ex.size();
+    if (n == 0) return nullptr;
+    lo = hasMin ? userMin : ex.min();
```
swahtz, line 1814: **🚩 issue:** ‏Suppose the grid extrema are [0, 1] and the user supplies only `min=2`. This produces lo=2, hi=1, returns null, and the caller reports that every voxel has constant value 2. The code should distinguish truly constant extrema from an effective range that excludes all values.

swahtz (commented):
Two notes to fix potential issues but otherwise lgtm

kmuseth:
Follow-up fixes on top of the histogram/stats commits

openvdb/math/Stats.h

Histogram::count, min, and max now take size_t instead of int, eliminating narrowing-conversion warnings (-Werror=conversion) that appeared when callers iterated with a size_t loop variable.
min(n) and max(n) now explicitly cast n to double before dividing by mDelta, preventing silent integer division.
Histogram::add now explicitly rejects NaN before the range comparison. Both val < mMin and val > mMax are false for NaN, so without this guard a NaN would slip through and trigger undefined behaviour when cast to size_t.
Histogram::add now clamps the computed bin index to the last bin rather than indexing one past the end. mMax is stored with a small epsilon so a sample exactly equal to the upper bound scales to numBins, which previously caused an out-of-bounds write.
openvdb/unittest/TestStats.cc

Added a test case for the upper-bound clamping fix (value == max must land in the last bin and be counted exactly once).
Added a test case for NaN rejection (add(NaN) must return false and leave size() at zero).
openvdb_cmd/vdb_tool/include/Tool.h — histogram() action

Fixed a logic error in the build lambda: when the user supplies only one range bound and it falls outside the grid's value range (e.g. grid extrema [0,1], user passes min=2), the old code returned nullptr and the caller reported "all N voxels have constant value 2" — which is wrong on both counts. The build lambda now sets a constant flag only when neither bound came from the user and ex.min() >= ex.max() (the grid is truly constant). Any other lo >= hi case falls through to the "none of the N active voxels fall within the requested range" message.

swahtz (approved):
lgtm, thanks for addressing my notes

pendingchanges/openvdb_cmd.txt
```diff
@@ -2,7 +2,8 @@ vdb_tool:
...
...ng a level set whose narrow band is not the expected ramp. Each bin prints its half-open value range, the sample count, the percentage of the total, and a bar. The "vdb" option selects grids (a comma-separated list of age indices, or "*" for all, the default), "bins" sets the bin count (default 10), "cols" sets the character width of the longest bar (default 40), and "min"/"max" override the range, which otherwise spans the grid's own extrema (either bound may be given on its own). "log=true" switches the bars to a logarithmic scale, which keeps highly peaked distributions -- common for level sets and fog volumes -- readable. Supported for Float, Double, Int32, Int64 and Bool grids, and for the vector grids (Vec3S, Vec3D, Vec3I), which are binned by vector magnitude |v| -- so "-grad -histogram" plots the distribution of gradient magnitudes, which for a true signed distance field should concentrate at 1 (the Eikonal condition). Other value types are skipped with a note. Grids with no active voxels, grids whose active values are all identical (no range to bin), and explicit ranges that exclude every voxel are each reported explicitly rather than printing an empty or misleading chart.
```
Idclip: A quick note - these release notes are way way too verbose, there's no way we can include this entire file in the notes as is. I would suggested trimming all of these down to a few concise sentences each and moving the rest into actual documentation

kmuseth: Shortened the release notes for "stats" and "histogram", as well as the notes on the bug fixes in math::Histogram::add()

PR #2297, opened 2026-08-24
fixed warnings that caused build to fail

kmuseth:
Fix deprecated CudaDeviceBuffer uses and NVCC unsigned-comparison warning

Replace three remaining uses of the deprecated nanovdb::CudaDeviceBuffer type alias with the current nanovdb::cuda::DeviceBuffer name, which was introduced to give the buffer its canonical namespaced location. The affected files are:

nanovdb/examples/ex_read_nanovdb_sphere_accessor_cuda/read_nanovdb_sphere_accessor_cuda.cu
nanovdb/examples/ex_nodemanager_cuda/nodemanager_cuda_kernel.cu
nanovdb/tools/cuda/GridStats.cuh
A \code block in nanovdb/tools/CreateNanoGrid.h and a test name in TestNanoVDB.cu that reference CudaDeviceBuffer are intentionally left unchanged (doc comment and string identifier respectively, not compiled code).

Additionally, fix a build error in openvdb/tree/ValueAccessor.h triggered by NVCC warning #186-D (-Werror=deprecated-declarations promotes it to an error). NVCC instantiates all branches of if constexpr before applying discarding semantics, so it evaluates Idx > NumCacheLevels+1 for every index including Idx = 0. When NumCacheLevels+1 wraps to zero (unsigned overflow in a degenerate instantiation), the comparison reduces to size_t > 0, which NVCC flags as a pointless unsigned comparison. The fix casts both operands to ptrdiff_t before comparing, which is safe since cache level counts are always small.


swahtz (approved):
lgtm

openvdb/openvdb/tree/ValueAccessor.h
```diff
@@ -997,7 +997,7 @@ class ValueAccessorImpl final :
         return this->evalFirstPred([&](const auto Idx) -> bool
             {
                 if constexpr(Idx < Start)                  return false;
-                else if constexpr(Idx > NumCacheLevels+1)  return false;
+                else if constexpr(static_cast<ptrdiff_t>(Idx) > static_cast<ptrdiff_t>(NumCacheLevels) + 1)  return false;
```
Idclip, line 1000: This should be size_t no? At least that's the type that `TSEvalFirstPredImpl` will pass

Idclip: Ah nvm I assumed this was a implicit convert issue not an unsigned wrap. Why does `-Werror=deprecated-declarations` elevate this to an error? But either way looks okay now

nanovdb/nanovdb/examples/ex_nodemanager_cuda/nodemanager_cuda_kernel.cu
```diff
@@ -31,7 +31,7 @@ extern "C" void launch_kernels(const nanovdb::NodeManager<float>* deviceMgr,
 
 // Simple wrapper that makes sure nanovdb::cuda::createNodeManager is initiated
 extern "C" void cudaCreateNodeManager(const nanovdb::NanoGrid<float> *d_grid,
-                                      nanovdb::NodeManagerHandle<nanovdb::CudaDeviceBuffer> *handle)
+                                      nanovdb::NodeManagerHandle<nanovdb::cuda::DeviceBuffer> *handle)
 {
     *handle = std::move(nanovdb::cuda::createNodeManager<float>(d_grid));
 }
```
harrism: todo: this file needs a newline at the end. (If we had clang-format, it would fix that automatically)

nanovdb/nanovdb/tools/cuda/GridStats.cuh
```diff
@@ -311,7 +311,7 @@ void GridStats<BuildT, StatsT, ResourceT>::update(NanoGrid<BuildT> *d_grid, cuda
     static const uint32_t threadsPerBlock = 128;
     auto blocksPerGrid = [&](uint32_t count)->uint32_t{return (count + (threadsPerBlock - 1)) / threadsPerBlock;};
 
-    auto nodeMgrHandle = nanovdb::cuda::createNodeManager(d_grid, CudaDeviceBuffer(), stream);
+    auto nodeMgrHandle = nanovdb::cuda::createNodeManager(d_grid, nanovdb::cuda::DeviceBuffer(), stream);
```
harrism: This PR conflicts with #2293 here, which deletes this line. Whichever PR merges first will cause a small conflict in the other. (no biggy).

harrism (approved):
The alias that errored,  `nanovdb::CudaDeviceBuffer`, has been deprecated since NanoVDB v32.7.0 (PR #1651). What did change recently: #2280 made `NANOVDB_CUDA_WERROR` default `ON`, so any `nvcc` diagnostic — including deprecated-use warnings on those three ancient in-tree `CudaDeviceBuffer` uses — is now a build error on any toolkit that emits it. CUDA 12.6 (what CI and my local builds use) demonstrably does not diagnose those uses. But your Blackwell box needs a newer toolkit than that, and that newer nvcc (13.x) evidently does emit the deprecation warning (plus the #186-D unsigned-comparison diagnostic in openvdb's ValueAccessor.h). 

So the ValueAccessor warnings are not related to the deprecation warnings, just to the fact that the header is included in a TU compiled by your newer NVCC. These are all needed fixes nonetheless, however one is in a line deleted by my currently open #2293, awaiting your review.

openvdb/openvdb/tree/ValueAccessor.h
harrism: It's not because of `deprecated-declarations`, it's because we enabled -Werror by default for CUDA TUs and the older CUDA compiler (12.6) used in CI does not warn on this line, while Ken's newer 13.x does.

Idclip: Okay thank you, the MR description is confusing (not sure why https://github.com/AcademySoftwareFoundation/openvdb/pull/186 is related too)

nanovdb/nanovdb/examples/ex_nodemanager_cuda/nodemanager_cuda_kernel.cu
kmuseth: newline added

nanovdb/nanovdb/tools/cuda/GridStats.cuh
kmuseth: True, but as you point out, it's a trivial one-line merge conflict. I'll keep it here (so I can finally build on my NVCC 13.x workstation), but in return, I promise to review #2293 next and resolve the conflict if you prefer.

kmuseth: Actually, I've changed my mind. I'll remove this change from this PR and first merge in #2293, which I've just approved.

PR #2298, opened 2026-08-25
Add In-Memory Conversion to No Offset Grids

danrbailey:
This fixes an issue raised by @apradhana (and a few related issues).

When using VDB files that have no offsets, because they were written using a stream, the VDB grids are cached in-memory. These in-memory grids were not being converted (to half, bool, etc) if requested, as conversion was only applied during read.

This is now addressed using in-memory conversion for cached grids. In addition, this fixes clipping of cached grids and instance grids (which are effectively also cached).

Finally, I have added diagnostics when the hints were not applied such as when returning a cached grid and requesting a topology-only grid.

### DIFF
--- a/openvdb/openvdb/io/File.cc
+++ b/openvdb/openvdb/io/File.cc
@@ -34,9 +34,10 @@ namespace {
 ///   produced had it been read through @a codec (looked up by the caller via
 ///   the protected @c Archive::findCodec(), since this is a free function).
 ///   Returns null if no conversion is needed or possible, in which case the
-///   caller keeps the original grid and @a diagnostics is set accordingly.
+///   caller keeps the original grid.  A conversion that was requested but cannot
+///   be done is reported to @a diagnostics and logged against @a filename.
 GridBase::Ptr convertGridForReadMode(const GridBase& source, const ReadOptions& readOptions,
-    Codec* codec, ReadDiagnostics& diagnostics);
+    Codec* codec, ReadDiagnostics& diagnostics, const std::string& filename);
 
 /// @brief Return the name used in diagnostics and log messages for @a mode.
 std::string readModeName(ReadMode mode);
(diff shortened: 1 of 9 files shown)


openvdb/openvdb/io/File.cc
```diff
@@ -608,6 +669,157 @@ File::endName() const
...
+
+    // No conversion codec registered for this grid type (falls through to the
+    // plain gridType codec), or the registry agrees the type is unchanged.
+    if (targetType.empty() || targetType == source.type()) {
+        diagnostics.addWarning(source.getName(),
+            "ReadMode::" + modeStr + " conversion is not supported for grid type '"
+            + source.type() + "'; reading as original type");
+        return GridBase::Ptr();
+    }
+
```
apradhana: Nitpick comment: for `targetType == source.type()`, we can say
```
if (targetType == source.type()) {
    // The grid already has the requested representation.
    return {GridBase::Ptr(), ConversionStatus::NoOp};
}
```

openvdb/openvdb/io/File.cc
```diff
@@ -608,6 +669,157 @@ File::endName() const
...
+        using SourceGridT = std::decay_t<decltype(typedSource)>;
+        using TargetGridT =
+            typename SourceGridT::template ValueConverter<TargetBuildT>::Type;
+        // No LeafNode conversion constructor exists from ValueMask to any
+        // other build type, so exclude it here; nested if constexpr avoids
+        // instantiating the ambiguous CanConvertType<ValueMask, ValueMask>.
+        if constexpr (!std::is_same_v<typename SourceGridT::BuildType, ValueMask>) {
+            if constexpr (CanConvertType<typename SourceGridT::BuildType, TargetBuildT>::value) {
+                if (TargetGridT::gridType() == targetType) {
+                    result = typename TargetGridT::Ptr(new TargetGridT(typedSource));
```
apradhana: Could we special-case `std::is_same_v<TargetBuildT, ValueMask>` here and use a topology copy with inactive `false` and active `true`? `TargetGridT(typedSource)` copies the source grid’s values into the mask. For example, a nonzero background becomes `true`. But `ReadMode::Mask` should ignore the source values and create a mask based only on which voxels are active.

apradhana (changes requested):
Thanks for adding this, @danrbailey . 

I noticed that `TopologyOnly` is still ignored for no-offset files. Should we add a `throw` for this and add a comment saying that it's not supported in the `ReadMode` [contract](https://github.com/AcademySoftwareFoundation/openvdb/blob/e4dbaaf905085c41685c72dcb30da873fb0c1cc4/openvdb/openvdb/io/Codec.h#L80-L90)?

danrbailey:
> Thanks for adding this, @danrbailey .
> 
> I noticed that `TopologyOnly` is still ignored for no-offset files. Should we add a `throw` for this and add a comment saying that it's not supported in the `ReadMode` [contract](https://github.com/AcademySoftwareFoundation/openvdb/blob/e4dbaaf905085c41685c72dcb30da873fb0c1cc4/openvdb/openvdb/io/Codec.h#L80-L90)?

I went back and forth on this. TopologyOnly only reads the topology and doesn't read the value data. However, it still allocates and populates the values in the VDB (with zeros), that is to ensure that we never have partially allocated data. 

In the case of no-offset files, we have a full grid already cached, so we could deep copy that and then zero out all the values just so that the two match. I'm not sure I see the value in doing that though? I think what the reader is saying is that they only care about the Topology. We don't make any guarantees about what is in the value data - both zero values and actual value data satisfy that?

danrbailey:
> Thanks for adding this, @danrbailey .
> 
> I noticed that `TopologyOnly` is still ignored for no-offset files. Should we add a `throw` for this and add a comment saying that it's not supported in the `ReadMode` [contract](https://github.com/AcademySoftwareFoundation/openvdb/blob/e4dbaaf905085c41685c72dcb30da873fb0c1cc4/openvdb/openvdb/io/Codec.h#L80-L90)?

@apradhana - this should be resolved now. All modes (TopologyOnly, Half, Bool, Mask) now work as expected with no-offset files.

openvdb/openvdb/io/File.cc
```diff
@@ -608,6 +670,166 @@ File::endName() const
...
+File::resolveCachedGrid(const GridBase::Ptr& cachedGrid, const io::ReadOptions& readOptions,
+    ReadDiagnostics& diagnostics) const
+{
+    if (readOptions.readMode == ReadMode::MetadataOnly) {
+        return cachedGrid->copyGridWithNewTree();
+    }
+
+    GridBase::Ptr grid = cachedGrid;
+
+    if (readOptions.readMode == ReadMode::TopologyOnly) {
```
apradhana, line 800: Please consider adding a comment a.k.a. contract where `ReadMode::TopologyOnly` is introduced in `Codec.h` that says that only topology is guaranteed and values may get background or already-cached data.

openvdb/openvdb/unittest/TestCodec.cc
```diff
@@ -348,6 +351,687 @@ TEST_F(TestCodec, testFloatToHalfCodecConversion)
...
+    ReadOptions readOptions;
+    readOptions.readMode = ReadMode::TopologyOnly;
+
+    io::File f(path);
+    f.open();
+    f.enableReadDiagnostics();
+
+    GridBase::Ptr grid;
+    EXPECT_NO_THROW(grid = f.readGrid(srcGrid->getName(), readOptions));
+    ASSERT_TRUE(grid);
```
apradhana, line 761: Please consider adding a test that the returned grid has the same topology as srcGrid.

apradhana (approved):
Thanks for adding support for no-offset grids path. It looks good. I added two non-blocking comments.

danrbailey:
> Thanks for adding support for no-offset grids path. It looks good. I added two non-blocking comments.

Thanks! I added a warning to the TopologyOnly section in Codec.h and extended the unit tests slightly.

Issue #2300, opened 2026-08-27
NanoVDB: cuda::SignedFloodFill's root-tile pass reads out of bounds on the device (confirmed with compute-sanitizer)


harrism:
## Summary

`tools::cuda::SignedFloodFill`'s `processRoot` copies the tree header + root + tiles from the device grid into a small managed-memory buffer, then evaluates `root->getChild(root->tile(idx))->getFirst/LastValue()` inside a `<<<1,1>>>` kernel **against that copy**. `RootData::Tile::child` is a signed byte offset **relative to the RootData it is resolved against** (`RootData::getChild` returns `util::PtrAdd<ChildT>(this, tile->child)`), and the child nodes were never copied — so the computed pointer lands far past the end of the managed allocation.

Confirmed on a minimal repro (two voxels of value -1 at (0,0,0) and (0,0,8192), i.e. two root children on one z-scanline with a gap):

```
Invalid __global__ read of size 8 bytes
  at ...processRoot<float>...[lambda]
  Address 0x... is out of bounds, 5801 bytes after the nearest allocation (48 bytes)
```

The faulting address decomposes exactly as managed-buffer base + 64 (tree) + 128 (tile[0].child) + 8216 (the childMask word `getLastValue` reads).

**The failure is usually silent.** Without the sanitizer the run produces no CUDA error on Ada: unified memory maps address space at coarse granularity, so the out-of-bounds read returns unrelated bytes, and the inside/outside scanline decision is based on garbage — the pass can silently skip or wrongly insert interior tiles. (In the repro it happens to fall through to the documented "missing internal tile" throw.)

Introduced with the root-tile pass itself (1db28418, Feb 2025).

## Why nobody has hit it

The kernel only runs for two lexicographically adjacent root children with identical (x,y) origins and a z-gap **greater** than 4096 — interior root-level tiles, i.e. level sets thousands of voxels across. A census found:

- **No test reaches it.** Both CUDA tests (`CudaSignedFloodFill`, `NonBlockingStreamSignedFloodFill`) build origin-centered spheres whose 8 root children are all direct neighbors (`d[2] == 4096`), so every pair `continue`s; the seam test early-returns with one child; the python tests mirror the sphere topology. The second test's own comment concedes the pass "needs a level set thousands of voxels across."
[...]

PR #2301, opened 2026-08-27
NanoVDB: migrate off the dual-space buffers and deprecate the DeviceBuffer name (CUDA)

harrism:
## What this PR is, and why

This is the migrate-and-deprecate step of the memory-management work under #2232. Grid storage is moving from the dual-space `DeviceBuffer` (one object holding a host copy and a device copy, moved with `deviceUpload`/`deviceDownload`) to the single-space `cuda::Buffer` with explicit transfers (`cuda::copyTo`). This PR migrates everything inside NanoVDB and then deprecates the old name — designed so that **nobody is broken and nobody gets a warning they cannot act on**:

- `cuda::DeviceBuffer` becomes a **deprecated alias** for the renamed implementation class. The GPU tools' default buffer types name the implementation directly, so code that relies on defaults — most user code — compiles **without any warning** until the defaults change at removal, one or more releases from now. Only code that spells `DeviceBuffer` sees the warning, and the message carries the migration recipe.
- `cuda::copyTo` **adopts the source handle's metadata** — validated once, when that handle was constructed from raw bytes — so transfers run no kernel and are **callable from host-only translation units** (`cuda/HandleStorage.h`). The validation kernel keeps guarding the real trust boundary: constructing a handle from raw bytes.
- Cross-stream ordering, which `DeviceBuffer` handled with internal events, is the documented four-line CUDA-event pattern in `cuda/Buffer.h` — matching `cuda::buffer` (CCCL) and RMM.
- `UnifiedBuffer` intentionally keeps its name here: its replacement (`ManagedResource`, added in this PR, plus the multi-GPU tool that defaults to it) completes in the multi-GPU follow-up in the same release, with the identical alias treatment.

## Commits, in review order

1. **Tool entry points accept single-space buffers** — the tools allocated results through the duck-typed `create()` that `cuda::Buffer` doesn't have; a small bridge (`cuda/HandleStorage.h`) dispatches to `create()` or the buffer's memory resource. Plus the friendly `copy()` diagnostic from the #2292 review.
2. **`ManagedResource`** — cudaMallocManaged-backed; a handle over it exposes both the host and device accessors over one allocation (the UnifiedBuffer migration target). Also restores four buffer headers missing from the install list.
3. **TopologyBuilder's two dual members split** into pinned host staging + resource-allocated device buffers, with the upload mechanics in one place instead of six tools.
4. **MeshToGrid's 15 internal buffers** onto resource-backed `cuda::Buffer` (it predated the earlier scratch conversions); fixes its sidecar being hard-coded to `DeviceBuffer`, which meant the `SidecarBufferT` parameter never compiled with any other type.
5. **SignedFloodFill's root pass** onto `ManagedResource` (the out-of-bounds read documented in #2300 is pre-existing and unchanged here; its fix belongs to that issue).
6. **All 15 CUDA examples migrated** — including three pre-existing example bugs found on the way (a grid-index example whose "GPU" output was computed on the CPU, a deprecated shim use, a dead `#if 0`).
7. **The deprecation itself** — rename + deprecated alias, with the tests keeping their coverage of the still-shipping dual paths under file-local warning suppression.
8. **CUDA headers out of example host files** — the topology drivers are plain `.cpp`; transfers route through their `.cu` companions (plus an old signature mismatch fixed).
9. **`copyTo` adopts metadata and becomes host-callable** — deletes the transfer-helper pattern the previous commit needed, removes a scratch allocation and two synchronizing round trips per device transfer (net −59 lines).
10. **Review-round polish** — comment style, the rename rationale documented at the class, one over-deletion restored.

## Testing

Full CPU/CUDA/buffer/memory-resource/mgpu suites green at every step (155+63+46+19+11 on RTX 6000 Ada); example programs run-verified where buildable; python module builds; a two-translation-unit link check confirms the rename preserves the cross-TU constructor instantiation; a host-only probe TU calling `copyTo` compiles under g++ and clang.


nanovdb/nanovdb/GridHandle.h
```diff
@@ -171,9 +180,9 @@ class GridHandle
     /// @brief Returns a pointer to the host data; not available for a
     ///        single-space device buffer, which has no host-readable bytes.
     /// @warning Note that the return pointer can be NULL if the GridHandle was not initialized
-    template<typename U = BufferT, typename util::disable_if<BufferHasDeviceSingle<U>::value, int>::type = 0>
+    template<typename U = BufferT, typename util::disable_if<BufferHasDeviceSingle<U>::value && !BufferHasHostSingle<U>::value, int>::type = 0>
```
swahtz: **💅 polish:** ‏We use this `BufferHasDeviceSingle<U>::value && !BufferHasHostSingle<U>::value` expression 16 times here and 3 times in `NodeManager.cuh`, we could push this predicate into something named to DRY this up/improve robustness.

nanovdb/nanovdb/tools/cuda/SignedFloodFill.cuh
```diff
@@ -25,7 +25,8 @@
 #include <nanovdb/NanoVDB.h>
 #include <nanovdb/cuda/Buffer.h>
 #include <nanovdb/GridHandle.h>
-#include <nanovdb/cuda/UnifiedBuffer.h>
+#include <nanovdb/cuda/Buffer.h>
```
swahtz: **⌨️ typo:** ‏`Buffer.h` is already included 2 lines above

nanovdb/nanovdb/cuda/DeviceBuffer.h
```diff
@@ -391,16 +401,29 @@ class DeviceBuffer
...
 
-}; // DeviceBuffer class
+}; // DualDeviceBuffer class
+
+/// @brief The dual-space device buffer under its long-standing public name.
+/// @deprecated Grid storage is moving to the single-space cuda::Buffer:
+///             build or read into a host handle and move it with
+///             cuda::copyTo (see cuda/GridHandle.cuh), or allocate the
+///             result of a GPU tool directly in a cuda::Buffer. Transfers
+///             construct the device handle with a validation kernel, so
```
swahtz: **❓ question:** ‏Isn't this description inaccurate?  Isn't `copyTo` kernel-free and host-callable?

nanovdb/nanovdb/cuda/HandleStorage.h
```diff
@@ -0,0 +1,239 @@
...
+                                     "that is not default-constructible");
+        }
+    }
+    DstBufferT dst = detail::makeTransferStorage<DstBufferT>(bytes, stream, proto);
+    const void* srcPtr;
+    if constexpr (srcDev) srcPtr = src.deviceData();
+    else                  srcPtr = src.data();
+    constexpr cudaMemcpyKind kind = srcDev ? (dstDev ? cudaMemcpyDeviceToDevice : cudaMemcpyDeviceToHost)
+                                           : cudaMemcpyHostToDevice;
+    cudaCheck(cudaMemcpyAsync(dst.data(), srcPtr, bytes, kind, stream));
```
swahtz: **🚩 issue:** ‏(From Claude)  `cuda::copyTo` returns with the H2D cudaMemcpyAsync still in flight for a stream-less (sync-resource) source such as a pinned-resource Buffer, leaving a use-after-free window when the source handle is destroyed promptly.

nanovdb/nanovdb/cuda/HandleStorage.h
```diff
@@ -0,0 +1,239 @@
...
+    if constexpr (!dstDev)
```
swahtz: **🚩 issue:** ‏(From Claude)  `cuda::copyTo` skips the documented pre-return synchronization for a ManagedResource destination: it is host-readable (hasHostSingle==true) yet classified dstDev==true, so host reads through the returned handle race the pending async copy.

nanovdb/nanovdb/GridHandle.h
swahtz, line 484: **🚩 issue:** ‏(From Claude) `GridHandle::copy()` has no compiling route to a host buffer for ManagedResource handles, and its two error messages point at each other in a circle.

nanovdb/nanovdb/cuda/NodeManager.cuh
```diff
@@ -49,7 +49,7 @@ inline BufferT makeNodeManagerStorage(cudaStream_t stream, const RefT& ref, uint
 /// @param buffer buffer from which to allocate the output handle
 /// @param stream cuda stream
 /// @return Handle that contains a device NodeManager
-template <typename BuildT, typename BufferT = DeviceBuffer>
+template <typename BuildT, typename BufferT = DualDeviceBuffer>
```
swahtz, line 52: **❓ question:** ‏Why does NodeManager have to use `DualDeviceBuffer`, is it not possible to use the single-space `cuda::Buffer`?

nanovdb/nanovdb/cuda/HandleStorage.h
```diff
@@ -0,0 +1,239 @@
...
+        } else {
+            throw std::runtime_error("cuda::copyTo: an empty handle cannot be copied to a buffer type "
+                                     "that is not default-constructible");
+        }
+    }
+    DstBufferT dst = detail::makeTransferStorage<DstBufferT>(bytes, stream, proto);
+    const void* srcPtr;
+    if constexpr (srcDev) srcPtr = src.deviceData();
+    else                  srcPtr = src.data();
+    constexpr cudaMemcpyKind kind = srcDev ? (dstDev ? cudaMemcpyDeviceToDevice : cudaMemcpyDeviceToHost)
```
harrism: Can't we just use cudaMemcpyDefault here?

nanovdb/nanovdb/GridHandle.h
harrism: Good call. Named it `BufferIsDeviceOnly` beside the other detectors in HostBuffer.h and replaced all 19 spellings across the handle headers. Fixed in bfc450d6.

nanovdb/nanovdb/tools/cuda/SignedFloodFill.cuh
harrism: Fixed in bfc450d6.

nanovdb/nanovdb/cuda/DeviceBuffer.h
harrism: Correct on both counts: the metadata-adoption commit made copyTo kernel-free and host-callable, and this doc predated it. Rewrote the @deprecated block and the alias message; they now point at cuda/HandleStorage.h, where copyTo lives. Fixed in bfc450d6 (a sibling message in GridHandle::copy caught in e8224fe7).

nanovdb/nanovdb/cuda/HandleStorage.h
harrism: Real, but resolved in the contract rather than with a hidden synchronization: cross-stream ordering stays the caller's, and for a source with a retained stream it is automatic (destruction frees on the same stream as the copy). The gap was the documentation, which claimed the retained-stream overload has "no such requirement" without qualifying stream-less *sources*: a synchronous-resource (pinned) source frees immediately on destruction, unordered against the still-asynchronous copy. The @warning now states that responsibility explicitly — synchronize the stream before destroying such a source (pageable sources are exempt; their copies degrade to synchronous). Fixed in ec26d38a.

nanovdb/nanovdb/cuda/HandleStorage.h
harrism: Confirmed, and this one is a code fix: the documented postcondition is "a host-readable destination synchronizes before returning", and a ManagedResource destination is host-readable (its handle exposes grid() immediately) but was classified by hasDeviceSingle alone. The sync condition is now `!dstDev || BufferHasHostSingle<DstBufferT>` — pinned and managed destinations synchronize, device-only destinations stay stream-ordered. Added TestBuffer.GridHandleCopyToManagedSynchronizes. Fixed in ec26d38a.

nanovdb/nanovdb/GridHandle.h
harrism: Half-confirmed: there is a compiling route — `cuda::copyTo<HostBuffer>(managedHandle)` works (managed handles satisfy the device-single side of its static_assert) — so the messages terminate rather than circle. But copy(pool)'s message did bounce you through copy()'s error to find that out; both messages now name cuda::copyTo and its header directly. Fixed in bfc450d6 and e8224fe7.

nanovdb/nanovdb/cuda/NodeManager.cuh
harrism: It doesn't have to be — the single-space route is the createNodeManager overload taking a memory resource (directly below this one): it returns a handle over `cuda::Buffer<std::byte, ResourceRef<ResourceT>>`, and GridStats uses it internally. The DualDeviceBuffer default on this overload is the legacy pool-buffer entry point, kept warning-free until the defaults flip at the removal step — the same policy as the tools' signature defaults. No change here; happy to add a cross-reference in the doc comment if that would help.

nanovdb/nanovdb/cuda/HandleStorage.h
harrism: Yes — every pointer copyTo touches is a UVA address (device, managed, pinned, or pageable host), so the runtime infers the direction, and DistributedPointsToGrid already relies on cudaMemcpyDefault for its cross-device copies. The constexpr kind ternary is gone; srcDev/dstDev remain only for pointer selection and the sync decision. Fixed in ec26d38a.

nanovdb/nanovdb/GridHandle.h
swahtz: Looks great, thanks

nanovdb/nanovdb/cuda/HandleStorage.h
```diff
@@ -0,0 +1,246 @@
...
+    if constexpr (srcDev) srcPtr = src.deviceData();
+    else                  srcPtr = src.data();
+    cudaCheck(cudaMemcpyAsync(dst.data(), srcPtr, bytes, cudaMemcpyDefault, stream));
+    if constexpr (!dstDev || BufferHasHostSingle<DstBufferT>::value)
+        cudaCheck(cudaStreamSynchronize(stream)); // the host-readable result is the postcondition; covers the both-space (managed) destination, whose handle exposes host accessors immediately
+    // A handle-to-handle copy adopts the source's metadata, which was
+    // validated when that handle was constructed from raw bytes -- no kernel
+    // runs here, which is what keeps this header host-includable. A device
+    // destination is stream-ordered: use it on @a stream, or synchronize.
+    return detail::HandleFactory::make(std::move(dst), detail::HandleFactory::meta(src));
```
swahtz, line 227: **[P2] Validate host-originated metadata before adopting it**

This assumes every source handle has already received the device constructor's validation, but the host-buffer constructor only checks the first header's `isValid()` and then calls `cpyGridHandleMeta()`. It does not check grid-chain bounds, grid indices/counts, or subsequent headers.

A local CPU probe constructed a `GridHandle<HostBuffer>` successfully after setting the raw header's `mGridSize` to twice the allocation and `mGridIndex` to 7: `bufferSize=2452256`, `gridSize=4904512`, `gridIndex=7`, with `isValid()` still returning true. Previously, uploading this handle with `copyTo` would reject it in `validGridChainHead`/`parseGridChain`; this path now copies the allocation and adopts the invalid metadata without those checks. Downstream operations trusting the reported grid size can then access beyond the device allocation.

Could the host constructor perform equivalent bounds and chain-consistency validation before metadata is considered trusted? That would preserve the kernel-free, host-callable transfer design. A regression test for a malformed host handle would cover this gap. The host-construction behavior was reproduced locally; GPU execution was unavailable.

pendingchanges/nanovdb.txt
```diff
@@ -15,6 +17,9 @@ NanoVDB:
...
   - The bug-fix to the nanovdb::ReadAccessor (see below) improves random-access performance in some use-cases (especially on the CPU).
 
+  Deprecations:
+  - nanovdb::cuda::DeviceBuffer is deprecated: grid storage is moving to the single-space nanovdb::cuda::Buffer, with explicit transfers via nanovdb::cuda::copyTo. Code that uses the GPU tools' default buffer type is unaffected until the removal (the defaults now name the transitional DualDeviceBuffer implementation); only code naming DeviceBuffer sees the warning, whose message carries the migration recipe. Note that constructing a device grid handle validates the grid with a kernel, so transfers belong in CUDA translation units; host-only source files hold the returned handle and delegate the copy to a small helper in a CUDA file (the CUDA examples demonstrate this). The tools' internal scratch, the examples and the NanoVDB tests are fully migrated; UnifiedBuffer follows in the multi-GPU work, in the same release.
```
swahtz: **[P3] Update the release notes to match the final transfer API**

This still says transfers belong in CUDA translation units and that host-only callers must delegate to a helper in a CUDA file. `cuda::copyTo` is now kernel-free and callable directly from ordinary C++ through `nanovdb/cuda/HandleStorage.h`, as the migrated examples demonstrate. Please update the migration guidance so users do not add unnecessary CUDA wrappers.

The new `ManagedResource` entry above also says a handle parses its metadata on the host, whereas construction from raw managed bytes uses the device validation path. That description should be updated as well.

nanovdb/nanovdb/cuda/HandleStorage.h
harrism: Confirmed and fixed — good catch, and it was a class, not an instance: the host-buffer constructor and the host side of the dual-buffer constructor both built metadata through the unvalidated `cpyGridHandleMeta` walk. Both now run `detail::parseHostGridChain`, the host counterpart of the device parse with the same checks: buffer floor, count > 0, count vs. bytes/sizeof(GridData), and per header validity, index/count consistency, and size bounds. Each header is read through an aligned copy rather than a cast pointer, so a forged size in a preceding header cannot even produce a misaligned dereference (the device kernel now does the same). The transfer design is unchanged: `copyTo` still adopts metadata kernel-free, because construction now guarantees that metadata is trustworthy for every buffer family. `cpyGridHandleMeta` is deleted. Regression tests: `TestNanoVDB.GridHandleRejectsMalformedHostChain` (forged size, index, count, magic — reproduced failing before the fix on exactly your size×2 / index=7 scenario) and `TestBuffer.GridHandleRejectsMalformedDeviceChain` (same forgeries uploaded as raw device bytes). Fixed in 8b20c01c.

pendingchanges/nanovdb.txt
harrism: Both fixed. The transfer guidance was already corrected in 7f945576 (the note now says `cuda::copyTo` adopts validated metadata, runs no kernel, and is callable directly from host-only translation units). The ManagedResource entry now says such a handle validates its metadata through the device — a host parse could race the kernels producing the grid — while exposing both accessor families; fixed in 895c05ef.

swahtz (approved):
Thanks for addressing my comments.  This is looking great.

PR #2304, opened 2026-08-27
Minor updates to vectorized PCA from feedback

Idclip:
Also fixed various casting precision warnings across different batch sizes and float types


richhones (approved):
lgtm

PR #2307, opened 2026-08-31
tools: fix wrong voxel offset in MeshToVolume checkNeighbours

sifakis:
## Summary
- In `openvdb/openvdb/tools/MeshToVolume.h`, the `checkNeighbours()` overload that scans a leaf node's raw `data` array has a `mask[6]` branch labeled `i+1, j, k-1` that checks `data[pos + NodeT::DIM * NodeT::DIM]` — the same offset used by the `mask[0]` (`i+1, j, k`) branch two lines above, missing the `- 1` for the `k-1` step.
- The analogous edge-adjacent branches immediately below (`mask[7]`, `mask[8]`, `mask[9]`) all correctly include the `±1` z-offset, confirming this is a copy/paste slip rather than intentional.
- Fixed the offset to `data[pos + NodeT::DIM * NodeT::DIM - 1]` so the correct voxel is checked.

## Test plan
- [ ] CI build/tests for `openvdb` core


sifakis:
Superseeded by #2306

PR #2308, opened 2026-08-31
RasterizeTrilinear Improvements

Idclip:
- Vectorized rasterizeTrilinear
  - Added support for integer and arbitrary sized openvdb vector and matrix attributes (for collocated transfers)
  - Split out PointTransfer tests from TestPointRasterizeTrilinear into TestPointTransfer
  - Improved the tests for RasterizeTrilinear

### DIFF
--- a/openvdb/openvdb/points/PointRasterizeTrilinear.h
+++ b/openvdb/openvdb/points/PointRasterizeTrilinear.h
@@ -5,7 +5,7 @@
 ///
 /// @file PointRasterizeTrilinear.h
 ///
-/// @brief Transfer schemes for rasterizing point data
+/// @brief Weighted trilinear rasterization of point data
 ///
 
 #ifndef OPENVDB_POINTS_RASTERIZE_TRILINEAR_HAS_BEEN_INCLUDED
@@ -19,6 +19,7 @@
 #include <openvdb/tools/Morphology.h>
 #include <openvdb/tree/ValueAccessor.h>
 #include <openvdb/util/Assert.h>
+#include <openvdb/simd/Simd.h>
 
 #include "PointDataGrid.h"
 #include "PointMask.h"
@@ -31,21 +32,42 @@ OPENVDB_USE_VERSION_NAMESPACE
 namespace OPENVDB_VERSION_NAME {
 namespace points {
 
-///
+/// @brief  Traits for staggered (MAC) trilinear rasterization. This method
+///   only works on scalar or Vec3 attribute types and only ever returns a Vec3
+///   tree. It is typically used for velocity rasterization in fluid
+///   simulations to avoid the pressure-velocity decoupling problem that occurs
+///   on collocated (cell-centered) grids. The resulting Vec3 tree effectively
+///   represents the velocity values on the cell faces rather than the cell
+///   centers.
 template <typename ValueT, bool Staggered = true>
 struct TrilinearTraits
 {
-    using ResultT = typename std::conditional<
-        VecTraits<ValueT>::IsVec, ValueT, math::Vec3<ValueT>>::type;
+private:
+    using FltT = typename types_internal::flt_t<sizeof(typename ValueTraits<ValueT>::ElementType)*CHAR_BIT>::type;
+    static_assert(ValueTraits<ValueT>::IsScalar ||
+        (ValueTraits<ValueT>::IsVec && ValueTraits<ValueT>::Size == 3),
+        "Source attribute type must be scalar or Vec3 for staggered rasterization.");
+public:
+    /// @brief  Resulting rasterized Tree ValueType (always Vec3)
+    using ResultT = math::Vec3<FltT>;
+    /// @brief  Resulting rasterized TreeType
     template <typename PointDataTreeT>
     using TreeT = typename PointDataTreeT::template ValueConverter<ResultT>::Type;
 };
 
-///
+/// @brief  Traits for collocated, cell-centered rasterization. Work for any
+///   scalar, vector or matrix attribute. Note that integer values are
+///   interpolated at their respective floating point precision and returned
+///   as trees at that precision.
 template <typename ValueT>
 struct TrilinearTraits<ValueT, false>
 {
-    using ResultT = ValueT;
+private:
+    using FltT = typename types_internal::flt_t<sizeof(typename ValueTraits<ValueT>::ElementType)*CHAR_BIT>::type;
+public:
+    /// @brief  Resulting rasterized Tree ValueType
+    using ResultT = typename ConvertElementType<ValueT, FltT>::Type;
+    /// @brief  Resulting rasterized TreeType
     template <typename PointDataTreeT>
     using TreeT = typename PointDataTreeT::template ValueConverter<ResultT>::Type;
 };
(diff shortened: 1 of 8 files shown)


openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
```diff
@@ -64,230 +91,430 @@ struct TrilinearTransfer :
...
     }
 
+    /// @brief  Multi point rasterization
+    /// @note   This is the only allowed entry point from the transfer schemes
+    inline void rasterizePoints(const Coord& ijk,
+                    const Index start,
+                    const Index end,
+                    const CoordBBox& bounds)
+    {
+        constexpr auto N2 = simd::SimdTraits<NativeT>::size;
```
richhones, line 120: nitpick: inconsistent naming with kBatchSize in Pca stuff (just merged)

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
```diff
@@ -64,230 +91,430 @@ struct TrilinearTransfer :
...
+        }
+#if defined(__GNUC__)  && !defined(__clang__)
+#pragma GCC diagnostic pop
+#endif
+
+        SimdT S[TArgsT::Staggered ? 3 : SourceSize];
+        for (size_t j = 0; j < SourceSize; ++j) {
+            S[j] = simd::load<N2>(cache.data() + (N2*j));
+        }
+        if constexpr (SourceSize == 1 && TArgsT::Staggered) {
```
richhones, line 273: is staggered rasterisation scalars well defined? I thought only vectors mapped to staggered rasterisation.

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
```diff
@@ -64,230 +91,430 @@ struct TrilinearTransfer :
...
-    using BaseT = TrilinearTransfer<TreeType, PositionCodecT, SourceValueT, SourceCodecT, FilterT>;
+    using BaseT = TrilinearTransfer<StaggeredTransfer<TArgsT>, TArgsT>;
     using RealT = typename BaseT::RealT;
-    using BaseT::value;
 
-    static_assert(VecTraits<typename TreeType::ValueType>::IsVec,
-        "Target Tree must be a vector tree for staggered rasterization");
+    static_assert(VecTraits<typename TArgsT::TreeType::ValueType>::IsVec &&
+        VecTraits<typename TArgsT::TreeType::ValueType>::Size == 3,
+        "Target Tree must be a Vec3 tree for staggered rasterization");
```
richhones, line 301: as above, do we need "Source Tree must be Vec3 for staggered rasterisation" too?

openvdb/openvdb/points/PointRasterizeTrilinear.h
```diff
@@ -31,21 +32,42 @@ OPENVDB_USE_VERSION_NAMESPACE
...
 namespace points {
 
-///
+/// @brief  Traits for staggered (MAC) trilinear rasterization. This method
+///   only works on scalar or Vec3 attribute types and only ever returns a Vec3
+///   tree. It is typically used for velocity rasterization in fluid
+///   simulations to avoid the pressure-velocity decoupling problem that occurs
+///   on collocated (cell-centered) grids. The resulting Vec3 tree effectively
+///   represents the velocity values on the cell faces rather than the cell
+///   centers.
```
richhones, line 41: scalars not 'stagger'able

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
```diff
@@ -64,230 +91,430 @@ struct TrilinearTransfer :
...
+        const Coord& ijk)
+    {
+        // Build area of influence depending on point position
+        // @note  We should only movemask once on a single mask but VCL doesn't
+        //   expose a nice portable way to deal with that. We should introduce
+        //   some kind of zero crossing API method in the simd namespace i.e:
+        //     auto m = movemask(P < 0.0);
+        //     m == 0x0 // all positive
+        //     m == 0xF // all negative
+        //     // else mixed
```
richhones, line 444: im not sure what this comment means but i assume you know :P

richhones (commented):
Looks good, minor comments about staggered scalar rasterisation (i.e. scalar source) being ill-defined which this PR doesnt actually change but seeing as you are updating, would be nice to fix.

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
Idclip: I mean, it's well defined, but whether or not its useful? Probably not. For some reason the API we wrote supports it. Are you suggested we remove support for it?

richhones: Ok after a little digging it might be something that has some use cases (viscosity, multi-phase sims), ignore this.

openvdb/openvdb/points/PointRasterizeTrilinear.h
```diff
@@ -65,8 +87,11 @@ struct TrilinearTraits<ValueT, false>
 /// @tparam Staggered whether to perform a staggered or collocated rasterization
 /// @tparam ValueT    the value type of the point attribute to rasterize
 /// @param points     the point tree to be rasterized
-/// @param attribute  the name of the attribute to rasterize. Must be a scalar
-///   or Vec3 attribute.
+/// @param attribute  the name of the attribute to rasterize. Must be a Vec3
```
richhones: following from the staggered scalar discussion, assuming we _do_ allow staggered scalar rasterisation, this can be Vec3 or scalar right?

openvdb/openvdb/points/PointRasterizeTrilinear.h
richhones: as elsewhere, ignore

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
Idclip: done

openvdb/openvdb/points/impl/PointRasterizeTrilinearImpl.h
Idclip: I'm doing two comparisons and horizontal reductions per component (e.g. Px < 0 && Px >= 0). This is because if any component is less than zero then the bounds needs to be -extended, and if any is greater-equal it needs to be +extended. But you have all this information into a single `movemask` operation. Issue is that you need to call the appropriate intrinsic depending on the vector type (i.e. Vec4f, `_mm_movemask`, Vec8f, `_mm256_testz_ps`, etc). Both of these methods return an int that we can check the bits to determine if any are less or any are greater - but VCL instead wraps these methods with more logic to compute the final bool. There is a `to_bits` method in vcl which extracts the bit representation of the comparison, but that isn't as efficient as the `movemask` instruction either. Anyway, it's minor, but was something I coudn't figure out how to do nicely.

openvdb/openvdb/points/PointRasterizeTrilinear.h
Idclip: fixed

PR #2309, opened 2026-09-02
Houdini 22 Weekly

danrbailey:
Add Houdini 22 to the weekly CI. Once this has run, I will follow up and add H22 builds, then finally drop H20.5.

Only small caveat here is that jemalloc no longer ships with Houdini. It might be wise to download/build/install it ourselves for Houdini.

(@jmlait)

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -70,6 +70,7 @@ jobs:
     strategy:
       matrix:
         config:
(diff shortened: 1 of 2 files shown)


Idclip:
> Only small caveat here is that jemalloc no longer ships with Houdini. It might be wise to download/build/install it ourselves for Houdini.

Funnily enough I sent a message to @jmlait about this today - lets discuss in the next TSC

Idclip (approved):

PR #2314, opened 2026-09-06
Add Houdini 22.0 to CI

danrbailey:
Drop Houdini 20.5.

(@jmlait)

### DIFF
--- a/.github/workflows/houdini.yml
+++ b/.github/workflows/houdini.yml
@@ -72,11 +72,11 @@ jobs:
     strategy:
       matrix:
         config:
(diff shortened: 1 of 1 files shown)


Idclip:
Should we bump to 21.5 too?

danrbailey:
> Should we bump to 21.5 too?

There is no Houdini 21.5. This is building against the two latest releases (21.0 and 22.0).

Idclip:
> > Should we bump to 21.5 too?
> 
> There is no Houdini 21.5. This is building against the two latest releases (21.0 and 22.0).

huh I just assumed there would be, nevermind!

Idclip (approved):

PR #2317, opened 2026-09-09
CI: pin nanobind to 2.5.0 on Windows

harrism:
## What

Pins the Windows CI's nanobind to 2.5.0 — the version every other CI platform already pins through `ci/install_nanobind.sh` — by building it from source at that tag instead of taking the floating vcpkg port. It installs into the vcpkg tree (nanobind's install layout matches the port's), so CMake discovery through the vcpkg toolchain is unchanged and no workflow files are touched.

## Why

The Windows runners' vcpkg baseline rolled to nanobind 3.0.0 this week. Its headers trip nvcc front-end diagnostics (#177/#550 in `nb_func.h`) that `--Werror=all-warnings` turns into errors, so every PR's `windows-nanovdb` job now fails in the python-binding CUDA translation units (first seen on #2301, but any PR triggering that job reproduces it, and master will on its next run). Pinning restores the version parity the other platforms already have.

Migrating all platforms to nanobind 3.x is deliberate future work, tracked separately.

## Test plan

The `windows-nanovdb` job on this PR is the test: it exercises the modified install script and must come back green where current PRs fail.

### DIFF
--- a/ci/install_windows.ps1
+++ b/ci/install_windows.ps1
@@ -17,8 +17,7 @@ $vcpkgPackages = @(
     "jemalloc",
     "boost-iostreams",
     "boost-interprocess",
-    "boost-algorithm",
-    "nanobind"
+    "boost-algorithm"
 )
 
 # Update vcpkg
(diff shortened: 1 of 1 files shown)


swahtz (approved):
Thanks for catching this.

apradhana (approved):

danrbailey (approved):
Thanks for fixing this temporarily. Agree we should make updates to support NanoBind 3.0.

The Houdini CI failures are because there is currently only one H22 install available and it was built from a more recent commit than the one that is the ancestor of this PR. Merging the master branch into this one should resolve that issue.

PR #2319, opened 2026-09-09
NanoVDB: per-device memory resources for DistributedPointsToGrid and deprecate UnifiedBuffer (CUDA)

harrism:
Completes the deprecation wave that #2301 opened: the DeviceBuffer and UnifiedBuffer names are deprecated together, in the same release, as #2301's release note promises. Three commits.

## What

1. **`tools::cuda::DistributedPointsToGrid` routes its device-local scratch through per-device memory resources.** New defaulted `ResourceT` template parameter (default `DeviceResource`, unchanged behavior) and constructor overloads taking one resource instance per device in the mesh, indexed by device id, validated with `std::invalid_argument` (count must match the mesh, entries non-null) — memory pools are per-device, hence one instance each. The CUB scratch pools and the segmented-sort tile offsets allocate through them; the tile offsets become a stream-ordered `cuda::Buffer`, replacing a synchronous `cudaFree` that stalled the device mid-pipeline. Storage the host and several devices address through one allocation (the shared `PointsToGridData`, the striped key/index arrays, per-device count/offset arrays) stays managed by design, now owned by `cuda::Buffer<T, ManagedResource>` members instead of ~49 hand-paired `cudaMallocManaged`/`cudaFree` calls split across the constructor, destructor and `processBBox`; the pinned merge intervals move to `cuda::Buffer<ptrdiff_t, PinnedResource>`. The multi-device sort/scan helpers take their pool type as a template parameter. `getHandle` allocates through `createDeviceStorage`, so a `cuda::Buffer<std::byte, ManagedResource>` grid handle works alongside the legacy default.
2. **Multi-GPU tests and examples migrate onto managed `cuda::Buffer` handles.** TestMultiGPU's grid tests and both mgpu examples move off `UnifiedBuffer` (the `deviceDownload` calls disappear — there is no second copy). `ex_make_mgpu_nanovdb` builds its index grid host-side and transfers with `cuda::copyTo`, replacing per-device `deviceUpload` replicas with one managed grid whose pages replicate through read-mostly advice; its feature buffers become `cuda::Buffer<float, ManagedResource>` with explicit `memAdvise`/`memPrefetchAsync`. FewTiles/ManyTiles stay on the default buffer type to keep the legacy path covered.
3. **`cuda::UnifiedBuffer` is deprecated** by the same mechanism as DeviceBuffer in #2301: the implementation is renamed `DualUnifiedBuffer` (transitional, deleted with the alias), `UnifiedBuffer` becomes a `[[deprecated]]` alias whose message carries the migration recipe, and DistributedPointsToGrid's signature defaults name the implementation so default-using callers stay warning-free until the defaults change at removal. The header keeps its name and include path. TestNanoVDB's UnifiedBuffer-surface tests deliberately keep exercising the deprecated name under that file's existing whole-TU suppression.

## Verification

- New test `TestNanoVDBMultiGPU.PerDeviceResources_DistributedCudaPointsToGrid` injects a counting resource per device and asserts every device's scratch routes through and returns to its own instance.
- Full NanoVDB gate green: 156 CPU / 65 CUDA / 12 mgpu on one GPU **and** 12 mgpu on two GPUs / 48 buffer / 19 memory-resource; both mgpu examples run on one and two GPUs. Full build with `NANOVDB_CUDA_WERROR` clean.
- Explore self-review and Codex review passes applied before opening.


nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
```diff
@@ -404,40 +411,57 @@ void inclusiveSumAsync(const nanovdb::cuda::DeviceMesh& deviceMesh, nanovdb::cud
...
+    ///        routing device-local allocations through the default instance of @c ResourceT on every device
     /// @param deviceMesh DeviceMesh on which to run/distribute the operation
     /// @param map Map to be used for the output grid
     DistributedPointsToGrid(const nanovdb::cuda::DeviceMesh& deviceMesh, const Map &map);
+    /// @brief Constructor that also supplies the per-device memory resources
+    /// @param deviceMesh DeviceMesh on which to run/distribute the operation
+    /// @param map Map to be used for the output grid
+    /// @param resources one non-null resource-instance pointer per device in the mesh,
+    ///        indexed by device id; each instance allocates on its device and must outlive this class
+    /// @throw std::invalid_argument if the resource count differs from the mesh's device count
```
swahtz, line 433: **💅 polish:** ‏Let's just have a test for the new `std::invalid_argument` throw; a case for an incorrect resource count and a null entry.

swahtz (commented):
Overall looks fine, just requesting a pair of test cases

nanovdb/nanovdb/tools/cuda/DistributedPointsToGrid.cuh
harrism: Added `TestNanoVDBMultiGPU.PerDeviceResourcesRejected_DistributedCudaPointsToGrid`: one resource too many, one too few, and a null entry at the correct count (so the count check can't mask it) — all three expect `std::invalid_argument`. Fixed in c11af5ba.

swahtz (approved):
Great, thanks for addressing my note

PR #2320, opened 2026-09-10
added fix to unit-test related to NaN on Ubuntu 26.04

kmuseth:
Fix flaky asin/acos unit test assertions on NaN sign bit

asin(2)/acos(2) are domain errors, and IEEE 754 doesn't specify the sign bit libm sets on the resulting NaN. On Ubuntu 26.04 the calculator's std::asin/std::acos path returns -nan while the test's asin(2.0f)/acos(2.0f) reference returns nan, failing four EXPECT_EQ string comparisons in unittest.cpp. Strip the sign before comparing so the test only checks for NaN, not a platform-dependent bit pattern.


apradhana (approved):
Looks good to me. This should resolve [this failure with clang](https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/34271138528/job/102213172973) and [this failure with gcc](https://github.com/AcademySoftwareFoundation/openvdb/actions/runs/34271138528/job/102213173248) on ubuntu 26.04.

danrbailey (approved):

PR #2321, opened 2026-09-10
Update NanoVDB CI Platform

swahtz:
Updates NanoVDB's CI job definition to:

- VFX Platform 2026 (with 2024 lower-bound build mirroring OpenVDB)
- Upgrade macos to 26
- Upgrade CUDA to 12.9 (VFX Platform 2026 docker image version)
- **nvcc ignores gtest's `[[maybe_unused]]` on `test_info_` in clang-host mode** and reports `#177`, which `--Werror=all-warnings` makes fatal.  The CUDA test classes now live in a named namespace so the front end can't prove the member unreferenced. This also retires the `--diag-suppress=177` workaround MSVC hosts needed for the same reason.
- **nvcc was silently using GCC as its host compiler** while `CXX` was clang.  Clang 18 and later mangle`GridHandle`'s dual-buffer move constructor differently from GCC (the non-type template parameter has a dependent type), so the definition emitted by the `.cu` object never matched the reference from the `.cc` objects and the Python module failed at import. `CUDAHOSTCXX` now points nvcc at `CXX`, per matrix entry. The 2024 image keeps g++ as the nvcc host: its clang 17 still mangles the way g++ does, and CUDA 12.6's front end has no `__is_scoped_enum`, which the CCCL headers assume under a clang host. CCCL is also pinned to v3.4.2 rather than tracking `main`.

### DIFF
--- a/.github/workflows/nanovdb.yml
+++ b/.github/workflows/nanovdb.yml
@@ -92,7 +92,7 @@ jobs:
           -DCMAKE_POSITION_INDEPENDENT_CODE=ON
           \'
       - name: test
-        run: cd build && sudo ctest -V -E ".*cuda.*|.*mgpu.*"
+        run: cd build && ctest -V -E ".*cuda.*|.*mgpu.*"
 
   windows-nanovdb:
     if: |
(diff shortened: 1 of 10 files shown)


apradhana (approved):
Looks good to me! Thanks, @swahtz!

danrbailey (approved):

PR #2322, opened 2026-09-10
NanoVDB: type the CUDA scratch buffers as cuda::Buffer<T> (CUDA)

### LINKED ISSUE #2312: NanoVDB: type the Mask scratch buffers as Buffer<Mask<N>> once Mask is trivially copyable
## Context

#2310 defaults `nanovdb::Mask`'s copy constructor, making it trivially copyable. That satisfies `cuda::Buffer<T, R>`'s `is_trivially_copyable<T>` requirement, so the scratch buffers that exist solely to hold arrays of `Mask<N>` can become typed `cuda::Buffer<Mask<N>, ...>` instead of `Buffer<std::byte>` (or a legacy buffer) plus a cast. A census of the tree found exactly seven such sidecar sites; recording them here so they don't need re-finding.

## Pure retypes (change the element type, delete the cast)

1. **`tools::cuda::MeshToGrid`** — the `retainMaskBuffer` in the first `getHandle` overload (allocated as bytes, `reinterpret_cast` to `Mask<3>*`, memset to 0xFF for all-on init).
2. **`tools::cuda::MeshToGrid`** — the per-subdivision-pass `maskBuffer` holding CTA hit results (`Mask<3>`). Its sibling `countsBuffer` is a separate allocation, so no offset arithmetic is affected.
3. **`tools::cuda::MeshToGrid`** — the `retainMaskBuffer` in the UDF-sidecar `getHandle` overload (same shape as 1).
4. **`examples/ex_dilate_nanovdb_cuda`** — `dstLeafMaskBuffer` in the kernels file (`Mask<3>` array).
5. **`unittest/TestNanoVDB.cu`** — the leaf-mask buffer in the dilation test (currently on the legacy dual-space buffer via `create()` + `deviceData()`; retyping folds into its migration to `cuda::Buffer`).

## Retype with a small refactor

6. **`tools::cuda::TopologyBuilder::mUpperMasks`** (`Mask<5>` array): the builder's single byte-buffer `ScratchT` alias is shared by ~9 members, so typing the mask members means per-member aliases; the `deviceUpperMasks()` accessor returns `void*` and each consumer in `util/cuda/Morphology.cuh` and `MeshToGrid.cuh` casts — the accessor should return `Mask<5>*` and the casts go.
7. **`tools::cuda::TopologyBuilder::mLowerMasks`** (`Mask<4>`, viewed as a 2D `Mask<4>(*)[Mask<5>::SIZE]`): same alias/accessor refactor; note the 2D array-shape cast at the use site remains even after the element type is fixed.

## Notes

- The zero/0xFF `cudaMemsetAsync` initializations stay as-is (a typed buffer's storage is still just bytes to initialize; `noInit` construction plus memset remains the right pattern).
- `VoxelBlockManager` has **no** mask sidecar despite being the intuitive candidate — its scratch arrays are `uint32_t`/`uint64_t` (typed-buffer candidates of their own, but not `Mask`).
- Sites 1–5 are all touched by #2301, so this work should land after it merges and be based on its post-merge shape to avoid conflicts.
- Out of scope: casts into grid internals (NanoVDB grids are byte blobs by design), shared-memory reinterpretations, and the union workaround in MeshToGrid's leaf-mask kernel — the latter is blocked by `Mask`'s user-provided zeroing default constructor, which #2310 deliberately leaves alone.

harrism:
Closes #2312. Also lands the sweep portion of #2327 (everything but VoxelBlockManagerHandle, which stays open there pending coordination with #2291).

## What

`Mask` became trivially copyable in #2310, which removed the last reason for the CUDA tools to hold typed arrays in byte buffers reinterpreted at every use. Two commits:

**1. The Mask sidecars (#2312)** — `TopologyBuilder`'s upper/lower masks become `cuda::Buffer<Mask<5>>` / `cuda::Buffer<Mask<4>>`; `deviceUpperMasks()` returns `Mask<5>*` and `deviceLowerMasks()` returns the row-shaped `Mask<4> (*)[Mask<5>::SIZE]`, so the stride is fixed once beside the allocation and the seven morphology functors and every caller take typed pointers (eight casts gone). MeshToGrid's two retain-mask buffers and hit-mask buffer, the dilation example, and the dilation test (off the deprecated dual `DeviceBuffer`) become `cuda::Buffer<Mask<3>>`.

**2. Every remaining single-type scratch buffer** — TopologyBuilder's upper/lower/leaf/voxel offsets, lower/leaf parents, the three enumerate-pass count arrays and its device `Data` staging; MeshToGrid's transformed triangles, box-triangle pairs, unique root origins and nine local key/count/offset/pair buffers (its byte `ScratchT` alias is gone); SignedFloodFill's root-child array. The lower/leaf offsets expose their per-upper-node row shape via `lowerOffsetRows()`/`leafOffsetRows()`, same pattern as the masks. TopologyBuilder's variable-length root staging stays a byte buffer (its size is `RootT::memUsage(tileCount)`, not an element count) and is now the only thing its alignment `static_assert` guards — strengthened to `NANOVDB_DATA_ALIGNMENT`, which is what the root reinterpret actually needs. MeshToGrid gains an equivalent assert for its 16-byte-aligned pair struct, which previously had none.

Allocations are sized in elements and memsets use `size_bytes()`, so the byte arithmetic disappears and `Buffer`'s count×sizeof overflow check applies where hand-multiplied sizes had none.

## Verification

Full build clean under `NANOVDB_CUDA_WERROR`; 156 CPU / 65 CUDA / 48 buffer / 19 memory-resource tests pass. Explore self-review and Codex review passes on both commits; every element type was checked against `Buffer`'s `is_trivially_copyable` requirement (including `RootChild`, whose converting constructor leaves its copy operations trivial).


swahtz (approved):
Looks good to me

kmuseth (approved):
Both Claude and I agree that this looks good :)

PR #2325, opened 2026-09-10
Add the missing NanoVDB namespaces to the doxygen input

### LINKED ISSUE #2324: [BUG] Some namespace missing from API documentation
### Describe the bug
Some namespaces and source files appear to be missing from the API docs hosted on openvdb.org, and trying to search for classes/functions in them does not return any results.

For example, `nanovdb::math` and `nanovdb::tools` appear to be missing, but there may be more

### To Reproduce
Search for, for example, `ZeroCrossing`, which should find `nanovdb::math::ZeroCrossing` but currently only finds `openvdb::math::ZeroCrossing`

### Expected behavior
All namespaces containing public API's show up in the API documentation and can be searched

swahtz:
Fixes #2324.

## The bug

`nanovdb::cuda`, `::io`, `::math`, `::putil`, `::tools` and `::util::cuda` are missing from the API docs on openvdb.org. Searching for `ZeroCrossing` finds only `openvdb::math::ZeroCrossing`.

`doc/CMakeLists.txt` lists only `nanovdb/nanovdb` and `nanovdb/nanovdb/util`, and `DOXYGEN_RECURSIVE` is `NO`. When NanoVDB was reorganised into per-namespace directories the list was never extended. @Idclip's list on the issue names four directories; `nanovdb/nanovdb/cuda` and `nanovdb/nanovdb/util/cuda` are missing too, so this adds six.

## Why it isn't a six-line change

Adding the directories surfaces 53 doxygen diagnostics in headers the docs build has never parsed, and `WARN_AS_ERROR` turns those into a red docs job. They are fixed here rather than suppressed:

| File | Fix |
|---|---|
| `tools/CreatePrimitives.h` | Nine shared doc blocks document `@param tolerance`/`@param ditherOn`, which exist only on the Fp overloads the block does not attach to. Moved into `@details`. |
| `tools/GridChecksum.h` | `@param grid` -> `gridData`; filled in the empty `crc32Head` stub; dropped a duplicate stub on `validateChecksum`; `@data` -> `@c data`. |
| `tools/CreateNanoGrid.h` | Added missing `@param srcGrid`; `@param String` -> `@param name`. |
| `cuda/DeviceMesh.h` | `@param DeviceMesh` -> `@param other`. |
| `cuda/DeviceBuffer.h`, `cuda/DeviceResource.h` | Documented `device`, `bytes`, `alignment`. |
| `cuda/UnifiedBuffer.h` | Two `//@{` openers are not doc comments, leaving their `///@}` unbalanced; both become `///@{`. `@param host` -> `@param device`. |
| `util/cuda/Timer.h` | `start()` takes no arguments, so its stale `@param`s go; a blank line splitting a doc block is closed up; `restart()` documents `os`. |
| `util/cuda/Util.h` | The ignored trailing stream parameter is unnamed, so `@param dummy` cannot bind; folded into the brief. |
| `io/IO.h` | `namespace Internal` wrapped in `@cond` -- the header already warns it must never be called by client code, and doxygen cannot match its `static` declarations to their definitions. |

Every NanoVDB change above is comment-only.

## The one code change (3 lines)

`RootNode::addChild` and `InternalNode::addChild` spelt their parameter type `ChildT` in the out-of-line definitions but `ChildType`/`ChildNodeType` in the declarations. Doxygen matches these textually. It coped while only OpenVDB's nodes were documented, but `nanovdb::tools::build` supplies identically named `RootNode` and `InternalNode` with their own `addChild`, and the definitions then resolve ambiguously.

Both definitions now spell the parameter type `ChildNodeType`, the public alias, and `RootNode`'s declaration does the same -- so the two classes agree with each other as well as with their own definitions. `ChildNodeType` aliases the template parameter in both classes, so no types change.

## Test plan

- [x] Built doxygen 1.14.0 (the version `ci/install_doxygen.sh` pins) and ran the CMake `doc` target with `-DOPENVDB_DOXYGEN_WARN_AS_ERROR=ON`: exit 0, empty warning log. Baseline on master is also 0, so nothing regressed.
- [x] `nanovdb::math::ZeroCrossing` now appears in the search index; namespace pages exist for `nanovdb::cuda`, `::io`, `::math`, `::tools`, `::util`. `nanovdb::io::Internal` is absent, as intended.
- [x] `ci/validate_docs_references.py` passes; `g++ -fsyntax-only` on a TU calling both `addChild` overloads.

## For reviewers

`nanovdb::tools::build` now documents its own `RootNode`, `InternalNode` and `LeafNode`, which sit alongside OpenVDB's identically named classes. That may be more confusing than helpful -- happy to `@cond` the builder's node classes if you agree.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -315,7 +315,13 @@ if(OPENVDB_DOXYGEN_NANOVDB)
     doc/nanovdb/HowToBuild.md
     doc/nanovdb/HelloWorld.md
     nanovdb/nanovdb
+    nanovdb/nanovdb/cuda
+    nanovdb/nanovdb/io
+    nanovdb/nanovdb/math
+    nanovdb/nanovdb/putil
+    nanovdb/nanovdb/tools
     nanovdb/nanovdb/util
+    nanovdb/nanovdb/util/cuda
     nanovdb/nanovdb/examples/ex_bump_pool_buffer
     nanovdb/nanovdb/examples/ex_collide_level_set
     nanovdb/nanovdb/examples/ex_make_custom_nanovdb
--- a/nanovdb/nanovdb/cuda/DeviceBuffer.h
+++ b/nanovdb/nanovdb/cuda/DeviceBuffer.h
@@ -247,6 +247,7 @@ public:
     /// @brief Returns an offset pointer of a specific type from the allocated host memory
     /// @tparam T Type of the pointer returned
     /// @param count Numbers of elements of @c parameter type T to skip
+    /// @param device Device whose buffer is returned, or cudaCpuDeviceId for the host buffer
     /// @warning might return NULL
     template <typename T>
     T* data(ptrdiff_t count = 0, int device = cudaCpuDeviceId) const
@@ -258,6 +259,7 @@ public:
 
     /// @brief Returns a byte offset void pointer from the allocated host memory
     /// @param byteOffset offset of return pointer in units of bytes
+    /// @param device Device whose buffer is returned, or cudaCpuDeviceId for the host buffer
     /// @warning assumes that this instance is not empty!
     void* data(ptrdiff_t byteOffset, int device = cudaCpuDeviceId) const
     {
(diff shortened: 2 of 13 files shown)


Idclip:
lgtm

> nanovdb::tools::build now documents its own RootNode, InternalNode and LeafNode, which sit alongside OpenVDB's identically named classes. That may be more confusing than helpful -- happy to @cond the builder's node classes if you agree.

They should probably be added and we should figure out how to disambiguate them. But I don't have visibility of what this looks like at the moment, so defer to you to decide.

Idclip (approved):

PR #2326, opened 2026-09-10
moved tools/PolySoupToLevelSet.h to vdb_tool/include/ShrinkWrap.h

kmuseth:
- Moved and renamed tools/PolySoupToLLvelSet.h to vdb_tool/include/Shrinkwrap.h

- Removed python bindings to the old polySoupToLevelSet function

- Added unit-tests of the new shrink-wrap function


openvdb_cmd/vdb_tool/include/Tool.h
```diff
@@ -2835,16 +2830,16 @@ void Tool::soupToLevelSet()
     throw std::invalid_argument("got points, expected mesh! Hint: use -points2ls instead!");
   }
   if (keep) mesh = mesh->deepCopy();// deep copy since mesh will be modified below
-  if (mParser.verbose) mTimer.start("Soup -> SDF");
+  if (mParser.verbose) mTimer.start("Shrinkwrap -> SDF");
 
   Spinner spin, *progress = mParser.verbose ? &spin : nullptr;
-  const tools::ShrinkWrapLimit D(nErode, thres);
-  tools::PolySoup poly{std::move(mesh->vtx()), std::move(mesh->tri()), std::move(mesh->quad()), mesh->bbox()};
+  const ShrinkWrapLimit D(nErode, thres);
```
apradhana: `static_cast<float>(nErode)`

openvdb_cmd/vdb_tool/include/Shrinkwrap.h
```diff
@@ -55,7 +55,7 @@ class ShrinkWrapLimit;
 /// @brief Class that implements the actual shrink wrap algorithm
 /// @tparam GridType Template parameter of the desired shrink wrap grids
 template <typename GridType = FloatGrid>
-class PolySoupToLevelSet;
+class Shrinkwrap;
```
apradhana: Previously, it there was `ShrinkWrapLimits`, which I think is correct. My suggestion is to use `ShrinkWarp.h` and `class ShrinkWrap` instead of `Shrinkwrap`. It's similar to `LevelSet` v.s. `Levelset` in my mind. Line `84` below, for example has the correct camel-case: `shrinkWrap`.

kmuseth: You mean camel-case spelling or something else?

openvdb_cmd/vdb_tool/include/Tool.h
kmuseth: nErode is now a float so no casting needed

openvdb_cmd/vdb_tool/include/Shrinkwrap.h
apradhana: Yes and make it consistent as in the [OpenVDB coding-style](https://www.openvdb.org/documentation/doxygen/codingStyle.html#sClassConventions). I think the class should be `ShrinkWrap`, functions are already correct with `shrinkWrap`, and file should be `ShrinkWrap.h`.

kmuseth: dressed your comment in [here](https://github.com/AcademySoftwareFoundation/openvdb/pull/2326/commits/27407c92bb0fdb358f285cfb7e5d50b5bc4b2a3d)

apradhana (approved):
Looks good to me!

danrbailey (approved):
Thanks for making these changes!

PR #2330, opened 2026-09-11
Update Latest Weekly CI Builds

danrbailey:
* Upgrade ASWF images - `ubuntu-22.04` large hosts to `ubuntu-latest` (which is currently `24.04`).
* Upgrade latest `c++20` build to CY2027 and `c++23`
* Upgrade `ubuntu-latest` to `ubuntu-26.04` (this is GCC 15 + Clang 22).
* Upgrade `macos-latest` to `macos-26`.
* Fix a tolerance in point frustum rasterization unit tests due to using a newer libm.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -61,7 +61,7 @@ jobs:
       (github.event_name != 'workflow_dispatch' ||
        github.event.inputs.type == 'all' ||
        github.event.inputs.type == 'houdini')
-    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-22.04-8c-32g-300h') || 'ubuntu-latest' }}
+    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-latest-8c-32g-300h') || 'ubuntu-latest' }}
     name: linux-houdini:${{ matrix.config.hou_hash }}
     env:
       CXX: clang++
@@ -137,7 +137,7 @@ jobs:
       github.event_name != 'workflow_dispatch' ||
       github.event.inputs.type == 'all' ||
       github.event.inputs.type == 'extra'
-    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-22.04-8c-32g-300h') || 'ubuntu-latest' }}
+    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-latest-8c-32g-300h') || 'ubuntu-latest' }}
     name: linux-extra:${{ matrix.config.name }}
     container:
       image: aswf/ci-openvdb:${{ matrix.config.image }}
@@ -291,7 +291,7 @@ jobs:
       github.event_name != 'workflow_dispatch' ||
       github.event.inputs.type == 'all' ||
       github.event.inputs.type == 'ax'
-    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-22.04-8c-32g-300h') || 'ubuntu-latest' }}
+    runs-on: ${{ (github.repository_owner == 'AcademySoftwareFoundation' && 'ubuntu-latest-8c-32g-300h') || 'ubuntu-latest' }}
     name: >
       linux-ax:${{ matrix.config.image }}-cxx:${{ matrix.config.cxx }}-${{ matrix.config.build }}
     container:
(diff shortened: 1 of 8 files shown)


swahtz (approved):

.github/workflows/weekly.yml
```diff
@@ -153,7 +153,7 @@ jobs:
           - { name: 'pygrid', image: '2026', build: 'Release', components: 'core,python,bin,view,render,test', cmake: '-DOPENVDB_PYTHON_WRAP_ALL_GRID_TYPES=ON' }
           - { name: 'asan',   image: '2026', build: 'asan',    components: 'core,test',                        cmake: '-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON -DNANOVDB_USE_OPENVDB=ON -DOPENVDB_AX_STATIC=OFF -DOPENVDB_CORE_STATIC=OFF -DUSE_BLOSC=OFF' } # We never called blosc_destroy(), so disable blosc to silence these errors
           - { name: 'ubsan',  image: '2026', build: 'ubsan',   components: 'core,test',                        cmake: '-DDISABLE_DEPENDENCY_VERSION_CHECKS=ON -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" ' }
-          - { name: 'c++20',  image: '2026', build: 'Release', components: 'core,test',                        cmake: '-DCMAKE_CXX_STANDARD=20' }
+          - { name: 'c++23',  image: '2027', build: 'Release', components: 'core,test',                        cmake: '-DCMAKE_CXX_STANDARD=23' }
```
Idclip: erm we still need to test explicitly with C++20. I'd leave this and move C++23 to the "latest" build job, which is the job we use to test the bleeding edge latest combination of everything

.github/workflows/weekly.yml
```diff
@@ -189,12 +189,10 @@ jobs:
     strategy:
       matrix:
         config:
-          # llvm-17 is currently the default which has symbol issues with the python module
-          - { runson: ubuntu-latest, cxx: g++, cmake: '-DOPENVDB_PYTHON_USE_AX=OFF' }
-          # Disable the clang job for now. See https://github.com/actions/runner-images/issues/8659
-          # - { runson: ubuntu-latest, cxx: clang++, cmake: '' }
+          - { runson: ubuntu-26.04, cxx: g++, cmake: '' }
+          - { runson: ubuntu-26.04, cxx: clang++, cmake: '' }
```
Idclip, line 195: wait, why is this one a specific version but all the rest have changed to latest? This is the _only_ job that should use the latest alias

Idclip (changes requested):
IMO, all the specific ubuntu tests should remain as anchored version numbers i.e. 22->24 or 26. The only job that tests on ubuntu-latest should be the "latest" job in the weekly run.

.github/workflows/weekly.yml
danrbailey: Because `ubuntu-latest` has not been updated to `ubuntu-26.04` yet, latest currently points at `ubuntu-24.04`. (Except `macos-latest` which _does_ point at `macos-26`, I'll update that).

I'll add a comment to that effect at least.

.github/workflows/weekly.yml
danrbailey: I can revert this change, I just wanted to put at least one of the container jobs up on to 2027.

danrbailey (commented):
> IMO, all the specific ubuntu tests should remain as anchored version numbers i.e. 22->24 or 26. The only job that tests on ubuntu-latest should be the "latest" job in the weekly run.

There is no `ubuntu-24.04-8c-32g-300h`, unfortunately these are just the names that the LF decided to call them. The only way to use these larger runners with Ubuntu 24.04 is to use `ubuntu-latest`. To be honest, this is the least important change here given that we run everything on these larger hosts inside a container anyway, but given that ubuntu-22.04 is being deprecated, felt sensible to pick the ubuntu-24.04 one while I was doing these updates.

.github/workflows/weekly.yml
Idclip: That's weird... looks like the image is still in "preview"? Probably worth just leaving this as `ubuntu-latest`

.github/workflows/weekly.yml
Idclip: Yeah that's fine, and technically all our ABI 13 jobs could be on C++20, but they're not - and we still need to test it explicitly (as in, testing against C++23 doesn't guarantee C++20 build success)

Idclip:
> > IMO, all the specific ubuntu tests should remain as anchored version numbers i.e. 22->24 or 26. The only job that tests on ubuntu-latest should be the "latest" job in the weekly run.
> 
> There is no `ubuntu-24.04-8c-32g-300h`, unfortunately these are just the names that the LF decided to call them. 

Ah that explains it, a bit annoying but makes sense now

Idclip (approved):
lgtm - I would personally still leave the latest job using "ubuntu-latest" so we don't have to worry about it, unless you need to test 26 now

danrbailey:
> lgtm - I would personally still leave the latest job using "ubuntu-latest" so we don't have to worry about it, unless you need to test 26 now

I'd like to leave it as 26 for now. It's only the weekly CI and I prefer that we have some testing for GCC 15 + Clang 22. (GCC 15 is over a year old now)

PR #2339, opened 2026-09-16
v13.1.0

Idclip:


apradhana:
One small release-version inconsistency: `pyproject.toml` still says `12.0.1`. Should we update it to `13.1.0` in this PR?

Idclip:
> One small release-version inconsistency: `pyproject.toml` still says `12.0.1`. Should we update it to `13.1.0` in this PR?

Ah, too late for 13.1 now, we can fix this in 13.2. I suggest we merge this, then I'll create an MR to bump everything to 13.1.1 and fix it there.

FWIW, this isn't in the release instructions and we should probably be auto-generating anything that relies on the OpenVDB Version during cmake configure time

apradhana (approved):

PR #2342, opened 2026-09-17
Limit various level set only operations in generically templated tools

Idclip:


openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Combine.cc
```diff
@@ -1298,24 +1282,30 @@ struct SOP_OpenVDB_Combine::CombineOp
...
-                MulAdd<GridT>(aMult).process(*aGrid, resultGrid);
-                MulAdd<GridT>(bMult).process(*bGrid, tempGrid);
-                doDifference(*resultGrid, *tempGrid,
-                        self->evalInt("prunedegenerate", 0, self->getTime()));
+                if constexpr (openvdb::RealGridTypes::Contains<GridT>) {
+                    MulAdd<GridT>(aMult).process(*aGrid, resultGrid);
+                    MulAdd<GridT>(bMult).process(*bGrid, tempGrid);
+                    openvdb::tools::csgDifference(*resultGrid, *tempGrid, /*prune*/true,
+                        /*prunedgen*/self->evalInt("prunedegenerate", 0, self->getTime()));
+                }
```
Idclip, line 1308: @jmlait the combine sop is the only one I'm unsure about - these all compile and do _something_ for non float types. I don't know if there's an early exit elsewhere in the sop for non floats or if Houdini today does process all non bool types for the csg ops?

Idclip: A cursory glance looks like it's only checking the metadata - so, in theory, you could have two int or even two vector grids in Houdini, change their metadata to levelset (I assume it's possible to hack that) and get an output from the sop

danrbailey (approved):

openvdb_houdini/openvdb_houdini/SOP_OpenVDB_Combine.cc
jmlait: This seems reasonable to drop support, whatever it is, for vector/int.

jmlait (approved):

PR #2343, opened 2026-09-17
Use the log4cplus::log4cplus target name exported by log4cplus

### LINKED ISSUE #2341: [BUILD] FindLog4cplus.cmake : uppercase vs lowercase confusion
### Environment
**Operating System:** Rocky Linux 9.8
**Version / Commit SHA:** 4d95e56ba7fb227dabea88b4a28c8ae0fca43392
**CMake Version:** 4.3.3
**Compiler:** gcc 14.2.1

### Describe the problem
OpenVDB's custom CMake module for log4cplus in `cmake/FindLog4cplus.cmake` names the projects `Log4cplus`, and for instance in https://github.com/AcademySoftwareFoundation/openvdb/blob/4d95e56ba7fb227dabea88b4a28c8ae0fca43392/openvdb/openvdb/CMakeLists.txt#L157 

But the exported CMake config from the log4cplus project calls itself `log4cplus::log4cplus`

https://github.com/log4cplus/log4cplus/blob/88bfae3f92ecb492f22e8ab48dd2ea33bda7270c/src/cmake/Config.cmake.in#L13

This causes issue when building in a CMake "config" environment which uses the CMake config installed by a build of log4cplus.

It would be useful if OpenVDB's custom `FindLog4cplus.cmake` module used the same symbols / casing convention as the CMake config from log4cplus, making it transparent whether CMake was using:

- an auto generated CMake config from Conan
- an installed CMake config from log4cplus
- OpenVDB's custom CMake module

### To Reproduce

The aswf-docker project will soon have a workaround for this issue, I will add a link to this issue once that is merged in.

itz-puneet:
Fixes #2341

`FindLog4cplus.cmake` creates an imported target named `Log4cplus::log4cplus`, but the CMake config installed by log4cplus exports `log4cplus::log4cplus` ([`src/CMakeLists.txt`](https://github.com/log4cplus/log4cplus/blob/master/src/CMakeLists.txt) sets the namespace from `PROJECT_NAME`), and so does the config Conan generates. When `find_package(Log4cplus)` resolves through a config package, only the lowercase target exists, and OpenVDB fails to configure because both `openvdb/openvdb/CMakeLists.txt` and `FindOpenVDB.cmake` link `Log4cplus::log4cplus`.

### Changes

- `FindLog4cplus.cmake` now creates `log4cplus::log4cplus` and adds `Log4cplus::log4cplus` as an alias, so existing users of the old name are unaffected. The docstring also listed a third spelling, `Log4cplus::Log4cplus`, which is corrected.
- `openvdb/openvdb/CMakeLists.txt` and `FindOpenVDB.cmake` link `log4cplus::log4cplus`, which now exists in both module and config mode.
- `FindOpenVDB.cmake` is installed and used by downstream projects, which may put their own older copy of `FindLog4cplus.cmake` first on `CMAKE_MODULE_PATH`. For that case it adds `log4cplus::log4cplus` as an alias when only `Log4cplus::log4cplus` is defined.
- Added `pendingchanges/log4cplus_target_name.txt`.

### Testing

No CI job builds with log4cplus, so I tested this with a small CMake project against two fake log4cplus installs. The project runs the `find_package(Log4cplus)` block copied verbatim from `FindOpenVDB.cmake`, calls `find_package(Log4cplus)` a second time as the core `CMakeLists.txt` does, and then checks that each target name OpenVDB links against exists.

| Mode | Setup | Before | After |
|---|---|---|---|
| module | `Log4cplus_ROOT` pointing at a header + library prefix | passes | passes; both names exist, the alias keeps `IMPORTED_LOCATION` |
| config | `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`, Conan-style `log4cplus-config.cmake` exporting only `log4cplus::log4cplus` | **fails**: `Log4cplus::log4cplus` is not a target | passes |
| legacy | the previous `FindLog4cplus.cmake` placed first on `CMAKE_MODULE_PATH` | n/a | passes via the alias in `FindOpenVDB.cmake` |

Tested with CMake 4.4.3. I haven't done a full OpenVDB build with log4cplus enabled.


danrbailey:
To be honest, I think we should remove the log4cplus dependency altogether. It is an optional dependency that is rarely used (perhaps only by Dreamworks?) and as you noted is not even tested by our CI. Not discussed among the rest of the TSC yet, so only my opinion at this stage, but I think it's a good candidate to be removed altogether.

itz-puneet:
That makes sense given it isn't exercised in CI. This change is small and backward compatible, so it could fix config-package builds now (the aswf-docker images are adding a workaround for it) while the TSC decides. If you'd rather go straight to removing the dependency, I'm happy to close this and put up a PR for that instead.

Issue #2345, opened 2026-09-17
[BUG] OpenVDB::nanovdb imported target lacks NANOVDB_USE_OPENVDB, so NanoVDB<->OpenVDB conversion does not compile against an installed OpenVDB


appearce-newentity:
**Symptom**: 
Configuring vdb_tool against an installed OpenVDB (OPENVDB_BUILD_CORE=OFF, OPENVDB_TOOL_USE_NANO=ON) succeeds - find_package(OpenVDB COMPONENTS nanovdb) reports the component found - but compilation fails in Tool.h:

  Tool.h:2400: error: 'openToNanoVDB' is not a member of 'nanovdb::tools'
  nanovdb/tools/CreateNanoGrid.h:333: error: '... openvdb::Grid<...>' has no member named 'gridClass'
  nanovdb/tools/CreateNanoGrid.h:332: error: '... openvdb::Grid<...>' has no member named 'map'
 
  i.e. CreateNanoGrid.h is being compiled with its OpenVDB support disabled, and treats an openvdb::Grid as a NanoVDB grid.

**Cause**: 
the macro is only ever set on the in-tree target. nanovdb/nanovdb/CMakeLists.txt lines 339–346:

  if(NANOVDB_USE_OPENVDB)
    ...
    target_link_libraries(nanovdb INTERFACE openvdb)      # or OpenVDB::openvdb
    target_compile_definitions(nanovdb INTERFACE -DNANOVDB_USE_OPENVDB)
  endif()

The installed package's equivalent, cmake/FindOpenVDB.cmake lines 799-802, creates OpenVDB::nanovdb as an INTERFACE IMPORTED target carrying only INTERFACE_INCLUDE_DIRECTORIES  - no NANOVDB_USE_OPENVDB definition and no link dependency on OpenVDB::openvdb. The in-tree and installed targets therefore have different contracts, and every consumer that worked in-tree breaks when built standalone.
  
**Reproduce**:
cmake --install <openvdb-build> --prefix <inst>, then configure any project with find_package(OpenVDB REQUIRED COMPONENTS openvdb nanovdb) and compile a TU that calls nanovdb::tools::createNanoGrid(openvdb_grid). vdb_tool with OPENVDB_BUILD_CORE=OFF -DOPENVDB_TOOL_USE_NANO=ON is a ready-made case. Observed on v13.1.0 (4d95e56b), Ubuntu 24.04, gcc 13.3, CMake 3.28.
  
[...]

Issue #2344, opened 2026-09-17
[BUG] typo: vdb_tool: OPENVDB_TOOL_USE_PDAL=ON cannot configure on Windows — find_package(libdpal …) typo


appearce-newentity:
openvdb_cmd/vdb_tool/CMakeLists.txt, lines 98–108 on master:

  if(OPENVDB_TOOL_USE_PDAL)
    target_compile_definitions(vdb_tool_common INTERFACE "VDB_TOOL_USE_PDAL")
    if(WIN32)
      find_package(libdpal CONFIG REQUIRED)      # <-- line 101
    else()
      find_package(PDAL REQUIRED)
    endif()
    message(STATUS "PDAL: ${PDAL_LIBRARIES} ${PDAL_INCLUDE_DIRS}")
    target_link_libraries(vdb_tool_common INTERFACE ${PDAL_LIBRARIES})
    target_include_directories(vdb_tool_common INTERFACE ${PDAL_INCLUDE_DIRS})
  endif()

Problem: The Windows branch asks for a CMake package named libdpal — a transposition of libpdal. No such package exists, so on Windows OPENVDB_TOOL_USE_PDAL=ON (and therefore OPENVDB_TOOL_USE_ALL=ON) always fails at configure time with Could not find a package configuration file provided by "libdpal".

Fixing the spelling alone is not enough. The variables consumed afterwards — PDAL_LIBRARIES, PDAL_INCLUDE_DIRS — are set by PDAL's own PDALConfig.cmake, whose package name is PDAL on every
  platform (PDAL's documented usage is find_package(PDAL REQUIRED CONFIG); the vcpkg port installs that same config). A libpdal package config would not define them either. The WIN32 special case appears to have no purpose.

Suggested fix: collapse the branch to a single find_package(PDAL CONFIG REQUIRED), or find_package(PDAL REQUIRED) matching the existing non-Windows line.

Introduced: f04f2f68 (2024-05-20, "Added feature to read points into vdb_tool using Point Data Abstraction Library (PDAL)"). Present in every release since, including v13.1.0.

Why unnoticed: CI never builds vdb_tool with PDAL on any platform (grep -ri pdal ci/ .github/workflows/ is empty), and no existing issue or PR mentions libdpal.

 Severity: low — it blocks one optional feature on one platform, loudly, at configure time. No effect on the default build.

Issue #2278, opened 2026-08-11
vdb_tool: move implementations out of headers into a compiled vdb_tool_common library


harrism:
### Summary

`vdb_tool` is implemented almost entirely in headers: `Tool.h`, `Geometry.h`, `Parser.h`, `Calculator.h` and `Util.h` under `openvdb_cmd/vdb_tool/include/` total ~10,800 lines, while the two `.cpp` files (`src/main.cpp`, `src/unittest.cpp`) are thin drivers. `vdb_tool_common` is declared as an `INTERFACE` library, so nothing is separately compiled.

`Tool.h` alone is 4,201 lines. This works today, but it has three costs worth addressing before the tool grows further. Nothing is currently broken and this is not urgent.

### Problem 1: the headers are not include-safe

`class Tool` is declared with a declaration-only body, and its members are then defined **out-of-line at namespace scope without `inline`** — currently 61 non-template member definitions in `Tool.h` alone (plus 4 template members, which are fine). `Geometry.h` and `Parser.h` follow the same pattern.

That is legal only because `Tool.h` is included by exactly one translation unit per executable:

- `src/main.cpp` → `add_executable(vdb_tool src/main.cpp)`
- `src/unittest.cpp` → `add_executable(vdb_tool_test src/unittest.cpp)`

Since `vdb_tool_common` is `INTERFACE`, it contributes no compiled TU, so each binary links exactly one definition of each member and the ODR is satisfied.

The moment anyone adds a second `.cpp` to either target — splitting `unittest.cpp`, adding a new entry point, adding a benchmark — every one of those 61 members becomes a duplicate-symbol link error. The failure is loud rather than silent, so this is a maintenance trap rather than a latent correctness bug, but it makes the headers effectively un-reusable and it is a surprising constraint that isn't documented anywhere.

The free helper functions in `Tool.h` (`printBanner`, `formatBytes`, `formatNodes`, etc.) *are* correctly marked `inline`, and there are no non-`inline` namespace-scope mutable globals in any of the five headers, so the problem is confined to the out-of-line member definitions.

### Problem 2: the same ~75s of compilation is performed twice

Measured on this machine (g++ 13.3.0, `-std=c++17 -O2`, x86_64, headers from `master`, optional features NanoVDB/USD/OpenEXR/PDAL/Alembic all OFF), compiling each TU to an object file:

| Translation unit | Time |
|---|---|
| `src/main.cpp` | **77.8s** |
[...]

PR #2346, opened 2026-09-17
vdb_tool: mark out-of-line definitions inline so headers are ODR-safe (#2278, part 1)

appearce-newentity:
Part 1 of a three-PR series for #2278 (`vdb_tool`: implementations live in headers, `vdb_tool_common` is an `INTERFACE` target). This PR is the minimal, mechanical step that makes the headers safe to include from more than one translation unit; the library conversion and the `Tool.h` split follow as separate PRs stacked on this one.

## What changes

`inline` is added to every out-of-line definition that has external linkage, and the two `auto` return types on `Tool::getGrid`/`getGeom` are spelled out (a prerequisite for moving them later). Nothing moves; no behaviour changes.

The set of definitions was found with the linker rather than by inspection: compile two TUs that include a header, link them, and count `multiple definition of` diagnostics. A regex census undercounts here — it misses constructors (no return type), free functions, and explicit specializations such as `vectorize<int>` and `Parser::get<std::string>`, which look like templates but are ordinary functions.

| Header | Duplicate symbols before | After |
|---|---|---|
| `Util.h` | 5 | 0 |
| `Parser.h` | 15 | 0 |
| `Geometry.h` | 43 | 0 |
| `Tool.h` (transitively) | **120** | **0** |

## Verification

- Two-TU link of every header: 0 duplicate symbols, 0 warnings (gcc 13.3, Ubuntu 24.04).
- `vdb_tool_test` 34/34; `-sphere r=10 -ls2mesh -write` output byte-identical.
- Compile time is unchanged by design (nothing has moved yet).

## What follows

- Part 2 converts `vdb_tool_common` to a static library and moves the `Geometry`/`Parser` implementations into `.cc` files (also fixes three small CMake defects and moves the tinygltf implementation so `USE_GLTF` builds link with the library).
- Part 3 splits `Tool.h` into ten subsystem sources: `main.cpp` compile drops from ~228 s / 6.2 GB to ~9.5 s / 1.0 GB, and a `-j12` build that was OOM-killed on a 27 GB machine completes.
- A fourth, include-what-you-use pass takes `main.cpp` to ~5 s.

Each part is verified against five configurations (all-off; PNG+JPG+MPEG+GLTF; Debug standalone; clang-18 standalone; AX with LLVM 18). Two pre-existing issues were found along the way and filed separately: #2344 (PDAL `find_package(libdpal …)` typo) and #2345 (`OpenVDB::nanovdb` imported target lacks `NANOVDB_USE_OPENVDB`).

Refs #2278.

Submitted as part of ASWF Dev Days 2026 (`devdays26`).


kmuseth (approved):
Thanks Andrew - this looks good though I might change back some of the (now very verbose) return types to auto.

PR #2347, opened 2026-09-17
Add tools::vectorFromScalar

tstraubinger:
This adds a `tools::vectorFromScalar` free function for combining three scalar grids as the components of a single vector-valued grid. This can later replace the implementation inside of openvdb_houdini's Vector Merge SOP. I've added fairly extensive unit testing as well.

This work was originally proposed to me offline by Dan Bailey, and there isn't an associated tracking issue.


openvdb/openvdb/tools/VectorFromScalar.h
```diff
@@ -0,0 +1,197 @@
...
+    using ScalarT = typename ScalarGridT::ValueType;
+    using VectorT = math::Vec3<ScalarT>;
+    using ScalarTreeT = typename ScalarGridT::TreeType;
+    using VectorGridT = typename ScalarToVectorConverter<ScalarGridT>::Type;
+    using VectorTreeT = typename VectorGridT::TreeType;
+
+    math::Transform transform = x.transform();
+    if (transform != y.transform() || transform != z.transform())
+    {
+        OPENVDB_THROW(ValueError, "vectorFromScalar: all input grids must have the same transform");
```
Idclip, line 159: why not make this method work on trees instead of grids - then you don't need to worry about this check

openvdb/openvdb/tools/VectorFromScalar.h
```diff
@@ -0,0 +1,197 @@
...
+    // Perform a topology union with each of the source scalar grids.
+    // This leaves the combined grid whose tile and voxel depth is the
+    // maximum of all the source grids.
+    vectorGrid->topologyUnion(x);
+    vectorGrid->topologyUnion(y);
+    vectorGrid->topologyUnion(z);
+
+    // Write the grid values
+    auto nodeManager = tree::NodeManager<VectorTreeT>(vectorGrid->tree());
+    if (copyInactiveValues)
```
Idclip, line 176: in the spirit of less template instantiation, maybe just leave this as a runtime switch, unless your profiling shows a benefit

openvdb/openvdb/tools/VectorFromScalar.h
```diff
@@ -0,0 +1,197 @@
...
+
+    auto background = VectorT(x.background(), y.background(), z.background());
+
+    // Create an empty grid
+    auto vectorGrid = createGrid<VectorGridT>(background);
+
+    // Perform a topology union with each of the source scalar grids.
+    // This leaves the combined grid whose tile and voxel depth is the
+    // maximum of all the source grids.
+    vectorGrid->topologyUnion(x);
```
Idclip, line 170: Is it possible to initialize a vector grid directly from a scalar grid? If it is, it might be worth choosing the largest scalar grid and using the vector grids copy constructor to initialize, then topology union into it with the other 2, then only copy over (or reset for inactive) the other 2's values. It would save you 1/3 of the subsequent lookup cost

openvdb/openvdb/tools/VectorFromScalar.h
```diff
@@ -0,0 +1,197 @@
...
+
+    using VectorRootT = typename VectorTreeT::RootNodeType;
+    using VectorLeafT = typename VectorTreeT::LeafNodeType;
+
+    VectorFromScalarOp(const ScalarTreeT* x, const ScalarTreeT* y, const ScalarTreeT* z)
+        : mXTree(x)
+        , mYTree(y)
+        , mZTree(z)
+    {}
+
```
Idclip, line 40: would it be worth using accessors in this op

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: Good suggestion, will do

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: Good idea, yes I think it should be straightforward to do just a topology copy of the first grid. I'll give that a try.

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: I'll try that and do some profiling in Houdini and come back with numbers. Dan and I had found previously that in some workflows, particularly visiting grids with identical transforms, probing internal nodes directly led to better performance than using accessors. I'll see if I can reproduce that here or not.

tstraubinger:
It looks like many CI failures are due to -Wall -Werror flags being added which aren't included in my local build. I've added them myself with `-DCMAKE_CXX_FLAGS="-Wall -Werror"` in my CMake command to avoid CI churn but I'd like to know why they're not added by default locally.

openvdb/openvdb/tools/VectorFromScalar.h
danrbailey: I actually think it might just be worth splitting this into two operators would be more readable and the boiler plate here is minimal. (ie `VectorFromScalarCopyInactiveOp` or something else if that is too long)

danrbailey:
> It looks like many CI failures are due to -Wall -Werror flags being added which aren't included in my local build. I've added them myself with `-DCMAKE_CXX_FLAGS="-Wall -Werror"` in my CMake command to avoid CI churn but I'd like to know why they're not added by default locally.

Try `OPENVDB_CXX_STRICT=ON`. I believe this setting is off by default because not all components of OpenVDB is build-able without warnings currently. We only enable this for the core library builds I think in the CI.

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: I tried this and it was marginally slower. Presumably the grid TopologyCopy constructor does a little more work than a topology union.

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: I tried using function-local ValueAccessors for the non-root `operator()` instead of probing tiles and it's vastly slower. Numbers will be below

tstraubinger:
I added a basic time measurement to the VDB Vector from Scalar SOP, to measure the duration taken by its `ScalarGridMerger` helper. I then replaced its implementation with the one here and tried the various suggestions in isolation. Input grids are the result of VDB-from-polygons on many randomly oriented overlapping cubes with different topologies between the x, y, and z grids. Each input grid has roughly 1 billion active voxels and the vector-merged grid has roughly 1.6 billion active voxels. Measurements were repeated 16 times in each case. Commit hashes point to my local dev repo but I can share the relevant code upon request. This was on a 32-core x86 Linux machine.

| Experiment                                     | Mean & StdDev (ms) |
|------------------------------------------------|--------------------|
| Baseline (master)                              | 4280 +/- 260       |
| Initial version (this pull request)                    |  453 +/- 26        |
| Topo copy + union (874efa26c)                  |  459 +/- 40        |
| Runtime copy inact. (b97d7bc06)                |  469 +/- 34        |
| Separate ops inact. (d77fe36d2)                |  454 +/- 23        |
| ValAcc not tile probe (957561b89)              | 2850 +/- 150       |

TL;DR I was not able to make the code go any faster with the suggestions given. ValueAccessors in particular when used as function-local variables in the functor given to `foreachTopDown` produced a very significant slowdown versus probing the corresponding tiles directly.

openvdb/openvdb/tools/VectorFromScalar.h
tstraubinger: I tried adding a separate `VectorFromScalarCopyInactiveOp` but it resulted in more duplicated than shared code. There's also exactly this approach (minus `if constexpr`) used in `tools/Activate.h`.

tstraubinger:
Some non-performance-related questions:
1. Should there be explicit template instantiation macros towards the end, i.e.  `#ifdef OPENVDB_USE_EXPLICIT_INSTANTIATION`?
2. Should there be an interrupt to the op, e.g. so that Houdini cooking can be aborted early?
3. Should the `VectorFromScalarOp` be visible to client code in the `tools` namespace or should it be hidden?

openvdb/openvdb/tools/VectorFromScalar.h
Idclip: I suspect the copy constructor is not multithreaded which is a pain

Idclip:
Thank you for your efforts in profiling the suggestions @tstraubinger, much appreciated and interesting to see. In regards to: 

> 1 Should there be explicit template instantiation macros towards the end, i.e. #ifdef OPENVDB_USE_EXPLICIT_INSTANTIATION?
2 Should there be an interrupt to the op, e.g. so that Houdini cooking can be aborted early?
3 Should the VectorFromScalarOp be visible to client code in the tools namespace or should it be hidden?

@danrbailey (1)

2) This is a pretty fast operation, I personally wouldn't bother with an interrupter
3) Unless you are expecting people to use this class directly, no, I'd wrap it in a private namespace. I'd also actually move it into the nested `impl` folder such that the API is clearer

But otherwise approved from me

Idclip (approved):

